PR #49, opened 2015-04-15
MASTER - added the oculus icon

maciejgryka:
As requested - but the functionality seems broken both when pressing 'o' and this.


jozefdobos:
Thanks! The 'o' press works in Chrome, Firefox reports an error. For full integration, download the latest Chrome custom build from here: http://webvr.info

Issue #719, opened 2018-02-07
Opening issues takes a long time


JamesMilnerUK:

JamesMilnerUK:
This still isn't quite perfect, but probably worth leaving for #705

PR #720, opened 2018-02-07
ISSUE #719 - Reduce issues load times to around ~1/3 of previous

JamesMilnerUK:
This should speed up showing of issues by about 3 times. It's still not perfect (2-3 seconds for bigger models with complex show/hide groups) but it doesn't mess with any core logic of TreeService.

Premises:

We reset the Tree state before doing certain operations (i.e. showAllTreeNodes) and this updates the model also. This isn't necessary as we immediately impose our new state on the tree and then show/hide the bits of the model we need to. Using a flag we can decide if we want to update the model or just the state.

I also unwound updateParentVisibility to not be recursive.

The next big bottle neck is we have no heuristic for not going up paths we've already checked so we traverse up the tree more times then we need to updating parents. Anyway, one for another time.

### DIFF
--- a/frontend/components/issues/js/issues.service.js
+++ b/frontend/components/issues/js/issues.service.js
@@ -531,7 +531,8 @@
 
 			// Remove highlight from any multi objects
 			ViewerService.highlightObjects([]);
-			
+			TreeService.clearCurrentlySelected();
+
 			// Reset object visibility
 			if (issue.viewpoint && issue.viewpoint.hasOwnProperty("hideIfc")) {
 				TreeService.setHideIfc(issue.viewpoint.hideIfc);
@@ -649,9 +650,6 @@
 
 			TreeService.getMap()
 				.then(function(treeMap){
-
-					// show currently hidden nodes
-					TreeService.clearCurrentlySelected();
 
 					for (var i = 0; i < objects.length; i++) {
 						var obj = objects[i];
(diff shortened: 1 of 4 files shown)


frontend/components/issues/js/issues.service.js
```diff
@@ -536,7 +536,7 @@
 			if (issue.viewpoint && issue.viewpoint.hasOwnProperty("hideIfc")) {
 				TreeService.setHideIfc(issue.viewpoint.hideIfc);
 			}
-			TreeService.showAllTreeNodes();
+			TreeService.showAllTreeNodes(false);
```
Charence: This might be causing a problem in the case where issue has no hidden objects and no selected objects as I believe this always expects something else to update the tree. Perhaps something to add in the else case of checking for groups, handleHighlights, handleHidden, handleTree?

e.g. http://127.0.0.1/charence/1556f6b2-7db9-4ce3-bb73-24b6ad82073c/5fbb96c8-6dbf-4ab1-b07e-57a36cb126f8 click on ISSUE "Architectural.7 ABSOLUTELY EVERYTHING NO SELECTION"

You can see the tree updating, but the viewer doesn't change

JamesMilnerUK: Is there a quick a check for that?

PR #738, opened 2018-02-13
Issue 689 - prune nodes for issues

Charence:


backend/models/group.js
```diff
@@ -55,78 +55,18 @@ groupSchema.statics.listGroups = function(dbCol){
...
-				)
-			);
-		});
-
-		return Promise.all(removePromises);
-
-	}).then(() => {
-
-		data.objects.forEach(obj => {
+	data.objects.forEach(obj => {
```
JamesMilnerUK: would use for loops here as it's performance sensitive

backend/models/group.js
```diff
@@ -150,8 +90,11 @@ groupSchema.methods.clean = function(){
 	cleaned._id = utils.uuidToString(cleaned._id);
 	cleaned.issue_id = cleaned.issue_id && utils.uuidToString(cleaned.issue_id);
 	cleaned.objects.forEach(object => {
```
JamesMilnerUK: for loops again

PR #744, opened 2018-02-16
Issue 737 - Upgrade to Unity 2017.3

JamesMilnerUK:


frontend/globals/d.ts
```diff
@@ -0,0 +1,6 @@
+declare global {
```
carmenfan: Any reason this ts is called d.ts?
Just not exactly clear what this does (to my non frontend developer brain at least!)

JamesMilnerUK: This basically tells TypeScript that these are global types that it shouldn't need to complain about when it sees them in files in this folder

carmenfan: I get what the code is doing, I'm questioning the naming of this file XD

JamesMilnerUK: I did a lot of rejiging in the TypeScript Home branch, it's called overwrites.d.ts there :)

PR #786, opened 2018-03-07
ISSUE #780 - Update to the latest version of moment.js 

JamesMilnerUK:
… libs


JamesMilnerUK:
@carmenfan you might want to review but I think it passed (at least the push). This will make GitHub hush about the security vuln.

carmenfan:
This is used for pro-rata calculation on subscriptions i think. I'll double check this after subscription refactoring is done.

Issue #763, opened 2018-02-28
Subscription refactoring


carmenfan:
**Moving Jobs out of subscription**
Before:
```` js
//in user bson object:
 "subscriptions" : [ 
                {
                    "_id" : ObjectId("579b30134d831e1ef8ea2081"),
                    "plan" : "THE-100-QUID-PLAN",
                    "assignedUser" : "carmen",
                    "active" : true,
                    "createdAt" : ISODate("2016-07-29T10:29:39.000Z"),
                    "updatedAt" : ISODate("2016-07-29T10:29:39.000Z"),
                    "expiredAt" : null,
                    "limits" : {
                        "spaceLimit" : 10737418240.0,
                        "collaboratorLimit" : 1
                    },
                    "job" : "Project Manager"
                }, 
]
...
"jobs" : [ 
            {
                "_id" : "Project Manager",
                "color" : "#e3bd1a"
            }, 
            {
                "_id" : "test",
                "color" : "#33a02c"
            }
        ],

````
after:
````js
//in collection db.jobs
{
    "_id" : "Project Manager",
    "colour": "#e3bd1a",
    "users": ["carmen"]
}

````
- [x] Database migration script to extract job assignment out of user, and job declaration out of user (should be in a newly created collection called jobs within their own database
- [x] Update job schema within the backend
- [x] Ensure new account get default jobs added correctly
- [x] Ensure add/remove/edit job/lookup APIs functions correctly
- [x] Move job assignment away from subscription API (jobs should probably use it's own endpoint
- [x] Ensure job assignment/removal functions correctly

**Subscription changes**
before:
````js
//in user bson
 "subscriptions" : [ 
                {
                    "_id" : ObjectId("579b30134d831e1ef8ea2081"),
                    "plan" : "THE-100-QUID-PLAN",
                    "assignedUser" : "carmen",
                    "active" : true,
                    "createdAt" : ISODate("2016-07-29T10:29:39.000Z"),
                    "updatedAt" : ISODate("2016-07-29T10:29:39.000Z"),
                    "expiredAt" : null,
                    "limits" : {
                        "spaceLimit" : 10737418240.0,
                        "collaboratorLimit" : 1
                    },
                    "job" : "Project Manager"
                }, 
]
````
after:
````js
//in user bson
"subscriptions": {
        "paypal": [
            {
                "plan": id_1,
                "quantity": 10,
                "expiryDate":  ISODate("2016-07-29T10:29:39.000Z")
            },
[...]

PR #835, opened 2018-03-26
ISSUE #763 - Subscription refactoring

carmenfan:
This will require a db migration update once it's merged into staging.


backend/models/invoice.js
```diff
@@ -208,17 +205,15 @@
 			// add items bought in the invoice
 			plans.forEach(plan => {
 				for(let i=0 ; i< plan.quantity; i++){
-					//console.log('init invoice add items', plan, Subscription.getSubscription(plan.plan).plan);
 					this.items.push({
-						name: Subscription.getSubscription(plan.plan).plan,
-						currency: Subscription.getSubscription(plan.plan).currency,
+						name: plan.plan,
+						currency: "GBP",
```
Charence: I guess we're not planning to have non-GBP pricing anytime soon?

carmenfan: No, we don't even have the facility to show different currencies, £ is hard coded everywhere.

So when we do support foreign currencies we'd need a massive overhaul anyway.

PR #836, opened 2018-03-26
ISSUE #762 - Zoom on on double click on Tree item

carmenfan:
@JamesMilnerUK take a look and see if you're happy with it.

I only changed 2 things on the front end:
1 - what I passed to unityUtils (now an array of objects(with model info) instead of a model ID and a mesh ID
2 - currentSelectedMap to a local variable in tree.service.ts


frontend/components/tree/js/tree.component.ts
```diff
@@ -416,14 +416,39 @@ class TreeController implements ng.IController {
 		this.TreeService.updateModelState(node);
 	}
 
+	public selectAndCentreNode(node: any) {
+		if(node.toggleState !== "invisible") {
```
JamesMilnerUK: For consistency space after if, you could also do early return here keep the code flatter

frontend/components/tree/js/tree.component.ts
```diff
@@ -416,14 +416,39 @@ class TreeController implements ng.IController {
...
+				const keys = Object.keys(currentSelectedMap);
+				keys.forEach((key) => {
+					meshIDArrs.push({
+						model: key.replace("@", "."),
+						meshID: currentSelectedMap[key]
+					});
+				});
+
+				this.ViewerService.centreToPoint(meshIDArrs);
+		});
```
JamesMilnerUK: Indentation

JamesMilnerUK:
Some comments on minor style changes but apart from that LGTM. TSLint should pick up those things in future so we shouldn't have to worry about it.

Issue #764, opened 2018-03-01
Objects modelled in root node/space


Charence:
See: https://github.com/3drepo/3drepo.io/issues/760#issuecomment-368842970

PR #840, opened 2018-03-27
ISSUE #764 Objects modelled in root node/space 

carmenfan:
This is fixed.

The problem is that these objects are under the root node; with no metadata associated with them(hence no IFC GUIDs), thus they were ignored and chucked away during the shared id to IFC GUID conversion.

So my fix is to add a set recording the set of shared IDs presented when we create this group, remove all the mapped shared IDs, anything that is left, add them into the object array.

@Charence  please make sure it's ok!

### DIFF
--- a/backend/models/group.js
+++ b/backend/models/group.js
@@ -79,7 +79,7 @@ groupSchema.statics.uuidToIfcGuids = function(obj) {
 
 function uuidsToIfcGuids(account, model, ids) {
 	const query = { type: "meta", parents: {$in: ids}, "metadata.IFC GUID": {$exists: true} };
-	const project =  { "metadata.IFC GUID": 1 };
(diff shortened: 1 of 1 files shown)


backend/models/group.js
```diff
@@ -221,6 +222,7 @@ groupSchema.methods.updateAttrs = function(data){
 			if (obj.shared_id) {
 				const ns = obj.account + "__" + obj.model;
 				if ("[object String]" === Object.prototype.toString.call(obj.id)) {
+					sharedIDSets.add(obj.shared_id);
```
Charence: Should this line be in here? Shouldn't obj.shared_id always be added to sharedIDSets? This IF is just checking that obj.id is a String before attempting to convert it to a UUID.

carmenfan: yup, good point

PR #850, opened 2018-04-03
Issue 803 - Groups

JamesMilnerUK:
**Backend**

- Automated code fixes by the linter (single quotes to double quotes, no var etc)
- Handling of new properties for groups (description, createdAt etc)

**Frontend**

- UI for groups (HTML / CSS)
- Frontend groups state, held in group.service.ts
- Refactorings to TreeService and TreeComponent; don't watch for changes instead just call the ViewerService at the end of updates to the tree state. This is also a reasonable performance enhancement as we were calling updates in Unity on each loop iteration which was what caused the issue with the BCF model crashing for Andrew
- Move tree logic out of IssuesService and into TreeService


frontend/components/groups/js/groups.service.ts
```diff
@@ -0,0 +1,502 @@
...
+
+	}
+
+	/**
+	 * Do any cleaning that's necessary for groups in the UI
+	 */
+	public cleanGroups(groups: any[]) {
+		groups.forEach((group) => {
+			if (!group.name) {
+				group.name = "No assigned name";
```
carmenfan: put it in brackets so it's clear(er) that someone didn't just name it as "No assigned name".

I would probably just write `(No name)`, but up to you

frontend/components/panel/js/panel.component.ts
```diff
@@ -129,7 +129,7 @@ class PanelController implements ng.IController {
 			// If we have clicked on a canvas, we are probably moving the model around
 			if (event.target.tagName === "CANVAS") {
 				this.activate = false;
-				this.$scope.$apply();
+				//this.$scope.$apply();
```
carmenfan: remove unless there is a reason why it needs to stay (in which case we need a comment here)

frontend/components/panel/js/panel.component.ts
```diff
@@ -138,7 +138,7 @@ class PanelController implements ng.IController {
 		*/
 		angular.element(document).bind("mouseup", () => {
 			this.activate = true;
-			this.$scope.$apply();
+			//this.$scope.$apply();
```
carmenfan: remove unless there is a reason why it needs to stay (in which case we need a comment here)

frontend/components/panel/js/panel.component.ts
```diff
@@ -214,9 +214,7 @@ class PanelController implements ng.IController {
 	public calculateContentHeights() {
 		const tempContentItemsShown = angular.copy(this.contentItemsShown);
 		this.assignHeights(this.maxHeightAvailable, tempContentItemsShown, null);
-		this.$timeout(() => {
-			this.$scope.$apply();
-		});
+		//this.$timeout(() => {});
```
carmenfan: remove unless there is a reason why it needs to stay (in which case we need a comment here)

backend/models/group.js
```diff
@@ -132,19 +141,86 @@ groupSchema.statics.findIfcGroupByUID = function(dbCol, uid){
 				if (groupObjectsMap && groupObjectsMap.length > 0) {
 					group.objects = [];
 					for (let id in groupObjectsMap) {
-						group.objects.push(groupObjectsMap[id]);
+						if (groupObjectsMap.hasOwnProperty(id)) {
```
Charence: Shouldn't it be `if (id)`?

backend/models/group.js
```diff
@@ -165,6 +241,9 @@ groupSchema.statics.findByUID = function(dbCol, uid){
 			}
 			
 			for (let namespace in ifcObjectByAccount) {
+				if (!ifcObjectByAccount.hasOwnProperty(namespace)) {
```
Charence:
```
if (namespace) {
   ....
}
```

backend/models/group.js
```diff
@@ -239,6 +315,9 @@ groupSchema.methods.updateAttrs = function(data){
 		}
 
 		for (let namespace in sharedIdsByAccount) {
+			if (!sharedIdsByAccount.hasOwnProperty(namespace)) {
```
Charence: Same as above
```
if (namespace) {
   ...
}
```

backend/models/group.js
carmenfan: I think it's this way because of the linter James is using that you have to explicitly check the groupObjectsMap has the field.

I suggest if statements don't need to be there and we do not impose this linting rule. Makes no sense to me

backend/models/group.js
```diff
@@ -270,26 +349,26 @@ groupSchema.methods.updateAttrs = function(data){
...
 		}
 
-
-
 	}
 
 	return Promise.all(ifcGuidPromises).then(() => {
-
-		this.name = data.name || this.name;
+		this.description = data.description;
```
Charence: What happens if any of these attributes are `null` in `data`?

frontend/components/docs/js/docs.component.ts
```diff
@@ -59,6 +59,25 @@ class DocsController implements ng.IController {
 	}
 
 	public watchers() {
+
+		this.$scope.$watch(() => {
+			return this.ViewerService.selection;
```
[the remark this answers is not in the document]
JamesMilnerUK: Not sure I entirely follow your thought process here.

To my understanding the viewer receives a callback from the viewer itself (never touching the TreeService as far as I can tell), which in turn is handled by ViewerService. Is there something I'm missing about how we get the Model/Account/ID?

Here we used to call DocsService directly (coupling DocsService to the ViewerService and also causing a circular dependency). Now we just watch the ViewerService from the component waiting for information on new selections, and call the DocsService from the component (gluing together services rather than coupling them). In short all we've really done here is decoupled the ViewerService from the DocsService. 

Can see the nicety of potentially removing the TreeService dependency from DocsService and passing in the Meta ID directly and do the conversion in the Docs component if you're concerned about that being coupled.

Charence:
screenshot 2018-04-03 16 04 08

Visibility cannot seem to be set

frontend/components/tree/js/tree.service.ts
```diff
@@ -499,45 +484,63 @@ export class TreeService {
...
-		const model = node.model || node.project;
-		const key = this.getAccountModelKey(node.account, model);
-		let meshes = idToMeshes[node._id];
-		if (key && idToMeshes[key]) {
-			// the node is within a sub model
-			meshes = idToMeshes[key][node._id];
+		const highlightMap = {};
+
+		let stack;
+		if (Array.isArray(nodes) === false) {
```
Charence: Can we require `nodes` to always be an array?

backend/models/group.js
carmenfan: Why is TSLint enabled when it's a js script?

carmenfan: I think linting should stick to tidying code instead of depicting what bugs to catch though. We are adding an extra branch prediction here (and in other places) for not a lot of reason.

frontend/components/tree/js/tree.service.ts
```diff
@@ -499,45 +484,63 @@ export class TreeService {
...
+			if (highlightMap[key] === undefined) {
+				highlightMap[key] = {};
+			}
+
+			if (highlightMap[key].colour === undefined) {
+				highlightMap[key].colour = colour;
+			}
+
+			// Check top level and then check if sub model of fed
+			const meshes = idToMeshes[childNode._id] || idToMeshes[key][childNode._id];
```
Charence: Visibility problem caused here

frontend/components/tree/js/tree.service.ts
```diff
@@ -905,29 +931,97 @@ export class TreeService {
...
 	 * @param node	Node to toggle visibility. All children will also be toggled.
 	 */
-	public updateModelState(node) {
+	public updateModelVisibility(node) {
 
 		this.ready.promise.then(() => {
-			const childNodes = {};
-			this.traverseNodeAndPushId(node, childNodes, this.treeMap.idToMeshes);
+
+			const childNodes = this.getMeshMapFromNodes(node, this.treeMap.idToMeshes);
```
Charence: This is the only place where getMeshMapFromNodes is given a single `node` rather than an array of `nodes`. We can pass in `[node]` to save getMeshMapFromNodes having to deal with different inputs.

backend/models/group.js
carmenfan: I vote for turning it off.  Charence as well

frontend/components/tree/js/tree.service.ts
```diff
@@ -1041,6 +1043,11 @@ export class TreeService {
 	 * Unselect all selected items and clear the array
 	 */
 	public clearCurrentlySelected() {
+
+		if (this.DocsService.state.show) {
```
carmenfan: Create a function to disable doc component showing instead of manipulating the variable directly

frontend/components/tree/js/tree.service.ts
```diff
@@ -1131,10 +1138,31 @@ export class TreeService {
...
 
+	/**
+	 * Show metadata in the metadata panel if necessary
+	 * @param lastNode the node to show the metadata for
+	 */
+	public handleMetadata(lastNode) {
+		// Handle showing metadata in the metadata panel if necessary
+
+		if (lastNode && lastNode.meta) {
+			if (this.DocsService.state.active && !this.ViewerService.pin.pinDropMode) {
```
carmenfan: Dont do this check here, do it on handleObjectSelected within Doc's service.

Tree Service should tell DocsService there is metadata available on current state. It should be up to DocsService what to do with that information (whether to display, or ignore because it's pinMode or it's inactive)

If you have to keep pinMode here because of potential circular dependencies, fine (not ideal but oh well). But it shouldn't check Doc's service active state here. (imagine if you have another service needed to call Doc Service to handle document in the future, they would all have to know to check its state first.)

frontend/components/tree/js/tree.service.ts
```diff
@@ -1131,10 +1138,31 @@ export class TreeService {
...
+				this.DocsService.handleObjectSelected(
```
carmenfan: This is potentailly badly named now that what we are doing has changed.

Should be called displayDocument or something

frontend/components/tree/js/tree.service.ts
JamesMilnerUK: That's a good point. Doesn't cause a circular dependency so i'll put it in there.

frontend/components/tree/js/tree.service.ts
JamesMilnerUK: Good point

Charence:
- [ ] Cannot delete group

screenshot 2018-04-04 15 28 00

carmenfan:
@JamesMilnerUK  have you ever tested this feature as a viewer? Just occurred to me that viewer only have a read only permission on these things and the UI may not respond correctly (I haven't tested it, so I don't know)

frontend/components/groups/js/groups.component.ts
```diff
@@ -0,0 +1,363 @@
...
+			this.GroupsService.getCurrentMeshHighlights().then((length) => {
+				this.selectedObjectsLen = length;
+				this.changed = true;
+			});
+
+		}, true);
+
+	}
+
+	public toggleColorOveride($event, group: any) {
```
carmenfan: override

frontend/components/groups/js/groups.service.ts
```diff
@@ -0,0 +1,545 @@
...
+		} else {
+			this.colorOveride(group);
+		}
+	}
+
+	/**
+	 * Color override the meshes of a given group
+	 * @param group the group to color override
+	 */
+	public colorOveride(group: any) {
```
carmenfan: override (it's in other places too, i'll stop pointing it out!)

frontend/components/panel/js/panel.service.ts
```diff
@@ -233,9 +244,20 @@ export class PanelService {
 
 	// 	}
```
carmenfan: delete if not needed

frontend/components/tree/js/tree.service.ts
```diff
@@ -499,45 +468,62 @@ export class TreeService {
...
 	/**
-	 * Add all child id of a node recursively, the parent node's id will also be added.
+	 * Get map of meshes and associated colours from an array of nodes
 	 */
-	public traverseNodeAndPushId(node: any, nodes: any, idToMeshes: any) {
+	public getMeshMapFromNodes(nodes: any, idToMeshes: any, colour?: number[]) {
 
-		if (!node) {
-			console.error("traverseNodeAndPushId node is null: ", node);
+		if (Array.isArray(nodes) === false) {
```
carmenfan: if(!Array.isArray(nodes))

frontend/components/tree/js/tree.service.ts
```diff
@@ -936,56 +993,105 @@ export class TreeService {
...
+	}
 
-			return this.ready.promise.then(() => {
-				const currentSelectedMap = {};
-				this.currentSelectedNodes.forEach((n) => {
-					this.traverseNodeAndPushId(n, currentSelectedMap, this.treeMap.idToMeshes);
-				});
+	public getMeshHighlights(nodes) {
+		return this.ready.promise.then(() => {
+			const currentSelectedMap = this.getMeshMapFromNodes(nodes, this.treeMap.idToMeshes);
```
carmenfan: return this.getMeshMapFromNodes(nodes, this.treeMap.idToMeshes);

frontend/components/tree/js/tree.service.ts
```diff
@@ -936,56 +993,105 @@ export class TreeService {
...
+	 * @param nodes	Nodes to select.
 	 * @param multi	Is multi select enabled.
 	 */
-	public selectNode(node: any, multi: boolean, final: boolean) {
+	public selectNodes(
+		nodes: any[], multi: boolean, additive: boolean, colour: number[],
+		forceReHighlight: boolean,
+	) {
+
+		forceReHighlight = (forceReHighlight === undefined) ? true : false;
```
carmenfan: default behaviour should be false. This line doesn't need to exist?

frontend/components/tree/js/tree.service.ts
```diff
@@ -936,56 +993,105 @@ export class TreeService {
...
+		for (let i = 0; i < nodes.length; i++) {
+			const node = nodes[i];
 
-		if (node) {
-			if (multi) {
+			if (!node) {
+				continue;
+			}
+
+			if (additive) {
```
carmenfan: I'm pretty sure additive === forceHighlight. Please investigate!

carmenfan: Actually this whole if block is probably a lot cleaner if we just do something like this:

````js
const shouldSelectNode = !multi || forceHighlight || !node.selected;
this.setNodeSelection(node, shouldSelectNode);
````

Currently we have 3 if statement blocks here 2 of which does something identical, the 3rd do something identical with a slight variation on the parameter

File: frontend/components/tree/js/tree.service.ts [hunk 23]
```diff
@@ -936,56 +993,105 @@ export class TreeService {
...
 		}
+
+		return this.highlightNodes(nodes, true, undefined, false);
 	}
 
 	/**
-	 * Select a node in the tree.
-	 * @param node	Node to select.
+	 * Select nodes in the tree.
+	 * @param nodes	Nodes to select.
```
carmenfan: this needs updating

frontend/components/tree/js/tree.service.ts
```diff
@@ -994,70 +1100,256 @@ export class TreeService {
...
-		return Promise.reject("No node specified");
 	}
 
 	/**
-	 * Select multiple nodes in the tree.
-	 * @param nodes	Array of nodes to select.
+	 * Call the highlighting in the viewer
+	 * @param nodes	Nodes to select.
 	 * @param multi	Is multi select enabled.
+	 * @param colour the colour to highlight
```
carmenfan: needs updating

frontend/components/tree/js/tree.service.ts
```diff
@@ -994,70 +1100,256 @@ export class TreeService {
...
 	 */
-	public updateModelStateHidden(node) {
-		if (node.toggleState === "invisible") {
-			this.clickedHidden[node._id] = node;
-		} else {
-			delete this.clickedHidden[node._id];
-		}
+	public getMeshHighlightsBySharedId(objects) {
+		return this.getNodesFromSharedIds(objects).then((nodes) => {
+			const currentSelectedMap = this.getMeshMapFromNodes(nodes, this.treeMap.idToMeshes);
```
carmenfan: return this.getMeshMapFromNodes(nodes, this.treeMap.idToMeshes);

frontend/components/tree/js/tree.service.ts
```diff
@@ -994,70 +1100,256 @@ export class TreeService {
...
 	}
 
 	/**
-	 * Toggle node from clickedShown collection.
-	 * @param node	Node to toggle.
+	 * Select a series of nodes by an array of shared IDs (rather than unique IDs)
+	 * @param objects	Nodes to select
+	 * @param multi	Is multi select enabled
+	 * @param additive whether the selection should reset or keep add selections
+	 * @param colour the colour to highlight
```
carmenfan: needs update

frontend/components/tree/js/tree.service.ts
```diff
@@ -994,70 +1100,256 @@ export class TreeService {
...
+
+	}
+
+	/**
+	 * Highlight a series of nodes by an array of shared IDs (rather than unique IDs)
+	 * @param objects objects to show
+	 */
+	public highlightsBySharedId(objects: any) {
+
+		// TODO: This could probably be refactored
```
carmenfan: this really should be rewritten to utilise the sharedID to node function and call select nodes with an array instead of calling select node separately!

frontend/components/groups/js/groups.service.ts
```diff
@@ -301,42 +302,47 @@ export class GroupsService {
...
+		}
 
-				// If we haven't saved don't update saved meshes
-				if (!this.state.selectedGroup.new) {
-					const total = this.getTotalMeshes(meshes);
-					this.state.selectedGroup.totalSavedMeshes = total;
-				}
+		// If multi is enabled we want to inverse selections
+		// rather than add them
+		let additive = (multi === true) ? false : true;
```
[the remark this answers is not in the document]
carmenfan: Don't think it needs to exist, but if it does, !multi

carmenfan:
- [ ] As a viewer, i can still modify the group title/description/colour (but not save)
group

- [ ] Related issue to above: if I return to the list without saving, it should reset the group data to last saved

frontend/components/tree/js/tree.service.ts
JamesMilnerUK: Yeah definitely. Also probably only need to call expand to selection on the last node

frontend/components/tree/js/tree.service.ts
JamesMilnerUK: Additive removed now as it was the same as forceHighlight. Not sure I am convinced about the one liner, you'd change the number of lines from 7 lines to 5 but I would say it makes it harder to follow as it's less explicit.

frontend/components/tree/js/tree.service.ts
JamesMilnerUK: This is sorted now

frontend/components/tree/js/tree.service.ts
carmenfan: From what I understand this should change from  7 lines to 2 lines? maybe you need to commit the change first and we can re-visit as i'm not really seeing it.

In general I'm pretty against the idea of having a multi conditional if statement that does the same thing but tweaking a single variable, because it gives the illusion that they're doing different things when they're not. At minimum the `setNodeSelection()` call should sit outside and the if statement should only define the variation of the 2nd parameter for the function call. but i'd argue that is no less clearer than a one liner.

frontend/components/groups/js/groups.service.ts
```diff
@@ -303,45 +304,78 @@ export class GroupsService {
...
-			}
+			});
 
-		});
+		} else if (!this.state.multiSelectedGroups.includes(group)) {
+
+			// selecting group that's not selected
+			this.state.multiSelectedGroups.push(group);
+
+			return this.TreeService.selectNodesBySharedIds(
```
Charence: This looks exactly the same as Line 324-337 above. Can we delete some code?

JamesMilnerUK: I can always delete code @Charence

JamesMilnerUK:
Viewer editing should be disabled now and also editing without saving should reset

frontend/components/tree/js/tree.service.ts
JamesMilnerUK: this should be resolved

frontend/components/groups/js/groups.component.ts
```diff
@@ -118,15 +120,14 @@ class GroupsController implements ng.IController {
...
 
-					const hasPerm = this.AuthService.hasPermission(
-						this.ClientConfigService.permissions.PERM_CREATE_ISSUE,
-						this.modelSettings.permissions,
-					);
-
-					if (hasPerm) {
-						this.canAddGroup = true;
-					}
+				if (hasPerm) {
```
carmenfan:
````js
this.canAddGroup = hasPerm;
````

or i would just do 
````js
this.canAddGroup = this.AuthService.hasPermission(
					this.ClientConfigService.permissions.PERM_CREATE_ISSUE,
					this.modelSettings.permissions,
				);
````

and remove hasPerm variable completely

File: frontend/components/groups/js/groups.component.ts [hunk 31]
```diff
@@ -143,6 +144,26 @@ class GroupsController implements ng.IController {
 
 	}
 
+	public compareSavedGroup() {
```
carmenfan: Not entirely sure what's the purpose of this...

Isn't this essentially overwriting selectedGroup's properties with savedGroup's? is there much point in checking if they're the same?

frontend/components/tree/js/tree.service.ts
```diff
@@ -1089,16 +1089,19 @@ export class TreeService {
...
-				this.setNodeSelection(node, !node.selected);
-			} else {
-				// If it is not multiselect mode, remove all highlights
-				this.clearCurrentlySelected();
-				this.setNodeSelection(node, true);
-			}
+			const shouldSelect = !multi || forceReHighlight || !node.selected;
+			this.setNodeSelection(node, shouldSelect);
+
+			// if (forceReHighlight) {
```
carmenfan: remove if no longer needed

frontend/components/groups/js/groups.component.ts
JamesMilnerUK: So basically we need a store of the group we're editing so if we go back we can reverse those changes. We could just straight up override each property, if you think that's better?

carmenfan: Yea I feel like the comparison is a bit useless. Because they are either identical, or you want to take whatever that's in savedgroup... in other words, you just want savedGroup's data.

Would probably want to rename the function though if that's the case.

Issue #848, opened 2018-04-03
Shared id is being passed over as LUUID format in groups


carmenfan:
Groups are broken if the model isn't IFC, the shared IDs are being passed back as LUUID

PR #851, opened 2018-04-03
ISSUE #848 sharedID is being saved as string

carmenfan:
Shared IDs are being saved as strings if it is not a IFC model


carmenfan:
no idea lol

PR #858, opened 2018-04-04
Issue 838 - Better handling of incorrect profile passwords

JamesMilnerUK:
Fix the issue with it saying incorrect user name and password when old password is wrong
Nicer error messages for trying to save password which is too short


JamesMilnerUK:
@carmenfan should be adjusted now

PR #862, opened 2018-04-04
Issue 823 - mailing list opt-out option on sign-up

Charence:
- UI for mailing list opt-out
- Backend changes to save flag to DB


backend/models/user.js
```diff
@@ -265,15 +266,15 @@ schema.statics.createUser = function(logger, username, password, customData, tok
 		}
 	
 
-		["firstName", "lastName", "email"].forEach(key => {
+		["firstName", "lastName", "email", "mailListOptOut"].forEach(key => {
 			if (customData && customData[key]){
```
carmenfan: As discussed, let's do an early reject on customData and we can remove all the if(customData) checks after that.

backend/models/user.js
```diff
@@ -240,126 +237,125 @@ schema.statics.updatePassword = function(logger, username, oldPassword, token, n
...
+			if (!this.usernameRegExp.test(username)) {
+				return Promise.reject({ resCode: responseCodes.INVALID_USERNAME});
 			}
-		}
-	
 
-		["firstName", "lastName", "email", "mailListOptOut"].forEach(key => {
-			if (customData && customData[key]){
-				cleanedCustomData[key] = customData[key];
+			for (let i = 0; i < C.REPO_BLACKLIST_USERNAME.length; i++) {
```
carmenfan: should probably convert new username to lower case (the list is in constant is all in lower case), and just do indexOf or something similar instead of looping

backend/models/user.js
```diff
@@ -1113,9 +1088,9 @@ schema.statics.teamspaceMemberCheck = function(teamspace, user) {
 			return Promise.reject(responseCodes.USER_NOT_ASSIGNED_WITH_LICENSE);
 		}
 	});
-}
+};
 
-var User = ModelFactory.createClass(
+let User = ModelFactory.createClass(
```
carmenfan: const

PR #869, opened 2018-04-06
Issue 866 - Fix issue status change display and status notifications

Charence:
1. Restore code for hiding closed issues.
2. Fix convertActionValueToText

### DIFF
--- a/frontend/components/issues/js/issues.service.ts
+++ b/frontend/components/issues/js/issues.service.ts
@@ -1272,6 +1272,8 @@ export class IssuesService {
 
 		let actionText = "";
 
+		value = value.toLowerCase();
+
 		if (actions.hasOwnProperty(value)) {
 			actionText = actions[value];
 		}
--- a/frontend/components/issues/js/issues.service.ts
+++ b/frontend/components/issues/js/issues.service.ts
@@ -1259,14 +1259,24 @@ export class IssuesService {
 	 * @param value
 	 */
 	public convertActionValueToText(value: string) {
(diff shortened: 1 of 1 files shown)


frontend/components/issues/js/issues.service.ts
```diff
@@ -1259,14 +1259,24 @@ export class IssuesService {
...
+			"high": "High",
+			"open": "Open",
+			"in progress": "In progress",
+			"for approval": "For approval",
+			"closed": "Closed"
+		};
+
+		let actionText = "";
+
+		if (actions.hasOwnProperty(value)) {
```
carmenfan: Actually thinking about it, should we do value.toLower() here? There is possibility that BCF would give us something like noNe. Unlikely but it's probably safer to do it this way

frontend/components/issues/js/issues.service.ts
```diff
@@ -1251,14 +1259,24 @@ export class IssuesService {
...
+			"low": "Low",
+			"medium": "Medium",
+			"high": "High",
+			"open": "Open",
+			"in progress": "In progress",
+			"for approval": "For approval",
+			"closed": "Closed"
+		};
+
+		let actionText = "";
```
Charence: @carmenfan Maybe it should default to the input `actionText = value` rather than setting it to empty?

carmenfan: agreed, sounds like we're not displaying anything we don't know about 😨

PR #870, opened 2018-04-06
Issue 860 - Fix invoice template to show country and don't cache on DB

Charence:
1. Show country on invoice;
2. Remove invoice PDF from DB.

### DIFF
--- a/backend/models/invoice.js
+++ b/backend/models/invoice.js
@@ -79,7 +79,6 @@
 		taxAmount: { type: SchemaTypes.Double, get: signAndRoundTo2DP, set: roundTo2DP },
 		info: billingAddressInfo,
 		state: {type: String, default: C.INV_INIT, enum: [C.INV_INIT, C.INV_PENDING, C.INV_COMPLETE]},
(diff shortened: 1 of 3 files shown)


backend/routes/invoice.js
```diff
@@ -85,7 +85,7 @@
 				return Promise.reject(responseCodes.BILLING_NOT_FOUND);
 			}
 
-			return invoice.getPDF({regenerate: regenerate});
```
carmenfan: get rid of the regenerate variable, it's no longer used i believe?

backend/models/invoice.js
```diff
@@ -478,16 +475,8 @@
...
-		options = options || {};
-
-		if (options.regenerate || !this.pdf) {
-
-			return this.generatePDF();
-
-		} else {
-			return Promise.resolve(this.pdf.buffer);
-		}
+	schema.methods.getPDF = function () {
```
carmenfan: make routes/invoice call generatePDF instead? or just rename generatePDF to getPDF

Issue #873, opened 2018-04-09
expandToSelection does not behave as expected


JamesMilnerUK:
This is currently in staging, dev and production. For reproduction see GIF

![expandtoselection](https://user-images.githubusercontent.com/8822075/38492614-6e246316-3be7-11e8-847e-58be8d944a49.gif)

PR #875, opened 2018-04-09
ISSUE #873 - fix expandToSelection

JamesMilnerUK:
This will allow people to click on issues and have the tree panel respond as would be expected; expanding all nodes and scrolling to/selecting the correct nodes


JamesMilnerUK:
I also accidentally committed a small fix for groups (console error about a variable being undefined). Sorry!

Issue #857, opened 2018-04-04
Lack of handling of network requests for quotas, billing, etc


JamesMilnerUK:
At the moment we don't handle the promise rejection or provide any user feedback

PR #877, opened 2018-04-10
ISSUE #857 - Better handling of failing API calls on Billing page

JamesMilnerUK:
Catch request failures (log them) and also show an error message for users

### DIFF
--- a/frontend/components/account/js/account-billing.component.ts
+++ b/frontend/components/account/js/account-billing.component.ts
@@ -62,10 +62,7 @@ class AccountBillingController implements ng.IController {
 	) {}
 
 	public $onInit() {
-		this.initSubscriptions();
-		this.initBillings();
-		this.initPlans();
-		this.initQuota();
+		this.initBillingData();
 		this.showInfo = true;
 		this.saveDisabled = true;
 		this.countries = this.ClientConfigService.countries;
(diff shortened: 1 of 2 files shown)


frontend/components/account/js/account-billing.component.ts
```diff
@@ -78,29 +79,47 @@ class AccountBillingController implements ng.IController {
...
+			.catch((error) => {
+				console.error(error);
 			});
 	}
 
 	public initPlans() {
-		return this.APIService.get("plans")
+		return this.APIService.get("/plans")
 			.then((response) => {
 				if (response.status === 200) {
```
carmenfan: Is this check necessary (I know you probably didn't write this)? according to angular docs:
> A response status code between 200 and 299 is considered a success status and will result in the success callback being called. Any response status code outside of that range is considered an error status and will result in the error callback being called. Also, status codes less than -1 are normalized to zero. -1 usually means the request was aborted, e.g. using a config.timeout. More information about the status might be available in the xhrStatus property.

So we can assume that if the promise resolves it's a success? The other API calls on this page don't check response status numbers,

frontend/components/account/js/account-billing.component.ts
```diff
@@ -78,29 +79,47 @@ class AccountBillingController implements ng.IController {
...
 	}
 
 	public initSubscriptions() {
 		return this.APIService.get(this.account + "/subscriptions")
 			.then((response) => {
+				this.billingError = false;
 				this.subscriptions = response.data;
+			})
+			.catch((error) => {
+				this.billingError = true;
```
carmenfan: all of these checks seems to overwrite  `billingError`.

let's say API request to `plan` is fetched before `subscriptions`... and `plan` failed but `subscriptions` succeed. `billingError` would then be set to false when it should be `true`.

I think this boolean needs to be set to `false` on init, and all these checks needs to be like written as something like:

````js
.then((response) => {
    this.billingError |= false; //so if the billingError was set to true already, it will stay true
    ...
})
.catch((error) => {
   this.billingError = true; //Here we don't need care about previous value, if it was false we want it to be true.
   ...
})
````

Alternatively, billingError should just be a function that returns `true` if any of the follows are undefined:
`this.subscriptions`
`this.quota`
`this.plans`

Then you don't have to worry about setting this flag... It's probably cleaner this way.

JamesMilnerUK: Alternatively, we store all the requests and could do a Promise.all on those, and set ```this.billingError = true;``` in the `then` block? That might be a nice solution.

carmenfan: ya could do. Sounds sensible

carmenfan: Actually I think you mean the catch block, since you only want to flag it if there's error

Issue #830, opened 2018-03-20
Multiselect on filtered tree


paloknapo:
Filtered tree currently supports only single selection. 

![image](https://user-images.githubusercontent.com/760452/37666105-a2d8d3ee-2c56-11e8-8601-121d8a7ad902.png)

PR #879, opened 2018-04-10
Issue #830 - Multi select and zooming on double click for filtered tree

JamesMilnerUK:


frontend/components/tree/js/tree.component.ts
```diff
@@ -124,10 +115,7 @@ class TreeController implements ng.IController {
 
 			} else if (event.type === this.EventService.EVENT.VIEWER.BACKGROUND_SELECTED) {
 				this.TreeService.clearCurrentlySelected();
-				if (this.currentFilterItemSelected !== null) {
-					this.currentFilterItemSelected.class = "";
-					this.currentFilterItemSelected = null;
-				}
+				this.nodes.forEach((n) => n.class = "");
```
carmenfan: What is this used for? (I know you probably didn't write it)

From a bit of crawling through the code it seems to be either set as "treeNodeSelected" or "unselectedFilterItem" ? 
Is there a better way to represent this? I feel like this should be a boolean or something

frontend/components/tree/js/tree.component.ts
```diff
@@ -138,10 +126,7 @@ class TreeController implements ng.IController {
 				this.setupInfiniteItemsFilter();
 				this.TreeService.expandFirstNode();
 				this.setContentHeight(this.fetchNodesToShow());
-				this.$timeout(() => {}).then(() => {
-					// this.TreeService.showAllTreeNodes();
-					// this.TreeService.setVisibilityOfNodes(this.TreeService.getHiddenByDefaultNodes(), "invisible");
-				});
+				this.$timeout(() => {}).then(() => {});
```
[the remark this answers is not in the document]
JamesMilnerUK: Yep, good spot

JamesMilnerUK:
Fixed the zoom bug in this branch

frontend/components/tree/js/tree.component.ts
```diff
@@ -366,27 +357,28 @@ class TreeController implements ng.IController {
 		);
 	}
 
-	public filterItemSelected(item) {
+	public filterItemSelected($event, item) {
+
+		// Prevent double click toggling
+		const doubleClick = $event.detail > 1;
```
JamesMilnerUK: DRY

carmenfan:
2 bugs found:
- [ ] Selection stopped working on the filtered tree all together
- [ ] Tree filter doesn't like it when you type too fast
search

Notice here, I typed "yrdy" but it only requested for search results of "y" and displayed them.

I've also seen it search the right thing but still displayed "y"'s result. But I'm struggling to reproduce it now. Might worth just looking through the code and make sure the search results are cleared when a new search triggered.

carmenfan:
Search doesn't work at all now. I can see the search results have returned from the server and the UI is stuck on Filtering tree for objects:
image

JamesMilnerUK:
My bad! Should be fixed now.

carmenfan:
Found more bugs:
- We need to either disable the Search button on filter mode or we need to revert to tree here:
treebig1
- Sometimes the tab expands randomly when you select other objects (i'm guessing it's because the non filter tree tab is still doing stuff and expanded the tab)
treebug2
- Seems like select all is broken on federations:
treebug3

frontend/components/panel/js/panel-card-option-filter.component.ts
```diff
@@ -19,24 +19,28 @@ class PanelCardOptionFilterController implements ng.IController {
 
 	public static $inject: string[] = [];
 
-	public showFilter;
+	public showFilter: boolean;
+	public showFilterButton: boolean;
```
carmenfan: showFilterButton is a duplicate of showFilter.
looks like showFilterButton always equals to !showFilter. (which makes sense, you don't want the filter button when filter is shown)

Issue #806, opened 2018-03-13
Sign up error: Email is invalid whenever there is an empty compulsory field and Make company non compulsary


carmenfan:

carmenfan:
@Charence  can you do this one while you're changing the sign ups?

JamesMilnerUK:
I'll doooo it

carmenfan:
looks like you've done changes for  #807 in here. going to close 807

PR #881, opened 2018-04-10
Issue #806 - Fix error always being 'Email required' at signup

JamesMilnerUK:
Also makes password label consistent with other required inputs

### DIFF
--- a/frontend/components/sign-up/js/sign-up.component.ts
+++ b/frontend/components/sign-up/js/sign-up.component.ts
@@ -182,13 +182,6 @@ class SignupController implements ng.IController {
 			}
 		});
 
-		this.$scope.$watch(() =>  this.$scope.signup.$error, (error) => {
-			if (error.email) {
-				this.emailInvalid = true;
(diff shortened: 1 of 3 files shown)


frontend/components/sign-up/js/sign-up.component.ts
```diff
@@ -185,6 +185,8 @@ class SignupController implements ng.IController {
 		this.$scope.$watch(() =>  this.$scope.signup.$error, (error) => {
 			if (error.email) {
 				this.emailInvalid = true;
+			} else if (!error.email) {
```
carmenfan: Instead of storing a global variable, can we just access  this on doRegister? (the only place that we check this)

I feel like a continuous watch variable just to flag a boolean that is only used in one place sounds a bit overkill. But then again I'm not really sure how these scope fields work.

JamesMilnerUK: Yep, that's a good call. We can just access it when we do the validation, no need for the watcher.

PR #882, opened 2018-04-10
Issue 824 - new T&Cs are privacy policy update pop-up

Charence:
1. Additional field in config file as shown in config_sample
2. Backend changes to detect if user has logged in since policy update
3. Dialog box prompt if they haven't seen new policy

### DIFF
--- a/frontend/globals/unity-util.ts
+++ b/frontend/globals/unity-util.ts
@@ -116,6 +116,15 @@ export class UnityUtil {
 	}
 
 	/**
+	 * Check if an error is Unity related
+	 */
+	public static isUnityError(err) {
+		const checks = ["Array buffer allocation failed", "Invalid typed array length", "Unity", "unity"];
+		const hasUnityError = !checks.every((check) => err.indexOf(check) === -1);
+		return hasUnityError;
+	}
+
+	/**
 	 * Handle a error from Unity
 	 */
 	public static onUnityError(errorObject) {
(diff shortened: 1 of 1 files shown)


frontend/components/home/js/home.component.ts
```diff
@@ -254,7 +254,7 @@ class HomeController implements ng.IController {
 						});
 					}
 
-				} else if (currentData.initialiser && !currentData.username) {
+				} else if (!currentData.username) {
```
JamesMilnerUK: Are you completely sure you can remove currentData.initialiser? I remember this being very fragile!

Charence: Line 243 is `if (!currentData.initialiser) {`, so I'd hope `else` means the opposite of that.

backend/routes/auth.js
```diff
@@ -86,8 +86,19 @@ function login(req, res, next){
...
+
+		let responseData = { username: user.user };
+		
+		req[C.REQ_REPO].logger.logInfo("User is logged in", responseData);
+		
+		responseData.roles = user.roles;
+		responseData.flags = {};
+		
+		if (!user.hasReadLatestTerms()) {
+			responseData.flags.termsPrompt = true;
```
carmenfan:
````js
responseData.flags.termsPrompt = !user.hasReadLatestTerms();
````

File: frontend/components/home/js/home.component.ts (hunk 1 above)
Comment (JamesMilnerUK) [reply frontend/components/home/js/home.component.ts]: Oh dear

File: backend/routes/auth.js [hunk 3]
```diff
@@ -88,17 +88,21 @@ function login(req, res, next){
...
-		
+
 		req[C.REQ_REPO].logger.logInfo("User is logged in", responseData);
-		
+
 		responseData.roles = user.roles;
 		responseData.flags = {};
-		
+
 		if (!user.hasReadLatestTerms()) {
```
carmenfan: you don't need this if statement XD

Or is it because you don't want to include this field if you dont' want the prompt? In that case then you should proabbly revert back to the old thing lol

Charence: I don't want to include the field, but I guess it doesn't hurt if it is included.

Issue #868, opened 2018-04-05
Error: Cannot read property 'shouldShowIssue' of undefined


JamesMilnerUK:
![spectacle m13622](https://user-images.githubusercontent.com/8822075/38380466-b8ddb702-38fb-11e8-8dd3-5bf31dfc2605.png)

Charence:
@JamesMilnerUK What did you do to get this?

PR #886, opened 2018-04-11
ISSUE #868 - Bind issue functions to the issues component scope

JamesMilnerUK:
Use bind to preserve scope. This is probably the cause of any current issue / chat server errors.

### DIFF
--- a/frontend/components/issues/js/issues.component.ts
+++ b/frontend/components/issues/js/issues.component.ts
@@ -15,7 +15,6 @@
  *	along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-
 class IssuesController implements ng.IController {
 
 	public static $inject: string[] = [
@@ -132,7 +131,6 @@ class IssuesController implements ng.IController {
 
 		this.watchers();
 
-
 	}
 
 	public getIssuesData() {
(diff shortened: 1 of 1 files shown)


frontend/components/issues/js/issues.component.ts
```diff
@@ -233,22 +232,21 @@ class IssuesController implements ng.IController {
 
 				this.issuesReady.then(() => {
 					const hasPerm = this.AuthService.hasPermission(
-						this.ClientConfigService.permissions.PERM_CREATE_ISSUE, 
-						this.modelSettings.permissions
+						this.ClientConfigService.permissions.PERM_CREATE_ISSUE,
```
carmenfan:
````js
this.canAddIssue = this.AuthService.hasPermission(....)
````

Does this have latest of staging? I feel like we've changed this already on a separate branch

JamesMilnerUK: That was for groups :P The reason this code is even highlighted is auto formatting by TSLint

carmenfan: no I mean i thought we've got rid of the hasPerm variable and just assigned this.canAddIssue directly somewhere already. I remember pointing this out elsewhere

JamesMilnerUK: Yes it was very similar looking code but different file;  ```this.canAddGroup```  vs ```this.canAddIssue```. Probably worth changing it here to, I can look into it.

Issue #852, opened 2018-04-04
Display "please check you're using chrome/firefox 64bit" for this error


carmenfan:
![image](https://user-images.githubusercontent.com/11945337/38298460-4c5dfbc6-37ef-11e8-8156-81926b448c5b.png)

Display the helper message that tells the user to check they have sufficient memory for this type of error

PR #888, opened 2018-04-11
ISSUE #852 - Add Invalid typed array length as unity error

JamesMilnerUK:
Also makes it cleaner / more extensible in future by moving check to a function

### DIFF
--- a/frontend/globals/unity-util.ts
+++ b/frontend/globals/unity-util.ts
@@ -116,6 +116,15 @@ export class UnityUtil {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


Charence:
I think the intent of #852 was for a more helpful error message that tells them they either need more memory or a 64-bit browser?

JamesMilnerUK:
@Charence  we have two separate error messages. A generic on and a Unity one, this PR forces it down the Unity path which explains they're probably out of memory / need a 64 bit browser

Issue #896, opened 2018-04-13
Expand/Issues oddity causing expand attribute state to be incorrect


Charence:
e.g. http://127.0.0.1/andrew/0a789552-da5d-4dac-ad6f-231faee7ebf8

![screenshot 2018-04-13 12 39 46](https://user-images.githubusercontent.com/1438332/38732863-caa9a2f8-3f17-11e8-8a92-9fdcee028063.png)

Steps to reproduce:

1. Expand nodes in tree
2. Click on an issue that will expand an already expanded node
3. "-" icon on expand node become "+" and its children go into a strange orphaned state

PR #897, opened 2018-04-13
ISSUE #896 - Fix expand attribute state being incorrect on tree

JamesMilnerUK:
I wasn't able to replicate the issue after this change.


JamesMilnerUK:
I think 1. is due to selection not traversing, so it gets the right node but it's not highlighted

Regarding 2. I can't reproduce that behaviour on that model having played with it for 10 mins. Is there anything specific you did to get that state?

carmenfan:
1. Why does it do expansion but forget about selection? I thought it's suppose to be expand to selection? If it doesn't select I suggest not to expand at all. there's not a lot of point doing it.

2. Go to any sample federation (like T_E_S_T/9197b38e-e1a5-4972-ae25-8be0756151b8):
Click on the tree, 
image


then click on anything on the house,
image

 check Lego_House_Landscape's toggle button:
image


Looks like it's collapsing it without changing the toggle

JamesMilnerUK:
Interesting, I can't reproduce that for some reason:
expand

carmenfan:
Well the Landscape branch should be collapsing when you clicked on something in the architecture. That stopped working it seems.

JamesMilnerUK:
This should be done now. There's seems no exact way to ensure that the scroll will move to the indexed node as there is no way to watch for it finishing updating the DOM. It seems to work 9/10 so I think it's about as good as it's going to get.

carmenfan:
Seems ok if we can get rid of this 
expansion

JamesMilnerUK:
I think I have a fix
fixtreexpand
:

JamesMilnerUK:
I found the source of the bug; the topIndex was being set to larger than the allowable maximum (i.e. a node element that couldn't scrolled into view in the virtual list). I wrote some logic to set it to the maximum should that be the case.

PR #911, opened 2018-04-20
Issue 903 - Fix bugs from development (2.9.4)

JamesMilnerUK:


carmenfan:
- Select all -> save issue , only one of the sub model is saved (might want to wait for Charence's change which might auto fix this)

frontend/components/tree/js/tree.service.ts
```diff
@@ -1226,7 +1231,7 @@ export class TreeService {
 			return Promise.resolve([]);
 		}
 
-		return this.getMap().then(() => {
+		return this.ready.promise.then(() => {
```
Charence: issues.service uses onReady(). Should the same be used here?

frontend/components/model/js/model.component.ts
```diff
@@ -162,44 +162,55 @@ class ModelController implements ng.IController {
...
-					this.PanelService.hideSubModels(this.issuesCardIndex, false);
-				} else {
-					this.PanelService.hideSubModels(this.issuesCardIndex, true);
-				}
+				this.settings = response.data;
+				this.PanelService.hideSubModels(this.issuesCardIndex, !this.settings.federate);
 
-				this.ViewerService.updateViewerSettings(data.updateViewerSettings);
-				this.ClipService.initClip(data.properties.unit);
+				this.ViewerService.updateViewerSettings(this.settings);
```
Charence: Shouldn't it be `this.settings.updateViewerSettings`?

carmenfan: no, that's exactly why we weren't passing down the units to unity before (thus maps were broken)

PR #912, opened 2018-04-23
ISSUE 909 - fail save if there is no viewpoint

carmenfan:
This shouldn't happen unless something crazy is happening. But it seems to happen when we are at the brink of running out of Unity memory.

So putting in a check to catch any problem in serialising the viewpoint and let it fail gracefully.


frontend/globals/unity-util.ts
```diff
@@ -313,7 +313,12 @@ export class UnityUtil {
 
 	public static viewpointReturned(vpInfo) {
 		if (UnityUtil.vpPromise != null) {
-			const viewpoint = JSON.parse(vpInfo);
+			let viewpoint = {} ;
+			try {
+				viewpoint = JSON.parse(vpInfo);
+			} catch {
+				console.error("Failed to parse viewpoint", vpInfo);
```
JamesMilnerUK: Not sure of the surrounding context but would it make sense to reject the promise so it can be handled upstream?

carmenfan: That depends if the user would rather to save a issue without a viewpoint or get a warning instead. I kind of assume the former.

Charence: If we can't get a viewpoint, can it be made to look at the centre of the model?

carmenfan: It should already do that, I had to implement this for Issues made from Node Red

PR #913, opened 2018-04-23
Issue 813 - Handle selection up and down the tree

JamesMilnerUK:
Logic is to traverse down the tree set the right selection, and then traverse up the parents and make sure the selection is set appopriately (unselected, selected or parentOfUnselected).

Let me know if you can figure out how to make it faster!


frontend/components/tree/js/tree.service.ts
```diff
@@ -1015,43 +1022,121 @@ export class TreeService {
...
+
+			if (select && notParentOfUnselected && nodeIndex === -1) {
+				this.currentSelectedNodes.push(currentNode);
+			} else if (nodeIndex > -1) {
+				this.currentSelectedNodes.splice(nodeIndex, 1);
+			}
+
+			if (currentNode.toggleState !== "invisible") {
+
+				if (currentNode.toggleState !== "parentOfInvisible" && select) {
```
carmenfan: again, rejig it and you can avoid checking select multiple times

```` js
if (!select) {
	currentNode.selected = this.SELECTION_STATES.unselected;
} else if(currentNode.toggleState !== "parentOfInvisible") {
    currentNode.selected = this.SELECTION_STATES.parentOfUnselected;
} else {
    currentNode.selected = this.SELECTION_STATES.selected;
}
````

File: frontend/components/tree/js/tree.service.ts [hunk 2]
```diff
@@ -1015,43 +1022,121 @@ export class TreeService {
...
+
+				if (currentNode.children && currentNode.children.length) {
+					nodes = nodes.concat(currentNode.children);
+				}
+			}
+
+		}
+
+		setParentNodes.forEach((node) => {
+			this.setSelectionOnParentNodes(node);
```
carmenfan: Again, this is going through many nodes multiple times.
See comment on 1029.

This is worse since it's not a force set so it has to do a lot more logic as well!

frontend/components/tree/js/tree.service.ts
```diff
@@ -1015,43 +1022,121 @@ export class TreeService {
...
+				parentNode.selected = forceState;
+				continue;
+			}
+
+			let allUnselected = true;
+			let allSelected = true;
+
+			parentNode.children.forEach((n) => {
+				allUnselected = allUnselected && (n.selected === this.SELECTION_STATES.unselected || n.selected === undefined);
+				allSelected = allSelected && n.selected === this.SELECTION_STATES.selected;
```
carmenfan: If you rewrite this into a standard forloop, you can potentially reduce the # iterations by breaking once `!allUnselected && !allSelected`

I.e. you can stop as soon as you realised it's parentOfUnselected

Whether it's worth it or not i'm not sure.

frontend/components/tree/js/tree.service.ts
```diff
@@ -1107,18 +1193,38 @@ export class TreeService {
...
 
-			const shouldSelect = !multi || forceReHighlight || !node.selected;
+			if (node.selected === this.SELECTION_STATES.parentOfUnselected) {
+				parentsOfUnselected.push(node);
+			}
+
+			const selected = (node.selected !== this.SELECTION_STATES.parentOfUnselected &&
+								node.selected !== this.SELECTION_STATES.selected);
+
+			const shouldSelect = !multi || forceReHighlight || !!selected; // && node.selected !== "parentOfUnselected");
```
carmenfan: ` !!selected`?

frontend/components/tree/js/tree.component.ts
```diff
@@ -126,7 +126,7 @@ class TreeController implements ng.IController {
 			} else if (event.type === this.EventService.EVENT.VIEWER.BACKGROUND_SELECTED) {
 				this.TreeService.clearCurrentlySelected();
 				this.GroupsService.clearSelectionHighlights();
-				this.nodes.forEach((n) => n.selected = false);
+				this.nodes.forEach((n) => n.selected = "notSelected");
```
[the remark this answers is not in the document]
JamesMilnerUK: :+1:

frontend/components/tree/js/tree.service.ts
```diff
@@ -1015,43 +1022,121 @@ export class TreeService {
 		this.ViewerService.clearHighlights();
 		this.DocsService.closeDocs();
 
-		if (this.currentSelectedNodes) {
-			for (let i = 0; i < this.currentSelectedNodes.length; i++) {
-				this.currentSelectedNodes[i].selected = false;
+		if (this.currentSelectedNodes.length) {
```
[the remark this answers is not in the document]
JamesMilnerUK: :+1:

frontend/components/tree/js/tree.service.ts
```diff
@@ -1015,43 +1022,121 @@ export class TreeService {
...
+			while (this.currentSelectedNodes.length) {
+				const currentNode = this.currentSelectedNodes.pop();
+				currentNode.selected = this.SELECTION_STATES.unselected;
+				this.setSelectionOnParentNodes(currentNode, this.SELECTION_STATES.unselected);
```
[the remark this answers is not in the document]
JamesMilnerUK: I've made it so setSelectionOnParentNodes returns a list of traversed parents then don't retraverse those parents if they've been set

frontend/components/tree/js/tree.service.ts
JamesMilnerUK: In reality this will only ever be of length 1

carmenfan: if this will only be a length of 1 then I think you need to change your function definition, if it can be more than 1 then you need to do something similar to 1029.

frontend/components/tree/js/tree.service.ts
```diff
@@ -1015,43 +1022,121 @@ export class TreeService {
...
+			const nodeIndex = this.currentSelectedNodes.indexOf(currentNode);
+			const notParentOfUnselected = currentNode.selected !== this.SELECTION_STATES.parentOfUnselected;
+
+			if (select && notParentOfUnselected && nodeIndex === -1) {
+				this.currentSelectedNodes.push(currentNode);
+			} else if (nodeIndex > -1) {
+				this.currentSelectedNodes.splice(nodeIndex, 1);
+			}
+
+			if (currentNode.toggleState !== "invisible") {
```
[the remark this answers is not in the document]
JamesMilnerUK: Yes I will do. I just tried but it broke everything :man_shrugging:

frontend/components/tree/js/tree.service.ts
JamesMilnerUK: Mostly just thinking for extensibility but probably YAGNI

frontend/components/tree/js/tree.service.ts
JamesMilnerUK: Do you mean:

```javascript
if (!select) {
	currentNode.selected = this.SELECTION_STATES.unselected;
} else if(currentNode.toggleState === "parentOfInvisible") {
    currentNode.selected = this.SELECTION_STATES.selected;
} else {
    currentNode.selected = this.SELECTION_STATES.parentOfUnselected;
}
```
Otherwise the logic is different

frontend/components/tree/js/tree.component.ts
```diff
@@ -349,7 +350,7 @@ class TreeController implements ng.IController {
 	 */
 	public ignoreSelection($event: any, node: any): boolean {
 		const doubleClick = $event.detail > 1;
-		if (doubleClick || node.toggleState === "invisible") {
+		if (doubleClick || node.toggleState === this.TreeService.VISIBILITY_STATES.invisible) {
```
carmenfan:
````js
return doubleClick || node.toggleState === this.TreeService.VISIBILITY_STATES.invisible
````

File: frontend/components/tree/js/tree.service.ts [hunk 10]
```diff
@@ -810,13 +817,13 @@ export class TreeService {
...
 	public showTreeNodes(nodes: any[]) {
-		this.setVisibilityOfNodes(nodes, "visible");
+		this.setVisibilityOfNodes(nodes, this.VISIBILITY_STATES.visible);
 		this.updateModelVisibility(this.allNodes[0]);
 	}
 
 	public setTreeNodeStatus(node: any, visibility: string) {
-
-		if (node && ("parentOfInvisible" === visibility || visibility !== node.toggleState)) {
+		console.log(node);
```
carmenfan: remove

frontend/components/tree/js/tree.service.ts
```diff
@@ -1027,51 +1034,70 @@ export class TreeService {
...
-			while (this.currentSelectedNodes.length) {
-				const currentNode = this.currentSelectedNodes.pop();
-				currentNode.selected = this.SELECTION_STATES.unselected;
-				this.setSelectionOnParentNodes(currentNode, this.SELECTION_STATES.unselected);
+		const visitedParents = [];
+		const visitedChildren = [];
+
+		while (this.currentSelectedNodes.length) {
+			const currentNode = this.currentSelectedNodes.pop();
+			currentNode.selected = this.SELECTION_STATES.unselected;
```
carmenfan: this line can be moved within the if statement in 1045. If you've visited this before surely you've flagged this as unselected.

frontend/components/tree/js/tree.service.ts
```diff
@@ -1027,51 +1034,70 @@ export class TreeService {
...
+
+			// Skip over any node that has already been visited
+			if (visitedChildren.indexOf(currentNode) === -1) {
+
+				// Skip over any parent that has been visited
+				if (visitedParents.indexOf(currentNode) === -1) {
+					const parents = this.setSelectionOnParentNodes(currentNode, this.SELECTION_STATES.unselected);
```
carmenfan: You are still visiting some parents multiple times here. Let's use the following example:

currentlySelectedNodes = ["A", "B"];
A's treePath is C D E F A
B's treePath is C D E F A B

loop#1 dealing with A:
`setSelectionOnParentNodes` would be processing C D E F
loop#2 dealing with B:
`setSelectionOnParentNodes` would be processing C D E F A (because visitedParents does not have B in it, C D E F A are in seen parents, but we don't check for it.)

I would just give up on using `setSelectionOnParentNodes` here, you have to loop through the parents to push them into visited parents anyway, you may as well just set their visibility here. less looping, less traversal, and you can break once you've found an ancestor you have visited.

frontend/components/tree/js/tree.service.ts
```diff
@@ -1027,51 +1034,70 @@ export class TreeService {
...
 
 		while (nodes.length) {
-			currentNode = nodes.pop();
+			const currentNode = nodes.pop();
 
 			const nodeIndex = this.currentSelectedNodes.indexOf(currentNode);
-			const notParentOfUnselected = currentNode.selected !== this.SELECTION_STATES.parentOfUnselected;
 
-			if (select && notParentOfUnselected && nodeIndex === -1) {
+			if (
```
carmenfan: flip this around and you don't have to check nodeIndex twice:
````js
if (nodeIndex > -1) {
    this.currentSelectedNodes.splice(nodeIndex, 1);
} else if(select && currentNode.selected !== this.SELECTION_STATES.parentOfUnselected) {
    this.currentSelectedNodes.push(currentNode);
}
````

File: frontend/components/tree/js/tree.service.ts [hunk 14]
```diff
@@ -1081,9 +1107,7 @@ export class TreeService {
 
 		}
 
-		setParentNodes.forEach((node) => {
-			this.setSelectionOnParentNodes(node);
-		});
+		this.setSelectionOnParentNodes(node);
```
carmenfan: If you decided this function only caters for a single node, not an array, then you should change the function definition and stop passing n an array. The while loop is not necessary anymore.

As it stands the function is a bit odd that it takes in an array, but if the user ends up passing in an array then the function doesn't do the right thing on setSelectionOnParentNodes.

frontend/components/tree/js/tree.service.ts
```diff
@@ -1092,12 +1116,14 @@ export class TreeService {
 	 * @param currentNode the node who's parents to traverse up
 	 * @param forceState a selected state to enforce upon parents
 	 */
-	public setSelectionOnParentNodes(currentNode: any, forceState?: string) {
+	public setSelectionOnParentNodes(currentNode: any, forceState?: string): any[] {
 		const parentPath = this.getPath(currentNode._id);
 		parentPath.pop(); // Remove the node itself
+		const seenParents = [];
```
carmenfan: Looks like you only store this to return to L1049.

If you do the changes at L1049 then we don't need to store this anymore.

frontend/components/tree/js/tree.service.ts
```diff
@@ -1034,36 +1045,38 @@ export class TreeService {
...
-		const visitedParents = [];
-		const visitedChildren = [];
+		const visitedNodes = {};
 
-		while (this.currentSelectedNodes.length) {
-			const currentNode = this.currentSelectedNodes.pop();
-			currentNode.selected = this.SELECTION_STATES.unselected;
+		for (const id in this.currentSelectedNodes) {
+
+			if (!id) { continue; }
```
carmenfan: this id is the same as currentNode._id right?

should just merge L1057 with this. ie:

````js
if(!id || visitedNodes[currentNode._id]) { continue; }
````
Should save some operation as iterations where they fall into the continue line at L1058 wouldnt perform the look up on L1054. Also should avoid multiple `continue` code where possible

File: frontend/components/tree/js/tree.service.ts [hunk 17]
```diff
@@ -1078,16 +1091,16 @@ export class TreeService {
...
 			const currentNode = nodes.pop();
 
-			const nodeIndex = this.currentSelectedNodes.indexOf(currentNode);
+			const nodeSelected = this.currentSelectedNodes[currentNode];
 
 			if (
 				select &&
 				currentNode.selected !== this.SELECTION_STATES.parentOfUnselected &&
-				nodeIndex === -1
+				!nodeSelected
```
carmenfan: Seen this mutliple times now... try to avoid if statements where both conditions checks the same boolean.

````js
if(nodeSelected) {
    delete this.currentSelectedNodes[currentNode._id];
} else if (select && currentNode.selected !== this.SELECTION_STATES.parentOfUnselected) {
    this.currentSelectedNodes[currentNode._id] = currentNode;
}
````

**Comment by carmenfan** [comment]
selectionhidden

Note how it doesn't select Clinic A once i've hidden clinic S

File: frontend/components/tree/js/tree.service.ts [hunk 18]
```diff
@@ -1049,14 +1049,9 @@ export class TreeService {
...
 
-			if (!id) { continue; }
-
 			const currentNode = this.currentSelectedNodes[id];
 
-			// Skip over any node that has already been visited
-			if (visitedNodes[currentNode._id]) {
-				continue;
-			}
+			if (!id || visitedNodes[currentNode._id]) { continue; }
```
carmenfan: Actually, move this up before L1052 and do:
````js
if (!id || visitedNodes[id]) { continue; }
````

provided currentNode._id is the same as id (which sounds like it should be?)

carmenfan:
double click to zoom on tree can be very slow, try http://example.org/T_E_S_T/1c4df068-1373-4290-8702-5476fe4ccdc3 and try to double click to zoom on `T_E_S_T:Skanska Architecture`

carmenfan:
Doesn't always work.

Occasionally it's calling zoomtoHighlightedMeshes with 0 selection. See Gif below:

selectionhidden


(double clicking on Skanska Architecture zooms to the bottom part of the carpark... which is the behaviour if you call zoomToHighlightedMeshes with no selection) But it worked on the final time when I tried it.

PR #925, opened 2018-04-26
Issue 819 - Add revision to groups endpoints

Charence:
- New endpoints for:
  - `'/groups/revision/:rid/` (listGroups)
  - `'/groups/revision/:rid/:uid` (findGroup)
- Frontend changes in issues and groups to make use of revisions

ifcGuidsToUUIDs now filters for UUIDs assoc. with a specific revision through History. This is supposed to be beneficial as it removes unnecessary data from models with many revisions.


backend/routes/group.js
```diff
@@ -26,21 +26,32 @@ const Group = require("../models/group");
 const utils = require("../utils");
 const systemLogger = require("../logger.js").systemLogger;
 
-router.get("/", middlewares.issue.canView, listGroups);
-router.get("/:uid", middlewares.issue.canView, findGroup);
-router.put("/:uid", middlewares.issue.canCreate, updateGroup);
-router.post("/", middlewares.issue.canCreate, createGroup);
-router.delete("/:id", middlewares.issue.canCreate, deleteGroup);
+router.get('/groups/', middlewares.issue.canView, listGroups);
+router.get('/groups/:uid', middlewares.issue.canView, findGroup);
```
carmenfan: We should get rid of these. And if it's head then make it call /revision/master/head so it's inline with every other end point

Charence: So both Line 29 and 30 should have `revision/master/head/` inserted?

carmenfan: yea, anything with out revision tagged should be revision/master/head i think

PR #951, opened 2018-05-02
ISSUE #906 - Remove dependency on graphicsmagick

### LINKED ISSUE #906: Remove Imagemagick dependency
Imagemagick is potentially an uncessary binary dependency. We could make resizeAndCropScreenshot  use [sharp](https://github.com/lovell/sharp) instead (supposedly it's 4-5x times faster too).

JamesMilnerUK:
This fixes #906 

#### Description
Removes dependency on graphicsmagick, removing from code, README and CI

#### Test cases
Create an issue with a screen shot
Create a comment with a screen shot


JamesMilnerUK:
@carmenfan this should be mergable now

PR #958, opened 2018-05-03
Issue 923 - Use chips for showing topic types in modelsettings

### LINKED ISSUE #923: List issue topics in a nicer way on Model Settings page
spectacle lh7320

JamesMilnerUK:
This fixes #923

#### Description
Before we had a big list in a text area, chips are a nicer way of displaying the topcis

#### Test cases
Add new topics and save
Delete topics and save
Change topic in an issue


frontend/components/account/js/account-modelsetting.component.ts
```diff
@@ -90,6 +90,13 @@ class AccountModelSettingController implements ng.IController {
 
 	}
 
+	public newTopicType(chip) {
```
carmenfan: is this used anywhere? 
I'm struggling to find the reference

JamesMilnerUK: L47 in  frontend/components/account/pug/account-modelsetting.pug  - it's an Angular function responsible for creating the new topic type for the array

frontend/components/account/pug/account-modelsetting.pug
```diff
@@ -40,18 +40,23 @@
...
 						div(flex="50")
-							md-input-container.modelSettingInput(layout-fill layout-margin)
-								label Issue topic types
-								textarea(type="text", ng-model="vm.topicTypes")
-								.inputHint One line per topic type
+							md-chips(
+								ng-model="vm.topicTypes"
+								name="topicTypes"
+								md-removable="true"
+								md-transform-chip="vm.newTopicType($chip)"
```
JamesMilnerUK: Here :)

carmenfan: can we call it convertTopicType or something?

It's mildly confusion that somewhere down the line you have a variable called newTopicTypes. And at one point I was wondering why you are trying to pass a function around lol

JamesMilnerUK: I'm not sure convertTopicType is a good name as it's not really converting anything. Maybe `createTopicType` ?

carmenfan:
- It's quite laggy when you type in that input bar to create a new chip. 
- Maybe move it down to the bottom of the page and make it take up the whole page width? it makes the form look very asymmetrical at the moment. (But I guess it was pretty asymmetrical before as well... But if your'e here to make it look better then maybe we should address this)
- I feel like the input bar should be above the chip list
- I feel like the chips are massive, but that may just be me.

All in all I'm not sure if it helped the page look better as is, maybe @paloknapo  or @jozefdobos  can give some input on how to format this properly.

frontend/components/account/pug/account-modelsetting.pug
carmenfan: whatever works, just anything but the singular of newTopicTypes.

JamesMilnerUK:
Not sure we can do much about the delay unfortunately as it's an Angular Material internal thing
Also unfortunately cant control the position of the input bar as far as I'm aware
Personally I think the size of the chips is okay, it's just the default size. We could override but I'd rather not
I put it on a separate row and I think it's better now

frontend/components/account/js/account-modelsetting.component.ts
```diff
@@ -90,10 +90,19 @@ class AccountModelSettingController implements ng.IController {
...
+			value: chip.toLowerCase().replace(/ /g, "_"),
+			label: chip
+		};
+	}
+
 	public fetchModelSettings() {
 		this.APIService.get(this.targetAcct + "/" + this.modelId + ".json")
 			.then((response) => {
 
+				console.log(response.data);
```
carmenfan: remove

carmenfan:
image
Clicking on the highlighted region doesn't focus onto the input box even though it looks like it should. We need to eitehr expand that space or reduce the line.

carmenfan:
https://material.angularjs.org/latest/demo/chips

I know it's not 100% the same thing but the lag is no where near as bad on their demo. Are you sure it's not doing something silly?

I can see this lag driving people mad when they want to add multiple topic types. I'm not sure chips is worth the compromise just to make it look a bit smarter.

PR #966, opened 2018-05-09
Issue 741 - Update navigation mode when it changes from inside Unity

### LINKED ISSUE #741: Allow change of navigation mode from within unity
as of #737 update, unity may change the navigation mode from within unity. (this is so gamepad users can just press a button to switch between modes)

With this, a new function is introduced within UnityUtil:
```` js
	public static navMethodChanged(newNavMode) {
		// TODO: do some front end magic to update the navigation button
		// newNavMode can currently be "Turntable" or "Helicopter"
	}

````

This will be called by unity when navigation control has changed internally. front end just needs to update the button accordingly.

### PR DESCRIPTION
**Pull request by JamesMilnerUK**
This fixes #741

#### Description
This updates the UI when unity updates it's navigation mode

#### Test cases
Update navigation using Unity directly

### CODE REVIEW

File: frontend/globals/viewer.ts [hunk 1]
```diff
@@ -69,7 +69,8 @@ export class Viewer {
 		UNITY_ERROR: "VIEWER_EVENT_UNITY_ERROR",
 		UNITY_READY: "VIEWER_EVENT_UNITY_READY",
 		UPDATE_CLIPPING_PLANES: "VIEWER_UPDATE_CLIPPING_PLANE",
-		VR_READY: "VIEWER_EVENT_VR_READY"
+		VR_READY: "VIEWER_EVENT_VR_READY",
+		NAV_MODE_CHANGED: "NAV_MODE_CHANGED"
```
carmenfan: Doesn't look like this is used anywhere?

frontend/components/bottom-buttons/js/bottom-buttons.component.ts
```diff
@@ -120,19 +124,34 @@ class BottomButtonsController implements ng.IController {
...
 
+		this.watchers();
+	}
+
+	public watchers() {
+		this.$scope.$watch(() => {
+			return this.ViewerService.getNavMode();
+		}, (newNav, oldNav) => {
+			if (newNav && newNav !== oldNav) {
+				if (newNav === Viewer.NAV_MODES.HELICOPTER) {
```
carmenfan: Let's define the NAV_MODES as "Helicopter" and "Turntable" in Viewer.NAV_MODES

Then we can just do 
````js 
this.selectMode = newNav
````
And It will also auto fix the problem we have where Unity viewer is actually passing back "Helicopter" not "HELICOPTER"

carmenfan:
good to merge once travis passes

JamesMilnerUK:
@carmenfan I changed it to update and pass Travis but you might want to double check with the Gamepad!

PR #980, opened 2018-05-14
Issue 841 - Clash revival

### LINKED ISSUE #841: Clash revival
yay...

requires https://github.com/3drepo/3drepounity/issues/108

JamesMilnerUK:
This fixes #841

#### Description
Bring back clash to the compare tool

#### Test cases
Clash works and doesn't necessarily load in models


carmenfan:
There's a bug of it not hiding one of the sub models  under the following sequence :
- Load clinic Fed in my account
 - Initiate clash with Struct as base and architectural as target (You have to change rev on the drop down for clinic A)
- Change target to the other revision of clinic A
- the old target is now reverted back as a base, but it doesn't hide it as of the base model tab.
nav1


I think you just have to make sure the base model state is applied before starting to load target model and initiating clash.

carmenfan:
I don't have time to look into this, so there's a new issue for that specific bug https://github.com/3drepo/3drepo.io/issues/985

PR #1030, opened 2018-06-14
ISSUE 975 - Saved views

### LINKED ISSUE #975: Views
### Description
Adding views to 3D repo to allow people to have specific view points of a model

### Goals
- [x] Allow the user to save views 
- [x] Allow the user to go back to a saved view
- [x] Allow user to update a saved view

### Tasks
- [x] Frontend interface for saving and viewing views
- [x] Backend endpoints for saving and viewing views
- [x] Remove Mongoose from backend code

Charence:
This fixes #975 

#### Description
- Adding views to 3D repo to allow people to have specific view points of a model
  - Allow the user to save views
  - Allow the user to go back to a saved view
  - Allow user to update a saved view

#### Test cases
- Create view (with/without clip)
- Edit view
- Delete view


frontend/components/views/js/views.component.ts
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	public resetEditState() {
+		this.canAddView = true;
+		this.editSelectedView = null;
+		this.$timeout();
+	}
+
+	public addView() {
+		this.newView = { name: "" };
```
carmenfan: default name?

frontend/components/views/js/views.service.ts
```diff
@@ -0,0 +1,220 @@
...
+	 * @param model the model id
+	 * @param originalView the original view that will be updated
+	 * @return promise
+	 */
+	public updateView(teamspace: string, model: string, originalView: any) {
+
+		const viewId = originalView._id;
+		const viewsUrl = `${teamspace}/${model}/views/${viewId}/`;
+
+		return this.generateViewObject(teamspace, model, originalView.name)
```
carmenfan: remove as it's not needed? Since we only allow the user to change the name

carmenfan:
image

If you try to update a view that is empty, it causes a internal server error (Because mongo won't let you update an object with no field).

carmenfan:
- Cancelling out of edit mode would still keep the new string in UI
- Error in log when you hit the delete button without any selected view
image
- Deleting a view, then pressing the delete button again without selecting another view will have an error pop up  (looks like selectedView isn't cleared after delete?)
image
- Can we get the text to wrap on the box? we have 3 lines worth but it only displays one
image
- We need an error message to tell the user that the name is too long at creation (or restrict the max length on the box like on edit mode
image

PR #1081, opened 2018-08-08
Issue 1074 - Hooking groups up to the notification service

### LINKED ISSUE #1074: Realtime collab groups state synchronization
### Description
Users connected to the same model wants to keep up to the actual current  state of the groups in realtime.

### Goals
- [x] Collaborators/admins can see groups creation by other users connected to the same model
- [x] Collaborators/admins can see groups deletion by other users connected to the same model
- [x] Collaborators/admins can see groups edition by other users connected to the same model

### Tasks
- [x] Copy the way realtime sync is implemented with issues

sanmont3drepo:
This fixes #1074

#### Description
Adds groups realtime synchronization (creation/update/deletion) between users.

Modifications in backend:
- Added the new events in chatEvent.js
- Trigger these events in model/group.js
- Gets the session_id in routes/group.js and pass it through to the model methods

Modifications in frontend
- Added the suscribe/unsuscribe methods fro the new events in notifications component
- Groups component hookups with the new notifications.
- Added feedback when another user deletes de group, this feedback is in group.pug
  
#### Test cases
Connect to the same model with different users

- [ ] Add a group, the other person should be able to see the creation
- [ ] Delete a group, the other person should be able to see the deletion 
- [ ] Feedback on group deletion should appear as a temporal "this group has been deleted" over the 
group.
- [ ] Update a group, the other person should be able to see the creation


carmenfan:
Looking good, a few bugs and Improvements:

- Can we add a notification if the person is in the edit page of a group whilst someone updated it? (similar to deleted)
- When a group is updated in the edit mode, save button is enabled even though the user didn't make any changes (looks like the edit data is changed but not the actual group data):
group bug 1
- When editing the group's objects, the object counts don't update and the highlights doesn't update until you hit reset:
group bug 2
- Updates are not reflected on the groups list (probably related to the earlier bug):
group bug 3
- updated new groups doesn't seem to contain objects (edit existing object seems fine though!):
group bug 4
- When override all is toggled on, new groups need to be automatically toggled:
group bug 5

carmenfan:
Sorry one more!

- If the edited group is highlighted/focused, then it is refreshed we lose the highlight/focus on the panel, but the model remains highlighted. 
group bug 5

frontend/components/groups/js/groups.component.ts
```diff
@@ -410,6 +394,10 @@ class GroupsController implements ng.IController {
 		this.GroupsService.selectGroup(group);
 	}
 
+	public updateChangeStatus(): void {
+		this.changed = !!this.savedGroupData && angular.toJson(this.savedGroupData) !== angular.toJson(this.selectedGroup);
```
Charence: What does !! do?

PR #1112, opened 2018-08-17
ISSUE #1111 - Fix the override all flag when you go back from editing a group

### LINKED ISSUE #1111: Groups -  after editing a group the override all flag is being reset
### Description
After editing a group the override all flag is being reset.

### Steps to replicate
- Open groups
- Check override all in the component windows dropdown, top right
- Edit a group
- Go back, if you check the dropdown

override all

sanmont3drepo:
This fixes #1111 

#### Description
 Fixes the override all flag when you go back from editing a group

#### Test cases
Set override all , edit a group, comeback. Should stay still with override all.

### DIFF
--- a/frontend/components/groups/js/groups.component.ts
+++ b/frontend/components/groups/js/groups.component.ts
@@ -219,7 +219,7 @@ class GroupsController implements ng.IController {
 		}
 
 		// We don't want color over ride when we're editing
-		this.GroupsService.removeColorOverride(this.selectedGroup._id);
+		this.GroupsService.removeColorOverride(this.selectedGroup._id, false);
 		this.showGroupPane();
 	}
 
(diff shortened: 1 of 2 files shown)


frontend/components/groups/js/groups.component.ts
```diff
@@ -54,6 +54,7 @@ class GroupsController implements ng.IController {
 	private savedGroupData: any;
 	private customIcons: any;
 	private lastColorOverride: any;
+	private lastColorOverrideAll: any;
```
carmenfan: I don't think this is the best approach.

There's 2 problems, 
- another class scope variable which is not desirable
- when you call colorOverrideAllGroups(), it runs the computation again for all the groups (even though we know for a fact that n-1 groups are already overridden

I think it's better if you have a flag passed into `GroupService.removeColorOverride()`, telling it to ignore resetting overrideAll, which wouuld only set to true if we are removing override because we entered the edit page

Let me know what you think! 😄

PR #1120, opened 2018-08-21
Issue 1119 - Forgot Password Input field

### LINKED ISSUE #1119: Forgot password
### Description

Allowing a user to reset password using their "username" or "email".

### Goals

- [x] User will be able to click forgot password link and reset password.
- [x] When resetting password the user will be able to enter username or password. 

### Tasks

- [x] Make forgot password form a single line of input.
- [x] Query database for username or email.
- [x] Update the user's record with correct reset token.

Suggested changes from pull request. 
When username is incorrect, the user is displayed an invalid username message, this could pose a security risk.
- [x] Remove username error validation for security purposes. 
- [x] Ask user to Confirm Password on password reset form.
- The user's reset link is still accessible even after resetting password, this could pose a security risk. 
- [x] Invalidate user's reset link by removing reset token after successful password reset.
- [x] Ensure disable button is only in an active state when the following
 - [x] Password is valid and passes the strength checker. 
 - [x] Password is not weak 
 - [x] Passwords are both the same 
- [x] Refactor code.

- [x] Found bug on password-change.component.ts, where value "result.feedback.suggestions" was returning an undefined value. 
password-bug

### Description
When a user clicks the forgot password link, they will have to enter their username or email.

### Steps to replicate
- Click on forgot password link on login page
- Enter username or email address, 
- Click 'Request Password Change' button.

### Steps to replicate for password reset.
- Enter password 
- Type password incorrectly (error message will appear)
- retype password (Tab away from input field, to remove error message.)
- Click 'Send password request'.

niblux:
This fixes #1119 

#### Description
- Reduced forgot password form to a single line of input.
- User can now submit a "username" or "password" to reset password.


backend/models/user.js
```diff
@@ -452,29 +466,37 @@ schema.statics.getForgotPasswordToken = function(username, email, tokenExpiryTim
...
+	let user;
+
+	return this.findUsernameOrEmail(userNameOrEmail).then(_user => {
+
+		user = _user;
 
-		if(user.customData.email !== email) {
-			return Promise.reject({ resCode: responseCodes.USER_EMAIL_NOT_MATCH});
+		if (!user) {
+			return Promise.reject(responseCodes.INVALID_USERNAME);
```
Charence: I'm wondering if we want to tell the user that they've entered an invalid username (or e-mail address).

Maybe it should always show "Thank you. You will receive an e-mail...." even if the username/e-mail is invalid. This is so people can't figure out which users exist on our system by just using the forgot password form. Internally, we'll still have to figure out if the username/e-mail is valid so we don't try to generate a token for a non-existent user.

frontend/components/password-forgot/pug/password-forgot.pug
```diff
@@ -15,13 +15,10 @@
...
-			p Please enter your username and email
+			p Please enter your username or email
 			md-input-container.md-block
-				label Username
-				input(type='text', ng-model='vm.username', ng-required, autofocus)
-			md-input-container.md-block
-				label Email
-				input(type='text', ng-model='vm.email', ng-required, ng-keypress="vm.requestPasswordChange($event)")
+				label Username or Email
+				input(type='text', ng-model='vm.usernameOrEmail', ng-required, ng-keypress="vm.requestPasswordChange($event)", autofocus)
```
Charence: Can we also have a "confirm password" field here? Similar to the sign up page.

niblux: I have added the confirm password fields to the password reset.
I have also tested to see whether, the user can submit the password if passwords do no match. 
The user can only submit if both password's are correct.

backend/models/user.js
niblux: I have removed all the error messages that appear , when entering an incorrect username or email address.
I have also changed the server side validation to only set a token if a username or email has been found.

frontend/components/password-change/pug/password-change.pug
```diff
@@ -15,6 +15,16 @@
 							ng-model='vm.newPassword'
 							ng-required
 							ng-keypress="vm.passwordChange($event)")
+					md-input-container.md-block
+						label(ng-if="vm.registerMessage !== undefined") Confirm Password {{vm.registerMessage}}
```
Charence: vm.registerMessage never seems to show anything?

niblux: The "Set Password" button is only activated when both passwords are correct and they pass the strength validator. 

I have also resolved a css bug i came across, which squashed the password change div on lower mobile breakpoints.

squashedpassworddiv
squashedpassworddivmobile

The images below show the final result of css bug.

resolvedcss
resolvedcss2

carmenfan:
image
@niblux issue tag!!!

frontend/components/password-change/js/password-change.component.ts
```diff
@@ -80,13 +80,19 @@ class PasswordChangeController implements ng.IController {
...
+		this.$scope.$watchGroup(["vm.newPassword", "vm.confirmPassword"], () => {
 			this.message = "";
 			if (this.newPassword !== undefined) {
 				const result = this.PasswordService.evaluatePassword(this.newPassword);
 				this.passwordStrength = this.PasswordService.getPasswordStrength(this.newPassword, result.score);
 				this.checkInvalidPassword(result);
 			}
+			if (this.confirmPassword !== undefined && this.confirmPassword === this.newPassword) {
+				this.confirmPasswordInput();
+				this.buttonDisabled = false;
```
Charence: I think **buttonDisabled** state should be based on both confirmPasswordInput and checkInvalidPassword().

At the moment, if you enter a weak password (e.g. "asdf") and then do the confirm password correctly, you can press the "Set Password" button, which should be disabled because the password is still weak.

frontend/components/password-change/pug/password-change.pug
```diff
@@ -11,10 +11,22 @@
...
+						label(ng-if="vm.registerMessage === undefined") Confirm Password
+						input(
+							name="confirm"
+							aria-label="password confirm"
+							type="password"
+							ng-model='vm.confirmPassword'
+							ng-required
+							ng-blur="vm.confirmPasswordInput()"
+							ng-keypress="vm.confirmPasswordInput()")
```
Charence: For ng-keypress, can we make it do both confirmPasswordInput() and passwordChange(event)? I think we want it to confirm the password as we type as well as setting the password if someone presses enter.

frontend/components/password-change/js/password.service.ts
```diff
@@ -31,6 +32,12 @@ export class PasswordService {
 		this.accountDefer = $q.defer();
 	}
 
+	public checkDuplicates(password, confirmPassword) {
+		const passwordValid = password === confirmPassword;
+		this.passwordMessage = passwordValid ? "" : "(Password mismatched)";
```
carmenfan: Why is this a class scope variable?

frontend/components/password-change/js/password-change.component.ts
```diff
@@ -105,36 +113,26 @@ class PasswordChangeController implements ng.IController {
 
 	}
 
+	public confirmPasswordInput() {
+		return this.registerMessage = this.PasswordService.checkDuplicates(this.newPassword, this.confirmPassword);
```
carmenfan: This is super overkill. We don't need to go all the way to password service to do an equality check.

avoid class scope variable unless it's absolutely justified. 

All these can be done in the pug with an ngIf. i don't think this function needs to exist.

niblux: I completley agree :) as i was refactoring i have already removed it from the service.

Issue #1095, opened 2018-08-10
User Management tab - Design improvements


simplypixi:
### Descprition
This is a issue that gathers all issues related to the permissions view redesign.

### Related issues
- [x] #1082 Users tab
- [x] #1083 Jobs tab
- [x] #1084 Project permissions
- [x] #1085 Models and federations permissions
- [x] #1086 Redirection from Teamspaces
- [x] #1087 User panel

PR #1136, opened 2018-09-06
Issue 1132: Migration - React setup

### LINKED ISSUE #1132: Migration: React setup
It'll be initial step to migrate from AngularJS to ReactJS.
This issue includes React environment setup.

### Tasks
- [x] Add necessary React packages
- [x] Extend webpack configuration
- [x] Add new generators to plop
- [x] Redesign a files` structure to split AngularJS and React code
- [x] Rewrite one, small component to test extended environment and fix potentially bugs

simplypixi:
This fixes #1132 

#### Description
It'll be initial step to migrate the app from AngularJS to ReactJS.

#### Changes
- Added necessary libs:
   - react
   - react-dom
   - redux
   - reduxsauce
   - reselect
   - material-ui
   - axios

- Added new react-oriented generators to plop.js
- Redesigned a files structure to split AngularJS and React code

   - **routes** - This directory will contain subdirectories - one for each route (view) of your application with components, containers and styles inside. I recommend using flat structure which means that you should put each route as a child of this directory regardless of view relationship.

   - **modules** - This folder will contain constants, helpers, sagas, reducers etc. grouped in modules which means that you should keep your business logic here

   - **services** - This directory will contain proxies for all services with which your app is going to communicate. Initially there's only api proxy, which returns an axios instance for default backend API.

   - **styles** - In this directory we will put any global styles that cannot be placed in routes.

   - **internals** - configs of webpack, plop and other tools
- Refactored webpack configuration
- Decreased production bundle size
   I added small performance improvement. If you want I could increase it even more, but I think it's not proper time for do that.

- Added examples:
   - component: _(react-button)_,
   - route: _(react-route)_
   - module: _(teamspace)_

- Added migration helpers in `helpers/migrations.ts`

#### Test cases
- Running app in development mode
- Running e2e tests
- App building
- Loading data on app init (GET ${username}.json) - I changed AngularJS API service to axios and I rewrite this API call to test new config. You have to check whether everything is ok when app is loading.


simplypixi:
@carmenfan @Charence I started this branch from #1095 branch. I think we should merge #1095 first to `staging`, a then I'll rebase this branch again. What do you think about it?

carmenfan:
@simplypixi 
1. agreed
2. it looks like it's failing on linter?

simplypixi:
@carmenfan 
Whaat.... I saw something different before:
image

But it's a good message! ;)

carmenfan:
@simplypixi it works fine on my laptop (windows), and i added a couple of debug log on travis to see

Could it be that linux is case sensitive on directories so it's not recognising it?
image
image


notice the b is in caps on the `ls`

simplypixi:
It was a typo...
I fixed it and removed changes from travis config

carmenfan:
looks fine, can merge once conflicts has been resolved!

PR #1144, opened 2018-09-17
Issue 1137 - Migration - Users Tab

### LINKED ISSUE #1137: Migration: Users (Users Management)
### Description
This issue includes migration of Users tab in Users Management view.

### Goals
Migration of Users tab in Users Management view to ReactJS

### Tasks
- Rewrite MembersService
- Rewrite important methods from DialogsService
- Move pug template to JSX
- Move styles to styled components

simplypixi:
#### Description
This fixes #1137 

#### Changes
- Added new components: NewUserForm, JobItem, UserItem, CustomTable (generic), FloatingActionPanel, Highlight, User, Search, CellSelect
- Added Users container
- Moved MembersService to Users saga
- Added basic Dialog container and module
- Moved pug template to JSX
- Moved styles to styled components
- Removed React examples and old AngularJS components (UsersList, NewMemberForm)


simplypixi:
> @simplypixi
> I just tried running this issue branch and I'm getting a blank page with an error about minified react code:
> image
> 
> Any ideas? I assume the build process hasn't changed?

Wooow... this is weird. I'll check it tomorrow morning.
I moved webpack configs to seperated directory (internals) and added new config to handle JSX syntax. 
Did you call `yarn start` or `yarn build`?

carmenfan:
@simplypixi 
I called `yarn run build` (after `yarn install`)

simplypixi:
> @simplypixi
> I called `yarn run build` (after `yarn install`)

The issue exists because of newer React version. It worked fine on my side until I ran `yarn install`.
So I have a two questions:

1. Why did you remove `yarn.lock`? :)
2. Should I define specific version of React in `package.json`?

carmenfan:
> Why did you remove yarn.lock? :)
We've had a lot of trouble with it in the past (esp when doing automatic update deployments on the servers) where there's a conflict in yarn.lock. So we fixated all the libraries specific versions on package.json and ditch the yarn.lock

> Should I define specific version of React in package.json? Previous or the new one?
Yes please! I guess the new one, unless there's any reasons you think it's better to use the older one.

simplypixi:
> Yes please! I guess the new one, unless there's any reasons you think it's better to use the older one.

I think the new version has a bug, which you found. 
React knows that component is mounted but DOM library (which is used by React) can't find proper node because component it's unmounted :)
Moreover this error appears in `componentDidMount` lifecycle method, so even method's name suggest that the component is properly mounted. 

I suggest to stay with the older version for now. In the meantime I'll report a new bug to React's dev team.
I can update React version later - when I finish whole Users Management view.

Is it ok for you?

simplypixi:
> Clicking on User Management should default to the Users tab, but it is showing Projects tab by default

I think this bug has already appeared before migration, because it occurs in Angular context. I'll try to fix it. 

> The styling on the pop up also changed, which is probably fine I think as we weren't consistent before, but just bear in mind for the future we should stick to one :)

Honestly, I forgot about the styling for dialogs, but I think default styling has more pros than yours e.g buttons should be placed on bottom-right corners (usually statistical user looking for action button in that place).

Maybe we should do something like that? It's a consensus between these solutions.  What do you think about it?
image

If you want I can add old styling to new dialog. It's not a problem.

> Sorting only changes direction if the user clicks twice (first click only focuses onto the button instead of changing it)

Are you sure about that? I think it works properly. If you click on inactive sorting it should be activated and data will be sorted using current (for this sorting) direction. Moreover it's default behavior for that component.

carmenfan:
> Maybe we should do something like that? It's a consensus between these solutions. What do you think about it?
Yea that looks quite nice, let's do that.

> Are you sure about that? I think it works properly. If you click on inactive sorting it should be activated and data will be sorted using current (for this sorting) direction. Moreover it's default behavior for that component.
The reason why I noticed it is that angular didn't used to do that. I think we don't actually want active sorting (the major pain point for this is when the sorting is activated and the user changed something, it will jump away because it gets sorted to somewhere else), we only want to sort it once per click so maybe that's not the right component to use?

simplypixi:
> The reason why I noticed it is that angular didn't used to do that. I think we don't actually want active sorting (the major pain point for this is when the sorting is activated and the user changed something, it will jump away because it gets sorted to somewhere else), we only want to sort it once per click so maybe that's not the right component to use?

I think we have misunderstood. :)
I thought about the sorting button (and it's behavior) only, because you said:

> Sorting only changes direction if the user clicks twice (first click only focuses onto the button instead of changing it)

The sorting when data was changed is different case and I'll fix this.

Summary: 
I think sorting button and changing sort direction works properly.
I have to disable sorting on data change.

carmenfan:
haha, I think I mean both.

> - I think sorting button and changing sort direction works properly.

I think it should take 1 click to activate and change the sorting direction. basicalyl what the angular 
version is currently doing:
filter

simplypixi:
> I think it should take 1 click to activate and change the sorting direction. basicalyl what the angular
> version is currently doing:

ok!

simplypixi:
@carmenfan Fixed

simplypixi:
@Charence Fixed ;)

carmenfan:
All good, thanks :)

PR #1163, opened 2018-09-20
Issue 1135 - Tree improvement

### LINKED ISSUE #1135: tree/selection function not good when structure is flat
### Description
We run into problems with really slow highlighting/unhighlighting and potentailly crashing when you select an item if the tree structure is extremely flat (like a large obj with a lot of meshes only attached to the root of the tree)

### Steps to replicate
- Load a super flat structured model with lots of meshes
- things like selection/isolate performs really slow
- eventually it may crash the viewer

carmenfan:
This fixes #1135

#### Description
Various effort to optimise tree selection code.

#### Test cases
using this model: (carmen/d0eba0cc-989b-4e84-a7a5-8a455e10ff68)
- selecting objects should not take forever anymore.
- expanding/collpasing the tree should have minimal delay
- clicking on objects before tree is ready should no longer through an error on console log(it's now waiting on treeReady promise before proceeding)
- Should work fine on other models too!

### DIFF
--- a/frontend/components/compare/js/compare.service.ts
+++ b/frontend/components/compare/js/compare.service.ts
@@ -455,7 +455,7 @@ export class CompareService {
 	private changeModelVisibility(nodeName: string, visible: boolean) {
 		const tree = this.TreeService.getAllNodes();
 		if (tree.children) {
-			nodes[0].children.forEach((node) => {
+			tree.children.forEach((node) => {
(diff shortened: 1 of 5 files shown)


carmenfan:
@Charence should be fixed now. I don't believe issue needs to call that function at all and getting rid of it seems to have made no difference with the functionality so I just removed it.

PR #1196, opened 2018-10-10
Issue 1158 - Delete Filtered Groups

### LINKED ISSUE #1158: Delete all filtered groups
### Description

Delete All in groups currently deletes all groups; change functionality to only delete shown groups if filter applied

### Goals

- [x] Delete all should delete all groups if no filter applied
- [x] Delete all should not delete filtered out groups

### Tasks

- [x] Delete all filtered groups
- [x] Consider renaming "Delete All" button

niblux:
This fixes #1158 

#### Description
When clicking the "Delete all" button on the Groups component all groups are deleted.
However when a list of groups is filtered and a new list is displayed, when clicking "Delete all" this still deletes all the groups.
The desired functionality was to only delete the filtered groups that are displayed, and not the groups that were filtered out.

#### Test cases

- Filter a list of groups and click "Delete all", this will only delete the filtered list.
- When no filter search query has been provided , and all groups are displayed the "Delete all" button will delete all groups in the current model. 

groupsfilterdelete


frontend/components/groups/js/groups.component.ts
```diff
@@ -258,7 +259,11 @@ class GroupsController implements ng.IController {
 		const escapable = true;
 		this.dialogService.confirm(`Confirm Delete`, content, escapable, "Yes", "Cancel")
 			.then(() => {
-				this.groupsService.deleteAllGroups(this.teamspace, this.model);
+				if (this.filterText) {
```
carmenfan: This is not necessary, see comment in GroupService

frontend/components/groups/js/groups.service.ts
```diff
@@ -329,6 +329,16 @@ export class GroupsService {
 		return this.deleteGroups(teamspace, model, groupsToDelete);
 	}
 
+	/**
+	 * Deletes filtered groups in the current model
+	 * @param teamspace the teamspace name for the group
+	 * @param model the model id for the group
+	 */
+
+	public deleteFilteredGroups(teamspace: string, model: string, filteredGroups: any) {
```
carmenfan: Instead of creating this function, we should rename `deleteAllGroups` to `deleteAllDisplayedGroups`. We always only want to delete the groups that are currently displayed. If there is no filter, `groups` would be teh same as `groupsToShow`

So a new function is unnecessary. (but it would be good to change all the prompts from `delete all` to `delete all displayed` or something

niblux: I opted for the deleteGroups method , much less code and only being called once in groups.component.

carmenfan: We need to delete this method as it's not used

frontend/components/groups/js/groups.component.ts
```diff
@@ -258,7 +258,7 @@ class GroupsController implements ng.IController {
 		const escapable = true;
 		this.dialogService.confirm(`Confirm Delete`, content, escapable, "Yes", "Cancel")
 			.then(() => {
-				this.groupsService.deleteAllGroups(this.teamspace, this.model);
+					this.groupsService.deleteGroups(this.teamspace, this.model, this.groupsToShow);
```
Charence: Maybe Line 349 of deleteGroups in groups.service needs a check to make sure groupsToShow is not null?

niblux:
Changes to the above requests have been completed :)

niblux:
Hi Carmen, 
 I took your suggestions on board and reverted back to the original approach i had , the only difference is i kept the changes that had "groupsToShow" as a seperate state array instead of creating two arrays. 

The deleting groups works, and does not affect the colour override and i didnt need to use a whole new function i just had to make sure the deleteStateGroup was using "groupsToShow" instead of just groups

niblux:
I just pushed a small bug fix i found when doing some further tests this morning.

carmenfan:
@niblux  There was only tiny issues left so I made a commit, this PR can be merged now.

> I took your suggestions on board and reverted back to the original approach i had , the only difference is i kept the changes that had "groupsToShow" as a seperate state array instead of creating two arrays.

That's fine, but to that that you should clone the array (see my commit). The issue I had was that we're doing a filter function unnecessarily. (imagine if you have 200million objects, we are looping through that when we know we just want to show the whole array, it's wasted computation.)


Also, start a habit of looking at the change list once you've made a PR/ declare a PR is ready for another review. You'd catch silly things like a function that you added but no longer need is still there!

PR #1205, opened 2018-10-15
Issue 1204: Missed property in axios config

### LINKED ISSUE #1204: Migration: HTTP library config - missed session cookie
### Description
The session cookie is missed for cross-origin requests.

simplypixi:
This fixes #1204

#### Description
- Added missed property in axios config


simplypixi:
@carmenfan Probably this is the solution to the problem, but I didn't test it locally (out of office).

simplypixi:
@carmenfan I checked the fix locally and everything seems to be fine. Could you confirm that on your machine?

PR #1207, opened 2018-10-16
ISSUE #992 - Fix group description height bug firefox

### LINKED ISSUE #992: Group descriptions semi-visible
### Description
Group description text cuts off vertically in Firefox, only the top half of text visible.

### Picture
image

niblux:
This fixes #992 

#### Description
Groups Description was only slightly showing on Firefox.

This has been fixed by slightly increasing the min and max height on the '.group' div.

#### Test cases
Group description should be clearly visible and not cut off.


frontend/components/groups/css/groups.css
```diff
@@ -136,8 +136,8 @@ groups .group:hover {
 
 groups .group {
     border-bottom: solid gainsboro 1px;
-    min-height: 95px;
-    max-height: 95px;
+    min-height: 120px;
+    max-height: 120px;
```
carmenfan: 25 px is probably too excessive, i think you need like another 5px maybe to make the desc visible?

I think it's better if we can reduce the spacing between the group name, author and description, as that's a lot of padding between them.

niblux:
I have reduced the min and max height on the group div.

PR #1208, opened 2018-10-17
Issue 1174 - Fix race condition bug

### LINKED ISSUE #1174: Race condition on mongo save causing federations/models to occasionally fail in saving
- [ ] [Can't save federation](https://github.com/3drepo/3drepo.io/issues/1174#issuecomment-425377207)

carmenfan:
This fixes #1174 

#### Description
- refactored `createAndAssignRole` into `createNewModel` and `createNewFederation`
- moved model settings creation from `model/helper/model.js` to `model/modelSettings.js`
- moved model name regex check from `model/helper/model.js` to `model/modelSettings.js`
- removed model name blacklisting - not sure why it was done in the first place.
- Minimised the amount of time we load and save settings (the actual fix for race condition)
- Only allow model creations if a valid project name has been provided
- Fixed travis tests

#### Test cases
- [x] Race condition should no longer happen (Kieron tested this)
- [x] create new model should still work
- [x] create new federation should still work
- [x] toy project import should still work
- [x] edit federation should still work


carmenfan:
Should be fixed now!

Issue #1183, opened 2018-10-04
Migration: Teamspaces


simplypixi:
#### Description
This issue includes migration of Teamspaces view.

#### Goals
Migration of Teamspaces view to ReactJS.
Sketch: https://xd.adobe.com/view/9180ddc9-5807-4b95-6985-2f573e452411-e7b1/

#### Tasks
- [x] Component to display data in tree view
- [x] Component to add/edit project
- [x] Component to add/edit model
- [x] Component to add/edit federation
- [x] Component to upload files
- [x] Component to display revisions
- [x] Model settings route
- [x] Refactor and move old logic to new components
- [x] Expand teamspace tab height to almost full height 
- [x] Move plus button to top right 
- [x] On load, auto expand to own teamspace, showing all projects (collapsed)
- [x] Federation should list before models
- [x] Remove icons from model and federation list
- [x] When click on a project, auto expand models and federations
- [x] Auto collapse any expanded teamspace when another teamspace is expanded
- [x] Tooltip on the new hover menu buttons
- [x] Split new model and model upload
- [x] Model upload component - show model name, latest revision info and units
- [x] Enforce revision name as compulsory on model upload
- [x] New federation pop up utilises tick-boxes to select and add-remove from/to federation
- [x] Edit federation pop up is the same as new federation, just units drop-down is frozen 
- [x] Remove the disable on going into the viewer when a new revision is being uploaded
- [x] Move button to right hand side on pop up components
- [x] Teamspace, Federation folder, Model folder shouldn't have a hover menu/hamburger button, they should only have a + button permanently showing on their rows (note how + button is now aligned with the hamburger)
- [x] Project, individual model/federations should have a hover menu, with hamburger button acting as a show/hide 
- [x] Project rows should no longer have a + button on there (as it doesn't make any sense - since the user still needs to choose whether they want a model or federation)

simplypixi:
@carmenfan Do you know already anything about changes in this view? Did Pavol prepare mockups?

carmenfan:
We're literally having a meeting about this now, so hopefully we can send you the info by the end of the day!

simplypixi:
@carmenfan Perfect! ;) In this case I'll start migration of Profile page.

simplypixi:
@carmenfan 
We have some errors while adding models.

@dleszcz gets:
[...]

PR #1248, opened 2018-11-21
Issue 1206 : Prevent i.dgn file upload

### LINKED ISSUE #1206: detect i.dgn and throw file unsupported error
### Description
We support dgn but not idgn. Apparently idgn's extensions are `.i.dgn`, so we need to treat it as special case and return file format not supported.

### Goals
- dgn files should pass
- idgn files should fail

### Tasks
- [x] Add i.dgn as a non-accepted format and run check to prevent uploading
- [ ] Ensure this is also applicable on the backend. 
- [x] Display the i.dgn file error "FILE NOT SUPPORTED"

niblux:
This fixes #1206 : Prevent i.dgn file upload

#### Description
Add a file check to prevent file's with an i.dgn format being accepted.

#### Test cases
- The user will not be able to upload a file with the extension "i.dgn"


niblux:
Please note some of the change on this commit were from a copy of the latest version of the 'uploadFileComponent'.

frontend/routes/teamspaces/components/uploadModelFileDialog/uploadModelFileDialog.component.tsx
```diff
@@ -68,14 +68,19 @@ export class UploadModelFileDialog extends React.PureComponent<IProps, IState> {
...
 			tag: values.revisionName,
 			desc: values.revisionDesc
 		};
-		uploadModelFile(teamspaceName, projectName, modelId, fileData);
+
+		const modelData = {
+			modelName, modelId
+		};
+
+		uploadModelFile(teamspaceName, projectName, modelData, fileData);
```
sanmont3drepo: I think this breaks the functionality. 

The uploadModelFile  function  that gets called is actually here:
https://github.com/3drepo/3drepo.io/blob/4743a3927cbf4cce777104739fae2be9e3f6f423/frontend/modules/model/model.sagas.ts#L101

And it expects the third parameter to be just the modelId

niblux: Resolved, as error when copying and not merging correctly.

niblux:
I have implemented a back end file format check, which is ready for review. 
Test Cases : When a user uploads a file of an i.dgn file type the upload will return "Format not accepted" error.

carmenfan:
waiting on #1183 to merge first

niblux:
@carmenfan Thank you carmen, i love what you done with the file check. Squeaky clean 👍

PR #1277, opened 2018-12-07
Issue 1259 Migration - Views

dleszcz:
### Description
- Created a new React Views component
- Created a Views module to use ViewerService methods
- Implement styles with styled-components
- Remove old AngularJS views component


frontend/routes/viewer/components/views/views.component.tsx
```diff
@@ -0,0 +1,300 @@
+/**
```
simplypixi: @dleszcz Please move view item code to separate component ;)

dleszcz: done

simplypixi:
@carmenfan Could you check this PR?

dleszcz:
- [ ] Pressing the cross should have the same affect as clearing the query
- [ ] When search bar appears, it adds a horizontal scroll bar on the bottom
- [ ] the search doesn't seems to understand capital letters
- [ ] Deleting the last result of a search should display "No viewpoints matched" instead of bouncing back to the unsearched list
- [ ] it should auto focus onto the new name textbox after pressing the + button
- [ ] It should not be saving the viewpoint straight away(at the moment it looks liek it's creating it at the point it creates the item then does an edit when you click save. It should only save when the user clicks save. Pressing the cross should remove it from the list as if it never existed.
- [ ] None of the tabs on the view should reload on show/hide. The state should remain the same as before it was hidden
- [ ] New viewpoints/edits/deletes coming from chat service should not reset the filter:

@carmenfan  Fixed.

sanmont3drepo:
I dont understand why there is a renderwhentrue function if there is a way for doing this in react already outof the box:

`{booleanVariable && <ReactComponent/>}
`

sanmont3drepo:
> > I dont understand why there is a renderwhentrue function if there is a way for doing this in react already out of the box:
> 
> 1. This is not React feature. ;)
> 2. renderWhenTrue is just a helper, which is clearer and more understandable. Moreover, it's related to functional programming, which is the future of developing apps in React. We use this approach successfully in our projects.
> 
> Trust me. You'll have wonder, why you have not used this approach never before. ;)

That mindset actually reminds me of the statemanager implementation in the angular version. 
I disagree with the "clearer and more understandable". 
Not to distrust you, but is there any other reason beyond that?

simplypixi:
> > > I dont understand why there is a renderwhentrue function if there is a way for doing this in react already out of the box:
> > 
> > 
> > 
> > 1. This is not React feature. ;)
> > 2. renderWhenTrue is just a helper, which is clearer and more understandable. Moreover, it's related to functional programming, which is the future of developing apps in React. We use this approach successfully in our projects.
> > 
> > Trust me. You'll have wonder, why you have not used this approach never before. ;)
> 
> That mindset actually reminds me of the statemanager implementation in the angular version.
> I disagree with the "clearer and more understandable".

Please, don't try to compare stateManager to this solution. stateManager is a creature, which was created (probably) by someone without any (or with small) experience in Angular. It's totally rash.

> Not to distrust you, but is there any other reason beyond that?
Yep. "Moreover, it's related to **functional programming**, which is the future of developing apps in React. **We use this approach successfully in our projects.**".
Usually, we use this approach with the Ramda library (which is a functional equivalent of Lodash).

And - honestly - there is a lot of different approaches to achieve the same result e.g `if/else`, `ternary operator`, `Short-circuit operator`, `IIFE`, `HoC` and many other custom solutions like `renderWhenTrue`. It depends on the developer which of these solutions will be used. We decided to use `renderWhenTrue` and `Short-circuit operator` depending on our current needs.

The example provided by you looks good because is simple. Please, try to handle some more complex example using the same approach.

simplypixi:
> > > > > I dont understand why there is a renderwhentrue function if there is a way for doing this in react already out of the box:
> > > > 
> > > > 
> > > > 
> > > > 1. This is not React feature. ;)
> > > > 2. renderWhenTrue is just a helper, which is clearer and more understandable. Moreover, it's related to functional programming, which is the future of developing apps in React. We use this approach successfully in our projects.
> > > > 
> > > > Trust me. You'll have wonder, why you have not used this approach never before. ;)
> > > 
> > > 
> > > That mindset actually reminds me of the statemanager implementation in the angular version.
> > > I disagree with the "clearer and more understandable".
> > 
> > 
> > Please, don't try to compare stateManager to this solution. stateManager is a creature, which was created (probably) by someone without any (or with small) experience in Angular. It's totally rash.
> 
> Cleary the stateManager is way worst, but what I meant is that is a reimplementation of an already built feature. We just want to be sure that we dont wanna go in that direction.
> 
> > > Not to distrust you, but is there any other reason beyond that?
> > 
> > 
> > Yep. "Moreover, it's related to **functional programming**, which is the future of developing apps in React. **We use this approach successfully in our projects.**".
> > Usually, we use this approach with the Ramda library (which is a functional equivalent of Lodash).
> 
> I do like functional programming, but I'm not sure that in every case is the right response just like with any programming paradigm (imho).
> 
> > And - honestly - there is a lot of different approaches to achieve the same result e.g `if/else`, `ternary operator`, `Short-circuit operator`, `IIFE`, `HoC` and many other custom solutions like `renderWhenTrue`. It depends on the developer which of these solutions will be used. We decided to use `renderWhenTrue` and `Short-circuit operator` depending on our current needs.
> > The example provided by you looks good because is simple. Please, try to handle some more complex example using the same approach.
> 
> Still if you feel more comfortable with this approach and you think it will be useful in the future lets go for it.

@sanmont3drepo  Roger that! ;)
We provide you only the solutions verified on "battlefield" and try to decrease the complexity level of migration to new React's features in the future. We don't reinvent the wheel - even if it looks so. Don't worry about it!

But we don't know everything (nobody knows!), so if you have some idea, let us know. We can discuss it as we did it before.

@dleszcz Could you look at the changes reported by @carmenfan ? Thanks!

dleszcz:
> @dleszcz Could you look at the changes reported by @carmenfan ? Thanks!

I'm doing it now :)

dleszcz:
> Can we style this message a bit better?
Done.

> You're still doing a post and put to create a new view. It should only do a post when the user finishes the edit and hit save.
But then there will be no picture. We have to use a placeholder instead?

> Is it possible to keep the state when you hide and unhide?
Yes, it's possible. We have to do it, right?

carmenfan:
> @carmenfan
> 
> > Can we style this message a bit better?
> 
> Done.

Thanks 😄 
> 
> > You're still doing a post and put to create a new view. It should only do a post when the user finishes the edit and hit save.
> 
> But then there will be no picture. We have to use a placeholder instead?
> 

You technically have the picture, because the screenshot originally comes from the unity viewer and you got it via getScreenShot method. So you could display that (just like how screenshot edit panel displays it), alternatively yes we could just use a placeholder, or just display nothing at all until saved.

> > Is it possible to keep the state when you hide and unhide?
> 
> Yes, it's possible. We have to do it, right?

That would be awesome :)

simplypixi:
> Can we make it (and this is true for all panels) to initialise at the point of the viewer being loaded instead of at the first time it is visible? Reason being is that we very often pre-load the model to ensure there's no connection problems during a demo. and as this list increases or for more complicated tabs (like tree) it will take a lot longer to load.

@carmenfan I think we don't understand what specifically we should do. :) Which initialization do you mean? Could you be more specific?
Thanks!

carmenfan:
> > Can we make it (and this is true for all panels) to initialise at the point of the viewer being loaded instead of at the first time it is visible? Reason being is that we very often pre-load the model to ensure there's no connection problems during a demo. and as this list increases or for more complicated tabs (like tree) it will take a lot longer to load.
> 
> @carmenfan I think we don't understand what specifically we should do. :) Which initialization do you mean? Could you be more specific?
> Thanks!

I mean the contents of the panels needs to be loaded before the user click to see them, see here, the apis are clearly called upon the first reveal of the panel:



In comparisong to staging:

simplypixi:
> > > Can we make it (and this is true for all panels) to initialise at the point of the viewer being loaded instead of at the first time it is visible? Reason being is that we very often pre-load the model to ensure there's no connection problems during a demo. and as this list increases or for more complicated tabs (like tree) it will take a lot longer to load.
> > 
> > 
> > @carmenfan I think we don't understand what specifically we should do. :) Which initialization do you mean? Could you be more specific?
> > Thanks!
> 
> I mean the contents of the panels needs to be loaded before the user click to see them, see here, the apis are clearly called upon the first reveal of the panel:
> 
> 
> In comparisong to staging:
> 

@carmenfan So, the loader is ok but you want to see the loaded content instantly (e.g without delayed displaying images). Am I right? Or maybe do you want to remove loader too?

carmenfan:
@dleszcz Up to you guys how to implement it, but I need it to have fetched all the data pretty much as the model is loading (i.e. if you look at the network tab, it doesn't do extra API requests when i click the button because data was fetched during model load)

simplypixi:
> @dleszcz Up to you guys how to implement it, but I need it to have fetched all the data pretty much as the model is loading (i.e. if you look at the network tab, it doesn't do extra API requests when i click the button because data was fetched during model load)

Got it, but we fetch the data only if the panel is opened for the first time. Model loading (and rendering) is very overloading for the browser. Are you sure about loading data for whole panels at the same time? I know, you do that currently, but honestly, my Mac is constantly overheating during development on Viewer view. So, I think the better approach is loading the data if are truly needed.

My idea is:
1. Loading data on the first panel opening (with a loader)
2. Prefetching content (like images) to display it instantly

But we can still move back to your previous solution.

@carmenfan What do you think about it?

carmenfan:
> @carmenfan What do you think about it?

@simplypixi I totally see where you're coming from and I had a chat with Pavol about it. In the end we still think we need the data prefetched (The UI doesn't need to be pre-rendered, and images are less important too, if you think you can decouple that ), for 3 main reasons:

- Users are very used to CAD softwares, where they load something, go get a coffee and once everything is loaded they expect everything to be there without much further load time (application style of experience)
- There are use cases where the user has very sporadic connections/even potentially in offline mode, where they need access to certain functionality in a read only kind of situations (e.g. us Demoing with a very dodgy client wifi, or clients going into underground tunnels wanting to see their models on site) . to tell them that "you can still do it, but make sure you've opened all your tabs at least once" may be a bit too much 😆 
- You can do this for all components apart from the tree (unless you rewrite the way the tree works, in which case you will still need to prefetch the data for the tree), otherwise any model interactions such as hide/highlighting zoom to object will not work. 

Your mac is probably overheating due to the unity game, less about the data around the game (as the data are relatively small).

simplypixi:
> @simplypixi I totally see where you're coming from and I had a chat with Pavol about it. In the end we still think we need the data prefetched (The UI doesn't need to be pre-rendered, and images are less important too, if you think you can decouple that ), for 3 main reasons:
> 
> * Users are very used to CAD softwares, where they load something, go get a coffee and once everything is loaded they expect everything to be there without much further load time (application style of experience)
> * There are use cases where the user has very sporadic connections/even potentially in offline mode, where they need access to certain functionality in a read only kind of situations (e.g. us Demoing with a very dodgy client wifi, or clients going into underground tunnels wanting to see their models on site) . to tell them that "you can still do it, but make sure you've opened all your tabs at least once" may be a bit too much 😆
> * You can do this for all components apart from the tree (unless you rewrite the way the tree works, in which case you will still need to prefetch the data for the tree), otherwise any model interactions such as hide/highlighting zoom to object will not work.
> 
> Your mac is probably overheating due to the unity game, less about the data around the game (as the data are relatively small).

Ok! ;)

simplypixi:
@carmenfan I've added fixes. ;)

> Can we make it (and this is true for all panels) to initialise at the point of the viewer being loaded instead of at the first time it is visible? Reason being is that we very often pre-load the model to ensure there's no connection problems during a demo. and as this list increases or for more complicated tabs (like tree) it will take a lot longer to load.
Done

> You're still doing a post and put to create a new view. It should only do a post when the user finishes the edit and hit save.
Done

> Is it possible to keep the state when you hide and unhide?
Done

I've changed a bit the way of loading initial data, so please verify the others panels too. I checked them, but I think you know them better than me.

Moreover, I noticed that, during rendering the model, the panels are not fully responsive. They are a bit slow - probably because of rendering, which takes all available resources to handle the model. Maybe the panels should be disabled for that time. What do you think? Instead of that, we could display a kind of loading/rendering checklist.
Here is a mock:
image
The panel will be displayed if the whole process will be ended.

jozefdobos:
> @carmenfan I've added fixes. ;)
> 
> > Can we make it (and this is true for all panels) to initialise at the point of the viewer being loaded instead of at the first time it is visible? Reason being is that we very often pre-load the model to ensure there's no connection problems during a demo. and as this list increases or for more complicated tabs (like tree) it will take a lot longer to load.
> > Done
> 
> > You're still doing a post and put to create a new view. It should only do a post when the user finishes the edit and hit save.
> > Done
> 
> > Is it possible to keep the state when you hide and unhide?
> > Done
> 
> I've changed a bit the way of loading initial data, so please verify the others panels too. I checked them, but I think you know them better than me.
> 
> Moreover, I noticed that, during rendering the model, the panels are not fully responsive. They are a bit slow - probably because of rendering, which takes all available resources to handle the model. Maybe the panels should be disabled for that time. What do you think? Instead of that, we could display a kind of loading/rendering checklist.
> Here is a mock:
> image
> The panels will be displayed if the whole process will be ended.

@simplypixi I'd prefer loading message in the panels instead but whatever you agree on at the end just don't put a big modal window in the middle of the screen while the model is loading as users need to be able to interact with the model even if not fully finished yet. This is because many have slow connection and the 3D models are massive.

carmenfan:
@simplypixi 
Thanks for the fixes! I'll take a look 😄 

> Moreover, I noticed that, during rendering the model, the panels are not fully responsive. They are a bit slow - probably because of rendering, which takes all available resources to handle the model. Maybe the panels should be disabled for that time. What do you think? Instead of that, we could display a kind of loading/rendering checklist.

Yes it would be the case when the models are loading and this is a known issue. I don't think we should disable the panels though as Jozef stated. I'd say leave it for now (for the purpose of this PR at least). I'll talk to Pavol and see what he thinks

simplypixi:
> @simplypixi
> Thanks for the fixes! I'll take a look 😄
> 
> > Moreover, I noticed that, during rendering the model, the panels are not fully responsive. They are a bit slow - probably because of rendering, which takes all available resources to handle the model. Maybe the panels should be disabled for that time. What do you think? Instead of that, we could display a kind of loading/rendering checklist.
> 
> Yes it would be the case when the models are loading and this is a known issue. I don't think we should disable the panels though as Jozef stated. I'd say leave it for now (for the purpose of this PR at least). I'll talk to Pavol and see what he thinks

Sure. It was just a proposal. ;)

simplypixi:
@carmenfan Fixed ;)

PR #1299, opened 2018-12-19
Issue 1268: Migration - Screenshot dialog

### LINKED ISSUE #1268: Migration: Screenshot dialog
### Description
Screenshot dialog component for Issues/SafetiBase panel in viewer.
Design: https://xd.adobe.com/view/8a6b23f1-36b6-4419-5fe3-78a60d7ee17d-fc5a/

### Tasks
- [x] colour picker in screenshot tools

simplypixi:
This fixes #1268

#### Description
Rewritten scrrenshot dialog.

I've added extra tool - Clear. How does it work? It's clearing canvas (e.g. if a user totally changes his mind) instead of opening screenshot dialog again.


frontend/components/viewer/js/viewer.service.ts
```diff
@@ -322,12 +322,18 @@ export class ViewerService {
...
-	public getScreenshot(promise) {
-		if (promise) {
-			this.initialised.promise.then(() => {
-				this.viewer.getScreenshot(promise);
-			});
-		}
+	public async getScreenshot() {
+		await this.initialised.promise;
+
+		// This is anti-pattern, but first we need to modify unity-util to use proper solution
```
simplypixi: @carmenfan I can modify `unity-util` if you explain to me how can we generate new `.unityweb` file.

carmenfan: @simplypixi Whatever that's generated in .unityweb can't be changed, as that's auto-generated code by unity (i.e. we don't have direct control over the code in that and it will get overwritten everytime we bake the game)

For screenshots, we need a promise (or some form of callback) because we can only take a copy of the render buffer after the next frame. As as it stands, it takes a defer object when the web request a screenshot. Then the unity game will call [this function](https://github.com/3drepo/3drepo.io/blob/master/frontend/globals/unity-util.ts#L321) with the screenshot when it has one to give back.

There's definitely better ways to do this instead of sending a defer object, but the flexibility of the refactoring is limited to the interactions within unity-utils unfortunately.

What's your plan on that?

carmenfan:
> I've added extra tool - Clear. How does it work? It's clearing canvas (e.g. if a user totally changes his mind) instead of opening screenshot dialog again.

Sounds like a good idea. I've made a gif for the business guys for some feedback.

How difficult would it be to add the ability to add text/basic shapes? (like a simple MS paint)

frontend/components/viewer/js/viewer.service.ts
simplypixi: Ok, I understand. ;)

The current solution is dirty, but it behaves exactly like `$q.defer`, which you used before. It's the same solution implemented in a non-angular way.
I think, in that case, is better than using callbacks to achieve a similar result. So I propose to leave this as it is. Is it ok for you?

carmenfan: @simplypixi what do you think about getting the unity to return a promise instead?

The one complaint we've had with people trying to integrate our stuff is that "our promise isn't the standard promise" on the unity-utils api. So I think if we return a promise like a "standard promise" instead of making the user create one it might be a better approach?

simplypixi:
> > I've added extra tool - Clear. How does it work? It's clearing canvas (e.g. if a user totally changes his mind) instead of opening screenshot dialog again.
> 
> Sounds like a good idea. I've made a gif for the business guys for some feedback.

ok ;)

> How difficult would it be to add the ability to add text/basic shapes? (like a simple MS paint)

It's not difficult but time-consuming, so it strongly depends on the scope. If you thought only about putting text and shapes (with custom sizes) it'll be about extra 4 days of work.
If these elements should be moveable/resizable, it's a good moment to find some extra react component or 2d library. I found a few react components with painting abilities, but each of them has some limits e.g no erasing, permanent layout etc. So, in this case, the best solution is creating a component, which will be dedicated to your app using an external library.

Can we back to this topic tomorrow? I'll do some research. ;)

carmenfan:
> 
> 
> > > I've added extra tool - Clear. How does it work? It's clearing canvas (e.g. if a user totally changes his mind) instead of opening screenshot dialog again.
> > 
> > 
> > Sounds like a good idea. I've made a gif for the business guys for some feedback.
> 
> ok ;)
> 
> > How difficult would it be to add the ability to add text/basic shapes? (like a simple MS paint)
> 
> It's not difficult but time-consuming, so it strongly depends on the scope. If you thought only about putting text and shapes (with custom sizes) it'll be about extra 4 days of work.
> If these elements should be moveable/resizable, it's a good moment to find some extra react component or 2d library. I found a few react components with painting abilities, but each of them has some limits e.g no erasing, permanent layout etc. So, in this case, the best solution is creating a component, which will be dedicated to your app using an external library.
> 
> Can we back to this topic tomorrow? I'll do some research. ;)

Sure :) I think if it's going to take 4 days then it's probably better to leave it as a future project. But let me know what you found :)

simplypixi:
> * Can we make the drop down for draw tool size smaller as it's currently taking up the whole panel height?

Yes, I forgot about that. ;)

simplypixi:
> > * Can we make the drop down for draw tool size smaller as it's currently taking up the whole panel height?
> 
> Yes, I forgot about that. ;)

Fixed ;)

frontend/components/viewer/js/viewer.service.ts
simplypixi: @carmenfan It's a definitely good idea. ;) I wanted to do that, but I understood that we can not modify this file because is generated by unity.

carmenfan: You can modify unity-utils (in `frontend/globals/unity-util.ts`), you just can't modify anything in `frontend/unity` folder 😃

simplypixi: Ohh... 🤦‍♂️ Awesome! I'll change it tomorrow morning. :D

simplypixi: Done ;)

simplypixi:
@carmenfan As I promised, I did some research about extending "screenshot dialog" capabilities. If you plan to add more features to this component I suggest to use some external library e.g:
- [two.js](https://two.js.org/examples/)
- [pixi.js](http://www.pixijs.com/) - we have a huge experience with this one ;)
- [konva](https://konvajs.github.io/)

carmenfan:
> @carmenfan As I promised, I did some research about extending "screenshot dialog" capabilities. If you plan to add more features to this component I suggest to use some external library e.g:
> 
> * [two.js](https://two.js.org/examples/)
> * [pixi.js](http://www.pixijs.com/) - we have a huge experience with this one ;)
> * [konva](https://konvajs.github.io/)

Sounds good. Thanks  for taking the time to have a look!
We will add that to our backlog :)

simplypixi:
@carmenfan I think, it's ready to review. ;)

@dleszcz Could you check this PR too?

simplypixi:
> @simplypixi I can't build this branch, any ideas?
> image

@carmenfan Everything works properly on my side :/. While ago I pushed some changes so please pull the latest commits and try again. If it will not work, I'll try to investigate the issue.

carmenfan:
@simplypixi pulled the latest and it still didn't work... In the end I got it working when I removed my node_modules & yarn.lock and rebuild!

Looks good to me, can merge if you guys are happy.

PR #1329, opened 2019-01-15
ISSUE #1328 - Changed session storage from filesystem to db

### LINKED ISSUE #1328: change session persistence from filesystem to database
### Description
Change the session persistence to a more scalable and secure one

### Goals
- [x] Login and logout should work as usual

### Tasks
- [x] Choose a package for using mongodb as persistnece in express-session

sanmont3drepo:
This fixes #1328


backend/handler/getDBUri.js
```diff
@@ -0,0 +1,39 @@
+/**
```
carmenfan: Do we really have to move this out?

The handler folder is suppose to hold handlers to external services (e.g. mongodb, S3 etc), it's a bit weird to have a file called `getDBUri`, which should technically reside in `db.js`

backend/services/session.js
```diff
@@ -37,7 +37,10 @@ module.exports.session = function(config) {
 			path: "/",
 			secure: config.using_ssl
 		},
-		store: new FileStore()
+		store: new MongoDBStore({
```
carmenfan: Would it be better to move MongoDBStore inside `handler/db.js` ?

Then you don't need to expose getDBUri() and MongoDBStore here.

I feel like those info are a lot closer to the dbHandler than sessions and thus it should stay overthere, with a function to get a DBStore.

backend/package.json
```diff
@@ -34,6 +34,7 @@
     "async": "1.5.2",
     "body-parser": "1.15.0",
     "compression": "1.6.1",
+    "connect-mongodb-session": "^2.0.6",
```
carmenfan: lock down the version to prevent random breakage when they upgrade

backend/services/session.js
sanmont3drepo: The mongodbstore is a package for using mongo as a storage for sessions it doesnt serve other purpose and wouldnt be used in any other different place ,  so I think is closer to sessions.

sanmont3drepo: Moved out to db.js

PR #1347, opened 2019-01-25
Issue 1345: GIS bugs

### LINKED ISSUE #1345: GIS Bugs
- [x] **Number values are being stored as strings**
This is what the payload looks like if I update all the values within survey point and hit save:
image
I also  ran into problems with my temporary fix trying to flip the signs of the numbers because it was a string. I think it's best to make sure the data model has the right type instead of leaving it as string until we need to make sure it's converted?

- [x] **Survey points being sent for mapInitialise is not converted**
I made a [temporary fix](https://github.com/3drepo/3drepo.io/commit/6a0af6fd1aa4701ebd3d91b77f5e9f16f92ce252) in #1342 to convert it just before calling `mapInitialise` as a temporary workaround. Instead of having the data model in DirectX format (what the clients see) and convert it when you need to, it should store it in OpenGL format(what we use internally, the converted one) and only display it as DirectX format. Otherwise we'll end up with this kind of mess a lot, and it will not be consistent throughout the platform (e.g. camera views, pin position, they're all in OpenGL format).

- [x] **Can we also make sure we spell things in UK spellings? (eg `Initialise` instead of `Initialize`)**

simplypixi:
This fixes #1345


simplypixi:
@carmenfan 
Please take a look at this if you have a while.

carmenfan:
@simplypixi View branch was already merged into staging, then I had this 3.1 bug branch which reverted it, that got merged into staging, then I reverted the reverted. So as of now it should be merged in (but the temp fix is also merged in)

Issue #1344, opened 2019-01-24
Bugs from View migration


carmenfan:
### Bugs
- [x] [Cannot create views if the views component has no entry](https://github.com/3drepo/3drepo.io/issues/1342#issuecomment-457215895)
- [x] [When you exit an existing model and enter a different mode, the last model is being loaded](https://github.com/3drepo/3drepo.io/issues/1342#issuecomment-457178131)

carmenfan:
@simplypixi Regarding [this](https://github.com/3drepo/3drepo.io/issues/1342#issuecomment-457178131). What I observe is that the culprit is the new function `mapPropsToThis`

First model it loads it's fine (let's call this modelA), everything works fine.
When you exit the viewer and select another model from teamspace (let's call this modelB), I see that `mapPropsToThis` triggers `handleSettingsChange` with modelA again, then it triggers it with modelB. The 2 commands seems to trigger pretty much simultaneously which causes some race conditiosn within the unity viewer. all in all `handleSettingsChange` should only be called once with modelB settings.

Side note:
`mapPropsToThis` is a pretty terrible naming, can we change it? It's also kind of inaccurate for maps to depict what model to be loading (perhaps it's a temporary setup before it's fully migrated?) The settings in question is the model settings, and it should really be the model component retrieving this and passing it to whoever need it.

simplypixi:
> Side note:
> `mapPropsToThis` is a pretty terrible naming, can we change it? It's also kind of inaccurate for maps to depict what model to be loading (perhaps it's a temporary setup before it's fully migrated?) The settings in question is the model settings, and it should really be the model component retrieving this and passing it to whoever need it.

I think this name is good, because is created in react/redux stack naming convention (mapStateToProps, mapDispatchToProps). We wanted to emphasize that this method bind values from React store to model component. And yes, is a temporary method.
Nevertheless, we can change the name if you have a better idea. :)

carmenfan:
> > Side note:
> > `mapPropsToThis` is a pretty terrible naming, can we change it? It's also kind of inaccurate for maps to depict what model to be loading (perhaps it's a temporary setup before it's fully migrated?) The settings in question is the model settings, and it should really be the model component retrieving this and passing it to whoever need it.
> 
> I think this name is good, because is created in react/redux stack naming convention (mapStateToProps, mapDispatchToProps). We wanted to emphasize that this method bind values from React store to model component. And yes, is a temporary method.
[...]

PR #1352, opened 2019-01-28
Issue 1259/1344: Views

### LINKED ISSUE #1259: Migration: Views
**Description**
Migration of Views panel.

**Tasks**
Create a new Views component

**Extras**
design: https://xd.adobe.com/view/8a6b23f1-36b6-4419-5fe3-78a60d7ee17d-fc5a/screen/ee904147-e7a4-4eb1-93d1-c809c00dcf5d/Views

simplypixi:
This fixes #1259 and #1344.


simplypixi:
@sanmont3drepo , @carmenfan I removed re-loading previous model, but I noticed two bugs, which appear on staging too:

1. When I load Lego Architecture model, go to Teamspaces and then load Lego Federation model the viewer throws some errors related to the pins. It can't to remove the pins, because they exists in pins map but not in unity context.
image

2. When I go to Teamspace while the model is loading, the viewer doesn't cancel request/model rendering, so it's appears in the Teamspaces background. I think Pavol reported exactly the same bug, but it seems to be an issue on viewer side (e.g $destroy in viewer.component never calls).

What do you think about it?

carmenfan:
> @sanmont3drepo , @carmenfan I removed re-loading previous model, but I noticed two bugs, which appear on staging too:
> 
> 1. When I load Lego Architecture model, go to Teamspaces and then load Lego Federation model the viewer throws some errors related to the pins. It can't to remove the pins, because they exists in pins map but not in unity context.

I'm guessing this only happens on initialisation? It's probably worth taking a look during issue migration, as the pins should be consistent with unity provided the model has started loading. I think I've seen this in the pass and it doesn't really affect the functionality as the pins were loaded after the model started loading.

> 2. When I go to Teamspace while the model is loading, the viewer doesn't cancel request/model rendering, so it's appears in the Teamspaces background. I think Pavol reported exactly the same bug, but it seems to be an issue on viewer side (e.g $destroy in viewer.component never calls).
> 
> What do you think about it?

This is a known issue. It's not to do with the component not getting destroyed. It's API requests that has been fired in unity, and before they come back reset has been called. So the whole dataset has been wiped in Unity, but then we get the callback.

We know this happens for some time now and if we were too address it we need to change i in Unity, so I wouldn't worry too much about it. (as long as it's clearing the viewer if you exit after the model has been loaded

simplypixi:
@carmenfan @sanmont3drepo @dleszcz Rebase and ready for review ;)

frontend/components/index.ts
```diff
@@ -60,6 +60,4 @@ import './utils/js/utils.filter';
 import './viewer/js/multi-select.service';
 import './viewer/js/viewer.component';
 import './viewer/js/viewer.service';
-import './viewpoints/js/viewpoints.component';
```
sanmont3drepo: Should delete the viewpoints angular component folder also.

frontend/modules/viewer/viewer.sagas.ts
```diff
@@ -30,7 +31,7 @@ export function* waitForViewer() {
 		const ViewerService = yield getAngularService('ViewerService') as any;
 		yield ViewerService.initialised.promise;
 	} catch (error) {
-		console.error(error);
+		yield put(DialogActions.showErrorDialog('initialise', 'viewer', error));
```
sanmont3drepo: - If the error is an enpoint error should use the DialogActions.showEndpointErrorDialog action instead. If it is a string its ok. Please check every call.

frontend/routes/viewer/components/views/components/viewItem/viewItem.component.tsx
```diff
@@ -0,0 +1,119 @@
...
+	viewpoint: any;
+	active: boolean;
+	editMode: boolean;
+	teamspace: string;
+	modelId: string;
+	onCancelEditMode: () => void;
+	onSaveEdit: (values) => void;
+	onDelete?: (event) => void;
+	onOpenEditMode?: () => void;
+	handleClick?: (viewpoint) => void;
```
sanmont3drepo: - Can we rename **handleClick** to **onClick** ?

simplypixi: Yes, definitely. I don't know why it was named like this. :/

frontend/components/index.ts
simplypixi: done

simplypixi:
> I think there is a bug with the resizing of the card.
> When you have an empty list of views the height is small and stays small until you refresh the page.
> 
> Now:
> resizing bad
> Empty list:
> resizing bad2
> 
> How it should look:
> resizing ok

Fixed ;)

sanmont3drepo:
I think everything is ok, can I merge this @carmenfan ?

carmenfan:
Let me put this through the @paloknapo test first :)

carmenfan:
@simplypixi 
@paloknapo says everything works, but the icons should be in one row:
image
image

Also the icons should be in grey, like this:
image

simplypixi:
> @simplypixi
> @paloknapo says everything works, but the icons should be in one row:
> image
> image
> 
> Also the icons should be in grey, like this:
> image

I changed these icons on Risks branch about a week ago. I think this PR was created before the views got new icons' look. Move this fix here? ;)

simplypixi:
> > @simplypixi
> > @paloknapo says everything works, but the icons should be in one row:
> > image
> > image
> > Also the icons should be in grey, like this:
> > image
> 
> I changed these icons on Risks branch about a week ago. I think this PR was created before the views got new icons' look. Move this fix here? ;)

Fixed ;)

simplypixi:
> @simplypixi
> 
> On the Angular side, when there's a deleted group (or closed issues / resolved risks) happened on another client and you get the update via the chat service, We get this little message before the item disappears. Can we do the same for views?
> 
> 

Sure. ;) Should it look the same?

simplypixi:
@carmenfan Fixed and rebased ;)

carmenfan:
@simplypixi looking good :)

But this seems to happen still?

> Selected view should unselect when you add a new view:
>

simplypixi:
> @simplypixi looking good :)
> 
> But this seems to happen still?
> 
> > Selected view should unselect when you add a new view:
> > 

Fixed ;)

PR #1357, opened 2019-01-30
Issue 1356: Remove cleaning dist before production build

### LINKED ISSUE #1356: Maintenance page does not stay up whilst we are building the front end
@simplypixi 
is there any reason why we need to do a clean before we build?
image

When we upgrade the servers with new releases, we put the server in maintenance mode (running a config where MaintenanceMode:true is set) , which will give the users a maintenance page if they were to visit the site during that time.

We then upgrade the code, build the frontend then restart the servers.

Problem is, because the `dist` folder is removed before the pages are built (and at the moment it takes a good 5-10mins), the maintenance page stop showing for a prolong period of time.

Would it be possible to move the removal just before you copy over the built pages (or do something of the similar effect to reduce the down time?)

simplypixi:
This fixes #1356 

@carmenfan I've changed webpack's config. It wasn't necessary to clean dist before doing the production build. I've added by mistake the cleaning to shared config. On the other hand, I have one idea to improve the building process a bit. From our experience, it's better to serve the maintenance page to the users from another app instance. It safer because:
- You don't serve unnecessary code - just maintenance page, which is usually plain HTML with some CSS
- You have more control during the "upgrade" process. If something messes up, you have still active maintenance mode and you able to fix the issues without a big impact for the users.
- You don't have to restart the servers

But it's only suggestion and probably a topic for your DevOps team. :)


carmenfan:
> From our experience, it's better to serve the maintenance page to the users from another app instance.

Agreed, but unfortunately we are stuck with it for now as it's not a super high priority thing to change 😆

PR #1358, opened 2019-02-01
Issue 1254 - Fix Viewpoint Axis 

### LINKED ISSUE #1254: JSON file Download
### Description
This is Issue is associated with the following Issues
#1252 and #1234 

This is a new issue improving the work done on the #1234 , the axis on the downloaded JSON file is incorrect and requires the values not to be mirrored.

Instead of [X, Y, Z] the axis is [Z,Y,Z].

Also when the user filters the Issues, groups or risks list the filtered results need to downloaded instead of the entire list. 

JSON file for issues contains mirrored XYZ Axis. 
"Position": [ X, Z, Y ]
X Values from JSON:
**3841
3819
-3835**
Y Values from JSON: 
**-3851
3828
3829**

Y values are mirrored

Easy fix with y = number * (-1)
x **3841**     y **3851**
x **3819**     y **-3828**
x **-3835**   y **-3829**

### Goals
- [ ] Fix X,Y,Z axis errors.
- [ ] Ensure Filtered results are download if applicable and not all results.

### Tasks

- [ ] A new Group panel
- [ ] List group component
- [ ] edit group component
- [ ] new API end point to remove groups
- [ ] New database schema for groups (present it here)

Corrected Values
dev pins
revit pins
json position

niblux:
This fixes #1254 

#### Description
The JSON file download feature , was incorrectly outputting the axis viewpoints when download a JSON file, downloading output as [X, Z, Y] instead of [X, Y, Z ].

- Find viewpoint in JSON Response
- Revert the axis to the correct position. 
- Apply this change to the JSON response.

#### Test cases
The user should be able to download a JSON file and any viewpoint axis should be in the correct position.


niblux:
I made the requested change :) , merged with staging and pushed.

backend/routes/group.js
```diff
@@ -257,11 +257,13 @@ const getDbColOptions = function (req) {
 function listGroups(req, res, next) {
 
 	const dbCol = getDbColOptions(req);
-	const place = utils.APIInfo(req);
+	const place = utils.APIInfo(req);	
 
 	let groupList;
 	if (req.params.rid) {
 		groupList = Group.listGroups(dbCol, req.query, null, req.params.rid);
+	} else if (req.query.ids) {
```
carmenfan: This needs to use the same `listGroups` function. As it stands you're ignoring the revision information if there's ID given,

List Groups do more than just getting the group information back and for consistency it should do it the same way.

You should add the id list to the list group method like how risks/issues are doing it

backend/routes/risk.js
```diff
@@ -219,12 +219,17 @@ function listRisks(req, res, next) {
...
+	let ids;
 	let findRisks;
+	
+	if (req.query.ids) {
+		ids = req.query.ids.split(",");
+	}
 
 	if (req.params.rid) {
-		findRisks = Risk.findRisksByModelName(dbCol, req.session.user.username, null, req.params.rid, projection);
+		findRisks = Risk.findRisksByModelName(dbCol, req.session.user.username, null, req.params.rid, projection, null, ids);
```
carmenfan: Why is there an extra `null` here?

backend/routes/issue.js
```diff
@@ -479,12 +505,17 @@ function listIssues(req, res, next) {
...
+		findIssue = Issue.findIssuesByModelName(dbCol, req.session.user.username, null, req.params.rid, projection, null, ids);
 	} else {
-		findIssue = Issue.findIssuesByModelName(dbCol, req.session.user.username, "master", null, projection, null, null, req.query.sortBy);
+		findIssue = Issue.findIssuesByModelName(dbCol, req.session.user.username, "master", null, projection, null, ids, req.query.sortBy);
 	}
 
 	findIssue.then(issues => {
+
+	if (req.query.convertCoords) {
+		sortViewpointAxis(issues);
```
carmenfan: that's not a very clear function naming, it sounds as if you're sorting the order.

backend/routes/issue.js
```diff
@@ -460,9 +460,35 @@ function updateIssue(req, res, next) {
...
+			viewpoint[key] = convertPoint(viewpoint[key]);
+		}
+	});
+	return viewpoint;
+}
+
+	
+function sortViewpointAxis(issues) {
+	issues.forEach((issue) => {
+		convertViewpoint(issue.viewpoint);
```
carmenfan: It's more than viewpoints. Pins are positions on the root object. (i.e `issue.position`)

You'd also have to convert the clipping plane normal.

backend/routes/group.js
niblux: I have moved the id lookup functionality into listGroups, to keep it the same as risks and groups.

backend/routes/risk.js
niblux: Fixed.

backend/routes/issue.js
niblux: Renamed to `changeIssueAxis();` i think that is more suited as it is changing the axis points within an issue. 

What do you think 👍 or 👎 ?

backend/routes/issue.js
niblux: I now loop through the root object and then the viewpoint object , ensuring all points are addressed, including the clipping plane normal. :)

PR #1400, opened 2019-02-25
Issue 1379 - Add regex to Project name 

### LINKED ISSUE #1379: Users should not be able to create project without a name
A user has managed to create a project without a name, considering this is what we use on endpoints, the project doesn't function properly. 

We should check and disallow this. 

Projects should be between 1-120 characters, restricting anything that would break URL formatting.

carmenfan:
This fixes #1379

#### Description
- Add regex to validate project names on add project/update project

#### Test cases
For cases of create new/update projecT:
  - Project name cannot be empty
  - Project name cannot end with a space
  - Project name cannot contain these characters: `/?=#+`
  - length of project name has to be between 1-120 (UI is blocking on 1, for some reason)

### DIFF
--- a/backend/models/project.js
+++ b/backend/models/project.js
@@ -35,6 +35,11 @@
 		}]
 	});
 
+	function checkProjectNameValid (project) {
+		const regex = "^[^/?=#+]{1,120}$";
+		return project && project.match(regex);
+	}
+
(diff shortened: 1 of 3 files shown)


carmenfan:
@Charence updated

PR #1403, opened 2019-02-27
ISSUE #1381 only clear password fields upon submit

### LINKED ISSUE #1381: Sign up form clears the form even with an error
### Description
If the sign up request comes back with an error, the entire form is cleared

### Steps to replicate
- fill in the sign up form with a invalid user name (e.g. "i-am-a-user")
- click on Sign up
- Error message will pop up saying invalid user name.
- The form is completely cleared so instead of modifying the user name, the user will have to write the whole thing out again.

carmenfan:
This fixes #1381 

#### Description
- Only clear the password fields upon submit, avoiding the user having to rewrite the whole form if they put in an erroneous input
- I removed some Refs that doesn't seem to be in used (@simplypixi @dleszcz - please confirm if that makes sense?)
- I removed resetForm since it's no longer used (@simplypixi @dleszcz - please confirm if that makes sense?)

#### Test cases
- The form no longer clears if API returns an error
- Successful sign up should be uneffected

### DIFF
--- a/frontend/routes/signUp/signUp.component.tsx
+++ b/frontend/routes/signUp/signUp.component.tsx
@@ -114,10 +114,6 @@ export class SignUp extends React.PureComponent<IProps, IState> {
 	};
 
 	public reCaptchaWrapperRef = React.createRef<any>();
-	public listRef = React.createRef<any>();
-	public inputRef = React.createRef<any>();
-	public innerRef = React.createRef<any>();
-	public myInnerRef = React.createRef<any>();
 
 	public componentDidMount() {
 		const defaultCountry = clientConfigService.countries.find((country) => country.code === 'GB');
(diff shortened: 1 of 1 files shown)


frontend/routes/signUp/signUp.component.tsx
```diff
@@ -129,16 +125,12 @@ export class SignUp extends React.PureComponent<IProps, IState> {
 	public handleSubmit = (values, form) => {
 		const data = omit(values, 'username', 'emailConfirm', 'passwordConfirm', 'termsAgreed');
 		this.props.onRegister(values.username, data);
-		this.resetForm(form);
-	}
-
-	public resetForm = (form) => {
-		form.resetForm();
+		values.password = '';
```
simplypixi: @carmenfan Sorry but you can't do that in this way. It probably works, but only because of object's mutability.
Instead, you can use `form.setFieldValue: (field: string, value: any, shouldValidate?: boolean)` or `form.resetForm: (nextValues?: Values)` to achieve effect, that you want. ;)

Let me know if these solutions won't work.

simplypixi:
> * I removed some Refs that doesn't seem to be in used (@simplypixi @dleszcz - please confirm if that makes sense?)
👍 

> * I removed resetForm since it's no longer used (@simplypixi @dleszcz - please confirm if that makes sense?)

I've added the review. :)

frontend/routes/signUp/signUp.component.tsx
carmenfan: Thanks! I tried for a while to find alternatives to do it, looked at Formik's Documentation and couldn't find their API for forms so I kind of gave up and see what you'd say 😆 

This is now changed.

PR #1408, opened 2019-03-01
ISSUE 1407 - Comment permissions flag

### LINKED ISSUE #1407: Add comment permissions flag to new comment form
NewCommentForm is missing a canComment flag.

Comment text field should be disabled if canComment is false. Panels that use NewCommentForm will need to set the flag.

@simplypixi @dleszcz 
This will affect Issues

Charence:
This fixes #1407 

#### Description
- Add canComment permission flag to NewCommentForm
- Update RiskDetails to make use of flag

#### Test cases
- Risks card should work exactly the same as before
- New comment area should be disabled if canComment not true
- To make the footer visible, apply the following changes:
```
--- a/frontend/routes/viewer/components/risks/components/riskDetails/riskDetails.component.tsx
+++ b/frontend/routes/viewer/components/risks/components/riskDetails/riskDetails.component.tsx
@@ -208,9 +208,6 @@ export class RiskDetails extends React.PureComponent<IProps, IState> {
                                screenshot={this.props.newComment.screenshot}
                                viewpoint={this.props.newComment.viewpoint}
                                innerRef={this.commentRef}
-                               hideComment={true}
-                               hideScreenshot={!this.isNewRisk}
-                               hidePin={!this.isNewRisk}
                                onTakeScreenshot={this.handleNewScreenshot}
                                onChangePin={this.handleChangePin}
                                onSave={this.handleSave}
@@ -227,7 +224,7 @@ export class RiskDetails extends React.PureComponent<IProps, IState> {
                                        {this.renderPreview(this.props.risk)}
                                        {this.renderLogs(logs.length)}
                                </ViewerPanelContent>
-                               {this.renderFooter(!this.riskData._id)}
+                               {this.renderFooter(true)}
                        </Container>
                );
        }
```

### DIFF
--- a/frontend/routes/teamspaces/components/tooltipButton/tooltipButton.component.tsx
+++ b/frontend/routes/teamspaces/components/tooltipButton/tooltipButton.component.tsx
@@ -24,15 +24,16 @@ interface IProps {
 	Icon: React.ComponentType;
 	color?: string;
 	action?: (event) => void;
+	disabled?: boolean;
 }
 
 export const TooltipButton = (props: IProps) => {
-	const { label, action = null, Icon, color = 'inherit' } = props;
+	const { label, action = null, Icon, color = 'inherit', disabled = false } = props;
 	const iconProps = { color, fontSize: 'small' } as any;
 
 	return (
 		<Tooltip title={label}>
-			<IconButton aria-label={label} onClick={action}>
+			<IconButton aria-label={label} onClick={action} disabled={disabled}>
 				<Icon {...iconProps} />
 			</IconButton>
 		</Tooltip>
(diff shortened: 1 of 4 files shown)


carmenfan:
@simplypixi @dleszcz  This change would affect issues (We have to ensure only people who are allowed to commetn can comment)

Please take a look!

Charence:
image

Not sure why the Pin button doesn't go the same shade of grey as Screenshot when disabled.

Charence:
I was getting warnings about disabled buttons inside tooltips, so I added a span as suggested by this example: https://material-ui.com/demos/tooltips/

Not sure if we want it done this way, which allows the tooltip to still appear for disabled buttons.

sanmont3drepo:
LGTM!

Issue #1387, opened 2019-02-21
Smart Groups - details/new


dleszcz:
- [x] Add/Edit form
- [x] Display selected group logic
- [x] Set color (ColorPicker) logic
- [x] Reloading logic
- [x] Save logic
- [x] Listening on click on object (Selection mode)
- [x] Permissions
- [x] Creating criteria component
   - [x] fields autocomplete (fetching used fields included)
   - [x] Creating a constant with possible operations + showing list
   - [x] Component for adding many fields
   - [x] Add criteria logic
   - [x] Options menu (copy, paste, deselect, clear)
- [x] Edit existing criteria
- [x] Switching between groups (header)
- [x] Chat Events

### Extra notes
General
- Slight change in the layout of the group details
- A new group type on ui : Selection (normal), by criteria (smart)
- UI changes depending on group type
- do not add bin/attachment icon on edit group tab.
- add previous/next buttons on the top header like issues/risks
- “reset” button  for selection type: it should re-highlight what was last saved
- for criteria: it should highlight base on what are the current criteria
- Upon save, smart groups should update highlight with the latest criteria.
- Reuse colour picker for groups colour.

Criteria display:
- Criteria area contains existing rules as chips, unlike issue/risk search, but user cannot type new ones directly.
- User selects a chip -> rule info gets filled in the grey area (edit rule mode)
- User can deselect a chip by clicking on the background of the criteria area, which empties the values in the grey area (new rule mode)
- There can only be 1 rule per field.
- See operations tab on sketch  to see how the chips should be formatted
- long chips should shorten with “...”
- User can copy rules clicking on “copy” in hamburger button, but instead of copy the chip string, it should copy the json array of criterias.
- User can paste JSON array that will populate the chips area with new rules (paste should be in hamburger menu, with a pop up text box to take in the json)

Rule creation edit area:
- Field names are based on a API endpoint that gives you the list of metadata fields, but user can also put in something that doesn’t exist.
- Field textbox should function similar like user search in user management.
- operation is a determined list as shown
- Hit confirm button to add new criteria in chips area
- grey area should be scrollable separately to the whole tab.

Operations characteristics:
- Exists/ not exists should not have a value textbox.
- text operations should start with 1 value text box, with the ability to add more multiple text boxes, which can be deleted (apart from the last one)
- Regex Only allow 1 textbox, with an “i” button which links to the blog about it.
- Greater/smaller should only have 1 text box.
- range values should only have 2 text box in the same line.

[...]

Issue #1386, opened 2019-02-21
Smart Groups - list


dleszcz:
- [x] Panel Groups 
    - [x] filters 
    - [x] menu 
    - [x] add button 
    - [x] counting in footer
- [x] Place for groups in Redux store
- [x] List item (UI) + Fetching groups from BE
- [x] Select / Multiselect group items
- [x] Colouring group (item level) / Colouring all groups (menu level)
- [x] Download JSON file
- [x] Removing group (item level) / Removing all groups (menu level)
- [x] Isolate group

Issue #1395, opened 2019-02-22
Login screen now has scrollbars appearing


simplypixi:
**Login screen now has scrollbars appearing**

Caused by #1273 #1343 

![screenshot 2019-02-21 15 48 10](https://user-images.githubusercontent.com/1438332/53183588-80753200-35f3-11e9-8744-d602ba233d01.png)

_Originally posted by @Charence in https://github.com/3drepo/3drepo.io/issues/1391#issuecomment-466071674_

PR #1410, opened 2019-03-04
Issue 1387: Smart groups migration

dleszcz:
- Groups list https://github.com/3drepo/3drepo.io/issues/1386
- Group details https://github.com/3drepo/3drepo.io/issues/1387
- Remove Angular code


simplypixi:
> * Delete is not removing the group from list until you refresh the page
> * Updated Groups are not being updated until refreshed
> * New group is not displayed on Group list until refreshed
@carmenfan Did you have ChatService enabled? Because response of create, update and delete are handle by this service. Should we handle these actions in two different ways? I mean, using http response and the chat service.

> Clicking on the rule chip should populate the rule editing area with the existing data
@dleszcz Did you change something here? It was working on the morning. :(

> None of the new groups are showing the created by user

@carmenfan Are you sure it is not an issue on backend side? I think we reported similar bug today, but we'll check it! ;)

carmenfan:
> @carmenfan Are you sure it is not an issue on backend side? I think we reported similar bug today, but we'll check it! ;)

Does sound likely to be a backend bug since old groups is displaying... I'll have a look!

simplypixi:
> @carmenfan Are you sure it is not an issue on backend side? I think we reported similar bug today, but we'll check it! ;)

@Charence BE returns old data on the response while saving/updating group.

image

image

carmenfan:
@carmenfan Did you have ChatService enabled? Because response of create, update and delete are handle by this service. Should we handle these actions in two different ways? I mean, using http response and the chat service.

I know you deleted this but to answer your question - the user who created the item (and this applies to risks/issues/anything else as well) needs to get it updated by http. Reason being is that we know Chat service does not work for some of our clients due to corporate firewall, so at the very least the states of the app need to be sound without it for the user who's doing the work.

simplypixi:
> > @carmenfan Did you have ChatService enabled? Because response of create, update and delete are handle by this service. Should we handle these actions in two different ways? I mean, using http response and the chat service.
> 
> @simplypixi
> I know you deleted this but to answer your question - the user who created the item (and this applies to risks/issues/anything else as well) needs to get it updated by http. Reason being is that we know Chat service does not work for some of our clients due to corporate firewall, so at the very least the states of the app need to be sound without it for the user who's doing the work.

Got it! ;)

carmenfan:
**It no longer let me save until I alter the Description field:**
image

carmenfan:
@simplypixi 

- The problem where the backend isn't giving you the updated info after a http put should be fixed
- The problem with user name not being shown should also be fixed. You no longer need to send `updatedBy` and `updatedAt` on your HTTP requests as that will be automatically generated based on session info/current time on server. (if you send it, it'll just be ignored)

simplypixi:
> @simplypixi
> 
> * The problem where the backend isn't giving you the updated info after a http put should be fixed
> * The problem with user name not being shown should also be fixed. You no longer need to send `updatedBy` and `updatedAt` on your HTTP requests as that will be automatically generated based on session info/current time on server. (if you send it, it'll just be ignored)

Awesome! ;)

carmenfan:
@simplypixi @dleszcz I've updated the backend so it now returns you sharedIDs on smart group post/put requests. The highlights are now showing :)

3 things:
1. To make this work, the backend needs to know what revision is the user currently viewing. So I've changed the end points to take in revision, i.e. instead of `/:teamspace/:model/groups/:groupid`, it's now `/:teamspace/:model/revision/:rid/groups/:groupid` or `/:teamspace/:model/revision/master/head/groups/:groupid` The old endpoint still works for backwards compatibility, but please change all your end point calls to this new one

2. When I was looking into this I realised groups had the url structure the wrong way round, it was putting `groups` before `revision` so I've now swapped them around (i.e. it was  `/:teamspace/:model/groups/revision/master/head/gid` instead of `/:teamspace/:model/revision/master/head/groups/gid` . I think I've changed all the frontend to the new version now. But please bear this in mind!

3. Now that it's showing the highlights, I found a bug. when you edit a smart group and click save, it is not removing the currently highlighted objects before it applies the new set:

simplypixi:
@carmenfan We have a problem with the backend. Probably after the latest changes, because we have never noticed that before.
It doesn't appear on each model. We noticed that is happening on model: `viewer/subscriptionTest/9929074c-52f7-4ca5-8dbe-02894aec09a4`, when we where logged into `carmen` account. 

It's a memory leak, which is related to `buildRule` function in the group model.
Could you look at that?
image

dleszcz:
@carmenfan @Charence 

One more error. Could you check it, please? It happens after group(normal) update.

screenshot 2019-03-06 at 09 27 09


FYI, Mariusz just updated URLs ;)

carmenfan:
> 
> 
> @carmenfan @Charence
> 
> One more error. Could you check it, please? It happens after group(normal) update.
> 
> screenshot 2019-03-06 at 09 27 09
> 
> FYI, Mariusz just updated URLs ;)

This is fixed!

carmenfan:
> -- EDIT --
> It seems to happen on each model with defined rules - which make sense :P
What do you mean by defined rules? Can you show me what the group looks like?

simplypixi:
> > -- EDIT --
> > It seems to happen on each model with defined rules - which make sense 📦
> 
> @simplypixi
> What do you mean by defined rules? Can you show me what the group looks like?
> 
> EDIT: I think I know what you mean... I managed to reproduce it in the exists rule. Let me know if it's something else!

That's exactly what I meant. ;)

carmenfan:
> 
> 
> > > -- EDIT --
> > > It seems to happen on each model with defined rules - which make sense 📦
> > 
> > 
> > @simplypixi
> > What do you mean by defined rules? Can you show me what the group looks like?
> > EDIT: I think I know what you mean... I managed to reproduce it in the exists rule. Let me know if it's something else!
> 
> That's exactly what I meant. ;)

Also fixed :)

dleszcz:
@carmenfan We just finished fixing your requests :)

carmenfan:
- [ ] `Number of objects` should only be showing what the group objects are, not the number of objects currently being highlighted (i.e. it should only change when user click save):


- [ ] Save button doesn't enable if I change the group name:


- [ ] Save button in smart group doesn't enable/disable like manual selection mode
image


- [ ] We need to disable override when we go into edit group (for that particular group), once the user exits, it should repaint the overrides:

This is what's happening in staging:


- [ ] it should be last update**d**, also can we please have the timestamp similar to the ones in risks?
image

- [ ] Overrides needs to clear when the group is deleted (there also seems to be some errors in the console):


- [ ] As a Viewer, I should not be able to edit existing groups. The description and group type are disabled, but I can still change colour and group name and try to submit my changes:

simplypixi:
@carmenfan I have a small question. Should be the user able to add/select object, while the group is a smart group?

I think about that situation:
image

carmenfan:
> @carmenfan I have a small question. Should be the user able to add/select object, while the group is a smart group?
> 

You should be able change the highlighting in the viewer (reset button will allow the user to see the original state should he needs to!), but upon save, it shouldn't try to save the selected ids if it's on the criterias type, it should only save the rules.

paloknapo:
- [ ] Selecting chips doesn't always update filter fields


- [ ] Paste filters should create chips directly without user's CTRL+V


- [ ] Rename copy "Paste criteria" to "Paste filters"
image

- [ ] Show "Set filed" and "Set operation" in empty form fields
image

- [ ] Please change "Criteria" to "Smart" in Group type
image

- [ ] Please rename Criteria to Filters 
image

- [ ] Display number of objects in list view
image

- [ ] Scroll bar appears without good reason 


- Group type icon doesn't update

simplypixi:
> * Paste filters should create chips directly without user's CTRL+V

@paloknapo @carmenfan We already discussed that. We are not able to implement a solution which behaves like native "paste" (with cross-browser compatibility), because of browser security reasons. We can't get into the clipboard if the user doesn't invoke paste event using a keyboard shortcut. That's why text input appears.

Nevertheless, there is an alternative, we can store data in local/session storage, but you lose cross-browser compatibility.

dleszcz:
Hello @carmenfan,

According to Pavol's request:
> Display number of objects in list view

Are we able to extract this data based on the group object?

```
calculateTotalMeshes(group.objects); // ?
```
Will it always be the correct objects number value?

Because now we get this value on this way:
```
Viewer.getObjectsStatus();
calculateTotalMeshes(objectsStatus.highlightedNodes);
```

simplypixi:
> * 2 filters with same operations are not allowed. If user tries this, Update button is disabled and message is displayed below: "Filter with this operation is already used"
>   image

I think we need to precise it a bit. :)  Do you mean two different filters whether two the same filters?

carmenfan:
> Viewer.getObjectsStatus();
> calculateTotalMeshes(objectsStatus.highlightedNodes);
> ```

@dleszcz This asks the viewer the list of objects that it is currently highlighting, which should be correct if it is done immediately after highlighting the group (assuming you clear the highlgihts before you highlight the group)

This number should *theoretically* be the same as you accumulating the length of `shared_id` arrays withing `group.objects`

So for pavols' request you can just display the total of shared_id entries within the group's data.

dleszcz:
Ok, thanks @carmenfan !

carmenfan:
> @paloknapo @carmenfan We already discussed that. We are not able to implement a solution which behaves like native "paste" (with cross-browser compatibility), because of browser security reasons. We can't get into the clipboard if the user doesn't invoke paste event using a keyboard shortcut. That's why text input appears.
> 
> Nevertheless, there is an alternative, we can store data in local/session storage, but you lose cross-browser compatibility.

Yes, I don't think Pavol read that conversation. Just ignore this :)

dleszcz:
@carmenfan @paloknapo @Charence fixes done

But, I found on more bug:
1. Go to some _smart_ group details
2. Change type to _normal_
3. Save

Backend returns group object with rules array, which causes the group type has not changed after the saving.

carmenfan:
> @carmenfan @paloknapo @Charence fixes done
> 
> But I found on more bug:
> 
> 1. Go to some _smart_ group details
> 2. Change type to _normal_
> 3. Save
> 
> Backend returns group object with rules array, which causes the group type has not changed after the saving.

thanks! I'll look into it!

carmenfan:
@dleszcz 

In the middle of fixing this, I found some problems:

Now that i've enforced that a group can only have either objects or rules but not at the same time, I can't update smart groups, as they're trying to send an empty object array:
image

- Also, I'm seeing override being toggled on everytime I enter the group:


- Finally, Can we make it say `1 object` when it's 1? or say` x object(s)`
image

dleszcz:
> Can we make sure we only send objects or rules?
OK, sure 👍

dleszcz:
@carmenfan I updated smart group without objects array, and then i got this:

Screenshot 2019-03-14 at 08 22 58

simplypixi:
> * Now that i've enforced that a group can only have either objects or rules but not at the same time, I can't update smart groups, as they're trying to send an empty object array:

@carmenfan I'm wondering, why you not save the group type in the database. It's more convenient, isn't it? And more scalable - it'll be easier to add next group type in the future if you will came up with some new idea.
What do you think?

carmenfan:
> @carmenfan I updated smart group without objects array, and then i got this:
> 
> Screenshot 2019-03-14 at 08 22 58

Sorry, that should be fixed

> @carmenfan I'm wondering, why you not save the group type in the database. It's more convenient, isn't it? And more scalable - it'll be easier to add next group type in the future if you will come up with some new idea.
What do you think?

@Charence  came up with the schema. So he's probably better positioned to answer why it's done that way. I think ti's fine to add a group type, but we'd still need to check arguments validity. It's not really for our own UI - we know it'll be doing the right thing, but there are customers trying to use direct API access so we need checks in place to make sure they're trying to do what they seem like they are trying to do :)

carmenfan:
@dleszcz @simplypixi 

I've just commited some code to check the duplicate rule problem (i.e. multiple rules checking the  same field names) I used Postman to test it and i think it's working. But do let me know if you run into problems!

Will probably do some more commits in the new couple of days to add some backend tests, but it shouldn't affect you guys :)

dleszcz:
hello again @carmenfan 
I tested switching between types and after changing from _smart_ to _normal_ i got this:

Screenshot 2019-03-14 at 14 14 50

no _shared_ids_ in objects item:
Screenshot 2019-03-14 at 14 14 19

carmenfan:
@dleszcz I'm struggling to reproduce this. Can you tell me the steps you took to see this?

> no _shared_ids_ in objects item:
> Screenshot 2019-03-14 at 14 14 19

This is request payload... so you are sending a PUT request where object entries doesn't have `shared_ids`? that's invalid. If this is the case the error is correct. If there are no shared_ids within the model/submodel you're sending then it shouldn't have an entry (i.e. if you want to create an empty normal group, `objects: []`) In this case your group is indeed malformed.

dleszcz:
@carmenfan 
1. I sent request with objects 
`[{account: "T_E_S_T"
ifc_guids: []
model: "0687ef98-52b8-4910-a211-4cef1cb7422c"
shared_ids: []}]`
2. In response i got objects:
`[{account: "T_E_S_T", model: "0687ef98-52b8-4910-a211-4cef1cb7422c"}]`
3. I sent request with some rules
4. In response i got rules AND OBJECTS:
`[{account: "T_E_S_T"
model: "0687ef98-52b8-4910-a211-4cef1cb7422c}]`
5. Then after PUT with these objects occur error

carmenfan:
> @carmenfan
> 
> 1. I sent request with objects
>    `[{account: "T_E_S_T" ifc_guids: [] model: "0687ef98-52b8-4910-a211-4cef1cb7422c" shared_ids: []}]`
> 2. In response i got objects:
>    `[{account: "T_E_S_T", model: "0687ef98-52b8-4910-a211-4cef1cb7422c"}]`
> 3. Then I sent request with some rules
> 4. In response i got rules AND OBJECTS:
>    `[{account: "T_E_S_T" model: "0687ef98-52b8-4910-a211-4cef1cb7422c}]`
> 5. Then, after PUT with these objects i got an error **group request malformed**

I'm failing to reproduce this , I see the empty array on step 2:
image


Are you sure you're at the latest of this branch?

Also, you shouldn't be sending ifc_guids with your post/put request (probably doesn't hurt, but it shouldn't be there)

dleszcz:
@carmenfan Yes, I am sure. Can you pull my last commit? It happens only when group dont have any objects

carmenfan:
@dleszcz  sanity question - did you restart the server after you update?

I've updated, but I'm still struggling 😆 

step 1, i created an empty group, my payload looks different (and also response):
image

image

Then I change it to smart group and the payload and response is as follows:
image
image

Updating back to normal groups with the selected objects:
image

image




All working 😱

carmenfan:
Unrelated note, When I try to delete all, I get an error:


And override all is automatically toggled if there's no groups:
image

Also, if override all is checked, when a new group is created, instead of unflagging override all, it should override the new group:



In conclusion, the override all boolean should only change if:
- User clicked on the button to toggle on/off
- Override all is on, and user clicked to turn off override in one of the groups

Otherwise, the state of the group should respect the override all property instead of changing it, unless we're currently editing that group. Hope this make sense 😆

dleszcz:
:(

ezgif com-video-to-gif (1)

carmenfan:
> :(
> 
> ezgif com-video-to-gif (1)

Ok I managed to reproduce it! 🎉  

I'll look into it lol

carmenfan:
> > :(
> > ezgif com-video-to-gif (1)
> 
> Ok I managed to reproduce it! 🎉
> 
> I'll look into it lol

@dleszcz Fixed!

dleszcz:
@carmenfan It works! ;) Also, I fixed your last requests related with deleting all and overriding groups.

carmenfan:
> @carmenfan It works! ;) Also, I fixed your last requests related with deleting all and overriding groups.

Yay! I think it's pretty much there, so I'll get @sanmont3drepo to code review this and @paloknapo  to do a last User test

carmenfan:
I've changed the target to staging as there isn't much point in merging it back to the old backend branch.

@dleszcz @simplypixi Can you solve the conflict? thanks!!

paloknapo:
- [ ] Can't save smart group if it has lot of objects
image

- [ ] Please change override all icon
image

- [ ] Please update "more about RegEx" link to: 
https://3drepo.com/regex-for-data-validation/

sanmont3drepo:
There is an error that's been thrown when you click on "reset to saved selection"

error reset

simplypixi:
> There is an error that's been thrown when you click on "reset to saved selection"
> 
> error reset

@dleszcz I think we missed it during recent rebasing

simplypixi:
> * [ ]  Can't save smart group if it has lot of objects.
>   image

simplypixi:
> > * [ ]   Can't save smart group if it has lot of objects.
> >   image

@carmenfan  It's happening on backend side, isn't it?

carmenfan:
@simplypixi yes definitely a backend bug, we'll look into it our side! :)

simplypixi:
> @simplypixi yes definitely a backend bug, we'll look into it our side! :)

Cool! ;)

simplypixi:
> There is an error that's been thrown when you click on "reset to saved selection"
> 
> error reset

Fixed! ;]

simplypixi:
@carmenfan 

PR will be ready when you fixed backend. ;)🤞

carmenfan:
image

scroll bar is fixed in staging (from https://github.com/3drepo/3drepo.io/issues/1395) but it's not in here :(

dleszcz:
> scroll bar is fixed in staging (from #1395) but it's not in here :(
Our scrolling group details issue fix was the reason... 👎 
Repaired.

sanmont3drepo:
- [ ] When I click on the background it doesn't deselect the already selected mesh

error deselect

dleszcz:
@sanmont3drepo your request fixed

PR #1414, opened 2019-03-05
ISSUE #1413 add migration script for risk viewpoints

### LINKED ISSUE #1413: Migration script - Risk viewpoint to viewpoints array
### Description
Make risk data inline with issues,  create a migration script to copy(for backwards compatibility sake?) viewpoint into viewpoints array.

carmenfan:
This fixes #1413

#### Description
Added a migration script that will copy the viewpoint to the points array, making it inline with issues


carmenfan:
> 
> 
> No `dryrun`?

Yea, it's so trival didn't see much point of having it (or what I should print out)

PR #1415, opened 2019-03-07
Issue 1316 - Viewer Settings

### LINKED ISSUE #1316: Settings menu
### Description
We need to add a settings menu to user. Configurations to add:

- 3drepo Lite mode
- Shadows (hard/soft/none)
- Show statistics
- XRay highlighting
- Memory for Unity

Advanced options (probably never used by users but may be useful exposed for admins/troubleshooting):
- minimum near plane 
- Sphere/box collider algorithm
- far plane sampling points

Things we don't have but it will be viable to add in:
- Quality option - allowing user to turn off anti aliasing
- Artistic rendering mode
- Different Skybox
- Colour scheme
- Ability to tweak light intensity(?)
(More to follow)

sanmont3drepo:
This fixes #1316


frontend/routes/components/topMenu/components/visualSettingsDialog/visualSettingsDialog.component.tsx
```diff
@@ -0,0 +1,215 @@
...
+				Show Statistics
+				<Field name="statistics" render={ ({ field }) => (
+					<Switch checked={field.value} {...field} value="true" color="secondary" />
+				)}/>
+			</FormListItem>
+			<FormListItem>
+				Memory for Unity
+				<Field name="memory" render={ ({ field, form }) => {
+					return (
+					<ErrorTooltip  title={form.errors.memory || ''} placement="bottom-end">
```
simplypixi: Extra space before `title` prop :P

frontend/routes/components/topMenu/components/visualSettingsDialog/visualSettingsDialog.component.tsx
```diff
@@ -0,0 +1,215 @@
...
+	memory: schema.integer(16, 2032)
+});
+
+const BasicSettings = (props) => {
+	return (
+		<List>
+			<FormListItem>
+				Shading
+				<Field name="shading" render={ ({ field }) => (
+					<Select {...field}>
```
simplypixi: Could you use `SelectField` instead of `Select`? `SelectField` pretend native behaviours like navigation between items using the keyboard. ;)

frontend/routes/components/topMenu/components/visualSettingsDialog/visualSettingsDialog.component.tsx
```diff
@@ -0,0 +1,215 @@
...
+		visualSettings: null,
+		flag: false
+	};
+
+	public handleTabChange = (event, selectedTab) => {
+		this.setState({ selectedTab });
+	}
+
+	public onSubmit = (values, { resetForm }) => {
+		values.nearPlane = Number(values.nearPlane);
```
simplypixi: I think you shouldn't mutate data here. Could you move this to `updateSettings` action?

frontend/routes/components/topMenu/components/visualSettingsDialog/visualSettingsDialog.styles.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  &&:hover {
+	background-color:  rgba(234, 57, 57, 1);
+  }
+`;
+
+export const NeutralActionButton = styled(Button)`
+  && {
+    color: ${COLOR.BLACK_60};
+	background-color:  rgba(256, 256, 256, 0);
```
simplypixi: Do you see these extra spaces too? If yes, please change it. If not, it's probably github bug...  
image

frontend/routes/components/topMenu/components/visualSettingsDialog/visualSettingsDialog.styles.ts
```diff
@@ -0,0 +1,106 @@
...
+import { COLOR } from '../../../../../styles';
+import { Button, DialogContent, ListItem, Tooltip, Input, Tabs, Tab } from '@material-ui/core';
+import * as React from 'react';
+import { TooltipProps } from '@material-ui/core/Tooltip';
+import {omit} from 'lodash';
+
+export const NegativeActionButton = styled(Button)`
+  && {
+    color: ${COLOR.WHITE_87};
+	background-color:  rgba(234, 57, 57, 0.87);
```
simplypixi: Could you move repeatable colors to shared colors or to some constant in this file?

frontend/routes/components/topMenu/components/visualSettingsDialog/visualSettingsDialog.styles.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export const FormListItem = styled(ListItem)`
+	display: flex;
+	&& {
+    justify-content: space-between;
+    height: 35px;
+	}
+`;
+
+export const ErrorTooltip = styled((prop: TooltipProps) => {
```
simplypixi:
```suggestion
export const ErrorTooltip = styled(Tooltip).attr({
    classes: {
       tooltip: 'tooltip'
    }
})
```
?


And why did you do that?

```
const props = omit(prop, 'className');
props.classes = props.classes = { popper: prop.className, tooltip: 'tooltip' };
```

Maybe you should create new component instead of this. What do you think?

frontend/routes/components/topMenu/components/visualSettingsDialog/visualSettingsDialog.styles.ts
sanmont3drepo: Yest, its because of the mixture between tabs and spaces. Its easy to mix up both tabbing styles (the one from the general ts and the one for the strings in the styled components ts).

frontend/routes/components/topMenu/components/visualSettingsDialog/visualSettingsDialog.component.tsx
sanmont3drepo: Sure, but why? Its just a conversion to pass the data to updateSettings with the correct datatypes, also its kind of the react's domain problem; which is that the forms output only strings.

frontend/routes/components/topMenu/components/visualSettingsDialog/visualSettingsDialog.styles.ts
sanmont3drepo: Because is the way you style a material-ui tooltip. If you try it the other way it will throw an error.
https://github.com/mui-org/material-ui/issues/11467

frontend/routes/components/topMenu/components/visualSettingsDialog/visualSettingsDialog.component.tsx
sanmont3drepo: Yest, that seems like a good idea

frontend/routes/components/topMenu/components/visualSettingsDialog/visualSettingsDialog.styles.ts
carmenfan: linter has indent only rule enforced on tslint. Why are these getting through the check? 😮

frontend/routes/components/topMenu/components/visualSettingsDialog/visualSettingsDialog.component.tsx
simplypixi: Doing this here, you potentially mutate the values object, which can cause some side effects.
Moreover, as you said, Formik returns always values as string. I think it's better to keep this behaviour unified across the app and change data type, when it's needed.

It's loose idea, so if you think it's not necessary just omit this comment.

frontend/routes/components/topMenu/components/visualSettingsDialog/visualSettingsDialog.styles.ts
simplypixi: Styles are checked using stylelint, so tslint config doesn't affect checking them.
I've checked styleint config and there is no indentation rule. I'll fix it. Do you want these changes here or can I provide it on e.g Issues PR?

carmenfan: Yea Issues PR will be great :)

PR #1444, opened 2019-03-26
ISSUE 1440 - Additional tests for groups

### LINKED ISSUE #1440: Additional tests for groups
### Description
See gaps identified here: https://github.com/3drepo/3drepo.io/pull/1410

### Goals
- [x] Check author
- [x] Check updateBy
- [x] More groups tests

Charence:
This fixes #1440 

#### Description
- Additional tests for groups
- Return complete group data for single group


backend/test/integrated/group.js
```diff
@@ -42,6 +42,7 @@ describe("Groups", function () {
 			{
 				"account":"groupUser",
 				"model":"4ec71fdd-0450-4b6f-8478-c46633bb66e3",
+				"ifc_guids":[],
```
carmenfan: It should really be one or the other no? either returns ifc_guids, or shared ids.

backend/test/integrated/group.js
```diff
@@ -114,7 +115,8 @@ describe("Groups", function () {
 		it("using master head revision should succeed", function(done){
 			agent.get(`/${username}/${model}/revision/master/head/groups/${groupID}`)
 				.expect(200 , function(err, res) {
-					expect(res.body).to.deep.equal(goldenData);
+					expect(res.body.color).to.deep.equal(goldenData.color);
```
carmenfan: why do we have to split this up? result should match the goldenData entirely, if it doesn't then something sounds wrong.

backend/test/integrated/group.js
```diff
@@ -146,7 +148,8 @@ describe("Groups", function () {
 		it("using revision ID should succeed", function(done){
 			agent.get(`/${username}/${model}/revision/b74ba13b-71db-4fcc-9ff8-7f640aa3dec2/groups/${groupID}`)
 				.expect(200 , function(err, res) {
-					expect(res.body).to.deep.equal(goldenData);
+					expect(res.body.color).to.deep.equal(goldenData.color);
```
carmenfan: same as above

backend/test/integrated/group.js
```diff
@@ -163,19 +166,21 @@ describe("Groups", function () {
...
 					agent.post(`/${username}/${model}/groups/`)
 						.send(data)
 						.expect(200 , function(err, res) {
+							groupId = res.body._id;
 							done(err);
 					});
 				},
 				function(done) {
-					agent.get(`/${username}/${model}/revision/master/head/groups/?noIssues=true`)
+					agent.get(`/${username}/${model}/groups/revision/master/head/${groupId}`)
```
carmenfan: this is the deprecated api - need to swap group with revision

Also this is calling the specific group's API insetad of groupList. is this intended?

backend/test/integrated/group.js
```diff
@@ -198,13 +204,14 @@ describe("Groups", function () {
 					agent.post(`/${username}/${model}/groups/`)
 						.send(newGroup)
 						.expect(200 , function(err, res) {
+							groupId = res.body._id;
 							done(err);
 					});
 				},
 				function(done) {
-					agent.get(`/${username}/${model}/revision/master/head/groups/?noIssues=true`)
+					agent.get(`/${username}/${model}/groups/revision/master/head/${groupId}`)
```
carmenfan: same as above

backend/test/integrated/group.js
```diff
@@ -227,13 +235,14 @@ describe("Groups", function () {
...
```
carmenfan: same as above

backend/test/integrated/group.js
```diff
@@ -256,13 +266,14 @@ describe("Groups", function () {
...
```
carmenfan: same as above

backend/test/integrated/group.js
```diff
@@ -285,13 +297,14 @@ describe("Groups", function () {
...
```
carmenfan: same

backend/test/integrated/group.js
```diff
@@ -314,13 +328,14 @@ describe("Groups", function () {
...
```
carmenfan: same

backend/test/integrated/group.js
```diff
@@ -347,13 +363,14 @@ describe("Groups", function () {
...
```
carmenfan: same

backend/test/integrated/group.js
```diff
@@ -363,20 +380,23 @@ describe("Groups", function () {
...
```
carmenfan: same

backend/test/integrated/group.js
```diff
@@ -484,20 +504,23 @@ describe("Groups", function () {
...
```
carmenfan: same

backend/test/integrated/group.js
Charence: It returns extra things like _v

backend/test/integrated/group.js
Charence: Yes, it's intended because I think it's more useful to check the contents of the group rather than the the length of the group list array.

carmenfan: Actually do you need this 2nd API request? I thought HTTP POST returns you the whole group

backend/models/group.js
```diff
@@ -730,7 +731,9 @@ function buildRule(rule) {
 		throw responseCodes.INVALID_ARGUMENTS;
 	}
 
-	if (clauses.length > 1) {
+	if (clauses.length > 1 && notOperators.includes(rule.operator)) {
```
Charence: @carmenfan I could also do `rule.operator.includes("NOT")` to detect this, but this would rely on the operator name.

Charence:
Ready for another review

backend/models/group.js
```diff
@@ -579,6 +601,21 @@ function clean(groupData) {
...
 
+	if (cleaned.rules && 0 === cleaned.rules.length) {
+		delete cleaned.rules;
+	}
+
+	for (let i = 0; cleaned.objects && i < cleaned.objects.length; i++) {
+		if (cleaned.objects[i].ifc_guids && 0 === cleaned.objects[i].ifc_guids.length) {
+			delete cleaned.objects[i].ifc_guids;
+		}
+		if (cleaned.objects[i].shared_ids && 0 === cleaned.objects[i].shared_ids.length) {
```
carmenfan: Can we add a rule where if both ifc_guids and shared_ids are both empty, remove the entry from the object array?

We need to avoid having entries that have neither. It's ok with have an empty object array, but things go haywire on the frotn end if we have object entries that doesn't have actual entries

backend/models/group.js
```diff
@@ -579,6 +601,21 @@ function clean(groupData) {
 		cleaned.updatedAt = cleaned.updatedAt.getTime();
 	}
 
+	if (cleaned.rules && 0 === cleaned.rules.length) {
+		delete cleaned.rules;
+	}
+
+	for (let i = 0; cleaned.objects && i < cleaned.objects.length; i++) {
```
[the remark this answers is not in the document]
sanmont3drepo: Maybe if this has to happen to every array from an object  you dont have to passs in the propName and just iterate the props from the object looking for arrays.

Charence:
Changes addressed

Issue #1430, opened 2019-03-19
Migration: UI buttons


dleszcz:
### Subtasks
- [x] Creating toolbar component
- [x] Creating toolbar buttons (with tooltips)
- [x] Restyling of buttons on the left to make their behaviour similar to buttons from toolbar
- [x] Creating toolbar dropdown
- [x] Adding new viewer's constants: 
    - [x] navigationModes, 
    - [x] clippingModes,
- [x] Extending viewer's store of new values: 
    - [x] navigationMode, 
    - [x] helicopterSpeed, 
    - [x] focusMode, 
    - [x] clippingMode
- [x] Moving some of ViewerService method to the new viewer
- [x] Checking the icon in material/font awesome fonts
- [x] Migrate functions connected with buttons to actions:
    - [x] extent
    - [x] turntable
    - [x] show all
    - [x] hide
    - [x] isolate
    - [x] clip
    - [x] focus
    - [x] BIM
    - [x] Measure
- [x] Removing buttons on the right side
- [x] Removing old code

### Extras
- [x] transform bottom toolbar
- [x] put right buttons into new bottom toolbar.
- [x] remove text below button in favour of tooltip
- [x] use colour scheme from sketch
- [x] add button on top right to go back to teamspace
- [x] get rid of teamspace button in top right menu
- [x] button tool bar - when tool is active, highlight in orange/yellow colour.
- [x] dot on bottom of the button denotes it expands.
- [x] some buttons have submenu (Clip, Turntable)

Prototype:
https://xd.adobe.com/view/dbafde66-da15-4ead-5a67-7d7723e3d04b-ad9e/

Design specs:
https://xd.adobe.com/spec/7cbf178d-b7b1-4cfd-47f2-a1f3ff621111-0a8e/

simplypixi:
@carmenfan 
We check your current code of bottom buttons and I found something like this:
![image](https://user-images.githubusercontent.com/5683490/54668299-db188580-4aee-11e9-9816-9a2891ecc677.png)

I suppose that the bottom buttons should close on background click, but... when I tried to do this, it didn't work. Should we fix this whether omit?

carmenfan:
> @carmenfan
> We check your current code of bottom buttons and I found something like this:
> ![image](https://user-images.githubusercontent.com/5683490/54668299-db188580-4aee-11e9-9816-9a2891ecc677.png)
> 
[...]

PR #1446, opened 2019-03-27
ISSUE_1431 - Migration: Toolbar, Buttons and BIM card

### LINKED ISSUE #1431: Migration: Metadata(BIM) card
### Subtasks
- [x] Creating new BIM panel component (currently called "docs")
- [x] Adding tabs (All | Starred)
- [x] Adding list item (Star | Name | Value)
- [x] Adding filters to header (filtering by name or value)
- [x] Adding menu with single option to header (Clear starred)
- [x] Migrating rightPanel
- [x] Creating new module to handle BIM data and actions (migrating of docs service)
- [x] Binding the action to handler of model's click
- [x] Copying filters

### Extras
- [x] menu (option: Clear starred)
- [x] user can press on star - star turns yellow and available inn "starred" tab.
- [x] "favourites" will be saved in user profile - we will add API to edit/get this
- [x] when user unstar in starred tab, put a ~2sec delay to allow user to change their mind
- [x] add chip filter in meta tab. This is string only search, use metadata list from smart groups to give prompt for user.
- [x] Filter should work with both all and starred (name/value)
- [x] filter should look through both key and value for each entry
- [x] chips should work under OR logic i.e. list all results that matches any of the chips

Prototype:
https://xd.adobe.com/view/dbafde66-da15-4ead-5a67-7d7723e3d04b-ad9e/

Design specs:
https://xd.adobe.com/spec/7cbf178d-b7b1-4cfd-47f2-a1f3ff621111-0a8e/

simplypixi:
This fixes #1431 and #1430


frontend/modules/viewer/viewer.sagas.ts
```diff
@@ -47,6 +55,28 @@ export function* mapInitialise({surveyPoints, sources = []}) {
 	}
 }
 
+export function* initialiseToolbar() {
+	try {
+		Viewer.on(VIEWER_EVENTS.UPDATE_NUM_CLIP, (clipNumber) => {
```
simplypixi: I think we should remove this listener when the viewer is destroyed

frontend/modules/viewer/viewer.sagas.ts
```diff
@@ -47,6 +55,28 @@ export function* mapInitialise({surveyPoints, sources = []}) {
...
+			const isClipEdit = selectIsClipEdit(getState());
```
simplypixi: @dleszcz Please movie this to new action and the just dispatch it here. ;)
Because of that simple change, you will be able to avoid unnecessary uses of `getState()`. It's just a migration helper, so after the migration, we will have to remove all `getState()` occurrences.

frontend/routes/components/fontAwesomeIcon/fontAwesomeIcon.styles.ts
```diff
@@ -43,4 +43,12 @@ export const IconContainer = styled(Icon).attrs({
 	&.primary {
 		color: ${MuiTheme.palette.primary.main}
 	}
+
+	&.fontSizeSmall {
```
simplypixi: @dleszcz Why didn't you add `className` prop to this component? I think you can use `styled-components` here.

frontend/routes/components/topMenu/topMenu.styles.ts
```diff
@@ -33,3 +35,12 @@ export const Container = styled.div`
     transform: translateX(-50%);
   }
 `;
+
+export const BackIcon = styled(InputIcon)`
+	&& {
+		color: ${COLOR.WHITE};
+		font-size: 26px;
+    filter: drop-shadow(0 0 2px ${COLOR.BLACK_30});
```
simplypixi: @dleszcz indentation

frontend/routes/teamspaces/components/tooltipButton/tooltipButton.styles.ts
```diff
@@ -0,0 +1,59 @@
...
+  &.panelButton {
+    background-color: ${(props: any) => props.active ? COLOR.PRIMARY_MAIN : COLOR.REGENT_GRAY};
+    color: ${COLOR.WHITE};
+
+    &:hover {
+      background-color: ${(props: any) => props.active ? COLOR.PRIMARY_MAIN : COLOR.REGENT_GRAY};
+    }
+  }
+
+	&.toolbarButton {
```
simplypixi: @dleszcz indentation

frontend/routes/teamspaces/components/tooltipButton/tooltipButton.styles.ts
```diff
@@ -0,0 +1,59 @@
+import styled from 'styled-components';
+import IconButton from '@material-ui/core/IconButton';
+import { COLOR } from '../../../../styles';
+
+export const StyledIconButton = styled(IconButton)`
+  &.panelButton {
```
simplypixi: Why did you use classes directly? Why not `styled-component`? :O

frontend/routes/teamspaces/components/tooltipButton/tooltipButton.styles.ts
```diff
@@ -0,0 +1,59 @@
...
+    background-color: ${(props: any) => props.active ? COLOR.PRIMARY_MAIN : COLOR.REGENT_GRAY};
+    color: ${COLOR.WHITE};
+
+    &:hover {
+      background-color: ${(props: any) => props.active ? COLOR.PRIMARY_MAIN : COLOR.REGENT_GRAY};
```
simplypixi: It's should be helper

frontend/routes/viewer/components/bim/bim.styles.ts
```diff
@@ -0,0 +1,35 @@
...
+import { COLOR } from '../../../../styles';
+
+export const Container = styled.div`
+	border-top: 1px solid ${COLOR.BLACK_6};
+	margin-top: -1px;
+	overflow: auto;
+`;
+
+export const EmptyStateInfo = styled.p`
+  padding: 14px;
```
simplypixi: @simplypixi indentation

frontend/routes/viewer/components/closeFocusModeButton/closeFocusModeButton.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export class CloseFocusModeButton extends React.PureComponent<IProps, any> {
+	public render() {
+		return (
+			<Fade in={this.props.isFocusMode}>
+				<Container visible={this.props.isFocusMode}>
+					<TooltipButton
+						className="panelButton"
+						label={'Close focus mode'}
```
simplypixi: @dleszcz  You no need brackets

frontend/routes/viewer/components/closeFocusModeButton/closeFocusModeButton.styles.ts
```diff
@@ -0,0 +1,27 @@
...
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import styled from 'styled-components';
+
+export const Container = styled.div`
+  position: absolute;
```
simplypixi: @dleszcz indentation

frontend/routes/viewer/components/panelsMenu/panelsMenu.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+	COMPARE: 'Compare',
+	GIS: 'GIS'
+};
+
+export class PanelsMenu extends React.PureComponent<any, IState> {
+	public state = {
+		activePanels: {}
+	};
+
+	public handleClick = (panel) => {
```
simplypixi:
```suggestion
	private togglePanel = (panel) =>() => {
```

frontend/routes/viewer/components/panelsMenu/panelsMenu.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+			{ label: PANELS.RISKS, Icon: RisksIcon, action: () => this.handleClick(PANELS.RISKS) },
+			{ label: PANELS.GROUPS, Icon: GroupsIcon, action: () => this.handleClick(PANELS.GROUPS) },
+			{ label: PANELS.VIEWS, Icon: ViewsIcon, action: () => this.handleClick(PANELS.VIEWS) },
+			{ label: PANELS.TREE, Icon: TreeIcon, action: () => this.handleClick(PANELS.TREE) },
+			{ label: PANELS.COMPARE, Icon: CompareIcon, action: () => this.handleClick(PANELS.COMPARE)},
+			{ label: PANELS.GIS, Icon: GisIcon, action: () => this.handleClick(PANELS.GIS) }
+		];
+	}
+
+	public renderButtons = () => this.toolbarList.map((buttonProps, index) => (
```
simplypixi:
```suggestion
	private renderButtons = () => this.toolbarList.map((buttonProps, index) => (
```

frontend/routes/viewer/components/panelsMenu/panelsMenu.styles.ts
```diff
@@ -0,0 +1,26 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import styled from 'styled-components';
+
+export const Container = styled.div`
```
simplypixi: @dleszcz Indentation

frontend/routes/viewer/components/toolbar/toolbar.component.tsx
```diff
@@ -0,0 +1,308 @@
...
+	SubmenuDot
+} from './toolbar.styles';
+import { TooltipButton } from '../../../teamspaces/components/tooltipButton/tooltipButton.component';
+
+import { VIEWER_NAV_MODES, VIEWER_CLIP_MODES, VIEWER_TOOLBAR_ITEMS } from '../../../../constants/viewer';
+import {
+	INITIAL_HELICOPTER_SPEED, MIN_HELICOPTER_SPEED, MAX_HELICOPTER_SPEED
+} from '../../../../services/viewer/viewer';
+
+const MeasureIcon = () => <Ruler IconProps={{ className: 'fontSizeSmall' }} />;
```
simplypixi: Use `styled-components`

frontend/routes/viewer/components/toolbar/toolbar.component.tsx
```diff
@@ -0,0 +1,308 @@
...
+			const modelUI = document.querySelector('#modelUI') as HTMLElement;
+			const topMenu = document.querySelector('#topMenu') as HTMLElement;
+
+			modelUI.style.visibility = uiVisibility;
+			topMenu.style.visibility = uiVisibility;
+		}
+
+		if (!this.props.clippingMode && prevProps.clippingMode) {
+			this.setState({
+				activeSubMenu: ''
```
simplypixi: it could be one line

frontend/routes/viewer/components/toolbar/toolbar.component.tsx
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	public componentWillUnmount() {
+		this.props.deactivateMeasure();
+	}
+
+	public handleNavigationModeClick = (mode) => {
+		this.props.setNavigationMode(mode);
+		this.setState({
+			activeSubMenu: ''
```
simplypixi: it could be one line

frontend/routes/viewer/components/toolbar/toolbar.component.tsx
```diff
@@ -0,0 +1,308 @@
...
+		});
+	}
+
+	public handleClickAway = () => {
+		this.setState({
+			activeSubMenu: ''
```
simplypixi: it could be one line

frontend/routes/viewer/components/toolbar/toolbar.component.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+	public handleShowSubmenu = (label) => {
+		this.setState((prevState) => ({
+			activeSubMenu: prevState.activeSubMenu !== label ? label : ''
+		}));
+	}
+
+	public get toolbarList() {
+		return [
+			{
```
simplypixi: We need to refactor that somehow...

frontend/routes/viewer/components/toolbar/toolbar.component.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+	public renderSubmenu = (subMenu, label) => renderWhenTrue(() => {
+		const condition = this.state.activeSubMenu === label;
+		return (
+			<Fade in={condition}>
+				<ClickAwayListener onClickAway={this.handleClickAway}>
+					<Submenu>{subMenu.map((subButton, subKey) => (
+						<TooltipButton
+							key={subKey}
+							className={`toolbarButton toolbarSubButton ${subButton.specificOption && 'toolbarSpecificButton'}`}
```
simplypixi: :(

frontend/routes/viewer/components/toolbar/toolbar.styles.ts
```diff
@@ -0,0 +1,72 @@
...
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import styled from 'styled-components';
+import { COLOR, WHITE } from '../../../../styles';
+
+export const Container = styled.div`
+  background-color: ${COLOR.REGENT_GRAY};
+  border: 1px solid ${COLOR.SILVER_CHALICE};
```
simplypixi: @dleszcz indentation

frontend/routes/viewer/components/toolbar/toolbar.styles.ts
```diff
@@ -0,0 +1,72 @@
...
+export const ClipIconWrapper = styled.span`
+  position: relative;
+`;
+
+export const ClipNumber = styled.span`
+  position: absolute;
+  left: 8px;
+  top: 5px;
+  font-size: 9px;
+  font-weight: 900;
```
simplypixi: please add new field in `FONT_WEIGHT`

frontend/services/fontAwesome.ts
```diff
@@ -3,10 +3,14 @@ import {
...
+	faHelicopter,
+	faRulerHorizontal
 } from '@fortawesome/free-solid-svg-icons';
 
 library.add(faEraser);
 library.add(faEye);
 library.add(faTint);
 library.add(faHandPaper);
+library.add(faHelicopter);
+library.add(faRulerHorizontal);
```
simplypixi: https://material.io/tools/icons/?icon=straighten&style=baseline

backend/routes/user.js
```diff
@@ -72,4 +139,49 @@ function deleteApiKey(req, res, next) {
...
+
+	req.dataModel = await User.appendStarredMetadataTag(username,tag);
+	next();
+}
+
+async function replaceStarredMetadataTags(req, res, next) {
+	const username = req.session.user.username;
+	const tags = req.body;
+
+	if (!isArray(tags) || !tags.every(isString)) {
```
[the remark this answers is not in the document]
sanmont3drepo: Actually it would be the same thing. 
I need to know that every tag is efectivly a string. 
with some it would be 
tag.some(isNotString) 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

image

lets say that the element 10 isNotString , then it would stop there , because is a truthy value, effectivily iterating 10 elements. 

In case that every element is actually a string, it would iterate ALL elements because it didnt find a truthy value


From: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

image

In case of let say the element 10 is not a string , then isString will return false, therefore stoping at the 10nth element. Iterating the same amount of elements as with some.

In case that every element is actuallu a strng it would iterata ALL elements just like .some would.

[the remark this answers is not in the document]
simplypixi: Make sense! ;)

carmenfan:
image


Can we lower the BIM card? it's blocking the top right buttons. 

The panels on the left should align to the new lowered height of the right panels

backend/routes/user.js
```diff
@@ -72,4 +139,49 @@ function deleteApiKey(req, res, next) {
...
+	const username = req.session.user.username;
+	req.dataModel = await User.getStarredMetadataTags(username);
+	next();
+}
+
+async function appendStarredMetadataTag(req, res, next) {
+	const username = req.session.user.username;
+	const tag = req.body.tag;
+	if (!tag) {
+		res.status(400);
```
carmenfan: @sanmont3drepo  this should return `responseCodes.INVALID_ARGUMENTS`

backend/routes/user.js
```diff
@@ -72,4 +139,49 @@ function deleteApiKey(req, res, next) {
...
+	req.dataModel = await User.appendStarredMetadataTag(username,tag);
+	next();
+}
+
+async function replaceStarredMetadataTags(req, res, next) {
+	const username = req.session.user.username;
+	const tags = req.body;
+
+	if (!isArray(tags) || !tags.every(isString)) {
+		res.status(400);
```
carmenfan: @sanmont3drepo this should return responseCodes.INVALID_ARGUMENTS

backend/routes/user.js
```diff
@@ -72,4 +139,49 @@ function deleteApiKey(req, res, next) {
...
+
+	req.dataModel = await User.setStarredMetadataTags(username,tags);
+	next();
+}
+
+async function deleteStarredMetadataTag(req, res, next) {
+	const username = req.session.user.username;
+	const tag = req.body.tag;
+	if (!tag) {
+		res.status(400);
```
carmenfan: @sanmont3drepo this should return responseCodes.INVALID_ARGUMENTS

frontend/components/docs/js/docs.service.ts
```diff
@@ -14,6 +15,8 @@
  *  You should have received a copy of the GNU Affero General Public License
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+import { selectIsMetadataVisible } from '../../../modules/viewer';
```
carmenfan: @simplypixi @dleszcz any reason why we're not migrating this service?

simplypixi: Nope - removed ;)

simplypixi:
> See above comment for bugs found.
> 
> Also It seems like the whole interaction has slowed down a lot. See the following comparison:
> 
> 
> I think my computer needs a restart, but perhaps something worth investigating!

Even the progress component get some lag... Could you restart the computer? xD Everything seems to be ok on our side. It's just a simple list. ;)


> * [ ]  The tabs are doing this weird re-adjustment upon load, can we just not show it until it's ready?
>   
Do you mean cards/panels? Because it's not related to the tabs. ;)
The main problem is height algorithm which was implemented on Angular side. It's very slow and invokes unnecessary re-renders. We also had to plugin react components into it to get the calculated height, which - obviously - slows down things a bit.
Currently, we have a few options:
1. Rewriting height algorithm now and one more time after migration of whole panels.
2. Leaving it as it is and writing a new algorithm after migration
3. Finding a solution to show panel after full-render (but, the question is: how to detect which change in the component is the last one?)

> * [ ]  Can we have a confirmation pop up before we clear the starred array?
>   i.e. "Are you sure you want to clear your Starred items?" [YES/NO]
Sure. Should we implement it instead of 2s delay (about which @paloknapo mentioned on the meeting)?

simplypixi:
@carmenfan How often are the metadata change? I'm thinking about adding some cache to these requests. What do you think?

carmenfan:
> @carmenfan How often are the metadata change? I'm thinking about adding some cache to these requests. What do you think?

They're different on every upload, but is static for every revision (for now). But in the future we may allow users to edit them. So not sure how useful is caching.

carmenfan:
> Rewriting height algorithm now and one more time after migration of whole panels.
Leaving it as it is and writing a new algorithm after migration
Finding a solution to show panel after full-render (but, the question is: how to detect which change in the component is the last one?)

I guess it can stay like this for now 😆 ...

> Sure. Should we implement it instead of 2s delay (about which @paloknapo mentioned on the meeting)?

Yes. so the 2s delay is for removing individual starred items. (no confirmation needed for this)
We only want confirmation if the user decided to remove the whole lot in one go (i.e. hamburger menu -> clear starred)

dleszcz:
>  If I try to unhighight the object by doing alt + click, there is a massive lag, then it deselects the object, and re-selects it again
> Bim card should disappear if the object is unhighlighted

@carmenfan We think that the panels should have the same behavior and BIM panel should be toggled only when the user clicked on BIM button in a toolbar. 

Now when you click on background or deselect all nodes then we display information "No data". 

The lag on big models it's caused by the background model processing all the time. We can close this panel in this time, but it is not consistent with the rest of the panels. Please write what you think about it.

carmenfan:
> > If I try to unhighight the object by doing alt + click, there is a massive lag, then it deselects the object, and re-selects it again
> > Bim card should disappear if the object is unhighlighted
> 
> @carmenfan We think that the panels should have the same behavior and BIM panel should be toggled only when the user clicked on BIM button in a toolbar.
> 


The bim card is more like a pop up card with info than an interactive menu. So I'm not sure if the behaviour needs to be consistent with the panels on the left. This is why it's on the right.

I don't think it should say no data when no item is selected. Perhaps it should say no data when you select an object it actually has no metadata, but not when it hasn't selected anything. If the card stays up then it should probably say "select an object to view its bim data" or something.

@paloknapo  thoughts?

> Now when you click on background or deselect all nodes then we display information "No data".
> 
> The lag on big models it's caused by the background model processing all the time. We can close this panel in this time, but it is not consistent with the rest of the panels. Please write what you think about it.


As for the lag, I'll experiment a bit more, my concern is, when i tried it, it was significantly slower than production, and the model I was trying it on wasn't even a big model. None the less, it shouldn't be selecting itself again.

simplypixi:
> > > If I try to unhighight the object by doing alt + click, there is a massive lag, then it deselects the object, and re-selects it again
> > > Bim card should disappear if the object is unhighlighted
> > 
> > 
> > @carmenfan We think that the panels should have the same behavior and BIM panel should be toggled only when the user clicked on BIM button in a toolbar.
> 
> The bim card is more like a pop up card with info than an interactive menu. So I'm not sure if the behaviour needs to be consistent with the panels on the left. This is why it's on the right.
> 
> I don't think it should say no data when no item is selected. Perhaps it should say no data when you select an object it actually has no metadata, but not when it hasn't selected anything. If the card stays up then it should probably say "select an object to view its bim data" or something.
> 
> @paloknapo thoughts?
> 
> > Now when you click on background or deselect all nodes then we display information "No data".
> > The lag on big models it's caused by the background model processing all the time. We can close this panel in this time, but it is not consistent with the rest of the panels. Please write what you think about it.
> 
> As for the lag, I'll experiment a bit more, my concern is, when i tried it, it was significantly slower than production, and the model I was trying it on wasn't even a big model. None the less, it shouldn't be selecting itself again.

Ok! Fixed. Let us know if you change your mind - then I'll revert proper commit. ;)

simplypixi:
@carmenfan Are we waiting for something else before merging this PR? :P

carmenfan:
> @carmenfan Are we waiting for something else before merging this PR? :P

I need to get some feedback from @paloknapo , but we've been working on the last release (which released on Friday!) so we haven't had a spare testing box that I can put up for him to test.

I'll do it today 👍

carmenfan:
@dleszcz @simplypixi just Found a bug:
If I multi select several geometry, then unselects them, it works as intended unless I'm trying to unselect the lastly selected (the one which  has it's metadata currently displayed)

Could be wrong, but it kind of feels like we're telling the tree service to unhighlight everything if we're getting rid of the bim tab?

dleszcz:
> @dleszcz @simplypixi just Found a bug:
> 
> * [x]  If I multi select several geometry, then unselects them, it works as intended unless I'm trying to unselect the lastly selected (the one which  has it's metadata currently displayed)
>   Could be wrong, but it kind of feels like we're telling the tree service to unhighlight everything if we're getting rid of the bim tab?
>   
> * [x]  The bim card needs to be scrollable as it can get long
>   image
> * [x]  The bim button should disable when the user leaves the viewer (i.e. if I leave the viewer to load another model, it should resets itself back to disabled) - This applies to every other toggle button/panel
> * [x]  Failing to fetch metadata on federations - I think ti's using the federation model ID when it needs to use the sub model ID?
>   image

@carmenfan fixed ;)

frontend/services/viewer/viewer.ts
```diff
@@ -6,11 +6,16 @@ const MODE = {
 	PIN: 'PIN'
 };
 
+export const INITIAL_HELICOPTER_SPEED = 1;
```
Charence: Is there a reason to define these constants in frontend/services/viewer/viewer.ts instead of frontend/constants/viewer.ts?

simplypixi: Good point! It should be moved to viewer constants.
@dleszcz

frontend/modules/tree/tree.sagas.ts
```diff
@@ -49,7 +54,7 @@ export function* startListenOnSelections() {
 
 export function* getSelectedNodes() {
 	try {
-		yield call(delay, 0);
+		yield call(delay, 100);
```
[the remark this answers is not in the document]
simplypixi: Honestly, it's a hack. As far as I know, it's because of the specific behavior of tree service. 
@dleszcz Could you tell more about it?

frontend/services/viewer/viewer.ts
simplypixi: Fixed

frontend/modules/tree/tree.sagas.ts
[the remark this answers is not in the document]
dleszcz: @Charence It's temporary. We had to use some delay - similar to your fix in **groups.component** (old code with a comment below).

`this.$scope.$watchCollection(() => {
			return this.treeService.currentSelectedNodes;
		}, () => {
			this.$timeout(() => {
				/*
				 *	Temporary fix: This timeout is required because
				 * when currentSelectedNodes may be updated before unity
				 * (where groups get the object count) so without the delay
				 * it may have incorrect readings of how many objects are selected
				 * Proper fix would be to investigate this stupid control flow
				 * and make it work (and probably stop relaying on watches...)
				 */
				this.groupsService.updateSelectedObjectsLen().then(() => {
					this.groupsService.getSelectedObjects().then((currentHighlights) => {
						this.selectedNodes = currentHighlights || [];
						this.updateChangeStatus();
					});
				});
			});
		});` 

We will try to fix it and find a better solution when we will migrate TreeService.

dleszcz:
> The height of the BIM panel always seems to be not quite long enough for its contents.
Fixed

PR #1451, opened 2019-03-29
ISSUE #1445 - Added live chat

### LINKED ISSUE #1445: Live chat in 3drepo.io
### Description
Add livechat to 3drepo.io

### Goals
- [x] Users can use livechat

### Tasks
- [x] Add the livechat plugin

sanmont3drepo:
This fixes #1445

### DIFF
--- a/frontend/routes/app/app.component.tsx
+++ b/frontend/routes/app/app.component.tsx
@@ -21,6 +21,7 @@ import { DialogContainer } from '../components/dialogContainer';
 import { SnackbarContainer } from '../components/snackbarContainer';
 import { clientConfigService } from '../../services/clientConfig';
 import { isStaticRoute } from '../../services/staticPages';
+import LiveChat from 'react-livechat';
 
 interface IProps {
 	location: any;
@@ -29,6 +30,7 @@ interface IProps {
 	hasActiveSession: boolean;
 	authenticate: () => void;
 	logout: () => void;
+	currentUser: any;
 }
 
 interface IState {
@@ -111,6 +113,21 @@ export class App extends React.PureComponent<IProps, IState> {
 				this.setState({ referrer: DEFAULT_REDIRECT });
 			});
 		}
+
+		if (isAuthenticated) {
+			// @ts-ignore
+			window.LC_API = window.LC_API || {};
+			// @ts-ignore
+			window.LC_API.on_after_load = () => {
+				const {firstName, lastName, email } = this.props.currentUser;
+				const name = `${firstName || ''} ${lastName || ''}`.trim();
+				// @ts-ignore
+				window.LC_API.set_visitor_name(name);
+				// @ts-ignore
+				window.LC_API.set_visitor_email(email);
+			};
+
+		}
 	}
 
 	public toggleAutoLogout = (shouldStart = true) => {
@@ -161,6 +178,7 @@ export class App extends React.PureComponent<IProps, IState> {
 			<>
 				<DialogContainer />
 				<SnackbarContainer />
+				<LiveChat license={10338577} />
 			</>
 		);
 	}
--- a/frontend/routes/app/app.component.tsx
+++ b/frontend/routes/app/app.component.tsx
@@ -21,7 +21,7 @@ import { DialogContainer } from '../components/dialogContainer';
 import { SnackbarContainer } from '../components/snackbarContainer';
 import { clientConfigService } from '../../services/clientConfig';
 import { isStaticRoute } from '../../services/staticPages';
-import LiveChat from 'react-livechat';
+import { LiveChat } from '../components/liveChat';
 
 interface IProps {
 	location: any;
@@ -113,21 +113,6 @@ export class App extends React.PureComponent<IProps, IState> {
 				this.setState({ referrer: DEFAULT_REDIRECT });
 			});
 		}
-
-		if (isAuthenticated) {
-			// @ts-ignore
-			window.LC_API = window.LC_API || {};
-			// @ts-ignore
-			window.LC_API.on_after_load = () => {
-				const {firstName, lastName, email } = this.props.currentUser;
-				const name = `${firstName || ''} ${lastName || ''}`.trim();
-				// @ts-ignore
-				window.LC_API.set_visitor_name(name);
-				// @ts-ignore
-				window.LC_API.set_visitor_email(email);
-			};
-
-		}
 	}
 
 	public toggleAutoLogout = (shouldStart = true) => {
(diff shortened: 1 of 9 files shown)


frontend/routes/app/app.component.tsx
```diff
@@ -161,6 +178,7 @@ export class App extends React.PureComponent<IProps, IState> {
 			<>
 				<DialogContainer />
 				<SnackbarContainer />
+				<LiveChat license={10338577} />
```
simplypixi: What do you think about moving this license number to a new constant, which will contain main config of app?

Are you sure about adding an only single license number? Shouldn't be a license per environment (at least local and production)?
I don't know how exactly livechat works (we use Intercom), but - if you want to get some conversations' stats - this solution can obscure them.
Somebody can send some fake/test messages locally or on staging, which will make the stats corrupted.
If it's not that case, just omit this comment. ;)

frontend/routes/app/app.component.tsx
```diff
@@ -111,6 +113,21 @@ export class App extends React.PureComponent<IProps, IState> {
 				this.setState({ referrer: DEFAULT_REDIRECT });
 			});
 		}
+
+		if (isAuthenticated) {
+			// @ts-ignore
+			window.LC_API = window.LC_API || {};
```
simplypixi: Could you move this logic to a new service called `livechat`? It'll be small service, but it'll be easier to manage livechat code. It's similar (or even the same) approach that you used adding analytics to the app.

For example:
```
// services/livechat.ts

const initialise = ( { firstName = '' , lastName = '' , email = '' } = {}) => {
	const name = `${firstName} ${lastName}`.trim();
	// @ts-ignore
	window.LC_API.set_visitor_name(name);
	// @ts-ignore
	window.LC_API.set_visitor_email(email);
};

export const LiveChat = { initialise };
```

frontend/routes/app/app.component.tsx
carmenfan: I think this should go through Client Config like most other things?

carmenfan: I think this should go through Client Config like most other things?

simplypixi: Sounds good

frontend/routes/app/app.component.tsx
sanmont3drepo: I didnt wanted to add angular code. But I'll move it to a helper.

sanmont3drepo: The other thing that I could do is create a react component with the livechat code, binded to the login state and setting the license from the config.

sanmont3drepo:
The config option added is the property 

`{   liveChatLicense: Number }`

frontend/routes/app/app.component.tsx
simplypixi: @sanmont3drepo I didn't mean about angular service. I thought something like a helper (look at files in `services` directory) to handle external service. I also think that creating a new component is unnecessary. During the migration of top app component, we'll create a new module to set up things like this one during initialising process. So, probably, we'll have to remove this component and create service instead of it.

sanmont3drepo: @simplypixi  I tried today to convert it to just a service I need the props with the updated currentuser to tackle the async nature of configuring livechat. So its easier to have a component with the state binded as well as the license config. 
In the case of the service, its still needed to add the livechat from 'react-livechat' to some part of the react tree, so you wouldnt be able to get rid of the react tag so easy, unless we want to implement the import of the livechat script ourselves.

In the currentstate is just adding the custom livechat component to any place of the tree and that's it.

If you still have the opinion that it would be a simpler integration with the service approach, could you please expand more on the idea?

frontend/routes/components/liveChat/liveChat.component.tsx
```diff
@@ -0,0 +1,59 @@
...
+		// @ts-ignore
+		window.LC_API.set_visitor_name(name);
+		// @ts-ignore
+		window.LC_API.set_visitor_email(email);
+	}
+
+	public componentDidUpdate(prevProps) {
+
+		// @ts-ignore
+		window.LC_API = window.LC_API || {};
```
simplypixi: You should run this only if fields `firstName, lastName, email` have changed. So, you should add here an if statement.

frontend/routes/components/liveChat/liveChat.component.tsx
```diff
@@ -0,0 +1,59 @@
...
+		// @ts-ignore
+		if (Boolean(window.LC_API.set_visitor_email)) { // This is in case the chat was available before the email
+			this.setNameAndMail();						// in this case  the setNameAndEmail must be ran again
+		}
+
+	}
+
+	public render() {
+		const haveLicense = clientConfigService.liveChatLicense !== 0;
+		return renderWhenTrue(<ReactLiveChat license={clientConfigService.liveChatLicense} />)(haveLicense);
```
simplypixi: I think that in this case, the better approach is `haveLicense && <Component />` :P

frontend/routes/app/app.component.tsx
simplypixi: I thought about something like this:
```
// services/livechat.ts
window.LC_API = window.LC_API || {};

const initialise = ( { firstName = '' , lastName = '' , email = '' } = {}) => {
	const name = `${firstName} ${lastName}`.trim();
	// @ts-ignore
	window.LC_API.set_visitor_name(name);
	// @ts-ignore
	window.LC_API.set_visitor_email(email);
};

export const LiveChat = { initialise };


// app.component.tsx
import LiveChat from 'react-livechat`;
import LiveChatService from '../.../services/livechat';

class App {
    /* ... */
    componentDidUpdate() {
        if (isAuthenticated) {
             LiveChatService.initialise(this.props.currentUser)
        }
    }

   render() {
        /* ... */
        <LiveChat licence={123} />
   }
}
```

It's ok if you think it's an overcomplicated solution. We can figure out something different when the proper time comes. Now, it can stay as is - I mean the component is also good solution (for now), but (probably) we'll have to remove it in the future.

sanmont3drepo: Yeah, the problem is that that solution doesnt work, because you have to tackle the cases when window.LC_API has not being initialized (the .on_after_load handler, which in that context you need the reference to props to have the updated current user,  because if you use just currentuser and bind the initilize function to that you get the object of the previous state of current user)
and in the case that window.LC_API has being initialized  you need o just run it for now also taking in consideration that the mail has being set, because it might take a little to fetch the whole user data.
That's why it cant be that simple. I wish it could.

simplypixi: Got it ;)

PR #1467, opened 2019-04-12
ISSUE 1439 - Migration: Compare

dleszcz:


dleszcz:
Hello @carmenfan !

Travis build log:
> /home/travis/build/3drepo/3drepo.io/backend/models/helper/model.js
> 1010:10  error  'isUserAdmin' is defined but never used  no-unused-vars
>✖ 1 problem (1 error, 0 warnings)

You added this function. Can you take a look at this? ;)

paloknapo:
- [ ] Diff/Clash stays active when I leave the viewer and open different model, it should deactivate on leaving viewer
- card doesn't expand in federation and compare button behaves unexpectedly 

- in federation, Clash picks up all models for the base and the result is incorrect

simplypixi:
> * in federation, Clash picks up all models for the base and the result is incorrect
>   

@paloknapo Do you mean something like this?
screencast-app-3drepo com-2019 04 15-10-10-42

If not, could describe me a bit how the instant clash should work?

dleszcz:
>  Default clash revisions should be current revisions, not previous latest:
@carmenfan could you check if this is still happening? is this the initial state of your compare clash panel, right? :)

simplypixi:
> * card doesn't expand in federation and compare button behaves unexpectedly

@paloknapo Does it still happen? Tbh, we upgraded a bit heigh-setting algorithm in this PR. I also checked that a few minutes ago on Mac and PC with Windows (using Chrome). Which browser did you use while doing that test? Did you have the latest changes?

carmenfan:
> > Default clash revisions should be current revisions, not previous latest:
> 
> @carmenfan could you check if this is still happening? is this the initial state of your compare clash panel, right? :)

Doesn't look like it. It seems to work as expected now :)

Though It looks like the name isn't right if the revision had no name:
image

image

It's displaying the same date for both april and october one.

Also, the user shouldn't be able to start a clash if there is no target model:
image

dleszcz:
> > > Default clash revisions should be current revisions, not previous latest:
> > 
> > 
> > @carmenfan could you check if this is still happening? is this the initial state of your compare clash panel, right? :)
> 
> Doesn't look like it. It seems to work as expected now :)
> 
> Though It looks like the name isn't right if the revision had no name:
> image
> 
> image
> 
> It's displaying the same date for both april and october one.
> 
> Also, the user shouldn't be able to start a clash if there is no target model:
> image

It should be ok now ;)

frontend/routes/components/notifications/notifications.component.tsx
```diff
@@ -182,35 +260,15 @@ export class Notifications extends React.PureComponent<IProps, any> {
...
-
-			const hasThisWeekNot = thisWeek.length > 0 ;
-			const hasLastWeekNot = lastWeek.length > 0;
-			const hasOlderNot = older.length > 0;
-			this.setState({unreadCount, groupedByTeamspace, hasThisWeekNot, hasLastWeekNot, hasOlderNot });
-		}
-	}
-
-	public renderNotificationsPanel = (data) => {
+	private renderNotificationsPanel = (data) => {
```
sanmont3drepo: why?

simplypixi: I refactored the code - added some getters, set some methods as private, set proper order of fields and methods etc.

frontend/modules/compare/compare.sagas.ts
```diff
@@ -0,0 +1,414 @@
...
+		targetDiffRevision: targetRevision,
+		targetClashRevision: baseRevision,
+		revisions
+	};
+};
+
+function* getCompareModelData({ isFederation, settings }) {
+	try {
+		const revisions = yield select(selectRevisions);
+		const [teamspace, , currentRevision] = window.location.pathname.replace('/viewer/', '').split('/');
```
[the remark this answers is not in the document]
simplypixi: @dleszcz Please do it as @sanmont3drepo suggests. We need to connect react-router with redux store while migration of top-level component.

frontend/modules/compare/compare.sagas.ts
```diff
@@ -0,0 +1,414 @@
...
+			const selectedModelsMap = compareModels.reduce((map, model) => {
+				map[model._id] = true;
+				return map;
+			}, {});
+
+			yield put(CompareActions.setComponentState({
+				compareModels,
+				selectedDiffModelsMap: selectedModelsMap,
+				selectedClashModelsMap: selectedModelsMap,
+				targetDiffModels: selectedModelsMap,
```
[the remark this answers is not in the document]
dleszcz: Because this is just setting the initial state of our _Compare_ store data.

frontend/modules/compare/compare.sagas.ts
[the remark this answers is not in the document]
dleszcz: @sanmont3drepo @simplypixi done

frontend/modules/compare/compare.sagas.ts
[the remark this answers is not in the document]
carmenfan: What @sanmont3drepo was wondering is that why do we need 3 references to begin with. Unless we have misunderstood, it seems unnecessary to have all 3 if they are just references to the same object? It would make sense to separate them if you are making a copy of the objects. If I make changes to one, i am effectively making changes to all, so doing [things like this](https://github.com/3drepo/3drepo.io/pull/1467/files#diff-7b9b81947db4f121e3a64bfe6454017fR69) seems pointless.

I feel like we must be missing something! 😆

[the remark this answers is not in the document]
dleszcz: It didn't matter and works good because it was just initialState and called once on start, but you are right, we can use here the spread operator to make sure that everything will be clean :)

dleszcz:
> * [ ]  The revision names when tag is empty is still not quite right:
>   
> * [ ]  If User changes the revision within the clash/diff options while it's on clash/diff mode. We should exit clash/diff mode (see current version /new version behaviour):
>   

Hello @carmenfan. Can you give us access to this model? Because I can't reproduce this first bug :( Also, Can you reset our database? Many models are damaged or have a pending status, so this makes it difficult to look for models.

carmenfan:
> Hello @carmenfan. Can you give us access to this model? Because I can't reproduce this first bug :( 
I've fixed this bug. :)

> Also, Can you reset our database? Many models are damaged or have a pending status, so this makes it difficult to look for models.
Done. In the future you can run `.\scripts\dbMaintenance\resetProcessingFlag.js`in mongo shell to do this.

Found another bug:
- [ ] If the user decide to diff/class the same revision (i.e. base revision matches the target revision), you can't set base model as compare.. You will have to load a copy as if you're loading a different revision:
image

I.e. if (for somereason I want to) I diff myself with myself, it should highlight nothing.

frontend/routes/viewer/components/compare/components/compareDiffItem/compareDiffItem.component.tsx
```diff
@@ -0,0 +1,97 @@
...
+	onRevisionChange: (modelProps) => void;
+}
+
+export class CompareDiffItem extends React.PureComponent<IProps, any> {
+	public static defaultProps = {
+		selected: false
+	};
+
+	public get currentRevisionName() {
+		return this.props.baseRevision.tag || dayjs(this.props.baseRevision.timestamp).format('DD/MM/YYYY');
```
carmenfan: We have a lot of copies of `dayjs(...)`. Since we'd always want this to be consistent throughout the component, can we refactor this out to a utils function?

dleszcz: Do you want the same format everywhere? Because I see different formats in the app:
- 15/05/2019 (Filter panel, Compare)
- 15 May 2019 (Risks, Issues, Groups)
- 08:30 15 May (Groups details form)

dleszcz:
> I.e. if (for some reason I want to) I diff myself with myself, it should highlight nothing.
Do you really need to do that? On clash you can't do this currently.

This is a bug actually because in this case option should be disabled in revisions select component when you try to select the same revision.

frontend/routes/viewer/components/compare/components/compareDiffItem/compareDiffItem.component.tsx
carmenfan: just consistent throughout the particular component is fine! :)

carmenfan:
> > I.e. if (for some reason I want to) I diff myself with myself, it should highlight nothing.
> 
> Do you really need to do that? On clash you can't do this currently.
> 
> This is a bug actually because in this case option should be disabled in revisions select component when you try to select the same revision.

I don't need to do this. If it's easier just disable it. But if we let the user do it then the state needs to be right.

frontend/routes/viewer/components/compare/components/compareDiffItem/compareDiffItem.component.tsx
dleszcz: Done :)

dleszcz:
> I don't need to do this. If it's easier just disable it. But if we let the user do it then the state needs to be right.

Actually,  in select we disabling the same revision (revision with the same id). In your case did you have identical revisions created on one day? They have different IDs, but we can't say that they are identical :/ For us they are different so you can select them.

frontend/routes/viewer/components/compare/components/compareDiffItem/compareDiffItem.component.tsx
dleszcz: Hi @sanmont3drepo  This file was outdated. I moved dayjs function to services. Please check again, and let me know if it's ok for you :)

sanmont3drepo: @dleszcz  the code has been changed from 
`dayjs(child.date).format('DD/MM/YYYY');` to `formatDate(child.date, 'DD/MM/YYYY');`
What @carmenfan  meant was to change the `dayjs(child.date).format('DD/MM/YYYY');`  to somehing like  `formatShortDate(child.date);`

where `formatShortDate` is defined like 
`export const formatShortDate = (date) => dayjs(date).format('DD/MM/YYYY');`

dleszcz: done

Charence:
To improve usability, can the clickable areas for the slider states be enlarged? At the moment, you need to very carefully click on the text or slider line. The space between the line and the text does not listen to clicks.

May-28-2019 16-46-06

dleszcz:
> @maciejgryka @dleszcz
> 
> To improve usability, can the clickable areas for the slider states be enlarged? At the moment, you need to very carefully click on the text or slider line. The space between the line and the text does not listen to clicks.
> 
> May-28-2019 16-46-06

@Charence  done

dleszcz:
Hello @carmenfan 

When are you going to merge this PR to staging? ;)

carmenfan:
> Hello @carmenfan
> 
> When are you going to merge this PR to staging? ;)

@dleszcz It's tagged wait on merge because this is going into the next release. And the current release isn't ready yet. (It's waiting on the helicopter fix actually :P )

Issue #1468, opened 2019-04-12
Issue/Risk Warning


dleszcz:
https://github.com/3drepo/3drepo.io/pull/1388#issuecomment-481183628

> The bug occurs when you load an Issue or Risk directly using the URL query string.

<img width="1680" alt="55790882-3e298680-5ab5-11e9-960d-24affdb4d98c" src="https://user-images.githubusercontent.com/17221890/56027209-e473c700-5d15-11e9-8de8-5a17f0e7b191.png">

PR #1469, opened 2019-04-12
ISSUE #1468 - Risk/Issue Warning

dleszcz:
https://github.com/3drepo/3drepo.io/issues/1468

Always pass a boolean value as an active prop in panel button.


carmenfan:
@Charence Can you check this? (Since you found the bug 😆 )

PR #1510, opened 2019-05-13
ISSUE 1506 - Enforce jobs

### LINKED ISSUE #1506: Enforce jobs
### Description

All users added to a teamspace must be assigned a job

### Goals
- Universal jobs

### Tasks
- [x] Require job when adding user
- [x] Prevent job removal

Charence:
This fixes #1506 

#### Description
All users must be assigned a job when added to a teamspace and they cannot be made jobless.

#### Test cases
- Add user to teamspace without a job shouldn't work on io frontend or via API call
- Unassign menu option should be missing for existing user list
- It should not be possible to reassign a user with no job on io frontend or via API call


sanmont3drepo:
Maybe we should get rid of the 'unassigned' option. Or place a little error message when its being selected with "Users must get a job assigned" or something similar.
image

PR #1542, opened 2019-05-29
ISSUE #1541 - Errors are thrown if the account or the model are not found

### LINKED ISSUE #1541: When a teamspace account has not been found in permissions it should throw an error
### Description
When checking for permissions in the backend and the teamspace couldnt been found it should throw an error instead of a permission denied.

sanmont3drepo:
This fixes #1541

### DIFF
--- a/backend/middlewares/getPermissionsAdapter.js
+++ b/backend/middlewares/getPermissionsAdapter.js
@@ -40,7 +40,7 @@
 
 				return this.getUser().then(user => {
 					if(!user) {
-						throw ResponseCodes.ACCOUNT_NOT_FOUND;
+						throw ResponseCodes.RESOURCE_NOT_FOUND;
 					}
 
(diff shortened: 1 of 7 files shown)


backend/middlewares/getPermissionsAdapter.js
```diff
@@ -39,9 +39,8 @@
 			accountLevel: function(username) {
 
 				return this.getUser().then(user => {
-
 					if(!user) {
-						return [];
+						throw ResponseCodes.RESOURCE_NOT_FOUND;
```
Charence: Should the error be thrown by `User.findByUserName` or perhaps `this.getUser` instead?

Having the check here doesn't seem to catch all cases where the teamspace doesn't exist.

E.g. `http://127.0.0.1/api/Bogdan_asdfD/members`

PR #1547, opened 2019-05-31
Issue 1519: Model Options Response (Teamspaces)

### LINKED ISSUE #1519: Model Options Response (Teamspaces)
### Description
No response when hovering over the options bar for individual models in the teamspaces. Federation options seem to keep floating without disappearing..

### Steps to replicate
- Log in to your account and open a project in the teamspaces. 
- Hover over the pop-up options list for individual models

Model Options

simplypixi:
This fixes #1519

### DIFF
--- a/frontend/routes/teamspaces/components/modelItem/modelItem.component.tsx
+++ b/frontend/routes/teamspaces/components/modelItem/modelItem.component.tsx
@@ -139,7 +139,7 @@ export class ModelItem extends React.PureComponent<IProps, IState> {
 
 			if (!isHidden && !isPendingStatus(this.props.status)) {
 				return renderWhenTrue((
-					<Tooltip title={label} key={index}>
+					<Tooltip enterDelay={50} title={label} key={index}>
 						<IconButton aria-label={label} onClick={action}>
 							<Icon {...iconProps} />
 						</IconButton>
--- a/frontend/routes/teamspaces/components/projectItem/projectItem.component.tsx
+++ b/frontend/routes/teamspaces/components/projectItem/projectItem.component.tsx
@@ -104,14 +104,17 @@ export class ProjectItem extends React.PureComponent<IProps, IState> {
 		<RowMenu open={hovered} forceOpen={this.state.actionsMenuOpen} toggleForceOpen={this.toggleActionsMenuOpen}>
 			<TooltipButton
 				{...ROW_ACTIONS.EDIT}
+				enterDelay={50}
 				action={this.props.onEditClick}
 			/>
 			<TooltipButton
 				{...ROW_ACTIONS.PERMISSIONS}
+				enterDelay={50}
 				action={this.props.onPermissionsClick}
 			/>
 			<TooltipButton
 				{...ROW_ACTIONS.DELETE}
+				enterDelay={50}
 				action={this.props.onRemoveClick}
 			/>
 		</RowMenu>
--- a/frontend/routes/teamspaces/components/teamspaceItem/teamspaceItem.component.tsx
+++ b/frontend/routes/teamspaces/components/teamspaceItem/teamspaceItem.component.tsx
@@ -54,6 +54,7 @@ export const TeamspaceItem = (props: IProps) => {
 	const renderActions = () => renderWhenTrue(() => (
 		<TooltipButton
 			{...ROW_ACTIONS.ADD_NEW}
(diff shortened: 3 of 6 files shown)


frontend/routes/teamspaces/components/modelItem/modelItem.component.tsx
```diff
@@ -139,7 +139,7 @@ export class ModelItem extends React.PureComponent<IProps, IState> {
 
 			if (!isHidden && !isPendingStatus(this.props.status)) {
 				return renderWhenTrue((
-					<Tooltip title={label} key={index}>
+					<Tooltip enterDelay={50} title={label} key={index}>
```
simplypixi: I added this delay (which is not noticeable for an ordinary mortal) because the tooltip sometimes is getting the wrong position. It's because tooltip is setting the position before the menu's animation finished

simplypixi:
> Can we use css instead of react event handlers? Is less prone to error.
> This has already been done in notifications.

@sanmont3drepo No. This case is more complex than hover implementation on notifications. Primarily, the menu from teamspaces items can be permanently visible - it's a fallback for the touch devices.
Besides, I wouldn't like to mix react, styled-component and extra css classes to achieve the same effect.
Why do you think current solution is more error prune? Could you give me an example?

sanmont3drepo:
> > Can we use css instead of react event handlers? Is less prone to error.
> > This has already been done in notifications.
> 
> @sanmont3drepo No. This case is more complex than hover implementation on notifications. Primarily, the menu from teamspaces items can be permanently visible - it's a fallback for the touch devices.
- this can be taken care with media queries, if not you can pass a prop (like in avatar.styles.ts) marking as if its touch and use it in the styled component to not do the hover thing  :
https://www.styled-components.com/docs/advanced
https://www.w3schools.com/css/css_rwd_mediaqueries.asp

 also , where is the code for dealing with touch devices? It seems that I am not able to find it.

> Besides, I wouldn't like to mix react, styled-component and extra css classes to achieve the same effect.
- Me neither, that's why the piece of code Im referring to is in styled components ;-)

> Why do you think current solution is more error prune? Could you give me an example?
- My example is the bug to begin ;-). In css you woudnt have to deal with race conditions.

simplypixi:
> * this can be taken care with media queries, if not you can pass a prop (like in avatar.styles.ts) marking as if its touch and use it in the styled component to not do the hover thing  :
>   https://www.styled-components.com/docs/advanced
>   https://www.w3schools.com/css/css_rwd_mediaqueries.asp

Yes, I know about that. ;) It depends on which browsers you support. While we were implementing this view, pointer media queries weren't fully supported on each browser.
As I see, currently only IE doesn’t support that feature. There is another problem - what if, the user has a touchable screen and mouse (e.g laptop with touchable screen)? How the menu should behave?

I think it’s a good idea, but it shouldn't be the topic of this PR. You can create a new issue and we’ll check/fix that later. ;)

> also , where is the code for dealing with touch devices? It seems that I am not able to find it.

@PavolKnapo suggested adding an extra button which forces displaying the menu all the time.

> * Me neither, that's why the piece of code I'm referring to is in styled components ;-)

Yes, but your code (and case) is much simpler. We have one re-usable component, which is used on each list item and uses predefined transition component from MaterialUI. We can recreate this animation using styled-components/CSS - we’ll have to do that if we lose control of hover event. The question is… is it worth it?

> * My example is the bug to begin ;-). In css you woudnt have to deal with race conditions.

It depends. :) The tooltip still will need to „wait” for the end of menu animation - even if we’ll recreate this animation using CSS.

simplypixi:
@carmenfan @sanmont3drepo When do you want to merge this PR? Should I still change something?

carmenfan:
@sanmont3drepo wanted to take a deeper look into this. It's on his todo list

carmenfan:
@simplypixi 
@sanmont3drepo was going to look into this, but with new UI changes on teamspace it seems a bit irrelevant.

What should we do with this? merge it or just leave it?

simplypixi:
> @simplypixi
> @sanmont3drepo was going to look into this, but with new UI changes on teamspace it seems a bit irrelevant.
> 
> What should we do with this? merge it or just leave it?

Tbh, it could be removed. I'm rewriting teamspaces a bit right now.

carmenfan:
Then I'm gonna close this

PR #1552, opened 2019-06-03
ISSUE #1548 - Get heli speed on init

### LINKED ISSUE #1548: models/federations not honouring helicopter speed after a refresh
### Description
The user changes the  helicopter speed, It looks like the speed is saved and retrieved upon refresh, but it is not being set on the actual helicopter speed.

### Steps to replicate
- load a model
- Change into helicopter mode
- Increase/decrease speed significantly
- hit refresh
- change into helictoper mode, the speed is still set to 1

dleszcz:
This fixes #1548

### DIFF
--- a/frontend/components/model/js/model.component.ts
+++ b/frontend/components/model/js/model.component.ts
@@ -25,6 +25,7 @@ import { ViewpointsActions } from '../../../modules/viewpoints';
 import { JobsActions } from '../../../modules/jobs';
 import { RisksActions } from '../../../modules/risks';
 import { GroupsActions } from '../../../modules/groups';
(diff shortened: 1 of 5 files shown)


dleszcz:
> The init speed is working on positive speed settings, but not negative.
> 
> e.g:
> 
> * load a model
> * set speed to -10
> * refresh, speed still feels like 1 instead of what it suppose to be

Weird... because _HelicopterSpeedDown_ in UnityUtil is called in the same way like _HelicopterSpeedUp_ (and is called the correct number of times)

By the way, there is another way to set helicopter speed instead of calling these two methods? Setter method would be definitely better.

carmenfan:
Actually it looks like it's working... my bad!

PR #1559, opened 2019-06-05
ISSUE 1555 - Remove orphan files

### LINKED ISSUE #1555: Remove orphaned files
### Description
Script to remove orphaned files

Charence:
This fixes #1555 

#### Description
Script to remove orphan files and identify missing files


carmenfan:
@Charence  I changed it to be windows compatible. Please check it still works for you!

Things I changed:
- changed all `folder + filename` to os.path.join(x, y) -this should fix // problems (double check)
- use os.path.normpath - normalises the path, so it should fix any `/` and `\` problems.
- use `x in ignoreDirs` instead of turning it into a regex because regex hates `\`
- We now list/print orphans and missings in separate lists for clarity
- added the ref._id into the print for missings so it doesn't just show the the collection the ref was in.

Charence:
Script also needs to attempt to repair missing links

PR #1576, opened 2019-06-16
Issue 1573: Migration - API Service

dleszcz:


dleszcz:
Hello @Charence . I fixed fetching revisions. 

Was there anything changed on backend with downloading JSONs? It worked well before, but now it's not working also on staging. That's why I think the problem is not in this PR.

Screenshot 2019-06-19 at 10 35 34

frontend/modules/store.ts
```diff
@@ -14,7 +14,7 @@ export default function configureStore(initialState = {}) {
 	const enhancers = [];
 
 	if (IS_DEVELOPMENT) {
-		middlewares.unshift(require('redux-immutable-state-invariant').default());
+		// middlewares.unshift(require('redux-immutable-state-invariant').default());
```
Charence: What is this line for?

Charence:
> Was there anything changed on backend with downloading JSONs? It worked well before, but now it's not working also on staging. That's why I think the problem is not in this PR.

@dleszcz You are right! The JSON endpoint for issues and risks changed and frontend/services didn't get updated. I have done that now on this branch.

frontend/modules/store.ts
dleszcz: I will uncomment this line after migrating RevisionsService, because now errors are unnecessarily displayed by this library

PR #1586, opened 2019-06-24
Issue 1585 - Smart groups should return transformation IDs.

### LINKED ISSUE #1585: Smart groups should return the transformation shared IDs on search results
### Description
If I do a smart group query and only transformations satisfy the query, at the moment it doesn't return any results. This can be misleading.

### Steps to replicate
Set a smart group query that satisfy metadata of a transformation. the backend returns no results

carmenfan:
This fixes #1585


carmenfan:
Parked for now

PR #1603, opened 2019-07-04
Issue 1599 - Bugs on issues/risks

### LINKED ISSUE #1599: Issues and risks comments bugs
### Description
There are several bugs with comments:
- [x] When to users are connected to the same model and a comment is created manually the sent guid in the chatevent is wrong.
- [x] The system logs generated by the other person doing the changes appears to be empty
- [x] Deleting a comment in risks doesnt work
- [x] The order of the remote comments is wrong
- [x] After updating a issue the screenshots from the comments seems to hide away

sanmont3drepo:
This fixes #1599


frontend/helpers/issues.ts
```diff
@@ -15,34 +15,50 @@
...
+	if (issue.assigned_roles) {
+		const roleColor = get(jobs.find((job) => job.name === get(issue.assigned_roles, '[0]')), 'color');
+		issue = {...issue, roleColor};
+	}
+
+	if (issue.status) {
+		issue = {...issue, defaultHidden: issue.status === STATUSES.CLOSED};
+	}
+
+	renameFieldIfExists(issue, 'desc', 'description');
```
simplypixi: Isn't better to return proper field name from the backend? :)

frontend/helpers/issues.ts
```diff
@@ -15,34 +15,50 @@
...
  */
 
-import { get, omit } from 'lodash';
+import { get } from 'lodash';
 import { getAPIUrl } from '../services/api';
 import { STATUSES_COLOURS, STATUSES_ICONS, STATUSES } from '../constants/issues';
 import { isAdmin, hasPermissions, PERMISSIONS } from './permissions';
-import { sortByDate } from './sorting';
+
+const renameFieldIfExists = (issue, fieldName, newFieldName) => {
```
simplypixi: If you want to do that this way, it should be a pure function.

e.g
```suggestion
const renameFieldIfExists = (issue, fieldName, newFieldName) => {
   const updatedIssue = { ...issue };
   updatedIssue[newFieldName] = issue[fieldName];
   delete updatedIssue[fieldName];
   return updatedIssue;
}
```

frontend/helpers/issues.ts
```diff
@@ -15,34 +15,50 @@
...
 
 export const prepareIssue = (issue, jobs = []) => {
-	const thumbnail = getAPIUrl(issue.thumbnail);
-	const descriptionThumbnail = issue.viewpoint && issue.viewpoint.screenshot
+	if (issue.thumbnail) {
+		const thumbnail = getAPIUrl(issue.thumbnail);
+		issue = {...issue, thumbnail};
+	}
+
+	const descriptionThumbnail = get(issue, 'viewpoint.screenshot')
```
simplypixi: Is `get` necessary here?
`get` is much slower than the native solution: https://jsperf.com/get-vs-native

frontend/helpers/issues.ts
```diff
@@ -15,34 +15,50 @@
...
+const renameFieldIfExists = (issue, fieldName, newFieldName) => {
+	if (issue[fieldName]) {
+		issue[newFieldName] = issue[fieldName];
+	}
+};
 
 export const prepareIssue = (issue, jobs = []) => {
-	const thumbnail = getAPIUrl(issue.thumbnail);
-	const descriptionThumbnail = issue.viewpoint && issue.viewpoint.screenshot
+	if (issue.thumbnail) {
```
simplypixi: If you really need to split the code to all these statements, it's better to spread the issue once (at the beginning) and then add/update fields. You'll avoid unnecessary object spreading on each if statement.

```
const preparedIssue = { ...issue };

if (preparedIssue.thumbnail) {
    preparedIssue.thumbnail = getAPIUrl(preparedIssue.thumbnail);
}

return preparedIssue;
```

frontend/helpers/issues.ts
carmenfan: I think you guys renamed it when you migrated issues, so he was just sticking to the convention.

I do agree that it should just use the backend name though, but it does mean it has to be changed for the whole stack.

Should it be a separate issue?

frontend/helpers/issues.ts
sanmont3drepo: that would be less performant. I'll stick to your comments that makes this code more performant.

frontend/helpers/issues.ts
simplypixi: > I think you guys renamed it when you migrated issues, so he was just sticking to the convention.`
Yes, we did it, but @sanmont3drepo refactored this code. It was more like a suggestion. :)

> Should it be a separate issue?
👍

sanmont3drepo: @simplypixi   you authored that code:
https://github.com/3drepo/3drepo.io/blob/staging/frontend/helpers/issues.ts#L36

carmenfan: ok Let's create a new issue for this(https://github.com/3drepo/3drepo.io/issues/1604#issue-464626586). Marking it as resolved here.

@simplypixi  maybe you guys can take a look after all the migration is done?

Marking this as resolved in here.

simplypixi: @sanmont3drepo Yes, I know that. I'm sure I had some reason to do that, but I can't remember what it is - it's been while ago. :P

Issue #1522, opened 2019-05-16
Cache-Control on UnityWeb


keifgwinn:
### Description

It would be good if the client could cache the unityViewer as it's downloading 28.6 mb each time and the viewer's not changing often.

/unity/Build/unity.asm.code.unityweb

it appear to be setting cache-control to 0 

Cache-Control: public, max-age=0

### Goals

- [ ] UnityWeb is cached appropriately

### Tasks

- [ ] Update cache-control in the app?

simplypixi:
@carmenfan Resolved here #1590 (using service worker)

carmenfan:
closing in favour of #1590

Issue #1292, opened 2018-12-14
Rid of the "there's a new version of 3drepo" message once and for all.


carmenfan:
### Description
We want to get rid of this message and just load the latest when available. Things that may help: setting cache control to no-cache policy ?

My theory : we currently have no cache control policy, which means we're under the mercy of the browsers. Which result in occasionally, if we update the servers too frequently, there are times when the cache is not invalidated and thus the message doesn't pop up.

### Goals
- Is there a way to ensure the client doesn't load the cache if it's invalid? `no-cache` is suppose to check for validation before loading, so maybe that's the way
- if the above happens, we can get rid of the dialog asking for refresh.

simplypixi:
Three ideas:
1. Well-prepared service-worker.
2. Adding `chunkhash` or `contenthash` to production build. You can easily achieve that using webpack (15min of work ;)).
3. (old way) Adding `version` at the end of URL e.g. `/3drepo.js?123`, which changes on each build.

simplypixi:
@carmenfan Resolved here #1590 (using service worker)

carmenfan:
closing in favour of #1590

PR #1607, opened 2019-07-11
ISSUE 1590 - Migration: Viewer, Model and rest of services

### LINKED ISSUE #1590: Migration: Viewer & Model

simplypixi:
This fixes #1590 , #1292, #1522 , #1589 

#### Changes
- Migration of Model & Viewer migration
We create two components: ViewerCanvas and ViewerGui.
- Migration of top level of routing
- Removing the rest of old Angular code
- tsconfig update
- tslint update
- libraries upgrade
Now the app runs on the latest react and styled-components (v4), so it should be much faster.
Moreover, you can start using the newest React's features: hooks, memo, suspense, etc. I suggest to use them while creating new components.

#### What we didn't do

- material-ui upgrade - a lot of breaking changes. It'll make this PR even bigger.
- Binding CRA to handle the whole stack - It's a good idea, but we'll have to handle some custom solution used in your project, which will take some time.
- Merging global viewer with viewer service - let me know if you even want to do that. It's doable, but we tried to leave your globals as they are.

#### Known issues
The new version of styled-components sometimes throws warnings (development mode only) for custom event handlers e.g `onRootClick`. We checked that is the issue on the library side, but the good news is the next version is coming. It'll be faster (they rewrote styled-components using hooks) and contains a bugfix that we need. ;)

We went through for almost every place in the app during this part of the migration, so you probably will need to do the same while testing.


simplypixi:
@carmenfan @sanmont3drepo @Charence This PR is enormous - we know that, but we couldn't split that to smaller chunks, because these functionalities were related.
The only idea I have is moving the latest stack changes to separate branch - I ran auto-fix after re-configuring the linter, so a lot of files were updated.

Let me know what you think.

simplypixi:
> On some models I see the following errors:
> 
> Screenshot 2019-07-17 12 17 32
> 
> The above appears when I try to view the model. The other problem is the error message shouldn't display the tags such as `<br>`.
> 
> Screenshot 2019-07-17 12 19 50
> Screenshot 2019-07-17 12 19 05
> 
> The above appears when I try to proceed and click on an issue.

Could you check if it occurs on tree migration branch? Is it possible the model doesn't have a name? Because it looks like that (I mean about the error on the dialog). :P

dleszcz:
> On some models I see the following errors:
> 
> Screenshot 2019-07-17 12 17 32
> 
> The above appears when I try to view the model. The other problem is the error message shouldn't display the tags such as `<br>`.
> 
> Screenshot 2019-07-17 12 19 50
> Screenshot 2019-07-17 12 19 05
> 
> The above appears when I try to proceed and click on an issue.

@Charence These errors are fixed on Tree pull request in the last commit

simplypixi:
> The above appears when I try to view the model. The other problem is the error message shouldn't display the tags such as `<br>`.

And this one works too. It looks like that:
image

Did you test it on latest changes?

Charence:
> Did you test it on latest changes?

@simplypixi @dleszcz Which branch did you apply the changes to? ISSUE_1590?

simplypixi:
> > Did you test it on latest changes?
> 
> @simplypixi @dleszcz Which branch did you apply the changes to? ISSUE_1590?

Yes, `ISSUE_1590`.

I switched the `base branch` to ISSUE_1475 (Tree branch). ISSUE_1590 is a part of main feature branch but since we did rebase it's not necessary to merge it to the parent. I mean it contains all changes.

We'll resolve the conflicts as soon as possible.

frontend/globals/viewer.ts
```diff
@@ -287,6 +267,22 @@ export class Viewer {
...
+		if (progress === 1) {
+			this.emit(Viewer.EVENT.VIEWER_INIT_SUCCESS, progress);
+		} else {
+			this.emit(Viewer.EVENT.VIEWER_INIT_PROGRESS, progress);
+		}
+	}
+
+	public onModelProgress = (progress) => {
+		if (progress === 1) {
+			this.emit(Viewer.EVENT.MODEL_LOADED, progress);
```
carmenfan: I don't think this should be emitted here. We should stick to when unity actually sends you the model loaded signal.

frontend/index.html
```diff
@@ -98,40 +120,7 @@ <h4 style="font-weight: lighter;">
...
-      }
-
-    </script>
-
-
-
-    <!-- Entry point for all components-->
-    <div ui-view="main" flex layout="column"></div>
-    <app />
+    <div id="app"></div>
```
carmenfan: Why is this removed? I can no longer see the internet explorer message (right is staging):
image

frontend/internals/webpack/tools/loaders.js
```diff
@@ -105,5 +117,6 @@ module.exports = {
   FontLoader,
   ImageLoader,
   HTMLLoader,
-  PugLoader
+  PugLoader,
```
carmenfan: Do we still need PugLoader?

sanmont3drepo:
**Visual Settings**
- [ ] After reloading from changing visual settings the visual settings are not applied
- [ ] After changing the visual settings, saving them, then clicking the reset and saving, the app gets blank:
image

frontend/modules/dialog/dialog.redux.ts
```diff
@@ -104,20 +113,37 @@ export const showScreenshotDialog = (state = INITIAL_STATE, action) => {
...
+const hideDialog = (state = INITIAL_STATE) => {
 	return { ...state, isOpen: false, isPending: false };
 };
 
-export const setPendingState = (state = INITIAL_STATE, {isPending}) => {
+const setPendingState = (state = INITIAL_STATE, {isPending}) => {
 	return { ...state, isPending };
 };
 
+const setMuteNotifications = (state = INITIAL_STATE, { muteNotifications }) => {
```
carmenfan: I think we got rid of the dialog that uses this? Is it still needed?

simplypixi:
@carmenfan @dleszcz merged Tree branch yesterday (to resolve conflicts) and he had a problem with starting BE, so he pushed merged changes as a new branch (ISSUE_1590_copy). I checked that today and I have the same issue (required port is free). Did you change something recently? Could you look at it?
image

frontend/globals/viewer.ts
simplypixi: fixed

frontend/index.html
simplypixi: We did it in a "better way." Whole browser detection/unsupported page logic was moved to `support.ts` (it's not included directly in the template), but maybe we set wrong IE or Edge versions. Which versions of IE should we mark as unsupported? Do you support Edge?

image

frontend/internals/webpack/tools/loaders.js
simplypixi: I assume no

simplypixi: removed

frontend/modules/dialog/dialog.redux.ts
simplypixi: removed

simplypixi:
> * There seems to be an async problem. Sometimes if you hit refresh on a model(more likely on a hard refresh), you get this:

@carmenfan is it happening on certain models or all? I can't reproduce this bug :(

frontend/index.html
carmenfan: We dont support any version of ie (they generally don't support webgl),  but we need to make sure whatever code you use for this msg it is visible for anything as low as ie6. we don't support edge either but it should work (ie they can use it but if they run into bugs we dont fix them..) , I can't remember what we currently do with it. I think we just don't say anyhthing for edge and let the user through.

carmenfan:
@simplypixi the port problem is the problem I mentioned in the email. You need to update your chat config to define the port as 'port' instead of 'http_port' or 'https_port'

simplypixi:
> @simplypixi the port problem is the problem I mentioned in the email. You need to update your chat config to define the port as 'port' instead of 'http_port' or 'https_port'

Makes sense now. I totally forgot about that the Tree branch contains staging changes already.
Thanks!

frontend/index.html
simplypixi: @carmenfan Unsupported browser screen shouldn't be displayed for each version of IE now. Could you check it? If it still doesn't work properly, let me know which version od IE and Windows you use.

simplypixi:
> > * There seems to be an async problem. Sometimes if you hit refresh on a model(more likely on a hard refresh), you get this:
> 
> @carmenfan is it happening on certain models or all? I can't reproduce this bug :(

@carmenfan I'm still not able to reproduce this bug, but I've changed a bit logic of loading unityLoader script. Could you check if it's happening?

simplypixi:
@carmenfan Uff, I resolved conflicts. Please double-check the "resources" feature. Moreover, two unity changes were conflicted (8a814ed76, 1119f19bd) so probably you will have to prepare a new build.

@sanmont3drepo Could you migrate your change from staging? We discussed it before on Tree PR. If not, let us know. ;)

```
this.$scope.$watch(() => this.colorOverrides,  async (overrides, prevOverrides) => {
	const toAdd = overridesDiff(overrides, prevOverrides);
	const toRemove = overridesDiff(prevOverrides, overrides );

	removeColorOverrides(toRemove);
	addColorOverrides(toAdd);
});
```

carmenfan:
> @carmenfan I'm still not able to reproduce this bug, but I've changed a bit logic of loading unityLoader script. Could you check if it's happening?

I just tried it, and it seems like it's no longer happening!

> @carmenfan Uff, I resolved conflicts. Please double-check the "resources" feature. Moreover, two unity changes were conflicted (8a814ed, 1119f19) so probably you will have to prepare a new build.

OK.. I'm working on my unity chanage at the moment, depending on the timing of how everything is merged, you might find more conlicts... 😸 

resources seems to be still working 👍 

> @sanmont3drepo Could you migrate your change from staging? We discussed it before on Tree PR. If not, let us know. ;)

@sanmont3drepo  is on holiday until next monday, so he won't be able to do anything until then.

carmenfan:
@simplypixi is there anything outstanding on this?  It looks ok to me, perhaps we should merge this to tree migration?

simplypixi:
> @sanmont3drepo is on holiday until next monday, so he won't be able to do anything until then.

No problem, I can migrate that. 👍 

> @simplypixi is there anything outstanding on this? It looks ok to me, perhaps we should merge this to tree migration?

Tbh, it's up to you. We can merge it to tree migration :)
Did you go through the whole app? I mean, we changed "app core", so it'll be good to do some smoke tests at least.

The most important places to test: Tree panel, Viewer toolbar, app routing
The rest should be fine.

carmenfan:
> > @sanmont3drepo is on holiday until next monday, so he won't be able to do anything until then.
> 
> No problem, I can migrate that. 👍
> 
> > @simplypixi is there anything outstanding on this? It looks ok to me, perhaps we should merge this to tree migration?
> 
> Tbh, it's up to you. We can merge it to tree migration :)
> Did you go through the whole app? I mean, we changed "app core", so it'll be good to do some smoke tests at least.
> 
> The most important places to test: Tree panel, Viewer toolbar, app routing
> The rest should be fine.

I did a quick test and didn't really see anything wrong with the things you mentioned, I'd say lets merge this into the tree and give the tree more testing.

Otherwise i feel like we're testing a lot of versions of everything that may once again need more testing when it's merged together.. 😆

simplypixi:
> Otherwise i feel like we're testing a lot of versions of everything that may once again need more testing when it's merged together.. 😆
Hehe, I understand.... :D

PR #1616, opened 2019-07-19
Issue 1611 - public protocol support

### LINKED ISSUE #1611: Add a config to allow get the front end to talk in https when the server actually uses http
### Description
For deployment situations where we want to the load balancer to deal with SSL traffic and certificates, we want to spin up a web server that listens to http traffic, but the outside world (i.e. front end) still thinks we're using https.

Adding a new config for load balancer config that states the external port and protocol, that doesn't get influenced by internal traffic type/port

**Config changes**:
- following items in config.servers are no longer taken into account (i.e. it will be ignored):
    - http_port
    - https_port
    - port (unless it's chat server)
    - public_port
    - public_protocol
- Chat server configuration will take in a number in `port`  instead of `http_port`
- Chat now has a boolean flag `chatOnPublicPort` to allow chat to use the same public port as other servers (default: `false`)
- there are new configurations on the root level of the config file:
    - public_protocol: `http` or `https` (default: `https`  if ssl is configured)
    - public_port: the port the end users need to connect to (default: same as port)

carmenfan:
This fixes #1611

#### Description
- Allow the public protocol to differ from internal (e.g. https on the outside and http on the inside)
- Fixed a minor non fatal bug where we try to delete files without a corID (toy project import)

#### Notes
See #1611 for necessary config updates

### DIFF
--- a/backend/3drepo.js
+++ b/backend/3drepo.js
@@ -163,18 +163,10 @@ function setupSubdomain(mainApp, subdomain) {
 		if (!serverConfig.external) {
 
 			// Only load frontend server in maintenance mode
-
-			if(!config.maintenanceMode && serverConfig.service === "chat") {
-
+			if(!config.maintenanceMode || serverConfig.service === "frontend") {
+				logCreateService(serverConfig);
 				// chat server has its own port and can't attach to express
-				logCreateService(serverConfig);
-				createChat(serverConfig);
-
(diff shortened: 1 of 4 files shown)


Charence:
**Reminder:** Update `http_port` and `https_port` to `port` for chat.

PR #1628, opened 2019-07-23
ISSUE 1600 - Coloured issue pins

### LINKED ISSUE #1600: Issue pins should colour by Priority/status
### Description
Similar to Risks. issue pins should be coloured by priority. The only exception is if the issue is closed.

`None` or empty or unknown value(through BCF): Black
`Low`: Yellow
`Medum`: Orange
`High`: Red
Closed issue: Green

Charence:
This fixes #1600 

#### Description
Pins now have colours


carmenfan:
@Charence  found a bug: 
The closed issue pin doesn't disappear when I go back to the listing.

Charence:
> @Charence found a bug:
> The closed issue pin doesn't disappear when I go back to the listing.

Should be fixed

frontend/constants/risks.ts
```diff
@@ -82,34 +82,34 @@ export const RISK_LEVELS_ICONS = {
 
 export const RISK_LEVELS_COLOURS = {
 	[LEVELS.UNSET]: {
-		color: COLOR.DARK_GRAY,
-		pinColor: PIN_COLORS.DARK_GRAY,
-		selectedColor: PIN_COLORS.DARK_GRAY
+		color: COLOR.SOFT_BLUE,
```
carmenfan: @charence is there any reason why we need a color and pin color? they all seems to be the same.

frontend/helpers/issues.ts
```diff
@@ -63,12 +63,19 @@ export const prepareIssue = (issue, jobs = []) => {
 };
 
 export const getStatusIcon = (priority, status) => {
-  const statusIcon = {
-    Icon: STATUSES_ICONS[status] || null,
-    color: STATUSES_COLOURS[status] || STATUSES_COLOURS[priority] || null
-  };
+	const statusIcon = {
+		Icon: STATUSES_ICONS[status] || null,
+		color: (ISSUE_COLORS[status] ? ISSUE_COLORS[status].color : null) ||
```
carmenfan: Should this be the same as L77? 
That kind of looks cleaner.

```suggestion
		color: (ISSUE_COLORS[status] || ISSUE_COLORS[priority] || ISSUE_COLORS.NONE).color
```

frontend/constants/risks.ts
Charence: @carmenfan COLOR.DARK_GRAY is in hex `'#bfbfbf'` and PIN_COLORS.DARK_GRAY is parsed into an array of RGB values `[0.75,0.75,0.75]`.

I think this is done because HTML takes the hex value (perhaps it was easier to define the colour in hex as well?) and pin colour in unity takes the RGB value https://github.com/3drepo/3drepounity/blob/master/AssetBundleViewer/Assets/Utility/MessageParser.cs#L70

carmenfan: Considering pin colour is derived from colour, i don't think it makes sense to have 3 entries like this. In fact, we should probably just have color and selected color in hex and whatever component is colouring the pin should just do a conversion.

But probably not worth investigating. Do the other change and this can be merged.

Issue #1612, opened 2019-07-18
Teamspaces grid item


dleszcz:
Its a part of #1608

PR #1635, opened 2019-07-25
Issue 1631/1612 - Teamspaces structure update and new list item component (card)

### LINKED ISSUE #1631: Teamspaces: Components structure change

simplypixi:
This fixes #1631 and #1612


simplypixi:
@carmenfan @sanmont3drepo @Charence Starring models/federations will be finished on different branch ;)

simplypixi:
@carmenfan I edited your message by mistake :( sorry ;(

> * [ ]  We lost the status info (queued, processing etc), it should replace the timestamp when it is processing
> * [ ]  hamburger button on project bar now expand/collapse the project instead of controlling the toggle menu
> * [ ]  Can you make it so federations are listed before models?
> * [ ]  Can we make the cards bigger so we display longer model names in 1 line? i think 4 or 5 a row would be better than 6 + a lot of padding on the right.

Ok! ;)


> There's some elements in Pavol's wireframe that is not reflected here see below:
> 
> 
> 
> * [ ]  We're missing the + button on the project row:
> * [ ]  It's missing the search/sort function on the UI, is that going in on a different branch?

We only added new card component and changed the components and stored data structure (flat list instead of tree list ) - to speed up code/rendering a bit and make easier adding extra features like searching, sorting and so on. We'll change the rest on different branches. ;)

simplypixi:
@carmenfan Fixed ;)

carmenfan:
@simplypixi  Looking through the revision dialog right now and I found this:
image

As it stand, everything is disabled while the file is being processed/waiting to be processed.

Can we make it so I can view the model if there is an existing reivison? (We made this change a while back on the old version)

Also, can we see the settings menu, but disable:
- Upload
- delete.

There's no reason to stop the user from seeing permissions/history etc

carmenfan:
@simplypixi Instead of the add button we talked about earlier in this thread, can we  have a phantom model entry where the user clicks to add a new model/federation? something like this:
image

simplypixi:
> @simplypixi Instead of the add button we talked about earlier in this thread, can we have a phantom model entry where the user clicks to add a new model/federation? something like this:
> image

@carmenfan Yes, sure. What do you think about putting this element at the beginning? In my opinion it'll be more user-friendly - the user won't have to scroll down (if the grid is large) to add new element.

carmenfan:
> @carmenfan Yes, sure. What do you think about putting this element at the beginning? In my opinion it'll be more user-friendly - the user won't have to scroll down (if the grid is large) to add new element.

Sure, we can try it!

simplypixi:
> @simplypixi Looking through the revision dialog right now and I found this:
> image
> 
> As it stand, everything is disabled while the file is being processed/waiting to be processed.
> 
> Can we make it so I can view the model if there is an existing reivison? (We made this change a while back on the old version)
> 
> Also, can we see the settings menu, but disable:
> 
> * Upload
> * delete.
> 
> There's no reason to stop the user from seeing permissions/history etc

Done

> Instead of the add button we talked about earlier in this thread, can we have a phantom model entry where the user clicks to add a new model/federation? something like this:

and done ;)

carmenfan:
@simplypixi 

We lost the federation first ordering 😆 

Also, when I uploaded a file, the tooltip seems to hang around until I go to the hamburger button again, anything we can do about this?



Looking great otherwise! 👍

carmenfan:
@simplypixi  Also, can we make the stars yellow just like the BIM card?

simplypixi:
> @simplypixi Also, can we make the stars yellow just like the BIM card?

Hmm... Everything looks good on my side. Could you provide steps to reproduce this issue? :)
image

I think models/federations ordering disappeared after the last merge, but... I fixed it. ;)
Tooltip bug is also fixed.

dleszcz:
> Can we have all the teamspaces/projects expanded by default on the starred tab?

@carmenfan Is it possible to add project ID in _api/starredModels_ response? This would make it much easier to expand projects by default :)

carmenfan:
> > Can we have all the teamspaces/projects expanded by default on the starred tab?
> 
> @carmenfan Is it possible to add project ID in _api/starredModels_ response? This would make it much easier to expand projects by default in starred Tab :)

Hmm.. not easily, it's going to require quite a few of database queries. I would prefer not to do this unless it's absolutely necessary. Isn't it as simple as just make sure everything is expanded (since projects with no starred models are hidden)? I'm a bit confused why you need this.

dleszcz:
> Hmm.. not easily, it's going to require quite a few of database queries. I would prefer not to do this unless it's absolutely necessary. Isn't it as simple as just make sure everything is expanded (since projects with no starred models are hidden)? I'm a bit confused why you need this.

The reason is the current structure of our component. But don't worry, I'm just asking :) we probably found another solution and can do it without your changes, thanks!

dleszcz:
@carmenfan Done :)

dleszcz:
@carmenfan Take a look, please :)

PR #1637, opened 2019-07-26
ISSUE #1634 - Added a loading state to revision list

### LINKED ISSUE #1634: We need a loading screen for revision list
### Description
The revision history pop up currently display the old information until it loads in a new one. We can do with a loading screen so the user is not confused about what data it is currently displaying.

### Steps to replicate
- On the teamspace page, open revision history for model A
- close it, then open revision history for model B
- You will see the details for model A for ~1-2s before it changes to info for model B. 

Instead of seeing info for Model A, we should replace it with a loading screen.

sanmont3drepo:
This fixes #1634


carmenfan:
@simplypixi Does this affect your stuff? I'm guessing you're not changing the revision component so i guess not?

PR #1640, opened 2019-07-30
Issue 1623: Teamspaces - Revisions dialog

dleszcz:


simplypixi:
> Some thoughts:
> image
> 
> * Can we make the cancel button as `OK`? If I am cancelling an operation it sounds like I'm trying to revert what I just done, which we don't do.
> * Instead of hamburger menu, can we make a drop down that clearly shows the state of the revisions (`active`, `void`), and have `set as latest` as a button beside it? We can make the tab wider to fit, ti's currently very narrow
> * the font is really big and out of scale in comparison to the teamspace component. Can we make the font and padding smaller?
> * Can we format the date to be the same as the teamspace view (i.e. dd/MM/yy HH:mm)
> * Can we change the title of this tab to be `Revisions: <Model Name>`

5x yes

> Bug(?):
> 
> * It looks like the latest revision is no longer in bold?

I'll fix that ;)

frontend/modules/dialog/dialog.redux.ts
```diff
@@ -18,6 +18,8 @@
 import { get, omit } from 'lodash';
 import { createActions, createReducer } from 'reduxsauce';
 import * as Dialogs from '../../routes/components/dialogContainer/components';
+// tslint:disable-next-line
+import RevisionsDialog from '../../routes/components/dialogContainer/components/revisionsDialog/revisionsDialog.container';
```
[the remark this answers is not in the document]
simplypixi: @carmenfan I suppose it's here because of line length. But... only that rule should be disabled.

@dleszcz Why didn't you use `Dialogs` instead?

simplypixi:
@carmenfan I think @dleszcz forgot about one thing:
image

Could you add information about the file type to BE response? :)

simplypixi:
> * Instead of hamburger menu, can we make a drop down that clearly shows the state of the revisions (`active`, `void`), and have `set as latest` as a button beside it? We can make the tab wider to fit, ti's currently very narrow

@carmenfan Hmm the dropdown probably is not necessary. What do you think about that? 

image

carmenfan:
> > * Instead of hamburger menu, can we make a drop down that clearly shows the state of the revisions (`active`, `void`), and have `set as latest` as a button beside it? We can make the tab wider to fit, ti's currently very narrow
> 
> @carmenfan Hmm the dropdown probably is not necessary. What do you think about that?
> 
> image

Hmm reason we thought of using drop down is because  shows clearly the current status instead of assuming users are educated on the colour scheme.

We could do that, but i still think it'd be good to have the current status listed

What about a toggle? like the compare card clash one?

simplypixi:
> What about a toggle? like the compare card clash one?

👍 
With the same colors?

carmenfan:
> > What about a toggle? like the compare card clash one?
> 
> 👍
> With the same colors?

Maybe green for active and grey for void?

carmenfan:
image
file type added. (NB: it is possible for this to be undefined)


@simplypixi @dleszcz  found a bug, we shouldn't be showing void revisions on the viewer or diff.
image

image

simplypixi:
@carmenfan What do you think? (I change a bit information architecture)

image

carmenfan:
> @carmenfan What do you think? (I changed a bit information architecture)
> 
> image

How about this? (roughly, this is a terrible paint job)
image

And I think we should ditch the background colour scheme all together. Instead, void revisions should just have grey text. Red feels like it's a warning/error situation, which it isn't.

So bold = latest, normal = active but not latest, grey = void.

carmenfan:
Also if file type is `BIM` can you show `NAVIS` instead? (BIM is our internal format that we use when it came from a plugin from Navisworks

simplypixi:
> How about this? (roughly, this is a terrible paint job)
> image
> 
> And I think we should ditch the background colour scheme all together. Instead, void revisions should just have grey text. Red feels like it's a warning/error situation, which it isn't.
> 
> So bold = latest, normal = active but not latest, grey = void.

Looks good 👍 But... what about "Set as latest" button? :D

> Also if file type is BIM can you show NAVIS instead? (BIM is our internal format that we use when it came from a plugin from Navisworks

Ok 👍

carmenfan:
> But... what about "Set as latest" button? :D

The business team says if we can't think of a nice way to fit it then we can ditch it... so we can ditch it... 😆

simplypixi:
@carmenfan  done ;)

frontend/helpers/revisions.ts
```diff
@@ -0,0 +1,11 @@
+import { get } from 'lodash';
+import memoizeOne from 'memoize-one';
+
+export const getActiveRevisions = memoizeOne((revisions) => {
+	return revisions.filter((revision) => !revision.void);
```
carmenfan: Instead of doing this, shouldn't we just make sure it utilises the 2 different APIs provided (You don't get the void revisions if you don't add the query string) for when we need the void revisions or not?

In reality no one should need to know about void revisions other than the revisions dialog in the teamspace view. This feels like we're doing more processing on the front end and it's pretty prone to error in the future...

dleszcz:
@carmenfan All of your latest requests done :)

frontend/helpers/revisions.ts
dleszcz: I will use this function only for the case that you are noticed:

>  Active revision no longer bold if it's not the latest (we need to filter to ignore void)

dleszcz:
@carmenfan Done :)

carmenfan:
We can no longer select revisions on this page.. 😆 

image

carmenfan:
@dleszcz can merge after we reinstate the link button on the viewer revision dialog.

dleszcz:
> We can no longer select revisions on this page.. 😆

Oops, my bad... fixed!

PR #1642, opened 2019-08-07
Issue 1641 - Unity Upgrade and Unity caching

### LINKED ISSUE #1641: Unity Caching and Unity Update
### Description
Update Unity to 2018.4.x and investigate on cache policies

### Goals
- [x] Working webassembly version of the unity game engine
- [ ] Unity Caching working on demand
- [ ] Cache policy for things like tree

### Tasks
- [ ] Update code to support unity API changes in https://github.com/3drepo/3drepounity/issues/222
- [ ] Add new cache properties in viewer settings
- [ ] add cache policy settings on the backend

carmenfan:
This fixes #1641 

#### Description
- Unity upgraded to 2018.4.4
- Uses webassembly for the build
- Added new unity APIs to support configurations of the unity cache
- Added cache control policies for things we know won't change.

#### Test cases
- Should function as before
- When caching is one (and it has been cached previously) there should be no calls to fetch bundles/json_mpc files


yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Charence: :-(

carmenfan:
@simplypixi  @dleszcz As this is likely to get merged into staging first (since everything on your side is stuck pending on the tree merge), just to let you know that this branch will add 2 more settings into viewer settings, so it will likely to cause some conflicts as it'll try to write to deleted files!

Issue #1701, opened 2019-09-19
Pin Improvements 2


carmenfan:
- [x] Update unity with latest pins
- [x] Change all pin colours to just use the current selected colours
- [x] Call selectPin/deselectPin for pin change

requires https://github.com/3drepo/3drepounity/issues/232

PR #1708, opened 2019-09-24
Issue 1702 - pivot point 

carmenfan:


carmenfan:
This is overridden by #1701 merge.

PR #1713, opened 2019-09-27
Issue 1711 - Smart groups selecting descendants

### LINKED ISSUE #1711: Smart groups parent highlighting
### Description
When a transform matches the rule it should be selected and if there is something unmatching its descendants it should select everything but the unmatching descendants.

sanmont3drepo:
This fixes #1711


backend/models/group.js
```diff
@@ -785,46 +776,168 @@ function findObjectsByQuery(account, model, query) {
...
+	const eachRuleResults = await Promise.all(ruleQueries.map(ruleQuery => getRuleQueryResults(account, model, idToMeshesDict, history.current, ruleQuery)));
+
+	const allRulesResults = multiIntersection(eachRuleResults);
+
+	const ids = [];
+	for (const id of allRulesResults) {
+		ids.push(utils.stringToUUID(id));
+	}
+
+	return await idsToSharedIds(account, model, ids, convertSharedIDsToString) ;
```
carmenfan: shared_ids are guaranteed to be unique per object within the same revision.

If you keep all your results in shared_ids instead of _ids, then you wouldn't need this conversion?

backend/models/group.js
```diff
@@ -785,46 +776,168 @@ function findObjectsByQuery(account, model, query) {
...
-			return Promise.resolve([]);
+		} catch(e) {
+			reject(e);
 		}
 	});
+
+	return idToMeshes;
+}
+
+function batchPromises(promiseGenerator, dataToBatch, size) {
```
carmenfan: I would probably put this function in the helper folder as it has nothing to do with groups and potentially useful for some other modules

backend/models/group.js
```diff
@@ -785,46 +776,168 @@ function findObjectsByQuery(account, model, query) {
...
-						});
+/**
+ *
+ * @param {string} account
+ * @param {string} model
+ * @param {string} revId
+ */
+async function getIdToMeshesDict(account, model, revId) {
+	const treeFileName = `${revId}/idToMeshes.json`;
+	const {readStream: fileStream} = await FileRef.getJSONFileStream(account, model, treeFileName);
```
carmenfan: Would it be better/faster to use getJSONFile instead of the streaming version? (No idea myself, genuine question)

backend/models/group.js
```diff
@@ -152,6 +154,8 @@ groupSchema.statics.ifcGuidsToUUIDs = function (account, model, ifcGuids, branch
 groupSchema.statics.uuidToIfcGuids = function (obj) {
 	const account = obj.account;
 	const model = obj.model;
+
+	// @ts-ignore
```
carmenfan: ts-ignore?! backend uses eslint, i don't think this needs to be here? doesn't trigger linter on my local repo.

sanmont3drepo: I think is because I have something wrong in my visual studio condifguration. Ill get rid of it.

backend/models/group.js
sanmont3drepo: cant be done

backend/models/group.js
sanmont3drepo: done

backend/models/group.js
sanmont3drepo: done

PR #1725, opened 2019-10-07
Issue 1720 - Show update dialog

### LINKED ISSUE #1720: Service worker caching problem
@simplypixi this is from #1475

>  I'm currently seeing this problem when we have this branch on a dev server
> 
> We deployed this branch at commit A
> We then did further fixes to the branch (let's call this commit B)
> We then deployed commit B onto the server
> When I load the page i'm seeing commit A's version. Then I did a hard refresh and it becomes commit B.
> 
> This sounds very much like the service worker loading the cache version, which is why we had that "there's a new version of 3D Repo" dialog that ask if the user wants to refresh to get the latest version.
> 
> If this is what's happening, can we make it so it somehow doesn't load the cache version if it's out of date? if not, can we revive that dialog so the user knows and is given the option to refresh?

dleszcz:
This fixes #1720

### DIFF
--- a/frontend/modules/dialog/dialog.redux.ts
+++ b/frontend/modules/dialog/dialog.redux.ts
@@ -24,9 +24,9 @@ interface IDialogConfig {
 	title: string;
 	template?: JSX.Element;
 	content?: string;
+	data?: any;
 	onConfirm?: () => void;
 	onCancel?: () => void;
-	data?: any;
 }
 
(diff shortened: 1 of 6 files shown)


Charence:
@dleszcz @simplypixi I think it would be better for the changelog link to open in a new tab. Otherwise, good!

dleszcz:
> @dleszcz @simplypixi I think it would be better for the changelog link to open in a new tab. Otherwise, good!

Done!

Issue #1738, opened 2019-10-11
Issues/Risks management board 


dleszcz:
Choosen library: https://rcdexta.com/react-trello

[Issue Management Proposal.pdf](https://github.com/3drepo/3drepo.io/files/3717476/Issue.Management.Proposal.pdf)

- [x] Install & configure library
- [x] Create redux store
- [x] Create new route for board
- [x] Drag and drop issue to other line 
- [x] Add link to Trello board on model item card
- [x] Create new menu item on dashboard
- [x] Prepare data to use in trello library
- [x] Create Trello board 
    - [x] Switch between issues & risks
    - [x] Switch between another teamspace/project/model
    - [x] Search by phrase
    - [x] Filter buttons by: Status, Priority, Type, Assigned to, Created By
    - [x]  Filter menu (like on viewer)
    - [x] Create "new issue" button
    - [x] New issue form 
    - [x] Issue details popup 
    - [x] Add other issue options 
    - [x] Implement empty and default states
    - [x] Create single issue item

dleszcz:
Hello @BDavydov !

Two things:
1. I cant see Status & Priority fields in Risks. There are only in issues now. Should this be added in the app? Or You want to filter by another properties in Risks?
2. I think there is no reason to edit the "Created By" property. What is the use case?

BDavydov:
Hello @dleszcz 
Thank you for your question.

1. Can we filter risks by the following properties: 
- Level of Risk
-Level of Mitigated Risk
-Category 
-Mitigation Status
-Risk Owner

2."Created By" should automatically be filled out with the account name of a person creating an issue. Ideally, it mustn't be changed in order to trail the source of an issue/risk.

carmenfan:
> 2."Created By" should automatically be filled out with the account name of a person creating an issue. Ideally, it mustn't be changed in order to trail the source of an issue/risk.

To clarify @BDavydov 's response. We want to see it sorted in created by, but the user should not be able to move the cards around.

dleszcz:
Hello @carmenfan @Charence ,
In the form levels aren't updated directly. They are calculated on _likelihood_, or _consequence_ change.

What about updating _level_of_risk_, or _residual_level_of_risk_ on Board?
Now, we get this response:

![Screenshot 2019-10-28 at 14 08 06](https://user-images.githubusercontent.com/17221890/67680839-6793fc00-f98c-11e9-934a-fff60fb280de.png)

[...]

PR #1760, opened 2019-10-29
ISSUE #1738: Issues/Risks Management Board

dleszcz:
ISSUE #1738 
Choosen library: https://rcdexta.com/react-trello

[Issue Management Proposal.pdf](https://github.com/3drepo/3drepo.io/files/3717476/Issue.Management.Proposal.pdf)

- [x] Install & configure library
- [x] Create redux store
- [x] Create new route for board
- [x] Drag and drop issue to other line 
- [x] Add link to Trello board on model item card
- [x] Create new menu item on dashboard
- [x] Prepare data to use in trello library
- [x] Create Trello board 
    - [x] Switch between issues & risks
    - [x] Switch between another teamspace/project/model
    - [x] Search by phrase
    - [x] Filter buttons by: Status, Priority, Type, Assigned to, Created By
    - [x]  Filter menu (like on viewer)
    - [x] Create "new issue" button
    - [x] New issue form 
    - [x] Issue details popup 
    - [x] Add other issue options 
    - [x] Implement empty and default states
    - [x] Create single issue item


carmenfan:
@dleszcz @simplypixi @m1chalkubiak  fyi I added a couple more after you guys edited the post! 😄

dleszcz:
> @dleszcz @simplypixi @m1chalkubiak fyi I added a couple more after you guys edited the post! 😄

OK!

carmenfan:
> @carmenfan
> For board items (issue/risk) and edit/add forms we are using reusable components (we use the same ones in viewer issues/risks lists).
> So, every design change from your feedback will be visible also in viewer. This is ok for you? Or, you want to use another styles on Board?

Which one would be affected? for e.g. the comments part already has a grey background so i'm not entirely sure which ones you mean.

The screenshot element I think can change the same way, obviously the go to model icon shouldn't exist in the viewer as it won't make any sense..

dleszcz:
> Which one would be affected? for e.g. the comments part already has a grey background so i'm not entirely sure which ones you mean.

I mean:
> All the text can be smaller, it's a bigger than the font size we're using on teamspaces/user management etc
> Can we have ellipsis on overflow text (mostly used for title and descriptions) and the tooltip should show what the full text is:

carmenfan:
> All the text can be smaller, it's a bigger than the font size we're using on teamspaces/user management etc

I was more referring to things like the drop down menus and column headers etc. Let's see how that looks first :)

> Can we have ellipsis on overflow text (mostly used for title and descriptions) and the tooltip should show what the full text is:
Yes that's definitely useful on both

carmenfan:
@simplypixi @m1chalkubiak 
Can we allow the user to markup the image uploaded just like the screenshots taken in the viewer?

simplypixi:
@carmenfan 
> Can we add a cross on the top left of the following to allow the user to close these windows:
Yes, we can, but I think it should be a global change. What do you think?

> Can we also add the go to model button on the cards:
Are you sure the button should be placed on each issue? Whole board is related to specific model so maybe the best place to put this button is panel's header or a place near model's dropdown - let me know what you think. ;)

simplypixi:
> Can we get a bit more feedback when the user is not allowed to move the card? like for 2 seconds put an overlay on the card to say Insufficient permissions to perform this action

I think it's not possible without a "hacky" solution because the library doesn't emit any drag event if dragging is not active.

simplypixi:
> @simplypixi @m1chalkubiak
> Can we allow the user to markup the image uploaded just like the screenshots taken in the viewer?

It's feasible but we shouldn't do that before merging new markup tools because of potential conflicts or unpredictable bugs (in the worst case).

carmenfan:
@simplypixi 
> Yes, we can, but I think it should be a global change. What do you think?

Yes i think that would work well!

> Are you sure the button should be placed on each issue? Whole board is related to specific model so maybe the best place to put this button is panel's header or a place near model's dropdown - let me know what you think. ;)

Yes it should be in the issue cards, because the go to model link shouldn't just go to the model, it should go into the specific issue (i.e. the url should be /viewer/ts/model?issue=issue_id or whatever the format is to link to issue)

carmenfan:
found another bug(ish):

- [ ] Doing this should not trigger a issue update:
markup

Some UI stuff:
- [ ] Can we add an outline here?
image

- [ ] The text of the drop down is a bit misaligned with the text next to it:
image

- Can we use the same component for the drop downs as user management? so we get the header on the top, e.g.:
image

- Can we change `Sort by` to `Group by`?
image

carmenfan:
> Can we get a bit more feedback when the user is not allowed to move the card? like for 2 seconds put an overlay on the card to say Insufficient permissions to perform this action
> 
> I think it's not possible without a "hacky" solution because the library doesn't emit any drag event if dragging is not active.

What about we unable dragging and just bounce it back and give a message saying the error?

As it stand it just looks like it bugged out. I'm just trying to think what we can do to give the user a bit more feedback 😆

simplypixi:
> > Can we get a bit more feedback when the user is not allowed to move the card? like for 2 seconds put an overlay on the card to say Insufficient permissions to perform this action
> > I think it's not possible without a "hacky" solution because the library doesn't emit any drag event if dragging is not active.
> 
> What about we unable dragging and just bounce it back and give a message saying the error?
> 
> As it stand it just looks like it bugged out. I'm just trying to think what we can do to give the user a bit more feedback 😆

Yep, sounds good. I thought about the same solution, to be honest. ;)

simplypixi:
> * Doing this should not trigger a issue update:

I already fixed that a few hours ago. ;D

simplypixi:
@carmenfan Done :)

carmenfan:
@simplypixi 

- [ ] Alignment issues
image
I don't think this was addressed fully 😆 Can we make it so the 4 cards pretty much occupy the whole space? (Now with 4 tabs there's quite a lot of gap at the side). The spacing should be more like this:
image

- [ ] Can we hide the scroll bars by default :
image

- [ ] Can we make the columns make use of the whole height of the panel:
image

- [ ] Can we make these buttons appear on hover?
image

- [ ] Can we make those `go to model link` opens on a new tab with  `rel=noopener`

- [ ] Can we change the header to Model/Federation
image

- [ ] Still getting this problem when i'm not an admin
image
Seems like it's trying to call the quota API. This drop down should be populated with the same data as gridView. But it seems like it's doing a lot of unnecessary API requests? It shouldn't need to call anything at all upon selection. (definitely don't need quota/member etc)
image

- [ ] Can we add some unselectable headers on this drop down to differentiate models and federations?
image

- [ ] If you flip between issues -> risk -> issues, it crashes:
markup
This seems to be the error:
image

simplypixi:
@carmenfan 

> Still getting this problem when i'm not an admin

How it's possible that you were able to choose teamspace with non-admin access? :O

> Seems like it's trying to call the quota API. This drop-down should be populated with the same data as gridView. But it seems like it's doing a lot of unnecessary API requests? It shouldn't need to call anything at all upon selection. (definitely, don't need quota/member etc)

Yep, calling quota API is unnecessary, but we still need to make a `GET members` request to get users' data (first name, last name). We use them to show proper labels of lanes if grouping by Owner (Risks) is selected. Do you want to show the teamspace name (user login) instead? 

> Can we put the cross on the top left?

Hmmm... You requested that position before. :P I think, you thought about top right corner since the beginning, weren't you? :D

carmenfan:
> How it's possible that you were able to choose teamspace with non-admin access? :O

You should be able to as this is the same as just viewing issues within a model. As long as you have at least a viewer access you can see issues. So we cant' make the assumption that only admins view this board.

> We use them to show proper labels of lanes if grouping by Owner (Risks) is selected. Do you want to show the teamspace name (user login) instead?
Risk owner should grouped by it's job, not the name

>Hmmm... You requested that position before. :P I think, you thought about top right corner since the beginning, weren't you? :D

Yep 😆

simplypixi:
> Risk owner should grouped by it's job, not the name

It sounds like `created_role` field (which is implemented as Created By), not `owner` (which you requested last time). :P
Currently, BE returns user's login as `owner` ;)

image

image

carmenfan:
> > Risk owner should grouped by it's job, not the name
> 
> It sounds like `created_role` field (which is implemented as Created By), not `owner` (which you requested last time). :P
> Currently, BE returns user's login as `owner` ;)
> 
> image
> 
> image

Ah that one. What I meant is risk owner on the UI, which equates (confusingly) assigned_roles in the backend:
image

simplypixi:
@carmenfan Everything should be fixed ;)

PR #1782, opened 2019-11-06
ISSUE #1781 - Allow selecting objects on the model only after tree wa…

### LINKED ISSUE #1781: Error message if the user selects something on the model before the tree has finished loading
At the moment the user clicks on an object and an error message pops up if the tree hasn't finished loading:
image

I think selecting objects should just be ignored until the tree is loaded... 

(Blackfriars federation is a good test case to trigger this bug)

m1chalkubiak:
This fixes #1781

#### Description
-  Allow selecting objects on the model only after tree was been loaded

### DIFF
--- a/frontend/modules/tree/tree.sagas.ts
+++ b/frontend/modules/tree/tree.sagas.ts
@@ -31,6 +31,7 @@ import {
 	selectGetNodesByIds,
 	selectGetNodesIdsFromSharedIds,
 	selectIfcSpacesHidden,
+	selectIsTreeProcessed,
 	selectNodesIndexesMap,
 	selectSelectionMap,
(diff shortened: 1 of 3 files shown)


m1chalkubiak:
> This is working, but can you also look into this:
> image
> As the tree is still loading it should say "Loading tree.. " at the bottom. This is supposably done in [a0db6f6](https://github.com/3drepo/3drepo.io/commit/a0db6f62baf1273d305d8167f84d03e3a9733ad8)

@carmenfan We did have information regarding processing of tree, but there was no rendered information regarding that - for the user. I added "Loading tree..." message and now we are showing it when it is needed.

PR #1783, opened 2019-11-07
Issue 1779: Minor UI Issues

### LINKED ISSUE #1779: Minor UI Issues
- [x] **Helicopter Bug**
We discovered in 4.0.0 release that the API call to change helicopter speed was sending teamspace and model ID as `undefined`. It looks like the props did not contain the right information regarding teamspace and model ID. This is worked around with dac6bc08cc0acd547ca6ee3f2b1708886a6746ae but we need to make it work the way it should.

- [x] **[Object Object] API calls when you go try to view a federation model**
It only seems to happen with federations, but when you enter the viewer, you will see an API call with an incorrect URL: 
image

- [x] **Model code not updated in gridview**
If you go into model settings and update the model code, it doesn't appear until you refresh the page
markup

- [x] **Make a footer on model settings**
Can we have the save button on a footer so it's always shown without the need to scroll?
image

- [x] **On sign up, the frontend allows hyphens on the username but backend doesn't**
Backend uses this Regex:  `/^[a-zA-Z][\w]{1,63}$/`, which is alphanumeric and underscore. But frontend currently says hypens are allowed
markup

- [x] **Change this to a user friendly error**
I think majority of our userbase would not know what this means... Can we change it to a more friendly message?
image

- [x] **Remove default group description when the user focuses**
Just like the group name, the default description should also disappear when the user focuses on the text box
markup

- [x] Can do we do the same on views (see line above)?
image

simplypixi:
This fixes #1779

### DIFF
--- a/frontend/helpers/issues.ts
+++ b/frontend/helpers/issues.ts
@@ -22,8 +22,7 @@ import { hasPermissions, isAdmin, PERMISSIONS } from './permissions';
 export const prepareIssue = (issue, jobs = []) => {
 	const preparedIssue = {...issue};
 	if (issue.thumbnail) {
-		const thumbnail = getAPIUrl(issue.thumbnail);
-		preparedIssue.thumbnail = thumbnail;
+		preparedIssue.thumbnail = issue.thumbnail && issue.thumbnail.length ? getAPIUrl(issue.thumbnail) : '';
 	}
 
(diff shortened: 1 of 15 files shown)


simplypixi:
> * When editing the project name  and the user leaves a space in the end the validation signals that is an incorrect project name, nevertheless the error message doesnt clarify this.
>   We should allow the user write the name and just trim it when saving it or clarify it in the message.
> 
> image

Fixed

PR #1834, opened 2019-12-09
ISSUE 1833 - Callouts feature

### LINKED ISSUE #1833: Callouts feature
### Description
Callouts

Adding the Callouts feature into screenshot dialog, allowing the user to create additional notes in screenshots.

### Goals
- [ ] User can create callouts notes on the screenshot dialog

### Tasks
- [ ] A new Callout tool

Reference animation:

m1chalkubiak:
This fixes #1833 

#### Description
- Allow adding callouts notes on the screenshot dialog


m1chalkubiak:
Hi @carmenfan 
I've prepared the initial version of the new tool, which still has a couple of obvious bugs, but during development I came to a few thoughts that I would like to discuss with you.

In the reference animation we have several shapes on toolbox to choose from. I was trying to figure out how we can dynamically draw line from shape to text box. But its appear that library doesn't help us with that, the only solution that is suggested in documentation is based on calculated the position manually. 
It was quite easy to do for simple shapes like a circle or rectangle, but it is not easy for custom shapes and can create a lot of bugs in case of some shapes.
Therefore, I would like to suggest simplifying this tool only for the function of drawing from a specific point. Drawing shapes would be as before and at the end you could add a note anywhere. 

I am also not sure about showing text area box, this wasn't implemented before, and after playing with it I see that it also can be hard to achieve.

carmenfan:
> In the reference animation we have several shapes on toolbox to choose from. I was trying to figure out how we can dynamically draw line from shape to text box. But its appear that library doesn't help us with that, the only solution that is suggested in documentation is based on calculated the position manually.
It was quite easy to do for simple shapes like a circle or rectangle, but it is not easy for custom shapes and can create a lot of bugs in case of some shapes.

That's probably fine, but i think custom polygons shouldn't be too hard, if you treat every 2 points to be a line, then it's just calculating the intersection between 2 vectors, and take the one with the minimum distance. Cloud is probably the trickiest one, we can calculate the distance as if it's a rectangle, but if we draw the line from the bounding box then there would be a gap between the shape and the line...

> I am also not sure about showing text area box, this wasn't implemented before, and after playing with it I see that it also can be hard to achieve.

Not sure what you mean here, you mean having a rectangular box around the text? it looks like you've already done this?

**Things I spot:**
- [ ] Intersection point isn't calculated correctly on rectangles:
markup
- [ ] There's a split second where the text appears below the box then jumps into the box:
markup
- [ ] The text box should have no fill, instead of white
- [ ] Tool bar jumps into text mode when you've created the shapes
- [ ] Call out shapes should just be part of the shapes toolbar instead of having it's own icon
- [ ] We need better icons, @BDavydov  can you take a look? (point, circle, rectangle)
- [ ] I can't/don't know how to change the text on these shapes, it should just immediately allow me to start editing the text once I've created the shape.
- [ ] Undo/redo should just be a single step for the whole shape instead of textbox->line->shape
- [ ] Text box size should reshape depending on the font size: 
image
- [ ] Text is appearing in the wrong place when it's in edit mode
markup


Also my chrome crashed 3 times trying to generate images for this post (Seems to happen when i'm creating text). I'm not sure what's going on but maybe the changes on this branch has created some instability somewhere.

m1chalkubiak:
@carmenfan 
> * [ ]  The text box should have no fill, instead of white. Probably useful for it to have a white filling while it's hovering around, but once committed it needs to be transparent or a very translucent white.

For now, I left it with a very delicate background so as not to complicate the implementation. If this suits you, we could leave it so, if not, I will try to do it in a different way.

> * [ ]  Call out shapes should just be part of the shapes toolbar instead of having it's own slot on the main toolbar
> * [ ]  We need better icons, @BDavydov  can you take a look? (point, circle, rectangle)

If they should be in the same toolbar as the shapes we would need other icons for sure, that's why for now I left it in a separate toolbar.

> Also my chrome crashed 3 times trying to generate images for this post (Seems to happen when i'm creating text). I'm not sure what's going on but maybe the changes on this branch has created some instability somewhere.

I made a large refactor of these components, previously managing all the tools were in one large file, it seems to me that it could cause performance problems. Now all tools have their separate components, it will be easier to develop them and they should work more stably.

It would be good though to thoroughly test the whole drawing again.

BDavydov:
@m1chalkubiak 

Hello Michael, 
Please find attached a ZIP folder with callout icons. 
[Callout Icons.zip](https://github.com/3drepo/3drepo.io/files/3984545/Callout.Icons.zip)

carmenfan:
@m1chalkubiak in general i think it's workng pretty well now.

There's still a bit of a problem with the line on the rectangle shape:
markup

As you can see sometimes the line is crossing the rectangle area

How hard is it to make it work like the circle? if we see the rectangle as 4 lines then it's just a line/vector intersection calculation x4 then take the shortest distance? I think that may work nicer.

carmenfan:
@m1chalkubiak  I ran in to a [travis test error ](https://travis-ci.org/3drepo/3drepo.io/builds/632341899?utm_medium=notification&utm_source=email)on another branch which seems to be caused by the redux-undo version incompatibility.

It looks like they've released a new version which changed the function definition. I've temporary fixed it on the other branch by fixing the version, but since we're using a beta version it may worth looking at updating?

m1chalkubiak:
Hi @carmenfan,
I've come back from my holidays yesterday and immediately started working on Your feedback. 
I totally agree with updating the redux-undo, if there is a released stable version we should use it.
New custom icons for callout shapes were added and moved to the toolbar with the others.
According to your suggestion, I rewrote callout line drawing for rectangle shape, now it will work more similar as with the circle shape.

PR #1841, opened 2019-12-11
ISSUE #1840 create a script to remove unverified users with expired tokens

### LINKED ISSUE #1840: create a script to remove unverified accounts with expired tokens

carmenfan:
This fixes #1840

a mongo shell script that removes anyone with `inactive` flag and an expired email token.


keifgwinn:
This has been set up on development and production to run sunday mornings,
04:05am.

PR #1842, opened 2019-12-13
Issue #1837 - Markup fixes

### LINKED ISSUE #1837: Markup bugs
- [x] Horizontal scroll bar appearing on markup. It should scale to fit without the need to scroll:
markup

- [x] Cloud shape thickness is dependant on size, unlike other shapes. The following shapes are all drawn with the same thickness, as you can see the rectangle retains the line thickness whereas the cloud just varies depending on the area drawn. The cloud shape should behave like rectangle:
image

- [x] Text edit is not at the same place as where I clicked 
markup

- [x] Changing Shape size is effecting font size
markup

- [x] Remove the cancel button. We have a X on top right, this is enough.

- [x] "Sample Text" should not be a stage in the undo/redo (unless the user committed it with sample text)
markup

- [x] Can we change `Sample Text` to `Type to change text...`

m1chalkubiak:
This fixes #1837

All described bugs should be fixed right now. I also rewrote the way that text components are styled and react on typing. Now, created textarea has a placeholder, so no additional step in history will be created. During typing it will start expanding according to the text length (without any limitation for now) and creating of new line is available under enter + shift combination.


m1chalkubiak:
@carmenfan 
I think that finally, I managed to deal with the issue of history text changes, although I did not manage to change the logic of text creation, I came up to how to group actions that we consider as a single step in the history. This solution may be useful to us in the future, even in the case of the callout feature.

PR #1845, opened 2019-12-17
Issue 1820 - Opacity support for groups

### LINKED ISSUE #1820: Add option to set transparency for groups colour override
### Description
Allow users to set an optional transparency setting, which will be applied on groups override

See https://marvelapp.com/80ef7f5/screen/64222126 for mock up

### Tasks
- [x] Add unity-utils functions to change objects transparency value (Carmen)
- [x] Add transparency value into groups data schema (should be separate to colour, as it is optional)
- [x] Add set transparency UI to colour picker (should only be visible for groups for now). Default setting when enabled should be set to 100% (opaque)
- [x] If possible, it'd be nice to show the colour swatch being affected by the transparency value

sanmont3drepo:
This fixes #1820


carmenfan:
@sanmont3drepo  Can merge once Travis is happy with it

PR #1858, opened 2020-01-13
ISSUE #1827 Adjust height setting functionality for left panels in viewer

### LINKED ISSUE #1827: height setting is not functioning correctly
When more than 1 card is shown, the height should be split equally. However as it stands it's it not the case: 

image
image

image

image

The height shares seems to be very non deterministic.

Can we change it to only allow 2 cards open at one time. When a 3rd card opens, the oldest card gets replaced (unless it's pinned?)

m1chalkubiak:
This fixes #1827

#### Description
- 1 card: make use of full height
- 2 cards: share height at 50:50
- 3rd card: top card gets removed and new card adds to the bottom (Queue style)
- pinned card always goes on top
- new card will remove non pinned card instead of the pinned card.
- If another card is pinned then the prev pinned card will be automatically unpinned
- pinned card can only go away if the user unpinned it or click on the left menu to close it


m1chalkubiak:
Hi @carmenfan 
I made one small change regarding our original agreement. I used a different icon for pinning action.
Honestly, In the beginning, I did not notice that you provided the name of the icon from Font Awesome and I found a different suitable icon with I used.
It's a lock icon with have quite visible two-state. In my opinion, these icons are easier to read than the rotated thumbtack.

Let me know what You are thinking about this change and the whole functionality.

carmenfan:
@m1chalkubiak I think the lock icon is fine. The logic seems to be working too.

Couple of things:

- View card is missing a line after the last entry, that's very tell now since now longer scale the tab
image

- The lock icon is in a different place depending whether the panel has search/hamburger button...
image
I think we should have every icon has a set location and just hides if that panel doesn't support the function, With less padding between them. Something like this?
image
image

PR #1861, opened 2020-01-17
ISSUE 1855 Remove unnecessary issue API calls

### LINKED ISSUE #1855: issue API is being called 3 times when you go into issue details
when you click into an issue, it seems to call the issues API 3 times:
image

The third time it's not using the correct model ID if it is a sub model issue. We should take a look to see why.

m1chalkubiak:
This fixes #1855

#### Description
- [x] Remove unnecessary issue API calls while you click into an issue

### DIFF
--- a/frontend/modules/issues/issues.sagas.ts
+++ b/frontend/modules/issues/issues.sagas.ts
@@ -422,7 +422,6 @@ function* setActiveIssue({ issue, revision, ignoreViewer = false }) {
 		if (issue) {
 			const {account , model, _id} = issue;
 			yield put(IssuesActions.subscribeOnIssueCommentsChanges(account, model, _id));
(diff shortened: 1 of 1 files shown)


m1chalkubiak:
@carmenfan I double-checked it but this is exactly what I did in this PR. I removed fetching of data on selecting issue item on the list, now this fetching is done only when we go to issue details.

Please, wouldn't it be a problem if you could check it again?

carmenfan:
> @carmenfan I double-checked it but this is exactly what I did in this PR. I removed fetching of data on selecting issue item on the list, now this fetching is done only when we go to issue details.
> 
> Please, wouldn't it be a problem if you could check it again?

@m1chalkubiak Yep you're right, it's working!

PR #1865, opened 2020-01-20
ISSUE #1863 Fix layout issue

### LINKED ISSUE #1863: Last user missing on the project permissions tab
- Add enough users to the teamspace so that you have to scroll to see the last user
- Go to the projects tab, enter any project
- The last user is missing , also the 2nd from last user is showing through the footer

markup

m1chalkubiak:
This fixes #1863

#### Description
- Fix layout issue on Projects permissions tab
- Fix fetching jobs and colors on Jobs tab

### DIFF
--- a/frontend/modules/jobs/jobs.redux.ts
+++ b/frontend/modules/jobs/jobs.redux.ts
@@ -1,5 +1,5 @@
 /**
- *  Copyright (C) 2019 3D Repo Ltd
+ *  Copyright (C) 2020 3D Repo Ltd
  *
  *  This program is free software: you can redistribute it and/or modify
  *  it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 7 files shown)


carmenfan:
@m1chalkubiak  can you spare some time to take a look at this so we can merge it? :)

thanks!

m1chalkubiak:
@carmenfan
Of course, sorry I missed this comment before. I will look at this right away.

m1chalkubiak:
@carmenfan 
Got a fix for the crashing on this. But also I see that in color picker we are getting values in RGBA format but in fact, without an easy way of manipulating an Alpha value, only manual changing of value is available. Don't know if this should work like that, it isn't clear for me.

carmenfan:
> @carmenfan
> Got a fix for the crashing on this. But also I see that in color picker we are getting values in RGBA format but in fact, without an easy way of manipulating an Alpha value, only manual changing of value is available. Don't know if this should work like that, it isn't clear for me.

Alpha value support on the colour picker was added by @sanmont3drepo for https://github.com/3drepo/3drepo.io/issues/1820.
Alpha should only be editable on the groups right now, as there's not a lot of meaning to job colours to have an alpha value

m1chalkubiak:
@carmenfan 
Oh, I see, but from what I had check just a moment age the Alpha part is showed when we switch to it on color picker in the editing of groups. On jobs we always see the aplha value. So, is this a bug?

carmenfan:
@m1chalkubiak yes it is, @sanmont3drepo  will havea  look. Thanks for spotting!

sanmont3drepo:
> @m1chalkubiak yes it is, @sanmont3drepo will havea look. Thanks for spotting!

done!

PR #1872, opened 2020-01-28
Issue 1792 - Sequence viewer

### LINKED ISSUE #1792: Sequences support
### Description

Model sequences support

wireframe: https://marvelapp.com/9hh1aa6

### Goals

- [x] Tests
- [x] List all sequences
- [x] Update backend for new schema
- [x] UI milestone 1

sanmont3drepo:
This fixes #1792


frontend/modules/tree/tree.sagas.ts
```diff
@@ -554,13 +555,37 @@ function* zoomToHighlightedNodes() {
...
+			ov.unhidden[key] = currentOverrides[key];
+		}
+
+		return ov;
+	} , {hidden: [], unhidden: {}});
+
+	yield put (TreeActions.hideNodesBySharedIds([{shared_ids: overrides.hidden}]));
+	yield addTransparencyOverrides(overrides.unhidden);
+
+	// yield hideTreeNodes(overrides.hidden, true);
```
carmenfan: Remove commented out code

frontend/modules/store.ts
```diff
@@ -54,3 +58,14 @@ export const store = configureStore();
 export const dispatch = store.dispatch;
 
 export const getState = store.getState;
+
+if (IS_DEVELOPMENT) {
```
carmenfan: Do we still want this here or was it just for development?

It's fine either way, just checking!

Charence:
- [ ] Behaviour when switching between sequences - if the date ranges of 2 sequences do not match and the current time being viewed does not exist in the other sequence, I'm not sure if the current behaviour of the play button is clear enough. E.g. we're at `time = 20` and the range for seq. A is `0 - 10`, pressing PLAY the first time seems to take you to the end of the range, so `time = 10`, then you need to press again to make the sequence play from `0`. I'm not sure if it's obvious to the user why the 1st click of PLAY doesn't make the sequence animate.

Mar-04-2020 17-33-12

- [ ] Play bar doesn't seem to have equal padding on both sides.

Screenshot 2020-03-04 17 22 42
Screenshot 2020-03-04 17 22 48

- [ ] When the end of sequence is reached. I wonder if it will be better to show a REPLAY button rather than a PLAY button. Something like this: https://material.io/resources/icons/?icon=replay&style=baseline

backend/models/sequence.js
```diff
@@ -0,0 +1,92 @@
...
+			sequence.teamspace = account;
+			sequence.model = model;
+
+			if (cleanResponse) {
+				this.clean(sequence);
+			}
+		});
+
+		const submodelSequences = await Promise.all(submodels.map((submodel) => this.getList(account, submodel, "master", null, cleanResponse)));
+		submodelSequences.forEach((s) => sequences.push.apply(sequences,s));
```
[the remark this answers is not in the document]
sanmont3drepo: used the first option

Issue #1797, opened 2019-11-15
Incorrect revision count update


Charence:
### Description
At the moment, 2 + 1 = 5. May be caused by too many upload success notifications?

### Steps to replicate
- Upload model revision
- Watch the revision count

### Pictures/ Gifs of the error
![Nov-15-2019 10-45-10](https://user-images.githubusercontent.com/1438332/68938218-cda6be80-0795-11ea-831d-990d598eb169.gif)

carmenfan:
I think this is related to #1770, which are both related to the confusing logic around what is a model settings update & unreliability of the socket io hook

Issue #1770, opened 2019-11-04
timestamp does not always update when the upload is finished


carmenfan:
### Description
Timestamp sometimes doesn't update upon a successful upload

### Steps to replicate
- Create a new model
- Upload a revision
- wait until it's processed
- Sometimes, the timestamp doesn't appear (It seems like the change notification sometimes happens, sometimes not)

PR #1874, opened 2020-02-03
Issue 1866 - Block simultaneous login

### LINKED ISSUE #1866: Block simultaneous login
### Description
Block simultaneous login

### Goals
- Block simultaneous login for the same user when accessing through a browser
- Allow simultaneous login for desktop/mobile client
- Use chat to notify unseated session that they have logged in elsewhere

### Tasks
- [x] Identify login session source (web or other)
- [x] Invalidate older web sessions for user
- [x] Establish chat channel for session based on session ID
- [x] Notify user their session has been terminated

sanmont3drepo:
This fixes #1866 , #1770 and #1797


backend/routes/auth.js
```diff
@@ -543,22 +547,45 @@ router.put("/:account", middlewares.isAccountAdmin, updateUser);
...
 		req[C.REQ_REPO].logger.logInfo("Creating session for " + " " + user.username);
 		if(err) {
 			responseCodes.respond(place, responseCodes.EXTERNAL_ERROR(err), res, {username: user.username});
 		} else {
 			req[C.REQ_REPO].logger.logDebug("Authenticated user and signed token.");
 
-			req.session[C.REPO_SESSION_USER] = user;
-			req.session.cookie.domain				 = config.cookie_domain;
+			user.socketId = req.headers[C.HEADER_SOCKET_ID];
+			user.webSession = false;
```
carmenfan:
```suggestion
			user.webSession = req.headers && req.headers["user-agent"]
```
I think that should work and we can ditch 560-562?

That sounds like a massive assumption though to say anyone who sends http request with user-agent filled in = browser.... @sanmont3drepo  any thoughts?

backend/routes/auth.js
```diff
@@ -543,22 +547,45 @@ router.put("/:account", middlewares.isAccountAdmin, updateUser);
...
 
-			responseCodes.respond(place, req, res, next, responseCodes.OK, {username: user.username, roles: user.roles, flags: user.flags});
+			if (user.webSession) {
+				const query = {
+					"_id": { $ne: req.session.id },
+					"session.user.username": user.username,
+					"session.user.webSession": true
+				};
+
+				db.getCollection("admin", "sessions").then(_dbCol => {
```
carmenfan: @sanmont3drepo is this how you'd do clear the session? considering we manage it through express-session, it feels  abit weird that we never store it but we manually delete it like this. Is there a function we're suppose to be using?

Also I think this should live in sessionManager? at the point of creating a new session, it removes the old ones? feels a bit coupled for it to be living in the routes.

backend/routes/auth.js
```diff
@@ -581,6 +608,7 @@ function login(req, res, next) {
 
 			req[C.REQ_REPO].logger.logInfo("User is logged in", responseData);
 
+			// TODO Is this necessary?
```
carmenfan: you know it's not :P Check if it's being used, if yes - change it  and remove this, if not - not remove it.

backend/services/chat.js
```diff
@@ -82,41 +91,66 @@ module.exports.createApp = function (server, serverConfig) {
...
-						socket.join(`${data.account}${modelNameSpace}`);
-						socket.emit(joinedEventName, { account: data.account, model: data.model});
-					} else {
-						socket.emit(credentialErrorEventName, { message: `You have no access to join room ${data.account}${modelNameSpace}`});
-						systemLogger.logError(`${username} - ${sessionId} - ${socket.client.id} has no access to join room ${data.account}${modelNameSpace}`, {
-							username,
-							account: data.account,
-							model: data.model
-						});
+				let sessionUsername = _.get(socket, "handshake.session.user.username");
```
carmenfan: const? looks like readonly at a glance

PR #1901, opened 2020-02-14
Issue 1817 - front end test infrastructure

### LINKED ISSUE #1817: Frontend tests
### Description
Add frontend testing to ensure a basic level of stability when deploying 

### Goals
- [x] Test should run before deploying
- [x] Sign in test when entering correct credentials
- [ ] Sign in test when entering incorrect credentials

### Tasks
- [ ] Add frontend test runner
- [ ] Run tests in travis
- [ ] Add sign in test when entering correct  credentials
- [ ] Sign in test when entering incorrect credentials

sanmont3drepo:
This fixes #1817


run/run_app_pm2
```diff
@@ -1,10 +1,8 @@
...
-    echo "No config argument supplied"
-else
-    pm2 stop all
-    pm2 delete all
-    NODE_ENV=$1 NODE_CONFIG_DIR="./config" pm2 start "./run/3drepo.yml"
-fi
\ No newline at end of file
+pm2 stop all
+pm2 delete all
+pm2 start "../run/3drepo.yml"
```
carmenfan: Does this needs changing? We lost the config details from the run script here.

Also it looks like you're not using this script anyway?

I think the original idea is that you should have pm2 installed globally. Servers def do (though I don't think servers rely on this script, it just runs 3drepo start/stop which outright calls pm2 i think @keifgwinn ?)

keifgwinn: 3drepo start uses ./run_app_pm2 and these changes would break the multi-core useage on the servers and also break the use of the config file

the servers have pm2 installed globally (currently, docker doesn't use pm2 yet)

carmenfan: In which case the change either has to revive the config setting, or just revert back to the old script @sanmont3drepo

PR #1925, opened 2020-03-09
ISSUE 1916 - Merging the static viewer and the viewer service

### LINKED ISSUE #1916: Merging the static viewer and the viewer service
### Description
Getting rid of currently separated into two large classes viewers, which largely overlap each other, by merging them into one.

### Goals
- [ ] Viewer should still work properly, in the same way that it was before merging
- [x]  ~~Render viewer's canvas only on the viewer page (if it will be possible)~~

### Tasks
- [x] Rid of two large viewer classes
   - [x] Separate the methods responsible only for controlling Unity to another class (in the future it probably would be possible to get rid of this class)
- [x] Remove all duplicate methods
- [x] Re-organize code by grouping methods by features
- [x] Remove all unnecessary (this will clear up during debugging and testing) classes added into the global scope.
   - [x]  ~~UnityUtils~~
   - [x] Pin
      - [x] Additionally remove old duplicated Pin class
   - [x] Viewer
      - [x] Remove old Viewers
- [x] ~~Move viewer canvas to viewer page route~~
   - [x] ~~Handle of properly destroying of Viewer class instance when leaving the route~~.

m1chalkubiak:
This fixes #1916 

#### Description
Getting rid of currently separated into two large classes viewers, which largely overlap each other, by merging them into one.


frontend/globals/unity-util.ts
```diff
@@ -308,7 +308,7 @@ export class UnityUtil {
 				}
 			});
 		} else {
-			UnityUtil.onReady().then(() => {
+			return UnityUtil.onReady().then(() => {
```
carmenfan: I don't think this should be returning?

frontend/helpers/colorOverrides.ts
```diff
@@ -75,7 +75,7 @@ export const addColorOverrides = (overrides) => {
 
 				for (let j = 0; j < modelsList.length; j++) {
 					const { meshes, teamspace, modelId } = modelsList[j] as any;
-					Viewer.overrideMeshColor(teamspace, modelId, meshes, color);
+					await Viewer.overrideMeshColor(teamspace, modelId, meshes, color);
```
carmenfan: This shouldn't need await, overrideMeshColor of different models shouldn't require one to be processed before another.

Not to mention the unity doesn't return a promise upon it's done.

frontend/helpers/colorOverrides.ts
```diff
@@ -103,7 +103,7 @@ export const removeColorOverrides = (overrides) => {
 
 				for (let j = 0; j < modelsList.length; j++) {
 					const { meshes, teamspace, modelId } = modelsList[j] as any;
-					Viewer.resetMeshColor(teamspace, modelId, meshes);
+					await Viewer.resetMeshColor(teamspace, modelId, meshes);
```
carmenfan: Same as above

frontend/modules/tree/tree.sagas.ts
```diff
@@ -357,7 +357,7 @@ function* isolateSelectedNodes({ nodeId }) {
 	if (nodeId) {
 		yield isolateNodes([nodeId]);
 		const meshes = yield TreeProcessing.getMeshesByNodeIds([nodeId]);
-		Viewer.zoomToObjects({entries: meshes});
+		return Viewer.zoomToObjects({entries: meshes});
```
carmenfan: return isn't necessary?

frontend/modules/tree/tree.sagas.ts
```diff
@@ -546,7 +546,7 @@ function* goToRootNode({ nodeId }) {
 function* zoomToHighlightedNodes() {
 	try {
 		yield call(delay, 100);
-		Viewer.zoomToHighlightedMeshes();
+		return Viewer.zoomToHighlightedMeshes();
```
carmenfan: return isn't necessary?

frontend/modules/viewer/viewer.redux.ts
```diff
@@ -22,6 +22,7 @@ export const { Types: ViewerTypes, Creators: ViewerActions } = createActions({
 	updateSettings: ['username', 'settings'],
 	updateSettingsSuccess: ['settings'],
 	fetchSettings: [],
+	updatePins: ['prev', 'curr'],
```
carmenfan: @sanmont3drepo  Does this make sense here? With the latest changes you've done with the pin fix

frontend/modules/viewer/viewer.sagas.ts
```diff
@@ -82,7 +83,20 @@ export function* fetchSettings() {
 	yield put(ViewerActions.updateSettings(username, currentSettings));
 }
 
+function* updatePins({prev, curr}) {
```
carmenfan: @sanmont3drepo

frontend/routes/viewerCanvas/viewerCanvas.component.tsx
```diff
@@ -65,13 +66,9 @@ export class ViewerCanvas extends React.PureComponent<IProps, any> {
...
 
 	public renderPins(prev, curr) {
-		const { viewer } = this.props;
-
-		const toAdd = pinsDiff(curr, prev);
-		const toRemove = pinsDiff(prev, curr);
-
-		toRemove.forEach(viewer.removePin.bind(viewer));
-		toAdd.forEach(viewer.addPin.bind(viewer));
+		if (this.shouldBeVisible) {
```
carmenfan: @sanmont3drepo I think you touched this part of the code. Make snre t's in the place that you envisaged

frontend/services/viewer/viewerService/viewerController.tsx
```diff
@@ -0,0 +1,666 @@
...
+		}
+	}
+
+	/**
+	 * Pins
+	 */
+
+	public addPin = async ({account, model, id, type, pickedPos: position, norm, colours, viewpoint, isSelected}) => {
+			await this.isViewerReady();
+			this.pins[id] = new Pin({ id, type, position, norm, colors: colours, account, model });
```
carmenfan: I don't think the pin file needs to exist, just pass the info straight to unity. Same with any oother pin functions here that calls this.pins...
There's not a lot of point having an instance of the pin in the viewer unless it's meaningful, and the state of the pins seems to be captured outside of this anyway.

@sanmont3drepo do you agree?

m1chalkubiak:
I see that after the last merge, something went wrong, I'm working on it right now.

frontend/services/viewer/viewerService/viewerController.tsx
```diff
@@ -0,0 +1,666 @@
...
+			delete this.pins[id];
+		}
+	}
+
+	public removeUnsavedPin() {
+		this.removePin({ id: this.newPinId });
+		this.setPin({ data: null });
+	}
+
+	public highlightPin(id) {
```
carmenfan: I don't think we use this? I think the highlight state isn't kept here and we just outright call change color.

I think this can be removed (but check the dependency!)

frontend/services/viewer/viewerService/viewerController.tsx
```diff
@@ -0,0 +1,666 @@
...
+		if (on) {
+			MultiSelect.toggleAreaSelect(false);
+		}
+	}
+
+	public setPin(data) {
+		this.pinData = data;
+	}
+
+	public getPinData(): any {
```
carmenfan: I don't understand the purpose of this function.  (or the whole pinData variable). It doesn't sound like it needs to exist. a quick grep shows it's being used in risk comments, risk comments don't have pins so I don't know why it's passing through a position with the comment.

@sanmont3drepo  @Charence what's going on here?

m1chalkubiak:
@carmenfan I was not aware of changes regarding pins, I need to review them carefully and check how conflicts should be resolved. Probably, I did this merger too quickly, apologies for that.

frontend/modules/viewer/viewer.redux.ts
sanmont3drepo: Yes, it doesn't align with they idea behind how pins are working @m1chalkubiak . Thats why I got rid from the "updatepins" bit from sagas.

The general concept is this:
- in the state (redux) there is only the information of the issues and risks and positions (which will later become the pins.)
- in the selector the information from the issues and risks/ filtering of them and show/hide pins state is being consumed and returns which pins should be visible at that moment.
- this information is being used in the viewer react component to actually show and hide the pins, acting like any react component (representing visually the information from the state)
- the previous/current pins is part of the mechanism in the visual representation layer, it has nothing to do with state. Its actually how react components work (difference between the prev and current state and managing the dom and how that it would be represented) 

As a minor exercise to see clearly that curr and prev is from  the visual representation layer and has nothing to do with state, you could change it to instead of just checking the difference between the previous and current state and adding/removing the pins, you could just wipeout all the pins and add all the pins from the current  state.

frontend/services/viewer/viewerService/viewerController.tsx
sanmont3drepo: I agree, I saw that when i was fixing the pin bug and I had the same hunch.

frontend/services/viewer/viewerService/viewerController.tsx
sanmont3drepo: I think is being used when you edit/drop a pin. Im not sure is the best way to do it.

frontend/services/viewer/viewerService/unityController.ts
```diff
@@ -0,0 +1,511 @@
...
+
+import EventEmitter from 'eventemitter3';
+import { VIEWER_EVENTS } from '../../../constants/viewer';
+import { DialogActions } from '../../../modules/dialog';
+import { dispatch } from '../../../modules/store';
+import { PIN_COLORS } from '../../../styles';
+
+declare const UnityUtil;
+
+export class UnityController {
```
sanmont3drepo: I dont understand the reason for the existence of this file as well as the viewercontroller.

m1chalkubiak: The reason for creating unityController class was a need to separate the methods responsible only for controlling Unity, from the remaining actual logic part.

Of course, the old classes should be removed, I have it noted in the description https://github.com/3drepo/3drepo.io/issues/1916

Nevertheless, I am open to ideas about the names of these classes as I did not have good ideas for them.

sanmont3drepo: Yeah, well, if you see the methods in this file is almost a 1:1 relation to unity util,  it doesnt add much, it just feels like a proxy to unityutil.

carmenfan: Agree with @sanmont3drepo 

1. I'm not sure why ViewerService is renamed to ViewerController?
2. I think the inheritance doesn't make much sense, instead of calling super, you may as well just call UnityUtils directly. If the idea is that it allows for future implementations of viewers with different rendering engine then the inheritance seems to be done wrong. ViewerController should generic parent class that unityController extends. But I think this is overkill. Just merging the 2 is fine.

frontend/modules/viewer/viewer.sagas.ts
sanmont3drepo: this should be erased

frontend/routes/viewerCanvas/viewerCanvas.component.tsx
sanmont3drepo: Yeah, the "updatePins" props shouldnt exist.

frontend/routes/viewerCanvas/viewerCanvas.container.ts
```diff
@@ -37,6 +38,7 @@ const mapStateToProps = createStructuredSelector({
 });
 
 export const mapDispatchToProps = (dispatch) => bindActionCreators({
+	updatePins: ViewerActions.updatePins
```
sanmont3drepo: This shouldnt exist

m1chalkubiak:
@carmenfan @sanmont3drepo 
Summarizing what I have improved:
First of all, as You suggested I moved all methods to one common class. I removed old viewers, so there is no odd named viewer class anymore. I removed most of the method connected with pins, exactly as Santiago said, most of them, in fact, didn't have any sense/weren't used anywhere.

I found and fix a small bug regarding the showing of the loading info.
Unfortunately, I found a new bug, which appeared probably after the last merging. When trying to show map layer in GIS panel, we get Unity Error and application crashes.

So @sanmont3drepo, If You got time and can take a look at this, I would be more than happy.

carmenfan:
@sanmont3drepo  This is working, just need to conflicts resolved!

PR #1955, opened 2020-04-01
Issue 1873 - Measuring tool v2

### LINKED ISSUE #1873: Measuring tool v2
requires https://github.com/3drepo/3drepounity/issues/244

**UnityUtils**
The following methods needs to be exposed from Unity:
- [x] dropping a new type of pin
- [x] A callback that shows the measurement that has just been created
- [x] A callback to show if a measurement has been deleted
- [x] Turn on/off measuring mode
- [x] hide/show measuring toolbar
- [x] Setting measuring mode to 
      - point to point
      - Raycast
      - Min. distance
      - Surface on click
      - Polygonal surface
- [x] Toggle metres/mm display
- [x] Toggle snap to edges
- [x] Clear all measurements
- [x] Change the colour of a specific measurement
- [x] delete a specific measurement

**Frontend**
- [x] Remove measuring icon from toolbar
- [x] add measuring panel/button on the right hand side
- [x] If the panel is closed, it should disable the measuring tool, but measurements should stay on the screen
- [x] If there are no measurements it should display a message (see wireframe)
- [x] Section headers should only appear if there is at least 1 measurement under it.
- [x] Add measuring type icons (They should be toggle buttons - only one can be active at a time)
- [x] The pin is a bit of a special case, it's technically outside of the measuring tool feature so there isn't a "Drop pin mode" call in unity util. Refer to the way we drop issue/risk pins, it should be handled the same way with a different pin type.
- [x] Unity utils will have a callback to tell you when a new measurement is created, whenever this happens it should be added to the card
- [x] All meausrements should come with a colour, but the user should be able to change it with the colour picker (which will update unity with the new colour
- [x] Measurements are deleted via the X button or Delete all in the menu.
- [x] Selected total should display the sum of all the measurements selected
- [x] Newly added measurements should be checked by default and contributes to the selected total entry
- [x] Hamburger button menu:
     - [x] Edge Snapping - a toggle that calls a unity function
     - [x] Show X Y Z - a toggle that calls a unity function
     - [x] Units displayed in mm/m - it should say mm if it's displaying m and vice versa, also change the measurement units accordingly on the panel. If the units is ft, hide the option
     - [x] Reset colours should reset the colours do the default (should be a different default colour per type of measurement
     - [x] Delete all - A toggle that calls unity function, also remove all measurements from the card

**Backend**
- [x] End point to fetch the information from a mesh by ID

sanmont3drepo:
This fixes #1873


backend/models/helper/matrix.js
```diff
@@ -0,0 +1,110 @@
...
+		matrix[i] = new Array(size);
+		for (let j = 0 ; j < size; j++) {
+			matrix[i][j] = i !== j ? 0 : 1;
+		}
+	}
+
+	return matrix;
+};
+
+const scalarMultiply = (matrix, scalar) => {
```
carmenfan, line 50: @sanmont3drepo 
This doesn't look like it's used?

I don't think you'd ever multiply a matrix with a scalar like this.

If the purpose is to scale a vector it's only applied on numbers where i=j (apart from the last line).

backend/models/helper/model.js
```diff
@@ -1005,10 +1010,61 @@ function getMetadata(account, model, id) {
...
+}
+
+async function getMeshById(account, model, meshId) {
+	const historyRes =  (await History.findByObjectId(account, model, meshId, {current:1}));
+	if (!historyRes) {
+		throw responseCodes.RESOURCE_NOT_FOUND;
+	}
+
+	const revisionIds = historyRes.current;
+	const projection = {
```
carmenfan, line 1022: @sanmont3drepo  a mesh will never have a matrix in it?

backend/models/helper/model.js
```diff
@@ -1005,10 +1010,61 @@ function getMetadata(account, model, id) {
...
+		"faces": 1,
+		"_extRef":1
+	};
+
+	const mesh = await Scene.getObjectById(account, model, utils.stringToUUID(meshId), projection);
+	let parentMatrix = null;
+
+	if ((mesh.parents || []).length > 0) {
+		parentMatrix = await getParentMatrix(account, model, mesh.parents[0], revisionIds);
+		if (mesh.matrix) {
```
carmenfan, line 1035: don't think this is needed a mesh will never have matrix in it by specification.

backend/models/helper/model.js
```diff
@@ -1005,10 +1010,61 @@ function getMetadata(account, model, id) {
...
+	combinedStream.append(stringToStream(["{\"matrix\":", JSON.stringify(mesh.matrix), ",\"vertices\":["].join("")));
+	combinedStream.append(vertices.pipe(new BinToVector3dStringStream({isLittleEndian: true})));
+	combinedStream.append(stringToStream("],\"triangles\":["));
+	combinedStream.append(triangles.pipe(new BinToTriangleStringStream({isLittleEndian: true})));
+	combinedStream.append(stringToStream("]}"));
+	return 	combinedStream;
+}
+
+async function getParentMatrix(account, model, parent, revisionIds) {
+	const mesh = await Scene.getBySharedId(account, model, parent, revisionIds);
```
carmenfan, line 1055: I think all these changes here should be in scene. Mesh/transformations are nodes within Scene so it makes sense.
We really want to get rid of model/helper/model.

frontend/globals/unity-util.ts
```diff
@@ -673,6 +695,104 @@ export class UnityUtil {
...
+	 */
+	public static disableMeasureToolSnap() {
+		UnityUtil.toUnity('DisableMeasureToolSnap', undefined, undefined);
+	}
+
+	/**
+	 * Clear all measurements
+	 * @category Measuring tool
+	 */
+	public static clearMeasureToolMeasurements() {
```
carmenfan, line 752: Can we change this to clearAllMeasurements or something similar? Let's avoid similarly named functions with a difference of a plural. it gets hard to read 😆

backend/models/helper/matrix.js
sanmont3drepo: Yes, TBH I dont know why I added that.

sanmont3drepo:
- [ ] When the model is in feets, area measurement fails

sanmont3drepo:
- [ ] When going directly to measuring mode without having clicked first theviewer  when trying to measure a mesh it selects the mesh instead of measuring it

clickfirsterror

PR #1967, opened 2020-04-08
ISSUE #1964 - Fix chip input styling

### LINKED ISSUE #1964: Adjust the chip input lines to rest of the app
### Description
The lines on the chip input  are still not the right colour if `yarn build` is ran. It's the right colour if you run `yarn watch`. 

_Originally posted by @carmenfan in https://github.com/3drepo/3drepo.io/pull/1934#issuecomment-609879941_

m1chalkubiak:
This fixes #1964

#### Description
The lines on the chip input  are still not the right colour if `yarn build` is ran. It's the right colour if you run `yarn watch`.

### DIFF
--- a/frontend/routes/components/chipsInput/chipsInput.styles.ts
+++ b/frontend/routes/components/chipsInput/chipsInput.styles.ts
@@ -23,7 +23,7 @@ export const StyledChipInput = styled(ChipInput)`
 	&& {
 		margin-top: 10px;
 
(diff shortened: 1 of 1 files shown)


m1chalkubiak:
After checking on https://dev.3drepo.io/ I realized that none of the changed styles are working on this element. This was a very silly mistake from my side I forgot that the names of the classes are minified on production build so we cannot rely on they original values.

PR #1968, opened 2020-04-08
Issue 1917 - Store Issues screenshots as external refs

### LINKED ISSUE #1917: stop storing screenshots in base64 within a bson in the database
We're currently storing screenshots in issues/issue comments embedded into the bson object as base64 binary.

As bson has a size limit of 16MB, this will become problematic if users starts uploading HQ images or if the amount of comments within the issue gets very long.

We should store them externally via whatever file storage the server is configured to connect to (fs, s3, gridfs etc?)

sanmont3drepo:
This fixes #1917 

#### Description
Store screenshots externally via file storage the server is configured to connect to (fs, s3, gridfs etc?)

#### Test cases
- When creating an issue the viewpoints screenshots should be saved in a external ref
- When creating a risk the viewpoints screenshots should be saved in a external ref
- When creating a comment the viewpoint should work the same as when creating the issue/risk
- When showing the screenshot it should access the external ref in the new ones and the old ones should work as they do now
- When deleting a screenshot it should delete the new ones using the external ref and the old ones should work as they do now
- Check what happens on update
- Check that everything still works with BCF


backend/models/fileRef.js
```diff
@@ -142,6 +142,24 @@ FileRef.getOriginalFile = function(account, model, fileName) {
 	return fetchFileStream(account, model, collection, fileName, true);
 };
 
+FileRef.getFile = (account, model, collName, ref_id) => {
```
carmenfan, line 145: @sanmont3drepo  There's a bit of change since safetibase, so I think this branch needs a revisit upon merging latest of staging.

For e.g. fetchFile parameters changed on fetchFile, it takes account, model, colName

backend/models/fileRef.js
```diff
@@ -251,17 +269,22 @@ FileRef.storeMitigationsFile = async function(account, user, name, data) {
 	return await this.storeFile(account, collName, user, name, data, extraFields);
 };
 
-FileRef.storeFileAsResource = async function(account, model, user, name, data, extraFields = null) {
-	const collName = model + RESOURCES_FILE_REF_EXT;
+FileRef.storeFile = async function(account, model, collName, data, extraFields = null) {
```
carmenfan, line 272: these changes looks like it's reverted the refactoring effort done in safetibase? I don't know if it's because it's missing a merge with the latest code

backend/models/fileRef.js
carmenfan: I think it's a bit confusing to have a `getFile` and a `fetchFile`. If it's because this is an external facing function (to obscure the need to append `.ref`) then maybe we should call this `fetchFile` and rename the other one as `_fetchFile`

Issue #1966, opened 2020-04-07
4.5.1 bug list


carmenfan:
### Bugs
- [x] [Background colour for comments is inconsistent](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-610336958)
- [x] [Pins missing if you go into the model after looking at kanban](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-610874884)
- [x] [Lines should only appear below the name if it's in editing mode](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-610879496)
- [x] [The save/cancel button should appear as soon as it enters edit mode](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-610880013)
- [x] [CoordView is showing even though it's suppose to be disabled](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-610889735)
- [x] [Existing text not clearing in edit mode](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-610894832)
- [x] [Measure colour selector missing previously used options.](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-610907819)
- [x] [X Y Z should align](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-610949811)
- [x] [Polygon area tool not clearing](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-611009953)
- [x] [Measurements don't always match](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-611016762)
- [x] [We can no longer save views with the default name](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-611641597)
- [x] [Limit Measurement name to 15 characters per line](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-613301405)
- [x] [Rearrange selected total](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-613309131)
- [x] [Save the new name when you hit Enter](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-613320259)
- [x] [UI oddities when creating issues/risks](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-613325683)
- [x] [Issue Title and Description are not operating properly](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-613330066)
- [x] [Length Measurements are showing XZY values instead of XYZ](https://github.com/3drepo/3drepo.io/issues/1966#issuecomment-613334619)
[...]

Issue #1963, opened 2020-04-06
Safetibase travis tests


carmenfan:
Missing some travis testings for safetibase, and old tests that lives in model settings needs to be migrated to teamspace settings.

Removed From model settings test that needs putting back:
- [x] updating topic type with duplicated values

PR #1978, opened 2020-04-16
Issue 1971 - refactor topic types/risk categories into arrays

### LINKED ISSUE #1971: Store topic types and categories as array of strings
### Description
Store topic types and categories as array of strings instead of key/val pairs

**DB Migration:**
`mongo <host:port> -u <user> -p <password> --authenticationDatabase=admin scripts\db_migrations\v4.6\changeIssueTypesAndRiskCategoriesDatatype.js`

### Tasks
- [x] Backend udpate
- [x] Frontend update
- [x] Migration script
   - [x] Rename everything in teamspace settings
   - [x] Rename everything in comments

carmenfan:
This fixes #1971

**!!! This has #1966 merged in it needs to wait until #1966 is in staging. !!!**

Base branch is set to #1966 for an easier view for code review

#### Description
- DB script to convert risk categories/issue types [object] => [string]
- altered type checks to suit on the backend
- altered front end to understand the new type

#### Test cases
NOTE: subscriptionTest DB is migrated to this new format if you don't want to ruin your current db. (It has a very obvious password!)
- API should now take in [string] instead of [object] for issueTypes/risk cat
-  data should show correctly in risk/issues
-  drop down menu should show correctly in issues/risks details and new issues/risks
- search filters should function correctly
- kanban grouping should work

#### Notes:
- Auto test not updated as they only exist in #1963


backend/models/teamspaceSetting.js
```diff
@@ -129,37 +129,23 @@ class TeamspaceSettings {
...
-									value,
-									label: label.trim()
-								};
+					const arrayUpdated = [];
+					const foundKeys = {};
+					for (let i = 0; i < data[key].length; ++i) {
+						const entry = data[key][i];
+						if(utils.isString(entry) && entry !== "") {
+							const value = entry.trim();
+							if (utils.hasField(foundKeys, value.toLower())) {
```
Charence, line 138: I get the error: `TypeError: value.toLower is not a function`

backend/models/teamspaceSetting.js
```diff
@@ -129,37 +129,23 @@ class TeamspaceSettings {
...
+								throw responseCodes.DUPLICATED_ENTRIES;
 							}
+							foundKeys[value.toLower()] = 1;
```
Charence, line 141: I get the error: `TypeError: value.toLower is not a function`

frontend/constants/issues.ts
```diff
@@ -32,7 +32,7 @@ export const PRIORITIES = {
 // FIXME: first item should be taken in topicTypes should be taken
 // for default value
 const TOPIC_TYPES = {
-	FOR_INFORMATION: 'for_information'
+	FOR_INFORMATION: 'For Information'
```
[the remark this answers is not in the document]
carmenfan: It's been address months ago I think.

https://github.com/3drepo/3drepo.io/pull/1978/files#diff-c9def28852214f55a5794dbda6ae3e28R576

backend/models/teamspaceSetting.js
```diff
@@ -129,37 +129,23 @@ class TeamspaceSettings {
...
-									label: label.trim()
-								};
+					const arrayUpdated = [];
+					const foundKeys = {};
+					for (let i = 0; i < data[key].length; ++i) {
+						const entry = data[key][i];
+						if(utils.isString(entry) && entry !== "") {
+							const value = entry.trim();
+							if (utils.hasField(foundKeys, value.toLowerCase())) {
+								throw responseCodes.DUPLICATED_ENTRIES;
```
Charence, line 139: It seems like this error isn't thrown e.g. if topicTypes "Unknown" and "unknown" exist.

I also get this logged:

```
(node:6028) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'name' of undefined
    at Utils.mongoErrorToResCode (/Users/charencewong/Workspace/3drepo.io/backend/utils.js:146:10)
    at Object.responseCodes.respond (/Users/charencewong/Workspace/3drepo.io/backend/response_codes.js:397:19)
    at TeamspaceSettings.update.then.catch.err (/Users/charencewong/Workspace/3drepo.io/backend/routes/teamspace.js:523:18)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:6028) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
```

frontend/constants/issues.ts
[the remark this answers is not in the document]
carmenfan: I'm not sure why we need to have a topic type object at all, so i'm just removing it.

Setting issues as `For Information` is default where possible was a requirement when we did issue so we only take the first item as default value if we can't find it. So I think as it stands it's ok!

backend/models/teamspaceSetting.js
carmenfan: fixed

carmenfan:
@Charence All fixed now!

PR #1989, opened 2020-04-21
ISSUE #1952 - Handle both touch and mouse events in markup tool

### LINKED ISSUE #1952: Markup tool doesn't work at all with touch.
The toolbar buttons work fine with touch, but the canvas does not seem to listen to touch events

m1chalkubiak:
This fixes #1952

### DIFF
--- a/frontend/routes/components/screenshotDialog/components/drawingHandler/handleCalloutDrawing/handleCalloutDrawing.component.tsx
+++ b/frontend/routes/components/screenshotDialog/components/drawingHandler/handleCalloutDrawing/handleCalloutDrawing.component.tsx
@@ -101,11 +101,11 @@ export class HandleCalloutDrawing
 	}
 
 	public subscribeDotDrawingEvents = () => {
-		this.props.stage.on('mouseup', this.handleMouseUpDot);
+		this.props.stage.on('mouseup touchend', this.handleMouseUpDot);
 	}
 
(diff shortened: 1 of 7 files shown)


m1chalkubiak:
> @m1chalkubiak I haven't tested this yet but this looks it'll only listen to touch if it is deemed a touch device otherwise it'll listen to the mouse?
> 
> My laptop has a touch screen - does that mean it won't work if i use that?

You are right, in Your case, it will only handle touch event, so You won't be able to use mouse in the markup tool. Of course this is therefore not an acceptable solution. I will try to look for a more sophisticated solution.

m1chalkubiak:
Hi @carmenfan 
Yesterday after making a re-search on how both touch and mouse can be handled, I also checked what is suggested in the official documentation of the library that we are using for drawing in this feature, and they are suggesting using those events in pairs. I checked this and it seems to work ok. So I think that I had over-engineered this task from the very beginning. 

Of course, I was only able to simulate touch event using the browser devtools. Both of them are working, however, I don't have a physical device to check if everything will work fine if we start to use both mouse and touch at the same time.
Therefore, I would need to ask You to help with that. If You could check if everything is working fine on Your laptop, I would be grateful.

carmenfan:
> Therefore, I would need to ask You to help with that. If You could check if everything is working fine on Your laptop, I would be grateful.

@m1chalkubiak No problem, will do.

Thanks for looking into it!

PR #2040, opened 2020-05-18
ISSUE 1908 - BCF refactor

### LINKED ISSUE #1908: Revisit BCF import/export and refactor it to make it more robust
it's currently a mess that no one understands and breaks easily...

Charence:
This fixes #1908


Charence:
- [ ] fix issue number order

backend/models/bcf.js
```diff
@@ -408,816 +464,639 @@ function getBCFMarkup(issue, account, model, unit) {
...
 
-								let highlightedGroupObject;
+												for (let j = 0; j < vpComponents[i].Selection.length; j++) {
+													for (let k = 0; vpComponents[i].Selection[j].Component && k < vpComponents[i].Selection[j].Component.length; k++) {
+														let objectModel = model;
 
-								// TODO: refactor to reduce duplication?
-								if (vpComponents[i].Selection) {
+														if (settings.federate) {
+															objectModel = ifcToModelMap[vpComponents[i].Selection[j].Component[k]["@"].IfcGuid];
```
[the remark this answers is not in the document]
carmenfan: Also probbaly better to do 
```suggestion
const objectModel = settings.federate ? ifcToModelMap[vpComponents[i].Selection[j].Component[k]["@"].IfcGuid]: model;
```

so we can use const instead of let.

backend/models/bcf.js
```diff
@@ -408,816 +464,639 @@ function getBCFMarkup(issue, account, model, unit) {
...
 
-						if (_.get(vpXML, "VisualizationInfo.Components")) {
-							const groupDbCol = {
-								account: account,
-								model: model
-							};
+										for (let i = 0; i < vpComponents.length; i++) {
 
-							const vpComponents = _.get(vpXML, "VisualizationInfo.Components");
+											let highlightedGroupObject;
```
carmenfan, line 868: I think we should do const here and setup the basic group stuff instead of calling createGroupObject and doing all these if checks everytime. i.e.

```suggestion
const highlightedGroupObject = {issue,name, color: [255,0,0], objects: {}};
```

then within those loops, all you have to do is
````
group.objects[model] =  group.objects[model] || [];
group.objects[model].push(id)
````

File: backend/models/bcf.js [hunk 3]
```diff
@@ -408,816 +464,639 @@ function getBCFMarkup(issue, account, model, unit) {
...
+			group.color = color;
+		}
 
-					}).then(savedIssues => {
+		if (!group.objects) {
+			group.objects = {};
+		}
 
-						const notifications = [];
+		if (!group.objects[groupAccount]) {
```
carmenfan, line 573: i don't think there's any point of doing this, as account is static - we're always passing the same value in.

backend/models/bcf.js
```diff
@@ -408,816 +464,639 @@ function getBCFMarkup(issue, account, model, unit) {
...
-						const saveIssueProms = [];
-						const issuesToCreate = [];
+	return groupData;
+}
 
-						// sort issues by date and add number
-						issues = issues.sort((a, b) => {
-							return a.created > b.created;
-						});
+function createGroupObject(group, name, color, groupAccount, groupModel, ifc_guid) {
```
carmenfan, line 554: `ifc_guid` needs to be in camel case

backend/models/bcf.js
carmenfan: Or, since that highlghtedGroupObjects isn't exactly in group schema anyway (you still have to call createGroupData after....), just leave the fluff out till you're about to create the group, ie only store what you care about:
````
{
    "<model id>": [ ifc guids]
}
````

File: backend/models/bcf.js [hunk 5]
```diff
@@ -408,816 +464,639 @@ function getBCFMarkup(issue, account, model, unit) {
...
+													for (let k = 0; componentsToHide && k < componentsToHide.length; k++) {
+														let objectModel = model;
+
+														if (settings.federate) {
+															objectModel = ifcToModelMap[componentsToHide[k]["@"].IfcGuid];
+														}
+
+														// Exclude items selected
+														if (highlightedObjectsMap && -1 === highlightedObjectsMap.indexOf(componentsToHide[k]["@"].IfcGuid)) {
+															hiddenGroupObject = createGroupObject(
```
carmenfan, line 963: similar to the highlights

backend/models/bcf.js
```diff
@@ -408,816 +464,639 @@ function getBCFMarkup(issue, account, model, unit) {
...
+									}
+									issue.viewpoints.push(vp);
 
-										for (let k = 0; componentsToShow && k < componentsToShow.length; k++) {
-											let objectModel = model;
+								});
 
-											if (settings.federate) {
-												objectModel = ifcToModelMap[componentsToShow[k]["@"].IfcGuid];
+								return Promise.all(groupPromises).then(() => {
```
carmenfan, line 1026: await to reduce indentation

backend/models/bcf.js
```diff
@@ -408,816 +464,639 @@ function getBCFMarkup(issue, account, model, unit) {
...
+			issue,
+			viewpointsData: xml.Markup.Viewpoints
+		};
+	});
+}
 
-					if (!group.objects[groupAccount][groupModel]) {
-						group.objects[groupAccount][groupModel] = { ifc_guids: [] };
-					}
+bcf.readBCF = function(account, model, requester, ifcToModelMap, zipPath, settings) {
```
carmenfan, line 658: this doesn't need exporting... it's a private function. don't include it inside bcf (and any other private functions)

backend/models/issue.js
```diff
@@ -191,6 +188,99 @@ class Issue extends Ticket {
...
+		const settings = await ModelSetting.findById({account, model}, model);
+		const issues = await this.findByModelName(account, model, branch, revId, projection, ids, noClean, useIssueNumbers);
+
+		return BCF.getBCFZipReadStream(account, model, issues, settings.properties.unit);
+	}
+
+	async importBCF(requester, account, model, revId, zipPath) {
+		const settings = await ModelSetting.findById({account, model}, model);
+
+		return BCF.importBCF(requester, account, model, zipPath, settings).then((bcfIssues) => {
```
carmenfan, line 204: you've already declared this as async, may as well use await here.

backend/models/issue.js
```diff
@@ -191,6 +188,99 @@ class Issue extends Ticket {
...
+		let branch;
+
+		if (!revId) {
+			branch = "master";
+		}
+
+		const history = await History.getHistory({ account, model }, branch, revId, {_id: 1});
+
+		if (!history) {
+			return Promise.reject(responseCodes.MODEL_HISTORY_NOT_FOUND);
```
carmenfan, line 219: I would do this check prior BCF import, we don't really want to waste all this time to do the parsing then to realise it's all for nothing!

backend/models/issue.js
```diff
@@ -191,6 +188,99 @@ class Issue extends Ticket {
...
+		} else if (history) {
+			revId = history._id;
+		}
+
+		const existingIssues = await this.getList(account, model, branch, revId);
+		const existingIssueIds = existingIssues.map(x => x._id);
```
carmenfan, line 225: if the whole point of having this is to do indexOf @ L235, we should store this in a set or object (i.e dictionary).

since array isn't really the best for search.

i.e.
```suggestion
		const existingIssueMap = {};
		for(let i = 0; i <existingIssues.length; ++i) {
		     existingIssuesMap[existingIssues._id] = i;
		}
```

backend/models/bcf.js
```diff
@@ -408,816 +464,639 @@ function getBCFMarkup(issue, account, model, unit) {
 	});
 }
 
-bcf.getBCFZipReadStream = function(account, model, username, branch, revId, ids, useIssueNumbers = false) {
-
+bcf.getBCFZipReadStream = function(account, model, issues, unit) {
```
carmenfan, line 467: if there isn't much overlap between read/write, it might be good to just split them into 2 files.

backend/models/bcf.js
```diff
@@ -408,816 +464,639 @@ function getBCFMarkup(issue, account, model, unit) {
...
-	}).then(issues => {
+	issues.forEach(issue => {
 
-		const bcfPromises = [];
+		const issueAccount = (issue.origin_account) ? issue.origin_account : account;
+		const issueModel = (issue.origin_model) ? issue.origin_model : model;
 
-		issues.forEach(issue => {
+		bcfPromises.push(
+			// FIXME
```
carmenfan, line 482: what is this fixme for?

backend/models/bcf.js
```diff
@@ -408,816 +464,639 @@ function getBCFMarkup(issue, account, model, unit) {
...
-
-		settings = _settings;
-		return Issue.findByModelName(account, model, branch, revId, projection, ids, noClean, useIssueNumbers);
+	const bcfPromises = [];
 
-	}).then(issues => {
+	issues.forEach(issue => {
 
-		const bcfPromises = [];
+		const issueAccount = (issue.origin_account) ? issue.origin_account : account;
```
carmenfan, line 478:
```suggestion
		const issueAccount = issue.origin_account || account;
```

backend/models/bcf.js
```diff
@@ -408,816 +464,639 @@ function getBCFMarkup(issue, account, model, unit) {
...
+		const issueModel = (issue.origin_model) ? issue.origin_model : model;
```
carmenfan, line 479:
```suggestion
		const issueModel =issue.origin_model || model;
```

backend/models/bcf.js
```diff
@@ -80,15 +97,28 @@ function parseXmlString(xmlString, options) {
...
+
+	if (_.map(list).indexOf(dataSanitised) === -1) {
+		return data;
+	}
+
+	return dataSanitised;
+}
+
+function getIssueBCF(issue, account, model, unit) {
 	// FIXME
```
carmenfan, line 122: if we don't know what these are for maybe we should just remove them lol

backend/models/bcf.js
Charence: I can't remember, but this function seems much clearer when compared to the import, so I think I'll remove the FIXME.

backend/models/bcf.js
Charence: I think BCF can be kept as one file?

backend/models/bcf.js
```diff
@@ -408,816 +464,639 @@ function getBCFMarkup(issue, account, model, unit) {
...
+							// revId?
+							// console.log(issue);
+							// console.log(viewpointsData);
+							return parseViewpoints(utils.uuidToString(issue._id), files[guid], viewpointsData).then(viewpoints => {
+								// console.log(viewpoints);
+
+								const vpGuids = Object.keys(viewpoints);
+								const groupPromises = [];
+
+								vpGuids.forEach(vpGuid => {
```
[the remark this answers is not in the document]
Charence: vpGuids is used again further down

backend/models/bcf.js
```diff
@@ -408,816 +464,639 @@ function getBCFMarkup(issue, account, model, unit) {
...
+			zipfile.on("end", () => {
+
+				// console.log("end 1");
+				const endPromise = Promise.all(promises).then(() => {
+					const parsePromises = [];
+					// console.log(files);
+					// TODO: parse all XML first
+					Object.keys(files).forEach(guid => {
+						// console.log(guid);
+						const parsePromise = parseMarkupBuffer(files[guid][`${guid}/markup.bcf`]).then((markupData) => {
```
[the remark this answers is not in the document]
carmenfan: Not addressed, but a TODO for future

PR #2046, opened 2020-05-21
ISSUE #2027 update maintenance script to look through all collections ending in .ref

### LINKED ISSUE #2027: repairFileShare.py needs to catch all .ref tables
### Description

More .ref tables to be added in the future, so the script should detect all of them.

### Goals
- [x] Loop all .ref collections to establish all files correctly

### Description
Missing some ref tables can make this script remove valid data

carmenfan:
This fixes #2027

#### Description
instead of going through settings to find all the model ID, then look through specific extensions, it now takes in the list of collection names, find all collection names ending in .ref.

#### Test cases
- should now find all references and work accordingly.

### DIFF
--- a/scripts/dbMaintenance/repairFileShare.py
+++ b/scripts/dbMaintenance/repairFileShare.py
@@ -95,7 +95,6 @@ for database in db.database_names():
                                     missing.append(refInfo + ": No backup found.");
                         else:
                             fileList[filePath] = True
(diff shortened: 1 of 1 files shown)


carmenfan:
@keifgwinn I've tested this locally and it seems to work, can you play with it and verify that it is doing what you need?

PR #2048, opened 2020-05-27
ISSUE #2025 - Disable clipping tool if measurement is activated

### LINKED ISSUE #2025: Clip plane should be toggled off when measurement tool is toggled on
**EDIT by Carmen:**
What should happen is, if a measurement type is activated, it should automatically disable clipping tool and vice versa.

> 
> 
> ### Description
> User can apply a clip plane to their model, but if they do not deactivate it in the toolbar they will not be able to measure length/area or points. Can we make clip plane automatically inactive (not orange in the toolbar) as soon as the user opens measurements Card or activates measurement feature?
> 
> ### Steps to replicate
> - Open any 3D Model in 3drepo.io
> - Activate single clip mode
> - Move the clip plane through the model
> - click anywhere in space (not model mesh)
> - Open measurements Card
> - Activate point to point measurement feature
> - Notice how it is not working until you deactivate the clip plane
> 
> ### Pictures/ Gifs of the error
> **Measuring with Clip Plane active**
> Measuring with Section Active
> 
> **Measuring with Clip Plane inactive**
> Measuring with Section Inactive

m1chalkubiak:
This fixes #2025

### DIFF
--- a/frontend/routes/viewerGui/components/measurements/components/measuringType/measuringType.component.tsx
+++ b/frontend/routes/viewerGui/components/measurements/components/measuringType/measuringType.component.tsx
@@ -69,11 +69,11 @@ export const MeasuringType = ({
 	};
 
 	React.useEffect(() => {
-		if (isMetadataActive) {
+		if (isMetadataActive || isClipEdit) {
 			setMeasureMode('');
 			deactivateMeasure();
 		}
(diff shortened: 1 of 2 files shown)


m1chalkubiak:
@carmenfan fixed.

PR #2051, opened 2020-05-27
ISSUE #2011 - Improving styling for displaying model names in invitation user modal

### LINKED ISSUE #2011: Invite user option does not display full model names
### Description
When a user is inviting another person to their project they have an option to assign relevant model permissions before sending the invite. Unfortunately, in the list of models available for my project, I can not see the full name of my models. If you follow the ISO standard for BIM file naming convention, the important last bit of the model name is missing. Adding a tooltip to the model name on hover may be an option, however, some users may find it irritating if they have a long list of models and need to hover on every single model to read their full name. Perhaps we can start a new line for those name parts that are not visible or make it wider/scrollable across?

### Steps to replicate
- Log in to 3D Repo
- Go to User Management tab
- Select Invite User option
- Fill out their email address and select a project you want to add them to
- Scroll through the list of models with names that contain more than 22 characters

Can't see the model Name when inviting a user

m1chalkubiak:
This fixes #2011

### DIFF
--- a/frontend/routes/components/customTable/customTable.styles.ts
+++ b/frontend/routes/components/customTable/customTable.styles.ts
@@ -71,7 +71,7 @@ export const Container = styled.div`
 `;
 
 export const Row = styled.div`
-	height: 62px;
+	min-height: 62px;
 	display: flex;
 	flex-direction: row;
 	border-bottom: 1px solid ${COLOR.BLACK_6};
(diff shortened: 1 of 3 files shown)


m1chalkubiak:
@BDavydov @carmenfan 
My proposition to solve this issue is extending model name column from 180px to 220px and add tooltip to it in case of longer names, but please verify it and let me know does this solution suits you?

PR #2059, opened 2020-05-29
Issue #2057 - Add query parameter for filtering lists

### LINKED ISSUE #2057: Add query parameter for filtering lists in the api by updatedSince timestamp
### Description
Add an  updatedSince  query parameter in the API for listing groups , issues and risks.  If present the issues/risks/groups must has  been modified or created after the timestamp value of updatedSince.

### Tasks
- [x] Add updatedSince to group list
- [x] Add updatedSince to issues and risks list

sanmont3drepo:
This fixes #2057

### DIFF
--- a/backend/models/ticket.js
+++ b/backend/models/ticket.js
@@ -479,16 +479,6 @@ class Ticket {
 		});
 	}
 
-	async findByModelNameUpdateTimestamp(account, model, branch, revId, ids) {
-		const filter = await this.getIdsFilter(account, model, branch, revId, ids);
-		const coll = await this.getTicketsCollection(account, model);
-		return await coll.aggregate([
-			{ $match: filter},
-			{ $unwind: "$comments" },
-			{ $group: { _id: "$_id", lastUpdated: { $max: "$comments.created" } } }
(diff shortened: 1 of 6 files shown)


backend/models/ticket.js
```diff
@@ -589,16 +589,23 @@ class Ticket {
...
+
 		tickets.forEach(ticket => {
 			if (convertCoords) {
 				this.toDirectXCoords(ticket);
 			}
 
 			const timestamp =  lastUpdated.find(updatedTimestamp =>  utils.uuidToString(updatedTimestamp._id) === ticket._id);
 			ticket.lastUpdated =  (timestamp || {}).lastUpdated || ticket.created;
+
+			if(!updatedSince || updatedSince && ticket.lastUpdated >= updatedSince) {
```
carmenfan, line 602: can we not add this onto the query instead of filtering it in the js?

carmenfan: @sanmont3drepo should just be a query like `$or: [ created: {$gt: x}, comments.created: $gt:x]`?

Issue #2070, opened 2020-06-04
Add new Notifications for references within comments


carmenfan:
### Description
This is an extension to #2026

- When a user is tagged in a comment, it should create a notification for the user to say `xxx has tagged you in a comment`
- When a user tagged another issue within the same model, it should create a system message on the tagged issue saying `xxx has referenced this issue in #<issue number>` And the issue number should be clickable to go to that issue

### Goals
- [x] For each user reference, we should create a notification for users who are being referred.
- [x] For each issue being referenced, we should create a system comment which states `xyz referenced this issue in #yy`

### Tasks
- [x] Parse comments and determine if we should create a notification for a specific user
- [x] Parse comments and determine if we should be creating a system comment on another issue
- [x] Ensure UI understands and dispaly the user referenced notification properly
- [x] Ensure UI understands and display the issue reference properly
- [x] add test for parse comment and add notification for user tagging
- [x] add test for checking that quotes are not generating notifications when user tagging
- [x] add test for parse comment and add new system message for issue referencing
- [x] add test for checking that quotes are not generating system messages when referencing an issue

PR #2092, opened 2020-06-29
ISSUE 2066 - Orthographic projection mode

### LINKED ISSUE #2066: Orthographic viewpoint support
### Description
We want to add support for orthographic camera.

Note that similar work is happening in the issues plugin. To make that work, we have added a field `orthographicSize` to the viewpoint schema. This should match up nicely with both [Unity camera](https://docs.unity3d.com/ScriptReference/Camera-orthographicSize.html) and [BCF schema](https://github.com/BuildingSMART/BCF-XML/tree/master/Documentation#orthogonalcamera-optional)

Currently we have `type` parameter within the viewpoint schema that takes in `orthographic` or `perspective`. We can assume the lack of type to be perspective for backwards compatibility sake. 

### Goals
The user should be able to :
- Toggle between orthographic/perspective camera view while they are navigating
- If the viewpoint the user clicked to change to is of a different camera type as what they are currently using, the viewer should automatically change to the same type as the view to ensure the user sees exactly what was intended

### Tasks
In Unity:
- [x] Implement functions to toggle camera between perspective and orthographic
- [x] Check the current type and output fields accordingly when user request current viewpoint data
- [x] When user sets current viewpoint, check camera type and adjust accordingly

In 3drepo.io:
- [x] Add functionality to toggle between orthographic/perspective 
- [x] Ensure type parameter and zoom factor is saved into the database if it exists (I think this works already)
- [x] Ensure these values are being provided if available on a GET request   (I think this works already)
- [x] Ensure we are saving orthographic viewpoints properly on BCF exports
- [x] Ensure we're preserving `ViewToWorldScale` in BCF imports
- [x] Add tests

Charence:
This fixes #2066 

#### Description
- Support Orthographic projection mode
- BCF import/export support for orthographic view
- Store projection mode in DB (null = default perspective view)
- Update Unity to support setting projection mode and returning view status

#### Tasks
- [x] Re-build Unity viewer https://github.com/3drepo/3drepounity/issues/278

#### Test cases
- BCF import/export with orthographic/perspective project should still work
- User should be able to toggle projection mode from UI
- Loading issue/risk should switch to stored projection mode and correct mode should be reflected in UI
- Extent button should reset a consistent view of the whole model (resets orthographic size if changed)


backend/models/bcf.js
```diff
@@ -397,9 +397,24 @@ async function getIssueBCF(issue, account, model, unit) {
...
+						}
+
+						if (("perspective" === vp.type || !_.get(vp, "extras._noPerspective")) && vp.fov) {
+							viewpointXmlObj.VisualizationInfo.PerspectiveCamera = {
+								...camera,
+								FieldOfView: vp.fov * 180 / Math.PI
+							};
+						}
+					} else if (_.get(vp, "extras.OrthogonalCamera")) {
+						viewpointXmlObj.VisualizationInfo.OrthogonalCamera = _.get(vp, "extras.OrthogonalCamera");
```
carmenfan, line 417: what kind of things would be stored here?

frontend/globals/unity-util.ts
```diff
@@ -1415,6 +1415,28 @@ export class UnityUtil {
 		UnityUtil.toUnity('SetUnits', UnityUtil.LoadingState.MODEL_LOADING, units);
 	}
 
+	/**
+	 * Enable orthographic view
+	 */
+	public static enableOrthographic() {
```
carmenfan, line 1421: we should rename this (as suggested in unity PR)

frontend/globals/unity-util.ts
```diff
@@ -1415,6 +1415,28 @@ export class UnityUtil {
...
+	 */
+	public static disableOrthographic() {
+		UnityUtil.toUnity('DisableOrthographic', UnityUtil.LoadingState.MODEL_LOADING, undefined);
+	}
+
+	/**
+	 * Set orthographic size for orthographic view
+	 * @param {number} orthographicSize - Camera's half-size when in orthographic mode
+	 */
+	public static setOrthographicSize(orthographicSize: number) {
```
carmenfan, line 1436: I don't think this should be exposed, unless we have a UI for the user to tweak this

frontend/globals/unity-util.ts
```diff
@@ -1431,6 +1453,8 @@ export class UnityUtil {
 		up: [number],
 		forward: [number],
 		lookAt: [number],
+		isOrthographic?: boolean,
```
carmenfan, line 1456: viewpoint should send in `type`? unity should be outputing the exact schema with `type` so it should take in `type` as well.

frontend/services/viewer/viewer.tsx
```diff
@@ -908,9 +909,29 @@ export class ViewerService {
...
+	}
+
+	public async setCamera({ position, up, view_dir, look_at, type, orthographicSize, account, model }) {
+		await this.isModelReady();
+		UnityUtil.setViewpoint(
+			position,
+			up,
+			view_dir,
+			look_at,
+			type === VIEWER_PROJECTION_MODES.ORTHOGONAL,
```
carmenfan, line 930: just pass type down

carmenfan:
@Charence  these's also conflict in the backend tests, probably clash with #2070

backend/models/bcf.js
Charence: @carmenfan This is for legacy support in case anybody imported a BCF with an orthogonal camera in the past

frontend/services/viewer/viewer.tsx
Charence: @carmenfan A lot of older viewpoints don't seem to have `type` defined, so if `type` is passed down here, a default projection type would have to be determined either in Unity Utils or Unity.

frontend/globals/unity-util.ts
Charence: @carmenfan At the moment, Message Parser in Unity doesn't know about the strings `orthogonal` or `perspective`. The property `orthographic` for the camera is also a `boolean`.

frontend/services/viewer/viewer.tsx
carmenfan: if type is empty assume it's perspective. We can either code that logic inside unity.

Either way it needs to be in unity because it needs to work for embedded users without extensively playing with the schema

carmenfan:
closing this as it's awaiting for more unity changes

Charence:
Reopening PR

Charence:
@carmenfan Could you please rebuild the unity viewer

carmenfan (changes requested):
As discussed, we need to change all references of orthogonal to orthographic

carmenfan:
@3drepo/product-team orthographic view deployment : https://issue-2066.dev.3drepo.io/

carmenfan (approved):

mosment3DRepo:
### Bug

Switching back to  perspective from orthographic seems to lose the model until you zoom or move the camera, at which point the model pops in much closer to the camera

### Steps
1. Set model view to orthographic
2. Set model back to perspective
3. Move camera viewpoint

### GIF
Ortho1

carmenfan:
> ### Bug
> Switching back to perspective from orthographic seems to lose the model until you zoom or move the camera, at which point the model pops in much closer to the camera
> 
> ### Steps
> 1. Set model view to orthographic
> 2. Set model back to perspective
> 3. Move camera viewpoint
> 
> ### GIF
> Ortho1

@sebjf I guess this is due to the whole zoom in perspective vs viewpoint size in orthographic, where it makes the camera direction/positionings are very different when you switch type? is there something clever we can do to manipulate the viewpoint to be a bit closer visually to what it was looking at?

sebjf:
> @sebjf I guess this is due to the whole zoom in perspective vs viewpoint size in orthographic, where it makes the camera direction/positionings are very different when you switch type? is there something clever we can do to manipulate the viewpoint to be a bit closer visually to what it was looking at?

Possibly. The orientation of the camera should remain unchanged, but it is likely positioned, if anything, too close to the model (the ortho camera is constrained to a sphere around the model). Also the clip planes could be incorrectly set (that is probably what is causing the popping).
The challenge is its hard to say what the user wants to look at, since in ortho mode you can have elements of the model very close and very far look similar in size.
We couldn't just take the depth in the centre of the screen, because that might be a transparent part of the model.

How close would we be aiming for? It would be easy for example to make it fit the bounding box. It would change size but viewing direction wise would look good.

We may also be able to use the zoom parameter of the ortho camera to compute a sensible perspective depth. Either through finding an expression of the geometric relationship between ortho window size and perspective size at the near plane, or just trusting there is one and using a magic number.

I can investigate if you like.

carmenfan:
> > @sebjf I guess this is due to the whole zoom in perspective vs viewpoint size in orthographic, where it makes the camera direction/positionings are very different when you switch type? is there something clever we can do to manipulate the viewpoint to be a bit closer visually to what it was looking at?
> 
> Possibly. The orientation of the camera should remain unchanged, but it is likely positioned, if anything, too close to the model (the ortho camera is constrained to a sphere around the model). Also the clip planes could be incorrectly set (that is probably what is causing the popping).
> The challenge is its hard to say what the user wants to look at, since in ortho mode you can have elements of the model very close and very far look similar in size.
> We couldn't just take the depth in the centre of the screen, because that might be a transparent part of the model.
> 
> How close would we be aiming for? It would be easy for example to make it fit the bounding box. It would change size but viewing direction wise would look good.
> 
> We may also be able to use the zoom parameter of the ortho camera to compute a sensible perspective depth. Either through finding an expression of the geometric relationship between ortho window size and perspective size at the near plane, or just trusting there is one and using a magic number.
> 
> I can investigate if you like.

@mosment3DRepo agreed that we can leave it for now, as similar odd behaviours happen in Revit.

mosment3DRepo (approved):

PR #2099, opened 2020-07-02
Issue #2043 -  Send the embeded group for creation of issues and risks

### LINKED ISSUE #2043: User shouldn't need to call create group API before calling create issue API
At the moment, if the user wants to create an issue with highlighted objects/hidden objects, they have to first call groups API and get an ID from it.

Instead of this, the issues API should take in the highlighted objects/hidden object list, and create the group on the user's behalf. As the user shouldn't need to know that issues uses groups under the bonnet, and also this will allow us to use transactions when we upgrade mongo so we don't create orphaned groups should there was an error in creating the issues.

sanmont3drepo:
This fixes #2043


backend/models/ticket.js
```diff
@@ -428,6 +428,41 @@ class Ticket {
 		systemComments.push(comment);
 	}
 
+	async createGroupsIfNecessary(account, model, newTicket, branch, rid) {
+		// TODO: take in consideration the revision
+		const groupField = this.groupField;
+
+		const getGroupData = ({viewpoint, name, _id}, objects) => {
+			return {
+				color: viewpoint.color,
```
carmenfan, line 437: move `color` inside of highlighted_objects

backend/models/ticket.js
```diff
@@ -428,6 +428,41 @@ class Ticket {
 		systemComments.push(comment);
 	}
 
+	async createGroupsIfNecessary(account, model, newTicket, branch, rid) {
+		// TODO: take in consideration the revision
+		const groupField = this.groupField;
+
+		const getGroupData = ({viewpoint, name, _id}, objects) => {
```
carmenfan, line 435: move function declaration out of function scope

backend/models/ticket.js
```diff
@@ -428,6 +428,41 @@ class Ticket {
...
+			return {
+				color: viewpoint.color,
+				name,
+				objects,
+				[groupField]: utils.stringToUUID(_id)
+			};
+		};
+
+		const createGroup = async (ticket, objectsField, idField) => {
```
carmenfan, line 444: move function declaration out of function scope

PR #2111, opened 2020-07-17
ISSUE 1953 - Add pending status for save buttons

### LINKED ISSUE #1953: Action buttons need loading status when network speeds are low
### Description
When you create an issue and hit the save button (Green button) it does not change its current state. In case of low internet speed, it is hard to tell whether the issue was submitted. Users may click on the save button again several times, subsequently create multiple versions of the same issue. 
Issues

### Goals
- [ ] Make the Save button change its status (Similar to the login button)
Loading Button

Complete it for the following actions:
- [ ] Save an Issue
- [ ] Save a SafetiBase Issue
- [ ] Save a Group

m1chalkubiak:
This fixes #1953

- [x] Make the Save button change its status (Similar to the login button)
Loading Button

Complete it for the following actions:
- [x] Save an Issue
- [x] Save a SafetiBase Issue
- [x] Save a Group

### DIFF
--- a/frontend/modules/groups/groups.redux.ts
+++ b/frontend/modules/groups/groups.redux.ts
@@ -1,5 +1,5 @@
 /**
- *  Copyright (C) 2019 3D Repo Ltd
+ *  Copyright (C) 2020 3D Repo Ltd
  *
  *  This program is free software: you can redistribute it and/or modify
  *  it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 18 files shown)


m1chalkubiak:
Hi @BDavydov @carmenfan 
I added a pending state not only when creating a new issue/risk, but because the same button is used (for already existing issues) to adding comments I also added it while waiting for this action. For groups, I also added the pending state at the time of saving and updating.
Please, let me know if it works fine for You.

PR #2114, opened 2020-07-21
Issue 2101 - Load/offload models 

### LINKED ISSUE #2101: Model Onload/Offload
### Description
Part of the model visibility feature. The user should be able to offload a model from the viewer. This differs from hiding a model as it should completely remove the geometry from memory.

This issue requires https://github.com/3drepo/3drepounity/issues/283

### Goals
- [x] The user should be able to turn off a model when viewing a federation
- [x] The user should be able to turn on a model when viewing a federation
- [x] For any objects with colour override via groups, if the model is turned off when triggered, it should automatically apply the colour override if the model is turned on again.

sanmont3drepo:
This fixes #2101


carmenfan:
parked for now

PR #2119, opened 2020-07-27
ISSUE #2098 - Add uploading model status in upload model dialog

### LINKED ISSUE #2098: Only allow the user to leave the model upload card once the network request has been sent
At the moment the model upload pop up disappears as soon as you hit ok, then the model status becomes `uploading...` and turn into `queued` once the network request is sent

There's a danger of the user leaving the page while the model is still in the uploading status as it would look like to the user that he  has triggered a model import.

Instead of doing this, we should get rid of the `uploading` status, instead show the uploading status on the pop up card with a loading status so the user is aware that his request has not been made yet.

m1chalkubiak:
This fixes #2098

### DIFF
--- a/frontend/modules/model/model.redux.ts
+++ b/frontend/modules/model/model.redux.ts
@@ -30,7 +30,7 @@ export const { Types: ModelTypes, Creators: ModelActions } = createActions({
 	setPendingRevision: ['revision'],
 	resetRevisions: [],
 	downloadModel: ['teamspace', 'modelId'],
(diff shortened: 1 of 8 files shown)


m1chalkubiak:
> * Can we add a message saying something like `Please do not leave this page before upload has finished`?
> * Can we also disable the cross and stops the user from leaving this pop up until it has finished?

Apologie @carmenfan I've missed those comments before. Now it should work as described.

PR #2123, opened 2020-07-30
ISSUE 2093 - Standardise viewpoints in views/issues/risks

### LINKED ISSUE #2093: Standardise viewpoints in views/issues/risks
### Description

Standardise viewpoints in views/issues/risks

### Goals

- Viewpoints in views/issues/risks should have the same properties
- Backwards compatibility should be retained for now (deprecated parts should be marked)

### Tasks

- [ ] Update viewpoint to match tickets
- [ ] Update/add tests (mark tests for deprecated formats)
- [ ] DB migration script

Charence:
This fixes #2093

#### Description
- See issue

#### DB Migration
1. migrateViews
2. cleanViews*
3. cleanIssues
4. cleanRisks
5. addRiskNumbers

*Run migrateViews before cleanViews

#### Test cases
- Views, Issues, Risks should work as before
- Views, Issues, Risks should also work with new modified data structure


frontend/services/viewer/viewer.tsx
```diff
@@ -895,8 +895,7 @@ export class ViewerService {
 	 */
 
 	public async getObjectsStatus({ teamspace, model } = { teamspace: '', model: '' }) {
-		await this.isViewerReady();
-		return this.getUnityObjectsStatus(teamspace, model);
+		return this.isInitialised ? await asyncTimeout(1000, this.getUnityObjectsStatus, teamspace, model) : null;
```
carmenfan, line 898: why do we want to wait then return? what if it's still not initialised?

scripts/db_migrations/v4.10/addRiskNumbers.py
```diff
@@ -0,0 +1,37 @@
...
+    if database != "admin" and database != "local":
+        db = MongoClient(connString)[database]
+        print("--database:" + database)
+
+##### Get a model ID and find entries #####
+        for setting in db.settings.find(no_cursor_timeout=True):
+            modelId = setting.get("_id")
+            print("\t--model: " +  modelId)
+            riskNumber = 1
+            for risk in db[modelId + ".risks"].find():
```
carmenfan, line 31: I think we want to sort this, whilst it's mostly true that they should be ordered by time naturally I don't think we can guarantee this.

scripts/db_migrations/v4.10/cleanIssues.py
```diff
@@ -0,0 +1,48 @@
...
+        for setting in db.settings.find(no_cursor_timeout=True):
+            modelId = setting.get("_id")
+            print("\t--model: " +  modelId)
+            for issue in db[modelId + ".issues"].find({"$or":[{"comments._id":{"$exists":True}},{"viewpoints._id":{"$exists":True}},{"viewpoint":{"$exists":True}}]}):
+                issueId = issue.get("_id")
+                unsetFields = { "$unset": {} }
+                print("\t\t--issue: " +  str(issueId))
+##### Remove legacy viewpoint #####
+                if "viewpoint" in issue:
+                    unsetFields["$unset"]["viewpoint"] = ""
```
carmenfan, line 36: We're removing potentially valid viewpoints here. 2 options:
1. check the contents of the viewpoint, we only want to remove the viewpoint if it's empty (the toy project situation)
2. if there's a valid viewpoint, move it to the first entry of viewpoints.

backend/models/viewpoint.js
```diff
@@ -16,167 +16,333 @@
...
+	async deleteViewpoint(account, model, uid, sessionId) {
+		if (utils.isString(uid)) {
+			uid = utils.stringToUUID(uid);
+		}
+
+		const coll = await this.getCollection(account, model);
+		return coll.findOneAndDelete({ _id: uid }).then((deleteResponse) => {
 			if (!deleteResponse.value) {
-				return Promise.reject(responseCodes.VIEW_NOT_FOUND);
+				return Promise.reject(this.response("NOT_FOUND"));
```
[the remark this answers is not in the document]
Charence: It can also return ISSUE_NOT_FOUND/RISK_NOT_FOUND - except you can't delete those things...

backend/models/comment.js
```diff
@@ -162,11 +162,15 @@ const addComment = async function(account, model, colName, id, user, data) {
...
 	if (data.viewpoint) {
-		viewpoint = View.clean(data.viewpoint, fieldTypes.viewpoint);
+		data = View.clean(account, model, data, fieldTypes.viewpoint);
+		viewpoint = data.viewpoint;
 		viewpoint.guid = utils.generateUUID();
 
 		if (viewpoint.screenshot) {
-			viewpoint.screenshot = new Buffer.from(viewpoint.screenshot, "base64");
+			viewpoint.screenshot = new Buffer.from(
+				viewpoint.screenshot.substring(viewpoint.screenshot.indexOf(",") + 1),
```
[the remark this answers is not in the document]
Charence: https://github.com/3drepo/3drepo.io/pull/2123/files#diff-c9def28852214f55a5794dbda6ae3e28L145 and https://github.com/3drepo/3drepo.io/pull/2123/files#diff-c9def28852214f55a5794dbda6ae3e28L145 got removed from the frontend code, so it's now being done on backend. It removes the png header of the base64 string.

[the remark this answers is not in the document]
carmenfan: that's fine, we just need a comment here i think

backend/models/viewpoint.js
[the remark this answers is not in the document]
carmenfan: Instead of relying on there's a response code called X_NOT_FOUND. maybe we should utilise inheritance and define a variable for these common error codes which gets overridden on the subclass?

frontend/services/viewer/viewer.tsx
sanmont3drepo: If I remember correctly the  problem is that when you are in the kanban and you havent entered a model isViewerReady never finishes so this function gets unresponsive and you cant create issues without entering first a model

carmenfan:
@sanmont3drepo  front end bugs:

- [ ] sometimes highlight/override group info is lost:
    - Enter an issue with colour overrides:
image
    - Without changing the viewer state add a comment:

   - If you click on the comment, the color override disappears. Looking at the post comment API override group is missing:
image

backend/models/risk.js
```diff
@@ -179,7 +179,7 @@ class Risk extends Ticket {
 					oldRisk._id,
 					oldRisk.comments,
 					data,
-					createViewPoint(residualData.viewpoint)
+					createViewPoint(oldRisk._id, residualData.viewpoint)
```
carmenfan, line 182: @Charence  can you investigate what this is for?

This is the only place that is calling L145, I'm not sure why a mitigation comment is saving a viewpoint (surely that's a property change kind of update?)

Above all the function parameter doesn't even match, so I highly doubt this is used anywhere

Charence: @carmenfan Risks also has `addMitigationComment`, which I think should all live within `Comment`

carmenfan: @Charence 
I'm less worried about where the function is but whether does addMitigationComment ever comes in with a viewpoint.

because as it stands it never triggers as it's passing the wrong function parameters.

If it never has a viewpoint then we should just remove the createViewpoint function

Charence: @carmenfan Actually, there are a few places that make reference to a `residual` field, but none currently exists so this bit of code doesn't do anything. I'm not sure why we wanted to treat `residual` differently in the first place.

carmenfan:
@sanmont3drepo 

**Hidden objects seems to not operate when there's a change in colour override**
In this example I have 3 issues:
Issue 4 & 6 - some issue without colour override, just highlights/hidden objects
Issue 5 - issue with both hidden objects and colour override

GIF here: https://www.dropbox.com/s/e2u0jknvj3he9vy/1.gif

Observation:
-  5 is not dealing with hidden objects, the screenshot shows the state at which I saved the issue, but it's not how it looks when I open it
- if I'm on 5 then click on 6 or 4, the hidden objects aren't being processed. 
- if I click between 4 6 without going to 5, it is fine.

Seems like the colour override data is stopping me from hiding my objects somehow.

PR #2138, opened 2020-08-17
ISSUE 2028 - Custom error message for model code field

### LINKED ISSUE #2028: Model Code restriction message isn't clear
### Description
User starts typing model code and chooses to use a special character like - or /. Error message comes up saying code must match the following: "/^[A-Za-z0-9]+$/"
So, if it must match then it is allowed to contain these special characters? Or is it the case of Must NOT contains the following: "/^[A-Za-z0-9]+$/" ?

### Steps to replicate
- Login to 3D Repo
- Start creating a new model in any project
- Start typing model Code and add any special character
- Error message comes up and Save button becomes grey

image

m1chalkubiak:
This fixes #2028

### DIFF
--- a/frontend/routes/teamspaces/components/modelDialog/modelDialog.component.tsx
+++ b/frontend/routes/teamspaces/components/modelDialog/modelDialog.component.tsx
@@ -29,8 +29,8 @@ import { clientConfigService } from '../../../../services/clientConfig';
 import { schema } from '../../../../services/validation';
 import { CellSelect } from '../../../components/customTable/components/cellSelect/cellSelect.component';
 
-import { getTeamspacesList, getTeamspaceProjects } from '../../../../helpers/model';
-import { MODEL_SUBTYPES } from './../../teamspaces.contants';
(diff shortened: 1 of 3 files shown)


frontend/routes/teamspaces/components/modelDialog/modelDialog.component.tsx
```diff
@@ -52,6 +52,8 @@ interface IProps {
 	createModel: (teamspace, data) => void;
 }
 
+const getModelCodeFieldErrorMsg = (error) => error ? 'Model code can only contain alphanumeric' : '';
```
carmenfan, line 55: Can we rename this to `Can only contain alphanumeric characters`?

m1chalkubiak: > Can we rename this to `Can only contain alphanumeric characters`?

@carmenfan Done. And as we are using it in more than one place I've moved this function to model helpers file.

PR #2146, opened 2020-08-21
ISSUE #1976 - Add missing fields in risks filters

### LINKED ISSUE #1976: Safetibase search is missing the new fields
new fields added from v3 isn't on the chip search
image

Also, we're calling `/:ts/mitigation/criteria` every time we enter a risk detail. This field doesn't change and we're making too many database read for no a lot of reason. I think making the change to support search will avoid that as we will call this api at initialisation. But something to bear in mind!

m1chalkubiak:
This fixes #1976

#### Description
- As suggested call for `/:ts/mitigation/criteria` were moved to initialization of risk panel
- Additional filters were added base on `/:ts/mitigation/criteria` endpoint data

### DIFF
--- a/frontend/constants/risks.ts
+++ b/frontend/constants/risks.ts
@@ -112,7 +112,14 @@ export const RISK_FILTER_RELATED_FIELDS = {
 	CREATED_BY: 'creator_role',
 	RISK_OWNER: 'assigned_roles',
 	CATEGORY: 'category',
+	ELEMENT: 'element',
+	LOCATION: 'location_desc',
 	RISK_CONSEQUENCE: 'consequence',
+	RISK_FACTOR: 'risk_factor',
+	ASSOCIATED_ACTIVITY: 'associated_activity',
+	SCOPE: 'scope',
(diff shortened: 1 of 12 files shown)


m1chalkubiak:
@carmenfan 
Fixed the same filters that are loaded for the risk panel search are now also used in the kanban board for risks, previously I missed it.

carmenfan:
@m1chalkubiak  associated activity list is showing the risk factor data:



Can we also make the list scrollable? As demostrated on the gif the dropdown is getting cut off

m1chalkubiak:
@carmenfan 
Now it should be scrollable if there will be a need.

carmenfan:
> @carmenfan
> Now it should be scrollable if there will be a need.

@m1chalkubiak awesome thanks! Is this also addressed?

> @m1chalkubiak associated activity list is showing the risk factor data:

m1chalkubiak:
> @m1chalkubiak associated activity list is showing the risk factor data:

@carmenfan I just pushed a fix for this

carmenfan (approved):

PR #2170, opened 2020-09-03
ISSUE #1970 - Update of browser support checking 

### LINKED ISSUE #1970: Browser version check
We've been getting some users trying to access 3D Repo with very old safari versions which ends up with errors due to the lack of ES6 support. We should test the version of browser they are using and throw an error telling them to upgrade.

We don't know for sure which version we can't support, but we know we need webGL2 and ES6 support for the browser to have a chance to work.

ES 6 support:
image

WebGL2 support:
https://caniuse.com/#feat=webgl2

NOTE: If the browser does not support ES6, they're doomed, just tell them to upgrade (add to the IE check?). If the browser doesn't support Webgl2, they can access everything but the viewer, so maybe a warning should suffice.

m1chalkubiak:
This fixes #1970

#### Description
- [x] Make unsupported page unskippable (by removing ok button)
- [x] Adjust supported browsers versions
- [x] Sow pop up with a warning about lack of WebGL 2.0 support 

#### Test cases
- User should only see the pop up in case of lack of support of WebGL 2.0 in browser.
- The information about reading the warning should be saved per user, it should appear for him only once and be saved in a cookie.

### DIFF
--- a/frontend/services/unsupportedBrowserDetection.ts
+++ b/frontend/services/unsupportedBrowserDetection.ts
@@ -12,6 +12,8 @@ const DEFAULT_SUPPORTED_BROWSERS_CONFIG = {
 		browser: 'edge', minversion: 14,
 	}, {
 		os: 'mac os', minos: '10.12.0', browser: 'safari', minversion: 10.1,
+	}, {
+		browser: 'opera', minversion: 38,
 	}],
 	tablet: [{
 		os: 'ios', minos: '9', browser: 'mobile safari',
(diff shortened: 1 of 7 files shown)


frontend/services/unsupportedBrowserDetection.ts
```diff
@@ -3,15 +3,15 @@ import UAParser from 'ua-parser-js';
 
 const DEFAULT_SUPPORTED_BROWSERS_CONFIG = {
 	desktop: [{
-		browser: 'firefox', minversion: 41,
+		browser: 'firefox', minversion: 54,
```
Charence, line 6: Opera 38+ is missing from the list of supported browsers.

Screenshot 2020-09-10 at 11 23 40

Charence (changes requested):

frontend/services/unsupportedBrowserDetection.ts
m1chalkubiak: @Charence Thanks for pointing this. Opera with the supported version was added to the list.

Charence (approved):

PR #2229, opened 2020-09-29
ISSUE #2196 - Fix formatting and markdown support in descriptions

### LINKED ISSUE #2196: Display formatting and mark down support in descriptions
### Description
Formatting seems to be preserved but not displayed in issues description

I can see my new lines when I edit but it becomes a wall of text when i'm out of edit mode.

It would also be good if it's rendered with markdown like the comments, so things like bullet points would render nicely

I only tested this on issues, but it should apply for both risks and groups

### Goals
- [ ] Display intended formatting even when the user is out of edit mode
- [ ] Support markdown

MajaZubko:
This fixes #2196


har00n-haider:
Markdown in descriptions work for me on the branch build, on both risks and issues.

sanmont3drepo (approved):

PR #2234, opened 2020-10-02
ISSUE #2185 - Add character counter and abbreviate screenshot url in comments

### LINKED ISSUE #2185: Quoting a comment with an image prevents me from being able to write my comment
### Description
I am looking at Issue comments in 3D Repo. When I see a comment that I want to quote and add my notes to it, but I hit the comment character limit, because it includes the Image's URL link.

#### Tasks: _(Edit by CF)_
- [x] Add a character count on comments so the users are aware of the limit and how far they are from it.
- [x] Abbreviate references to screenshots of another comment (regex to detect the url and abbreviate it to something like `screenshot-<commentID>` )?

### Steps to replicate
- Create any issue in 3D Repo
- Add a comment with a model screenshot
- Quote the comment with a model screenshot
- Try and your message to the comment

Comment Quote

MajaZubko:
This fixes #2185

### DIFF
--- a/frontend/helpers/comments.ts
+++ b/frontend/helpers/comments.ts
@@ -22,10 +22,12 @@ import { getRiskConsequenceName, getRiskLikelihoodName } from './risks';
 import { sortByDate } from './sorting';
 
 export const INTERNAL_IMAGE_PATH_PREFIX = `API/`;
+export const INTERNAL_VIEWPOINT_ID_REGEX = new RegExp('#SS-[\\w-]+', 'gi');
 export const MARKDOWN_USER_REFERENCE_REGEX = new RegExp('@\\w+', 'gi');
 export const MARKDOWN_TICKET_REFERENCE_REGEX = new RegExp('#\\d+', 'gi');
 export const MARKDOWN_RESOURCE_REFERENCE_REGEX = new RegExp('#res.[\\w-]+', 'gi');
 export const MARKDOWN_INTERNAL_IMAGE_PATH_REGEX = new RegExp(`${INTERNAL_IMAGE_PATH_PREFIX}`, 'gi');
+export const VIEWPOINT_ID_REGEX = new RegExp('/viewpoints/[\\w-]+', 'gi');
 
 export interface IComment {
 	action?: {property: string, from: string, to: string, propertyText: string, text: string};
@@ -34,6 +36,12 @@ export interface IComment {
 	guid: number;
 	owner: string;
 	sealed: boolean;
+}
+
+export interface IDetails {
+	account: string;
+	_id: string;
+	model: string;
 }
 
 export const createAttachResourceComments = (owner: string,  resources = []) =>
@@ -289,7 +297,7 @@ const convertActionValueToText = (value = '') => {
 	return actionText;
 };
 
-export const transformCustomsLinksToMarkdown = ( comment: IComment, tickets, type ) => {
+export const transformCustomsLinksToMarkdown = ( details: IDetails, comment: IComment, tickets, type ) => {
 	let text = comment.comment;
 
 	if (!text || (Boolean(comment.action)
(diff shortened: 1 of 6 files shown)


frontend/helpers/comments.ts
```diff
@@ -300,6 +308,30 @@ export const transformCustomsLinksToMarkdown = ( comment: IComment, tickets, typ
 	const usersReferences = text.matchAll(MARKDOWN_USER_REFERENCE_REGEX);
 	const ticketsReferences = text.matchAll(MARKDOWN_TICKET_REFERENCE_REGEX) || [];
 	const resourcesReferences = text.matchAll(MARKDOWN_RESOURCE_REFERENCE_REGEX);
+	const viewpointReferences = text.matchAll(INTERNAL_VIEWPOINT_ID_REGEX);
+
+	if (viewpointReferences) {
+		const teamspace = details.account;
+		const projectId = details.model;
+		const issueId = details._id;
+		let referenceType;
```
m1chalkubiak, line 317:
```js
const referenceType = type === 'risk' ? 'risk' : 'issues';
```
We don't have any third option and it will be simpler and clearer this way.

frontend/helpers/comments.ts
```diff
@@ -300,6 +308,30 @@ export const transformCustomsLinksToMarkdown = ( comment: IComment, tickets, typ
...
+
+		if (type === 'issue') {
+			referenceType = 'issues';
+		} else if (type === 'risk') {
+			referenceType = 'risks';
+		}
+
+		const uniqViewpointReferences = uniqBy([...viewpointReferences], 0);
+		uniqViewpointReferences.forEach(({0: viewpointReference}) => {
```
m1chalkubiak, line 326: `{ 0: viewpointReference }`
There should be additional spaces, it just our syntax convention, not an error.

frontend/helpers/comments.ts
```diff
@@ -300,6 +308,30 @@ export const transformCustomsLinksToMarkdown = ( comment: IComment, tickets, typ
 	const usersReferences = text.matchAll(MARKDOWN_USER_REFERENCE_REGEX);
 	const ticketsReferences = text.matchAll(MARKDOWN_TICKET_REFERENCE_REGEX) || [];
 	const resourcesReferences = text.matchAll(MARKDOWN_RESOURCE_REFERENCE_REGEX);
+	const viewpointReferences = text.matchAll(INTERNAL_VIEWPOINT_ID_REGEX);
+
+	if (viewpointReferences) {
+		const teamspace = details.account;
```
m1chalkubiak, line 314: Destructuring with variable renaming is something that will be useful to you many times

```js
const { account: teamspace, model: projectId, _id: issueId } = details;
```

m1chalkubiak (changes requested):

m1chalkubiak (dismissed):

har00n-haider:
Implementation looks good. I can see the word count tick up for input characters, and the short links work really well.

#### Side notes

As a side note would it be better to block the user from posting the comment if the char limit is exceeded? Especially seeing as the user cannot edit comments that have already sent. 

Pasting the generated short link does not work across tickets. Not sure if this is ever likely to happen:
image

m1chalkubiak:
Hi @har00n-haider 

> As a side note would it be better to block the user from posting the comment if the char limit is exceeded? Especially seeing as the user cannot edit comments that have already sent.

I think this was a small issue with presenting error, We are blocking sending comments above 220 char. Previously, the information about the number of characters turned red also when it was equal to 220. I corrected it.
 
> Pasting the generated short link does not work across tickets. Not sure if this is ever likely to happen:

Currently, our shortcode contains only the last part, i.e. `viewpointId`, and code assume that the remaining information matches the place where we are trying to call it. In order to be able to use this shortcode between tickets, the shortcode would also need to contain a `ticketId`, if we wanted to use it between different risk / issue ticket types, we would also have to pass this information, etc.
 Which inevitably leads us to the point where the shortcode is similar in length to the real link, which no longer makes sense. So the question is what do we want to allow? @carmenfan 

```js
${teamspace}/${projectId}/${referenceType}/${ticketId}/viewpoints/${viewpointId}/screenshot.png
```

carmenfan:
@m1chalkubiak Nah I think it'll be overkill. Considering we currently don't allow users to quote a picture from a different issue/risk via the UI, the user will have to be some what markdown competent to do it, if they really want to do it, they can just share the full url of the screenshot.

carmenfan:
@m1chalkubiak  i'm assuming you can still display a picture with the right markdown syntax and a valid url to an image?

m1chalkubiak:
> @m1chalkubiak i'm assuming you can still display a picture with the right markdown syntax and a valid url to an image?

Of course, we did not change any of native markdown behavior.

carmenfan (changes requested):
Looks good in general, just a couple of minor points:

- [ ] The word count is really really faint in colour, can we darken it? (maybe the same font colour as the comments (not the current comment (black) but existing (darker grey)?

image

- [ ] Can we move it to the right? 
image

carmenfan (approved):

PR #2276, opened 2020-11-06
ISSUE #2198 - APM investigations and stats

### LINKED ISSUE #2198: Investigate Elastic APM
### Description
This is to start Testing Elastic APM

### Goals

- [x] Application logs presented in Elastic
- [x] Runtime statistics collected from docker images
- [x] Exceptions traceable

### Tasks

- [x] Get the APM module installed
- [x]  Learn how to configure it

keifgwinn:
This fixes #2198 

#### Description
To enable elastic APM logging the NodeJS Elastic module is added and optional configuration made available.

#### Test cases
System should boot as normal without elastic configuration, and emit elastic tracing when configured.


keifgwinn:
> This fixes #2198
> 
> #### Description
> To enable elastic APM logging the NodeJS Elastic module is added and optional configuration made available.
> 
> #### Test cases
> System should boot as normal without elastic configuration, and emit elastic tracing when configured.

Also contains the database reporting to elastic scripts

.azure/Docker/src/init.sh
```diff
@@ -16,7 +16,8 @@ if [ "$1" = 'web' ]; then
     fi
 
     umask 2
-    exec gosu $NODE_USERNAME node /home/node/3drepo.io/backend/3drepo.js
+    cd /home/node/3drepo.io
+    exec gosu $NODE_USERNAME ./run/run_app $NODE_ENV
```
keifgwinn, line 20: related to the first change, @carmenfan  asked that the run scripts be present in the docker to be able to modify them later if we need.

scripts/reporting/statsGenElastic/importOldStats.js
```diff
@@ -0,0 +1,174 @@
...
+			try {
+				// Get the files as an array
+				const files = await fs.promises.readdir( importFolder );
+				const seen = [] // array to catch previously seen id's
+
+				for( const file of files ) {
+					const fullPath = path.join( importFolder, file );
+					const stat = await fs.promises.stat( fullPath );
+					
+					if( stat.isFile() && file.includes(".csv") && !file.includes("com.dropbox.attrs")  ){
```
carmenfan, line 107: more reliable to check for endsWith (assuming .csv is still going to be an extension)
```suggestion
					if( stat.isFile() && file.endsWith(".csv") && !file.includes("com.dropbox.attrs")  ){
```

scripts/reporting/statsGenElastic/importOldStats.js
```diff
@@ -0,0 +1,174 @@
...
+						console.log( "'%s' is a csv file.", fullPath );
+
+						fs.createReadStream(fullPath)
```
carmenfan, line 110: This whole block could be it's own function really, just to reduce the tabbing to improve readability

scripts/reporting/statsGenElastic/importOldStats.js
```diff
@@ -0,0 +1,174 @@
...
+						.pipe(csv({
+								headers: ["Teamspace","LastLogin"]
+								}
+							))
+						.on('data', (data) => {
+							if ( !Utils.isUndefined(data.Teamspace) && !Utils.isUndefined( Utils.clean(data.LastLogin) ) ) {
+
+								const lastLogin = {
+									"Teamspace" : String(data.Teamspace),
```
carmenfan, line 119: not a biggie, but JS object is not the same as JSON, you don't need quotes for keys (unless you're using numbers as your key and don't want it to turn into an array)

scripts/reporting/statsGenElastic/importOldStats.js
```diff
@@ -0,0 +1,174 @@
...
+									"Last Login" : String(data.LastLogin),
+									"DateTime" : Utils.formatDate(new Date(data.LastLogin))
+								}
+
+								const id = Utils.hashCode( Object.values(lastLogin).toString() )
+
+								if (!seen.includes(id) ) {
```
carmenfan, line 126: includes is O(n), as this gets big your script will run slower.

unless you need it to be an array for some reason, it'll be much more effective to store this as an object, which will give you O(1) performance on this.

i.e. 
````js
   // declare it like this
   const seen = {};
   //check it like this
   if(!seen.id) { }
   //update it like this
   seen.id = true;
````

File: scripts/reporting/statsGenElastic/importOldStats.js [hunk 6]
```diff
@@ -0,0 +1,174 @@
...
+									}, {
+										ignore: [404],
+										maxRetries: 3
+									})
+									request
+									.then(result => {
+										if (result.body.hits.total.value === 0) {
+												console.log("creating",seen.length,":::",id)
+												Utils.createElasticRecord( ElasticClient, Utils.teamspaceIndexPrefix + "-login", lastLogin, id)
+												seen.push(id) // log hash is created to save calls
```
carmenfan, line 143: avoid duplication, move seen.push(id) outside of the if condition since you're pushing it anyway.

scripts/reporting/statsGenElastic/package-lock.json
```diff
@@ -0,0 +1,269 @@
+{
```
carmenfan, line 1: we use yarn over npm (yarn is generally faster), not a big deal in general, but might be better to use yarn instead for consistency.

scripts/reporting/statsGenElastic/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "3DRepoStatsGenElastic",
+  "version": "1.0.0",
+  "description": "Statistics exporter base on a 3D Repo database",
+  "main": "genStats.js",
+  "author": "Keif Gwinn",
+  "license": "AGPL3",
+  "private": false,
+  "dependencies": {
```
carmenfan, line 9: lock all the versioning here if you don't want things to randomly break  😆

scripts/reporting/statsGenElastic/utils.js
```diff
@@ -0,0 +1,137 @@
...
+	} catch (error) {
+			console.log("[ELASTIC] ERROR:" + Index)
+			console.log(Mapping)
+			console.log(error)
+			console.log(error.body.error)
+			Promise.reject()
+		}
+	}
+
+Utils.createElasticRecord = async ( ElasticClient, Index, elasticBody, id = Utils.hashCode( Object.values(elasticBody).toString() ) ) => {
```
carmenfan, line 108: This function is async but you never wait on it on the files that call it.

This is a bit dangerous, you might think the script is done before it has finished.

scripts/reporting/statsGenElastic/utils.js
```diff
@@ -0,0 +1,137 @@
...
+		concat(files, dest, (err) => {
+			if(err) error(err);
+			else resolve();
+		});
+	});
+}
+
+Utils.formatDate = (date) => {
+	return date.toISOString()
+	// return dateFormat(date, 'dd-mm-yy'); // old format
```
carmenfan, line 40: remove if not needed?

scripts/reporting/statsGenElastic/utils.js
```diff
@@ -0,0 +1,137 @@
...
+}
+
+Utils.generateFileName = (prefix) => {
+	const date = Utils.formatDate(new Date());
+	return `${prefix}_${date}.csv`;
+}
+
+Utils.mkdir = (dir) => {
+	return new Promise((resolve, error) => {
+		fs.mkdir(dir, {recursive: true}, (err) => {
```
carmenfan, line 50: this is just `fs.promise.mkdir`?

scripts/reporting/statsGenElastic/utils.js
```diff
@@ -0,0 +1,137 @@
...
+	return username === 'adminUser' || username === 'nodeUser' || username === 'undefined';
+}
+
+Utils.teamspaceIndexPrefix = 'io-teamspace'
+Utils.statsIndexPrefix = 'io-stats'
+
+Utils.isUndefined = (value) => {
+    // Obtain `undefined` value that's
+    // guaranteed to not have been re-assigned
+    var undefined = void(0);
```
carmenfan, line 67: const

carmenfan: this function is also a bit overkill... 😆

scripts/reporting/statsGenElastic/utils.js
```diff
@@ -0,0 +1,137 @@
...
+					})
+				console.log("[ELASTIC] Created index " + indexName )
+				await ElasticClient.indices.putMapping({
+					index: Index,
+					body: { properties: Mapping }
+					}
+				)
+				console.log("[ELASTIC] created mapping " + indexName  );
+		}
+		return Promise.resolve()
```
carmenfan, line 97: you don't need this line. by the virtue of declaring async on the function it'll return a promise.

scripts/reporting/statsGenElastic/utils.js
```diff
@@ -0,0 +1,137 @@
...
+		
+	} catch (error) {
+			console.log("[ELASTIC] ERROR:" + Index)
+			console.log(Mapping)
+			console.log(error)
+			console.log(error.body.error)
+			Promise.reject()
```
carmenfan, line 104: can use throw here.

scripts/reporting/statsGenElastic/utils.js
```diff
@@ -0,0 +1,137 @@
...
+					index: Index,
+					body: { properties: Mapping }
+					}
+				)
+				console.log("[ELASTIC] created mapping " + indexName  );
+		}
+		return Promise.resolve()
+		
+	} catch (error) {
+			console.log("[ELASTIC] ERROR:" + Index)
```
carmenfan, line 100: missing `;`

Probably worth setting up the linter to format this project.

scripts/reporting/statsGenElastic/utils.js
```diff
@@ -0,0 +1,137 @@
...
+			console.log(Mapping)
+			console.log(error)
+			console.log(error.body.error)
+			Promise.reject()
+		}
+	}
+
+Utils.createElasticRecord = async ( ElasticClient, Index, elasticBody, id = Utils.hashCode( Object.values(elasticBody).toString() ) ) => {
+	try {
+		const indexName = Index.toLowerCase() // requirement of elastic that indexs be lowercase
```
carmenfan, line 110: this bit looks identical to createElasticIndex. Maybe make it call that (with Mapping set to undefined so it doesn't call the last bit of createElastIndex?

carmenfan (changes requested):

backend/3drepo.js
```diff
@@ -16,15 +16,29 @@
 */
 
 "use strict";
+const config = require("./config.js");
+
+if ("elastic" in config) {
```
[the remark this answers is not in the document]
keifgwinn: https://github.com/3drepo/3drepo.io/blob/63e555b34b4a5e733de3c960a1856cc75feff1b1/backend/3drepo.js#L54-L55 is where I took the style from ;)

The instruction manual for APM says that it should be the first piece of code running, but if nothing else is executing before runServer it's safe to move there.

scripts/reporting/statsGenElastic/genStats.js
```diff
@@ -0,0 +1,167 @@
...
+*	along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+'use strict'
+const MongoClient = require('mongodb').MongoClient;
+const UserList = require('./userList');
+const DBStats = require('./dbStats');
+const Utils = require('./utils');
+
+try {
```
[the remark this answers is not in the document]
keifgwinn: it shouldn't be here... missed it cleaning up.

scripts/reporting/statsGenElastic/genStats.js
```diff
@@ -0,0 +1,167 @@
...
+
+async function start() {
+	console.log(`Trying to connect to ${url}...`);
+	try {
+		const db = await client.connect();
+		console.log('Connected successfully!');
+
+		// initialise indicies if missing
+
+		const activityMapping = {
```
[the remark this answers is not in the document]
keifgwinn: ok making an elastic module

scripts/reporting/statsGenElastic/genStats.js
```diff
@@ -0,0 +1,167 @@
...
+	if ( errors.length > 0 ) { throw errors }
+} catch (e) {
+    for (let i=0; i<e.length; i++) {
+		console.log(e[i].message);
+	  }
+	process.exit(-1);
+}
+
+try {
+	process.env.ELASTIC_CLOUD_AUTH.split(":")
```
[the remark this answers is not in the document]
keifgwinn: the format is servername:base64 encrypted stuff, if a split doesn't error then I need to test the length of the returned array

backend/3drepo.js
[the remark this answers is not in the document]
carmenfan: > https://github.com/3drepo/3drepo.io/blob/63e555b34b4a5e733de3c960a1856cc75feff1b1/backend/3drepo.js#L54-L55
> 
> is where I took the style from ;)

yea. legacy shit 😆 ... There's plenty of that unfortunately

> The instruction manual for APM says that it should be the first piece of code running, but if nothing else is executing before runServer it's safe to move there.

runServer is the only thing that's called in 3drepo.js (everything else is function declarations), should keep it that way so it's easier to navigate.  Can make it the first thing it does on the runServer

scripts/reporting/statsGenElastic/genStats.js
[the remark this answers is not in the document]
carmenfan: yea split doesn't error as long as it's type string, it'll just return you the an array of 1 if it can't find the split value 
image

scripts/reporting/statsGenElastic/genStats.js
```diff
@@ -0,0 +1,167 @@
...
+		if ( Utils.clean(process.env.STATS_RUN_DAILY ) ){
+			await UserList.createUsersReport(db, ElasticClient) 
+		} else { 		
+			await Promise.all([
+				UserList.createUsersReport(db, ElasticClient),
+				DBStats.createDBReport(db, ElasticClient)
+			]);
+		}
+
+		await client.close();
```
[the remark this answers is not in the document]
keifgwinn: It's just a thin http wrapper so not required but will do for neatness.

scripts/reporting/statsGenElastic/utils.js
keifgwinn: I didn't write this bit, don't need it at all in this version but came through from the file based one.

scripts/reporting/statsGenElastic/utils.js
keifgwinn: ok refactored to a single function with extra options. It's like this because initially I wasn't creating mappings until I ran into problems later.

scripts/reporting/statsGenElastic/importOldStats.js
keifgwinn: done

scripts/reporting/statsGenElastic/importOldStats.js
keifgwinn: I did it for consistency as there's fields with spaces

keifgwinn:
/azp run 

actually testing clone now we've purged branches

scripts/reporting/statsGenElastic/importOldStats.js
keifgwinn: done

.gitignore
```diff
@@ -86,7 +86,14 @@ npm-debug.log.*
 .vscode/
 *.swo
 
-submodules/tests
+submodules/tests
 scripts/reporting/statsGen/yarn.lock
 scripts/reporting/statsGen/*/*.csv
 scripts/reporting/statsGen/*/*/*.csv
+scripts/reporting/statsGenElastic/yarn.lock
```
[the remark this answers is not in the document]
carmenfan: in general for 3drepo.io, we removed yarn.lock because there was some libraries that we use which would always conflict with different machine types (what worked on a windows machine produced an incompatible yarn.lock file for mac)

So basically everytime we merged in code, depending on who committed it you may have to remove the yarn.lock file because it's incompatible. So we removed it in favour of locking down every version in package.json

scripts/reporting/statsGenElastic/elastic.js
```diff
@@ -0,0 +1,155 @@
...
+		console.log("[ELASTIC] -- Client Health --",resp);
+	});
+	return ElasticClient;
+};
+
+Elastic.createElasticRecord = async (ElasticClient, Index, elasticBody, id, mapping) => {
+	try {
+
+		if (elasticBody) {
+			id = id || Utils.hashCode(Object.values(elasticBody).toString());
```
carmenfan, line 49: Can just get rid of the if/else statement in favour of this

```suggestion
			id = id || Utils.hashCode(Object.values(elasticBody || {}).toString());
```

scripts/reporting/statsGenElastic/elastic.js
```diff
@@ -0,0 +1,155 @@
...
+		if (elasticBody) {
+			await ElasticClient.index({
+				index: indexName,
+				id: id,
+				refresh: true,
+				body: elasticBody
+			});
+			console.log("[ELASTIC] created doc " + indexName + " " + Object.values(elasticBody).toString());
+		}
+		return Promise.resolve();
```
carmenfan, line 80: don't really need this line, in an async function you're automatically returning this when you exit

scripts/reporting/statsGenElastic/elastic.js
```diff
@@ -0,0 +1,155 @@
...
+	} catch (error) {
+		console.error("[ELASTIC] ERROR:" + Index, elasticBody, error);
+		Promise.reject();
```
carmenfan, line 83: This line doesn't do anything, you're creating a rejected promise, but you're not returning it anyway. It's superseded by L84

scripts/reporting/statsGenElastic/elastic.js
```diff
@@ -0,0 +1,155 @@
...
+		throw(error.body.error);
+	}
+};
+
+Elastic.createElasticIndex = async (ElasticClient, Index, mapping) => {
+	try {
+		Elastic.createElasticRecord (ElasticClient, Index, undefined, undefined, mapping);
+	} catch (error) {
+		Promise.reject();
```
carmenfan, line 92: redundant line

scripts/reporting/statsGenElastic/elastic.js
```diff
@@ -0,0 +1,155 @@
...
+	} catch (error) {
+		console.error("[ELASTIC] ERROR:" + Index, elasticBody, error);
+		Promise.reject();
+		throw(error.body.error);
+	}
+};
+
+Elastic.createElasticIndex = async (ElasticClient, Index, mapping) => {
+	try {
+		Elastic.createElasticRecord (ElasticClient, Index, undefined, undefined, mapping);
```
carmenfan, line 90: don't you need await here? Currently this function is declared async without handling any promises

scripts/reporting/statsGenElastic/elastic.js
```diff
@@ -0,0 +1,155 @@
...
+		request_timeout: 60
+	});
+
+	ElasticClient.cluster.health({},function(err,resp) {
+		console.log("[ELASTIC] -- Client Health --",resp);
+	});
+	return ElasticClient;
+};
+
+Elastic.createElasticRecord = async (ElasticClient, Index, elasticBody, id, mapping) => {
```
carmenfan, line 45: function parameters should really be in camel case. naming things as `ElasticClient` and `Index` gives the illusion that these came from `required(...)` or constants instead of function parameters. 


This applies with the rest of the files

```suggestion
Elastic.createElasticRecord = async (elasticClient, index, elasticBody, id, mapping) => {
```

scripts/reporting/statsGenElastic/newUsersPerMonth.js
```diff
@@ -0,0 +1,105 @@
...
+				writeNewElasticDocument(ElasticClient, month, year, counter, total);
+				counter = 0;
+				year = month === 12 ? year + 1 : year;
+				month = month + 1 > 12 ? 1 : month + 1;
+				if(year > 2018) {
+					break;
+				}
+			} while(!(month === nextM && year === nextY));
+		} else {
+			writeNewElasticDocument (ElasticClient, currentM, currentY, count, total);
```
carmenfan, line 50: await here?

scripts/reporting/statsGenElastic/newUsersPerMonth.js
```diff
@@ -0,0 +1,105 @@
...
+	users.forEach((user) => {
+		const createdAt = user.customData.createdAt;
+		const month = createdAt.getMonth() + 1;
+		const year = createdAt.getFullYear();
+
+		if(currentMonth === month && currentYear === year) {
+			++currentCount;
+		} else {
+			total += currentCount;
+			writeNewUserEntry(ElasticClient, currentYear, currentMonth, currentCount, total, month, year);
```
carmenfan, line 74: await?

scripts/reporting/statsGenElastic/newUsersPerMonth.js
```diff
@@ -0,0 +1,105 @@
...
+			currentYear = year;
+			currentCount = 1;
+		}
+	});
+
+	if(currentMonth !== -1) {
+		const now = new Date();
+		const targetDate = new Date(now.getFullYear(), now.getMonth() + 1, 1);
+		total += currentCount;
+		writeNewUserEntry(ElasticClient, currentYear, currentMonth, currentCount, total, targetDate.getMonth() + 1, targetDate.getFullYear());
```
carmenfan, line 86: await.?

scripts/reporting/statsGenElastic/newUsersPerMonth.js
```diff
@@ -0,0 +1,105 @@
...
+	}
+};
+
+const NewUsers = {};
+
+NewUsers.createNewUsersReport = (dbConn, ElasticClient) =>{
+	return new Promise((resolve, reject) => {
```
carmenfan, line 93: You don't need a promise around this. `reportNewUsersPerMonth` returns you a promise. you can just return that.
```suggestion
	return reportNewUsersPerMonth(dbConn, ElasticClient).then(() => {
```

or

```suggestion
       await reportNewUsersPerMonth(dbConn, ElasticClient);
       console.log("[DB] Generated NewUsersReport");
```

scripts/reporting/statsGenElastic/importOldStats.js
```diff
@@ -0,0 +1,174 @@
...
+*/
+
+'use strict'
+const MongoClient = require('mongodb').MongoClient;
+const Utils = require('./utils');
+const fs = require( 'fs' );
+const path = require( 'path' );
+const csv = require('csv-parser');
+
+var errors = [];
```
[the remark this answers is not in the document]
carmenfan: none of the points on this file are addressed. My opinion is still the same:
-  if it's a run and ditch and never used again then don't even bother committing it. 
- If it's committed then it means there's an intention to be reused and maintained, in which case it needs tidying up. Even if you don't do everything, it needs to be at least readable.

scripts/reporting/statsGenElastic/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "author": "Keif Gwinn",
+  "license": "AGPL3",
+  "private": true,
+  "scripts": {
+    "lint": "eslint ./",
+    "lint:fix": "eslint --fix ./"
+  },
+  "dependencies": {
+    "@elastic/elasticsearch": "7.9.1",
+    "concat-files": "^0.1.1",
```
carmenfan, line 15: this one needs locking still

scripts/reporting/statsGenElastic/teamspaceActivity.js
```diff
@@ -0,0 +1,149 @@
...
+	const stats = await Promise.all(modelProm);
+
+	const finalStats = stats.reduce(accumulateStats, {});
+	await printStatsToElastic(ElasticClient, finalStats, ts, licenseType);
+}
+
+const TS = {};
+
+TS.createTeamspaceActivityReport = (dbConn, ElasticClient, ts, licenseType) =>{
+	return new Promise((resolve, reject) => {
```
carmenfan, line 138: Same here new Promise isn't needed as reportActivity returns you a promise

scripts/reporting/statsGenElastic/teamspaceQuota.js
```diff
@@ -0,0 +1,105 @@
...
+	console.log ("[QUOTA] Writing Licenced Teamspaces Quota information");
+	const enterpriseTS = await writeQuotaDetails(dbConn, col, ElasticClient, true);
+	const discretionaryTS = await writeQuotaDetails(dbConn, col, ElasticClient, false);
+	return [...enterpriseTS, ...discretionaryTS];
+};
+
+const TS = {};
+
+TS.createTeamspaceReport = async (dbConn, ElasticClient) =>{
+	return new Promise((resolve, reject) => {
```
carmenfan, line 93: same here

carmenfan (changes requested):
@keifgwinn I stopped pointing it out, but all the ElasticClient function parameters needs to have a lowercase e as it's not a import/constant.

scripts/reporting/statsGenElastic/importOldStats.js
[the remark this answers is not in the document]
keifgwinn: removed

carmenfan (approved):

PR #2278, opened 2020-11-06
ISSUE #2277 remove default background colour

### LINKED ISSUE #2277: Remove background colour from unity default settings
No matter how hard you try, unity builds will always come with a default background colour.

There's no way to remove it on the unity build configs.

So the only way we can work around it is remove the property when we load it into unityutils.

carmenfan:
This fixes #2277
#### Description
- Completely removing unity background should no longer show the default grey background.


carmenfan:
@m1chalkubiak  this should solve your problem on https://github.com/3drepo/3D_Send/pull/45

Let me know if it works?

carmenfan:
This isn't what I think it is, need further investigation. @m1chalkubiak

PR #2293, opened 2020-11-12
Issue #2284 - Removed unnecessary fetchs for viewpoints groups

### LINKED ISSUE #2284: Remove unnecessary fetchs of viewpoint groups for recently created views/issues/risks/comments
### Description
After a view/issue/risk/comment/group has been created, it fetches data that was already known in the creation process.

sanmont3drepo:
This fixes #2284

### DIFF
--- a/frontend/modules/viewpoints/viewpoints.sagas.ts
+++ b/frontend/modules/viewpoints/viewpoints.sagas.ts
@@ -15,14 +15,16 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { groupCollapsed } from 'console';
 import copy from 'copy-to-clipboard';
 import { get, take } from 'lodash';
 import { all, put, select, takeLatest } from 'redux-saga/effects';
-import { selectSelectedViewpoint } from '.';
+import { selectSelectedViewpoint, selectViewpointsGroups } from '.';
 import { CHAT_CHANNELS } from '../../constants/chat';
 import { ROUTES } from '../../constants/routes';
 import { UnityUtil } from '../../globals/unity-util';
 import { createGroupsByColor, createGroupsByTransformations, prepareGroup } from '../../helpers/groups';
+import { mergeGroupsDataFromViewpoint, setGroupData } from '../../helpers/viewpoints';
 import * as API from '../../services/api';
 import { Viewer } from '../../services/viewer/viewer';
 import { ChatActions } from '../chat';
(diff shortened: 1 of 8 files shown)


carmenfan (changes requested):
@sanmont3drepo 

this works in general, but a little suboptimal with comments. We're re-fetching them if the user exits the risk/issue and re-enter, as the issue/risks data itself has refreshed.

I think it's better if we store a cache somewhere (i.e. `{groupid: data}`). perhaps on the code that fetches the group (with a boolean to ignore cache if needed), so the API would just return the data immediately if it's in the cache, and we don't need to worry manually gluing the data into the viewpoints.

frontend/modules/viewpoints/viewpoints.sagas.ts
```diff
@@ -214,17 +214,16 @@ export function* showViewpoint({teamspace, modelId, view, ignoreCamera}) {
 			yield Viewer.isViewerReady();
 
 			const viewpoint = view.viewpoint;
-			if (!viewpoint) {
```
carmenfan, line 217: if theres' no viewpoint there's nothing to do?

I don't think it's possible to have clipping planes but no view. (esp after the refactoring with views/issues/risks.)

Any reason why this was changed?

frontend/modules/viewpoints/viewpoints.sagas.ts
```diff
@@ -214,17 +214,16 @@ export function* showViewpoint({teamspace, modelId, view, ignoreCamera}) {
...
-				return;
-			}
 
 			if (viewpoint?.up && !ignoreCamera) {
 				yield put(ViewerGuiActions.setCamera(viewpoint));
 			}
 
 			const clippingPlanes = view.clippingPlanes || get(view, 'viewpoint.clippingPlanes');
 
-			yield put(TreeActions.setHiddenGeometryVisible(viewpoint.hideIfc !== false));
```
carmenfan, line 227: I think we need the original logic.

If the viewpoint didn't specify hideIfc (which is possible for viewpoints created by user via API I believe), it should be true by default. The modified version wouldn't capture this.

carmenfan (changes requested):

carmenfan:
> > @sanmont3drepo
> > this works in general, but a little suboptimal with comments. We're re-fetching them if the user exits the risk/issue and re-enter, as the issue/risks data itself has refreshed.
> > I think it's better if we store a cache somewhere (i.e. `{groupid: data}`). perhaps on the code that fetches the group (with a boolean to ignore cache if needed), so the API would just return the data immediately if it's in the cache, and we don't need to worry about manually gluing the data into the viewpoints.
> 
> Im not very fond of re-fetching the risks and issue data TBH. And because the groups are created for each particular issue/risk/viewpoint they cant be related to another issue/risk/viewpoint, so it wouldn't be shared between them.
> 
> **BUT** because we cant rely so much on the chat connection to keep the issues/risks updated , I still believe that we have to refetch those in case that their group data changed.
> 
> **So I'll change it to have a cache of groups in the viewpoints redux state part (it would complicate and add nothing if we add them to the general groups redux slice)**

My whole thinking stem from the `mergeGroupsDataFromViewpoint` calls on the risks/issues/views saga. I think it's a bit... manual? 

I think this should just be managed by the viewpoint saga that handles dealing with processing a viewpoint, and seamless to the features that utilises viewpoints. It's very coupled and if there's a new feature that utilises viewpoint, it will also have to manually call that on demand.

But I'm unsure how much I care about this. so I'm happy to drop it if you disagree.

frontend/modules/viewpoints/viewpoints.sagas.ts
carmenfan: reason I ask is because I remember putting this in to plug a white screen problem. so I'm not sure if we're recreating it again.

carmenfan (approved):

carmenfan:
@sanmont3drepo  looks fine, just waiting on Travis

PR #2303, opened 2020-11-20
ISSUE #2237 - Fix screenshots updating

### LINKED ISSUE #2237: Issue or SafetiBase Screenshot disappears if you update the viewpoint without updating the screenshot
### Description
When I open an existing Issue or SafetiBase risk, I can update its current viewpoint. If I do so, I have an option to update the screenshot. If I press no, then I only update the viewpoint. Next time open the same Issue or Risk, it shows the new viewpoint camera position, however, the screenshot is gone! ☹

### Steps to replicate
- Open the model viewer
- Create an Issue/Risk
- Come out of the new Issue/Risk
- Open the new Issue/Risk and see the screenshot in the card
- Update the viewpoint (without updating the screenshot)
- Go back to the Teamspace
- Open the same model again
- Open the Issue/Risk you have created and update the viewpoint 
- Notice how the screenshot is missing in the card now
- Bye Bye my beautiful markup it's been a pleasure ( ಥ_ʖಥ )
**Update the Viewpoint**

Update Risk Viewpoint

**Open the Risk/Issue again**
Updated Risk

m1chalkubiak:
This fixes #2237

### DIFF
--- a/backend/models/viewpoint.js
+++ b/backend/models/viewpoint.js
@@ -52,8 +52,6 @@ const clean = function(routePrefix, viewpointToClean, serialise = true) {
 
 		if (serialise) {
 			setViewpointScreenshotURL(routePrefix, viewpointToClean);
-			viewpointToClean.screenshot_ref = undefined;
-
 		}
 	}
 
--- a/backend/models/viewpoint.js
+++ b/backend/models/viewpoint.js
@@ -52,6 +52,7 @@ const clean = function(routePrefix, viewpointToClean, serialise = true) {
 
 		if (serialise) {
 			setViewpointScreenshotURL(routePrefix, viewpointToClean);
+			viewpointToClean.screenshot_ref = undefined;
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


backend/models/viewpoint.js
```diff
@@ -52,8 +52,6 @@ const clean = function(routePrefix, viewpointToClean, serialise = true) {
 
 		if (serialise) {
 			setViewpointScreenshotURL(routePrefix, viewpointToClean);
-			viewpointToClean.screenshot_ref = undefined;
```
m1chalkubiak, line 55: @Charence Can You take a look at this branch? Removing this line, I've fixed the issue with updating the viewpoint without passing a new screenshot. However, I broke the backend tests. Probably this line was here for a reason, but in my cases, its absence helps. When is this entry needed?

carmenfan: This is the internal reference on where the screenshot is stored, it shouldn't ever be exposed to the public domain so removing this should be fine. The backend probably have golden data that contains it and thus erroring out. I'll have a look. Thanks!

carmenfan (dismissed):

carmenfan (approved):

PR #2324, opened 2020-11-30
ISSUE #2314 - Add highlight all similar element & copy rule to bim panel

### LINKED ISSUE #2314: Select Elements based on parameter
### Description
Once the user has the BIM Card open, if they select a value against a parameter they can choose to either:
- Copy the rule in JSON format so it can be pasted into the Smart Groups rules box. eg.
`[{"field":"BrickName","operator":"IS","values":["Plate 2X4"],"_id":"criterion-36"}]`
- Highlight all elements with the same parameter value. This is enabled by the new API #2275 

Object selection

### Goals

- [x] As a user I can select multiple elements which match a found criteria
- [x] As a user I can save the rule to my clipboard so that I can use it in groups either immediately, or at a later date

m1chalkubiak:
This fixes #2314

#### Description
- As a user I can select multiple elements which match a found criteria
- As a user I can save the rule to my clipboard so that I can use it in groups either immediately, or at a later date


frontend/modules/bim/bim.sagas.ts
```diff
@@ -41,6 +45,28 @@ export function* fetchMetadata({ teamspace, model, metadataId }) {
...
 
+export function* highlightsAllSimilar({ rules }, colour?) {
+	try {
+		const teamspace = yield select(selectCurrentModelTeamspace);
+		const revision = yield select(selectCurrentRevisionId);
+		const model = yield select(selectCurrentModel);
+		const { data } = yield API.getMeshIDsByQuery(teamspace, model, rules, revision);
+
+		return Promise.all(data.map(({ model: modelId, mesh_ids }) => {
+			return Viewer.highlightObjects(teamspace, modelId, colour, true, true, mesh_ids);
```
carmenfan, line 56: Highlighting needs to go through the tree saga, otherwise it's going to go out of sync with the tree.

frontend/services/api/meta.ts
```diff
@@ -66,3 +66,25 @@ export const removeStarredMeta = (metaRecordKey) => {
...
+};
+
+/**
+ * Get Metadata by query
+ * @param teamspace
+ * @param modelId
+ * @param rules
+ * @param revision
+ */
+export const getMetadataByQuery = (teamspace, modelId, rules, revision = 'master/head') => {
```
carmenfan, line 88: Is this used?

carmenfan (changes requested):

frontend/services/api/meta.ts
carmenfan: Yea I'm not sure if we'll ever use it on the front end. This was more added for API users in mind

frontend/modules/bim/bim.sagas.ts
m1chalkubiak: Thanks for pointing this out, I tested cases with the tree panel and in fact, it didn't work well. However, using Tree sagas action caused a different issue from the user perspective, so whenever we were calling select similar action from bim panel, bim panel was reloaded with new element selected. From what I've understood it was selecting the last node in the provided nodes array. I added an additional property to tree action to omit this in our case.

Additionally, I added tooltips, as I got the feeling that It wasn't really obvious what these new icons do. 

Feel free if You see anything else that can be improved.

carmenfan (approved):
This looks fine, the only problem I have is that the query is super slow. But this is out of scope for this feature. I've created a separate issue here: https://github.com/3drepo/3drepo.io/issues/2352

Awaiting @mosment3DRepo's approval.

mosment3DRepo:
In light of the delay, as Carmen mentioned above, I think we should reduce the scope a little and keep it with just the Groups related content for now.

Are we able to replace the highlight button with a "Create Group" button instead, this would resolve the speed issue

eg. The add group button is selected, the groups card is activated with a new group and the filter relating to the parameter is automatically applied

carmenfan:
> In light of the delay, as Carmen mentioned above, I think we should reduce the scope a little and keep it with just the Groups related content for now.
> 
> Are we able to replace the highlight button with a "Create Group" button instead, this would resolve the speed issue
> 
> eg. The add group button is selected, the groups card is activated with a new group and the filter relating to the parameter is automatically applied

If you want the highlight functionality over create group. I suggest we just leave it as it is. I plan to look into the speed issue during the maintenance release anyway, so this may be wasted work.

So if you're happy with the functionality then I suggest we just merge this and if it comes to the point where we decided we can't resolve the speed issue, we change this to create group?

mosment3DRepo:
I think I am having a selection issue caused by #2366 

When I open model https://issue-2314.dev.3drepo.io/viewer/mosment3ddev/cfe03e40-f2ba-11ea-a248-318ee49d34a4 and select a door. If I use the select elements with similar parameter on the Category = Doors it crashes:

doorcrash

Am I right in thinking the bug is still active on this branch. In which case lets merge it and test it along with everything else

mosment3DRepo (approved):
Accepted subject to greater testing in release QA process

Charence:
@mosment3DRepo @carmenfan I guess this problem on windows is the same as the one for doors?

Dec-23-2020 12-02-32

PR #2325, opened 2020-11-30
Issue 2322 - 4.11.4  bug fixes

### LINKED ISSUE #2322: 4.11.4 bug list
- [x] [When the next frame of a sequence isn't preloaded the view jumps to the default configuration whilst it's loading](https://github.com/3drepo/3drepo.io/issues/2322#issuecomment-734995683)
- [x] [Issue isn't transforming meshes until you go into the issue](https://github.com/3drepo/3drepo.io/issues/2322#issuecomment-734996147)

carmenfan:
This fixes #2322

#### Description
- store previous Date in the state, and only use the state information of the current date if it is available
     - this stops the break when the state of the current date is loading.
- views transformation take precedence over sequence transformation so the user can see the transformations of the issues when they have a sequence open.

### DIFF
--- a/frontend/modules/sequences/sequences.redux.ts
+++ b/frontend/modules/sequences/sequences.redux.ts
@@ -45,6 +45,7 @@ export const INITIAL_STATE = {
 	selectedSequence: null,
 	lastSelectedSequence: null,
 	selectedDate: null,
+	lastSelectedDate: null,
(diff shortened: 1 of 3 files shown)


carmenfan:
@3drepo/product-team available on https://issue-2322.dev.3drepo.io/login

PR #2330, opened 2020-12-02
ISSUE #1199 - Add teamspace chat notifications & handle model create event

### LINKED ISSUE #1199: New model messages aren't processed by Front end
### Description
When user A creates a model in project, online collaborators do not get instant notification of this. This is because new model notifications are sent in the new model's modelID channel, which no one would be listening to as no one knew it existed!

We need to look into this and probably create project level (or teamspace level?) chat notifications.

### Steps to replicate
- Open 2 browsers log in as 2 different user
- User 1 creates a new model in a shared project
- User 2 will not see new model being created.

m1chalkubiak:
This fixes #1199

#### Description
I based on uncommented `ChatEvent.newModel` from BE. What I do not like about this is the fact that the event is sent to a channel named after a given teamspace and as a consequence, I have to assign myself to all teamspace channels on FE, hence the loop in the `teamspaces.sagas.ts` file.
It seems to me that a better solution would be to send this event to the channel of a user who should know about creating such a model, but it would require more work on the BE site.
If there is someone free to help with the BE part, we could handle the deletion and update of the model - these things are currently not working, no such events are sent from BE.

### DIFF
--- a/backend/models/helper/model.js
+++ b/backend/models/helper/model.js
@@ -307,10 +307,11 @@ function createNewModel(teamspace, modelName, data) {
 						model:  settings._id,
 						name: modelName,
 						permissions: C.MODEL_PERM_LIST,
-						timestamp: undefined
+						timestamp: undefined,
+						projectName
 					};
 
-					// ChatEvent.newModel(data.sessionId, teamspace, modelData);
+					ChatEvent.newModel(data.sessionId, teamspace, modelData);
 					return {modelData, settings};
 				});
 			});
--- a/frontend/constants/chat.ts
+++ b/frontend/constants/chat.ts
@@ -5,5 +5,6 @@ export const CHAT_CHANNELS = {
 	VIEWS: 'views',
 	MODEL: 'model',
 	NOTIFICATIONS: 'notifications',
-	RESOURCES: 'resources'
+	RESOURCES: 'resources',
+	TEAMSPACES: 'teamspaces',
 };
--- a/frontend/modules/chat/channel.ts
+++ b/frontend/modules/chat/channel.ts
@@ -22,6 +22,7 @@ import { ModelChatEvents } from './models.chat.events';
 import { NotificationsChatEvents } from './notifications.chat.events';
 import { RisksChatEvents } from './risks.chat.events';
 import { Subscriptions } from './subscriptions';
+import { TeamspacesChatEvents } from './teamspaces.chat.events';
 
 const getEventName = (teamspace: string, model: string, keys: string, event: string) => {
 	const eventName = [teamspace];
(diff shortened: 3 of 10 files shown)


carmenfan:
@m1chalkubiak 

> It seems to me that a better solution would be to send this event to the channel of a user who should know about creating such a model, but it would require more work on the BE site.

Not sure if this is really any better. at the moment we rely on the user to subscribe to what they're interested in, so the burden is on front end to connect to the channel by topic (i.e. teamspace, model etc

If we do it the other way round where users only have their own channels to subscribe to, then the very same burden lies on the backend, but a even more complex. On the front end, when you subscribe to the teamspaces, you know very well what you're subscribing to. But if we do it the other way round, when , for e.g., a user creates a new model, the backend will now need to pull out the list of users who would have access to this model, and loop through it to send a message to each user channel. This is a piece of information that backend currently doesn't need. And if there's 100 users that have access, there could be only 1 user who's online, and yet we'd have to send 100 messages through the channels, just incase.

So unless there's a cleverer way of doing it, I feel like the burden is actually a bit less keeping it this way

> If there is someone free to help with the BE part, we could handle the deletion and update of the model - these things are currently not working, no such events are sent from BE.

We don't really have the man power to look into this atm. Is this good enough as it is  + put the other functionality in a separate issue?

m1chalkubiak:
@carmenfan 
> > It seems to me that a better solution would be to send this event to the channel of a user who should know about creating such a model, but it would require more work on the BE site.
> 
> Not sure if this is really any better. at the moment we rely on the user to subscribe to what they're interested in, so the burden is on front end to connect to the channel by topic (i.e. teamspace, model etc
> 
> If we do it the other way round where users only have their own channels to subscribe to, then the very same burden lies on the backend, but even more complex. On the front end, when you subscribe to the teamspaces, you know very well what you're subscribing to. But if we do it the other way round, when , for e.g., a user creates a new model, the backend will now need to pull out the list of users who would have access to this model, and loop through it to send a message to each user channel. This is a piece of information that backend currently doesn't need for uploading a model. And if there's 100 users that have access, there could be only 1 user who's online, and yet we'd have to send 100 messages through the channels, as we'd have no idea who's online and who's not online until the point it hits the chatserver code.
> 
> So unless there's a cleverer way of doing it, I feel like the burden is actually a bit less keeping it this way

I understand, so let's leave it as it is now.

> > If there is someone free to help with the BE part, we could handle the deletion and update of the model - these things are currently not working, no such events are sent from BE.
> 
> We don't really have the man power to look into this atm. Is this good enough as it is + put the other functionality in a separate issue?

Sure, the issue created: https://github.com/3drepo/3drepo.io/issues/2343

carmenfan (approved):

PR #2333, opened 2020-12-02
Issue #2332 - Fixed multiple fetchs for the same frame in sequences

### LINKED ISSUE #2332: The same frame state is being fetchetd multiple times
### Description
The same frame is being fetched when you enter the sequence in this model:
https://staging.dev.3drepo.io/viewer/mosment3ddev/5a74b8c0-33dd-11eb-973d-01976be89052

sanmont3drepo:
This fixes #2332

### DIFF
--- a/frontend/modules/sequences/sequences.selectors.ts
+++ b/frontend/modules/sequences/sequences.selectors.ts
@@ -133,7 +133,7 @@ export const selectNextKeyFramesDates =  createSelector(
 				date = getDateByStep(date, scale, interval);
 				nextFrame = getSelectedFrame(frames, date);
 
-				while (lastFrame === nextFrame && nextFrame !== null) {
+				while (lastFrame === nextFrame && date <= maxDate) {
 					date = getDateByStep(date, scale, interval);
 					nextFrame = getSelectedFrame(frames, date);
 				}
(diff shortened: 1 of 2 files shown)


frontend/modules/sequences/sequences.selectors.ts
```diff
@@ -120,16 +120,26 @@ export const selectSelectedStartingDate = createSelector(
...
 			const keyFrames = [];
 			keyFrames[0] = new Date(Math.min(maxDate, (startingDate || new Date(0)).valueOf()));
+			let lastFrame = getSelectedFrame(frames, keyFrames[0]);
+			let nextFrame = null;
 			let date = getDateByStep(startingDate, scale, interval);
 			maxDate = new Date(maxDate);
 
 			for (let i = 0; i < 3 ; i++) {
 				date = getDateByStep(date, scale, interval);
+				nextFrame = getSelectedFrame(frames, date);
```
carmenfan, line 134: @sanmont3drepo  I think this can be simplified.

If I understood this correctly, we're just trying to find the next 3 key frames right, regardless of the step?

This can be achieved by just looking at the frames array in the sequence. Since we've moved out the activitiy entries, each frame entry in sequence.frames is guaranteed to be a unique frame state.

sanmont3drepo: The problem with doing that is the step, if you have a big step that spans through multiple frames, you would actually want just the steps that you are going through instead of fetching the inbetweens.

PR #2356, opened 2020-12-11
ISSUE 1972 - Remove Mongoose from groups and counter

### LINKED ISSUE #1972: Remove Mongoose
Remove mongoose dependency from backend

- [x] accountPermission (santiago) https://github.com/3drepo/3drepo.io/issues/2391
- [x] billingAddress(santiago) https://github.com/3drepo/3drepo.io/issues/2391
- [x] counter (santiago)
- [x] group (charence)
- [x] history (santiago) https://github.com/3drepo/3drepo.io/issues/2381
- [ ] invoice (santiago) https://github.com/3drepo/3drepo.io/issues/2379
- [ ] ipn (santiago) https://github.com/3drepo/3drepo.io/issues/2422
- [x] job (charence) https://github.com/3drepo/3drepo.io/issues/2370
- [ ] modelSetting (charence) https://github.com/3drepo/3drepo.io/issues/2387
- [x] permissionTemplate https://github.com/3drepo/3drepo.io/issues/2391
- [x] project (charence) https://github.com/3drepo/3drepo.io/issues/2345
- [x] ref (charence) https://github.com/3drepo/3drepo.io/issues/2372
- [ ] repo (santiago) https://github.com/3drepo/3drepo.io/issues/2423
- [x] role (santiago) https://github.com/3drepo/3drepo.io/issues/2383
- [x] scene (charence) https://github.com/3drepo/3drepo.io/issues/2377
- [x] user (santiago) https://github.com/3drepo/3drepo.io/issues/2391
- [x] userBilling (santiago) ( paypal part is being ignored)

Charence:
This fixes #1972


backend/models/group.js
```diff
@@ -406,271 +430,90 @@ groupSchema.statics.listGroups = function (dbCol, queryParams, branch, revId, id
...
+	const group = await Group.findByUID(account, model, branch, revId, groupId);
+
+	const convertedObjects = await getObjectsArrayAsIfcGuids(data, false);
+	const toUpdate = {};
+	const toUnset = {};
+	const fieldsCanBeUpdated = ["description", "name", "rules", "objects", "color", "transformation", "issue_id", "risk_id"];
+
+	let typeCorrect = !(data.rules && data.objects);
+
+	typeCorrect && fieldsCanBeUpdated.forEach((key) => {
```
carmenfan, line 466: Also similar to create, we should just ditch this approach and turn it into a switch?

backend/models/group.js
```diff
@@ -406,271 +430,90 @@ groupSchema.statics.listGroups = function (dbCol, queryParams, branch, revId, id
...
-);
-
-Group.deleteGroups = function (dbCol, sessionId, ids) {
-	const groupsIds = [].concat(ids);
-
-	for (let i = 0; i < ids.length; i++) {
-		if ("[object String]" === Object.prototype.toString.call(ids[i])) {
-			ids[i] = utils.stringToUUID(ids[i]);
-		}
+		return Promise.reject(responseCodes.INVALID_ARGUMENTS);
```
carmenfan, line 510: jus throw

backend/routes/group.js
```diff
@@ -466,19 +466,14 @@ router.post("/revision/:rid/groups/", middlewares.issue.canCreate, createGroup);
...
-const getDbColOptions = function (req) {
-	return { account: req.params.account, model: req.params.model, logger: req[C.REQ_REPO].logger };
-};
-
 function listGroups(req, res, next) {
-
-	const dbCol = getDbColOptions(req);
 	const place = utils.APIInfo(req);
+	const { account, model } = req.params;
+	const rid = req.params.rid ? req.params.rid : null;
```
carmenfan, line 472: does it have to be null? if we can use undefined then we can just merge it in wit L471

backend/routes/group.js
```diff
@@ -504,25 +493,13 @@ function listGroups(req, res, next) {
 }
 
 function findGroup(req, res, next) {
-
-	const dbCol = getDbColOptions(req);
 	const place = utils.APIInfo(req);
+	const { account, model, uid } = req.params;
+	const rid = req.params.rid ? req.params.rid : null;
```
carmenfan, line 498: same question here

carmenfan (changes requested):

keifgwinn:
/azp run

I'm testing this works after cluster refresh

backend/models/group.js
```diff
@@ -246,132 +257,145 @@ groupSchema.methods.getObjectsArrayAsIfcGuids = function (data) {
...
+	if (typeCorrect) {
+		const groupsColl = await getCollection(account, model);
+		await groupsColl.insert(newGroup);
 
-				return objectId;
-			}));
+		newGroup._id = utils.uuidToString(newGroup._id);
+		newGroup.objects = await getObjectIds(account, model, branch, rid, newGroup, true, false);
+
+		if (!data.isIssueGroup && !data.isRiskGroup && sessionId) {
```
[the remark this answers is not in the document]
Charence: @carmenfan Frontend only seems to be sending isRiskGroup from risks saga nowadays. Can't find isIssueGroup or anything similar for views.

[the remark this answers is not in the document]
carmenfan: it shouldn't even be sending isRiskGroup. Front end should be sending the group contents directing with the issues

backend/models/group.js
```diff
@@ -246,132 +257,145 @@ groupSchema.methods.getObjectsArrayAsIfcGuids = function (data) {
...
+	const project = { parents: 1, _id: 0 };
 
-			if (!group) {
-				return Promise.reject(responseCodes.GROUP_NOT_FOUND);
+	return db.getCollection(account, model + ".scene").then(dbCol => {
+		return dbCol.find(query, project).toArray().then(results => {
+			if(results.length === 0) {
+				return [];
 			}
+			return History.getHistory({ account, model }, branch, revId).then(history => {
```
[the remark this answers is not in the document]
Charence: @carmenfan I agree. However, I've noticed that sometimes (e.g. issue, history) it returns MODEL_HISTORY_NOT_FOUND instead of INVALID_TAG_NAME. I guess we should pick one. I'll get rid of MODEL_HISTORY_NOT_FOUND?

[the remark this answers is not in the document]
carmenfan: sure, agreed

backend/models/group.js
```diff
@@ -53,133 +49,148 @@ const embeddedObjectFields = {
...
+				_revId,
+				ifcGuids
+			).then(sharedIdResults => {
+				for (let j = 0; j < sharedIdResults.length; j++) {
+					objectIdsSet.add(utils.uuidToString(sharedIdResults[j].shared_id));
+				}
+
+				if (objectIdsSet.size > 0) {
+					objectId.shared_ids = [];
+					objectIdsSet.forEach(id => {
```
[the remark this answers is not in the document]
Charence: @carmenfan They're added to the set as strings so the comparator works. If they're added as UUIDs then it'll only be checking they're not the same instance of an object.

backend/models/group.js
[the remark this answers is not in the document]
Charence: @carmenfan I'm removing the check for isIssueGroup and isRiskGroup since frontend shouldn't be sending them anymore.

backend/models/group.js
```diff
@@ -255,56 +242,51 @@ function getObjectsArrayAsIfcGuids(data) {
...
+					case "rules":
+						if (data.rules && !checkRulesValidity(data.rules)) {
+							typeCorrect = false;
+						}
+						newGroup[key] = cleanEmbeddedObject(key, data[key]);
+						break;
+					case "color":
+						newGroup[key] = data[key].map((c) => parseInt(c, 10));
+						break;
+					case "transformation":
```
carmenfan, line 278: Should these just be `default`?

Persumably we already prune out any values that are not recognised at L255

backend/models/group.js
```diff
@@ -335,46 +316,30 @@ Group.create = async function (account, model, branch = "master", rid = null, se
...
-	ChatEvent.groupsDeleted(sessionId, account, model, groupIds);
+	ChatEvent.groupsDeleted(sessionId, account, model, ids);
 };
 
 Group.deleteGroupsByViewId = async function (account, model, view_id) {
 	return await db.remove(account, getGroupCollectionName(model), { view_id });
 };
 
-Group.findByUID = async function (account, model, branch, revId, uid, showIfcGuids = false, noClean = true) {
+Group.findByUID = async function (account, model, branch, revId, uid, showIfcGuids = false, noClean = true, convertToIfcGuids = false) {
```
carmenfan, line 329: what's the difference between showIfcGuids and convertToIfcGuids?

backend/models/group.js
```diff
@@ -439,29 +398,45 @@ Group.update = async function (account, model, branch = "master", revId = null,
...
+						break;
+					case "objects":
+						toUpdate.objects = cleanEmbeddedObject(key, convertedObjects);
+						toUnset.rules = 1;
+						group.rules = undefined;
+						break;
+					case "color":
+						toUpdate[key] = data[key].map((c) => parseInt(c, 10));
+						break;
+					case "transformation":
```
carmenfan, line 431: don't turn transformations to int, considering rotations are cos/sines it's almost guaranteed to lose precision.... This is a bug I fixed create function a while back 😆

backend/models/group.js
```diff
@@ -439,29 +398,45 @@ Group.update = async function (account, model, branch = "master", revId = null,
...
+						toUpdate[key] = data[key].map((c) => parseInt(c, 10));
+						break;
+					case "issue_id":
```
carmenfan, line 434: `default`?

backend/models/ticket.js
```diff
@@ -450,13 +450,15 @@ class Ticket extends View {
...
 
 		// Assign rev_id for issue
-		const history = await History.getHistory({ account, model }, branch, newTicket.revId, { _id: 1 });
-
-		if (!history && (newTicket.revId || (newTicket.viewpoint || {}).highlighted_group_id)) {
-			throw responseCodes.INVALID_TAG_NAME;
-		} else if (history) {
-			newTicket.rev_id = history._id;
-		}
+		History.getHistory({ account, model }, branch, newTicket.revId, { _id: 1 }).then((history) => {
```
carmenfan, line 453: any reason why this' is written as a promise?

It's not awaited for either, which means L463 may happen before this is done.

carmenfan (changes requested):

backend/models/group.js
Charence: @carmenfan They're flags for different things. It was introduced when I merged `findByUID` with another function that was there before. `convertToIfcGuids` decides if it goes through the `getObjectsArrayAsIfcGuids` or `getObjectIds` function, and `showIfcGuids` is used within the functions that `getObjectIds` calls.

Easiest solution would probably be renaming it (and possibly renaming all the other functions because they all sound similar).

Another possibility is rewriting the functions that set `group.objects` so that it makes more sense.

carmenfan (approved):

PR #2371, opened 2020-12-29
ISSUE 2370 - Remove Mongoose from Job

### LINKED ISSUE #2370: Remove Mongoose - Job
### Description

Remove Mongoose backend dependency from Job

### Goals

- [x] Update tests
- [x] Remove Mongoose
- [x] Clean up

Charence:
This fixes #2370


backend/models/job.js
```diff
@@ -18,155 +18,205 @@
...
+		_id: jobData._id,
+		users: []
+	};
 
-	});
+	if (jobData.color) {
+		newJobEntry.color = jobData.color;
+	}
 
+	return jobsColl.insert(newJobEntry);
```
carmenfan, line 67: I think it's a good idea to base this branch on the groups on so we have the changes on dbHandler to just call insert instead of doing awaits on collection

backend/models/job.js
```diff
@@ -18,155 +18,205 @@
...
 
-schema.statics.removeUserFromAnyJob = function(teamspace, user) {
-	return Job.findByUser(teamspace, user).then(job => {
-		if(job) {
-			return job.removeUserFromJob(user);
-		}
-	});
+Job.addUserToJob = async function(teamspace, jobName, user) {
+	// Check if user is member of teamspace
+	const User = require("./user");
```
carmenfan, line 72: is there a cyclic dependency?

We should avoid local requires where possible. And going forward we should do it like the front end and only import the functions we need.

backend/models/job.js
```diff
@@ -18,155 +18,205 @@
...
+	}
+
+	return foundJob;
 };
 
-schema.statics.findByJob = function(teamspace, job) {
-	return this.findOne({account: teamspace}, {_id: job});
+Job.findUsersWithJobs = async function(teamspace, jobNames) {
+	const jobsColl = await getCollection(teamspace);
+	const foundJobs = await (await jobsColl.find({ _id: { $in: jobNames } })).toArray();
```
carmenfan, line 113: you don't need an await on the find. It returns a cursor. It only returns a promise when you call .toArray()

https://docs.mongodb.com/drivers/node/usage-examples/find

backend/models/job.js
```diff
@@ -18,155 +18,205 @@
...
-				return Promise.reject(responseCodes.JOB_NOT_FOUND);
-			}
-
-			return Job.removeUserFromAnyJob(teamspace, user).then(() => {
-				job.users.push(user);
-				return job.save();
-			});
+Job.getAllJobs = async function(teamspace) {
+	const jobsColl = await getCollection(teamspace);
+	const foundJobs = await (await jobsColl.find()).toArray();
```
carmenfan, line 125: same here

backend/models/job.js
```diff
@@ -18,155 +18,205 @@
...
+Job.removeUserFromAnyJob = async function(teamspace, user) {
+	const job = await Job.findByUser(teamspace, user);
+	let result;
 
-	return this.findByJob(teamspace, jobName).then(jobFound => {
-		if(!jobFound) {
-			return Promise.reject(responseCodes.JOB_NOT_FOUND);
-		}
+	if (job) {
+		result = Job.removeUserFromJob(teamspace, job._id, user);
```
carmenfan, line 161: This function returns a promise which we're not awaiting here?

The result also doesn't seem important considering it's an update?

backend/models/notification.js
```diff
@@ -18,7 +18,7 @@
 "use strict";
 const { hasWriteAccessToModelHelper, hasReadAccessToModelHelper } = require("../middlewares/checkPermissions");
 const modelSettings = require("../models/modelSetting");
-const job = require("./job");
+const Job = require("./job");
```
carmenfan, line 21: Can we change this to only import the functions needed in this file? this will reduce the chance of cyclic dependencies (I think)

backend/routes/job.js
```diff
@@ -240,35 +225,23 @@
...
 	function updateJob(req, res, next) {
-
-		if(!req.body._id) {
-			return Promise.reject(responseCodes.JOB_ID_INVALID);
-		}
-		Job.findByJob(req.params.account, req.body._id).then(job => {
-			return job.updateJob(req.body).then(() => {
-				responseCodes.respond(utils.APIInfo(req), req, res, next, responseCodes.OK, {});
-			});
+		// FIXME: should consider using req.params.jobId instead of body._id
```
carmenfan, line 233: Probably irrelevant if we're defining the schema to require the job schema to be in the body.
But maybe we should check that params.jobId === body._id, otherwise throw invalid arugments

carmenfan (changes requested):

backend/models/job.js
Charence: Yes, I was hoping to update it once the other branch was merged

backend/models/job.js
Charence: Yes, there's a cyclic dependency

carmenfan:
@Charence  i merged the groups PR to staging and i'm seeing conflicts here, can you take a look?

Charence:
> @Charence i merged the groups PR to staging and i'm seeing conflicts here, can you take a look?

Resolved

carmenfan (approved):

PR #2376, opened 2021-01-05
ISSUE 2372 - Remove Mongoose from Ref

### LINKED ISSUE #2372: Remove Mongoose - Ref
### Description

Remove Mongoose backend dependency from Ref

### Goals

- ~~Update tests~~
- [x] Remove Mongoose
- [x] Clean up

Charence:
This fixes #2372


backend/models/helper/model.js
```diff
@@ -503,7 +503,7 @@ function searchTree(account, model, branch, rev, searchString, username) {
 
 			});
 
-			return Ref.find({account, model}, {
+			return findRef(account, model, {
```
carmenfan, line 506: This is very high coupling. This should be calling getRefNodes instead. ids to pass in is `history.current`. From what I can tell it's doing the same thing

backend/models/ref.js
```diff
@@ -16,48 +16,35 @@
...
-		}
-		return [];
-	});
+function getRefCollectionName(model) {
+	return model + ".scene";
+}
+
+const Ref = {};
+
+Ref.findRef = async function(account, model, query, projection) {
```
carmenfan, line 28: As mentioned in the previous comment, this should be removed

backend/models/meta.js
```diff
@@ -241,7 +241,7 @@ class Meta {
 		models.add(model);
 
 		// Check submodels
-		const refs = await Ref.find({account, model}, {type: "ref"});
+		const refs = await findRef(account, model, {type: "ref"});
```
carmenfan, line 244: A lot of these queries are not checking current IDs, considering we're passing in branch/revId it's probably worth considering if we actually have a bug here.... 😆 If it's a federation with revisions where we removed/added some sub models, are we still searching through some sub models that are no longer a submodel?

carmenfan (changes requested):

backend/models/meta.js
```diff
@@ -142,7 +137,7 @@ class Meta {
 	}
 
 	async getMetadataFields(account, model) {
-		const subModelRefs = await Ref.getRefNodes(account, model);
+		const subModelRefs = await getRefNodes(account, model);
```
carmenfan, line 140: I think we want to get impose history.current here as well.

I don't think we're interested in getting metadataFields form some submodels that are no longer a submodel

carmenfan (changes requested):

Charence:
@carmenfan If #2377 for Scene can get merged in, getRefNodes can make use of the function in Scene which takes in branch/revision?

carmenfan:
> @carmenfan If #2377 for Scene can get merged in, getRefNodes can make use of the function in Scene which takes in branch/revision?

@Charence  merged now

carmenfan (approved):

PR #2389, opened 2021-01-08
ISSUE_2388 testing unity player

### LINKED ISSUE #2388: Test 3GB Unity Player
### Description
Might be possible to hack emcc to use more memory.

### Goals
- [ ] Does it work
- [ ] can we load more models
- [ ] what's the max mem limit?

keifgwinn:
This fixes #2388 - testing only to get a spin up


keifgwinn:
unfortunately, doesn't seem that unity wants to use more than 2gb even when it is available.

keifgwinn:
it doesn't seem to work properly

PR #2390, opened 2021-01-12
Issue 2381 - Remove mongoose from history

### LINKED ISSUE #2381: Remove Mongoose - History
### Description

Remove Mongoose backend dependency from Ref

### Goals

- [x] Update tests
- [x] Remove Mongoose
- [x] Clean up

sanmont3drepo:
This fixes #2381


backend/models/helper/model.js
```diff
@@ -608,7 +608,26 @@ function downloadLatest(account, model) {
 	});
 }
 
-function uploadFile(req) {
+const checkTag = async (account, model, tag) => {
```
carmenfan, line 611: Can we move this to history and rename this to isvalidTag() or something?

Since everything is doing is related to history it seems better fit that way

backend/models/history.js
```diff
@@ -118,88 +93,92 @@ historySchema.statics.findByBranch = function(dbColOptions, branch, projection,
...
 
-historySchema.statics.updateRevision = async function(dbColOptions, modelId, data) {
-	if(utils.hasField(data, "void") &&
-		Object.prototype.toString.call(data.void) === "[object Boolean]") {
-		const rev = await History.findByUID(dbColOptions, modelId);
-		if(!rev) {
-			return Promise.reject(responseCodes.INVALID_TAG_NAME);
-		}
+History.updateRevision = async function(account, model, revId, voidValue) {
+	if(Object.prototype.toString.call(voidValue) === "[object Boolean]") {
```
carmenfan, line 125: add a `isBool` function in `utils.js` and use that?

backend/models/history.js
```diff
@@ -118,88 +93,92 @@ historySchema.statics.findByBranch = function(dbColOptions, branch, projection,
...
+		throw responseCodes.INVALID_ARGUMENTS;
 	}
 };
 
-historySchema.statics.findByTag = function(dbColOptions, tag, projection) {
-	projection = projection || {};
-	return History.findOne(dbColOptions, { tag, incomplete: {"$exists": false }}, projection);
-};
+History.renameRevisionTag = async function(account, model, revId, tag) {
+	if(Object.prototype.toString.call(tag) === "[object String]") {
```
carmenfan, line 142: `utils.isString()`

backend/models/history.js
```diff
@@ -118,88 +93,92 @@ historySchema.statics.findByBranch = function(dbColOptions, branch, projection,
...
-		return Promise.reject(responseCodes.INVALID_ARGUMENTS);
+		throw responseCodes.INVALID_ARGUMENTS;
 	}
 };
 
-historySchema.statics.findByTag = function(dbColOptions, tag, projection) {
-	projection = projection || {};
-	return History.findOne(dbColOptions, { tag, incomplete: {"$exists": false }}, projection);
-};
+History.renameRevisionTag = async function(account, model, revId, tag) {
```
carmenfan, line 141: Is this used anywhere? I don't think we allow users to change revision tags.... 😆 😟

carmenfan (changes requested):

backend/models/history.js
sanmont3drepo: Is for doing this:
https://github.com/3drepo/3drepo.io/blob/99833d59dde119c13aa98784013932e68299ef52/backend/routes/history.js#L205-L223

Maybe the function name is wrong?

carmenfan: I think this may be a legacy API that we don't use (and don't want to let user use). Unless you can see a reference to use this API on the front end we should remove it.

sanmont3drepo: I dont see any references in the frontend, I'll remove it.

carmenfan (approved):

PR #2392, opened 2021-01-13
Issue 2375 - supporting changes for bouncer worker refactoring

### LINKED ISSUE #2375: Changes for unity queue
Supporting changes for https://github.com/3drepo/3drepobouncer/issues/366

carmenfan:
This fixes #2375

This is related to https://github.com/3drepo/3drepobouncer/issues/366

#### Description
- Add support for new error code
- Instead of using the full path of the shared directory, replace it with a place holder (so the user is no longer restricted to putting the shared space in the same place on both boxes)
- Support new status messages


backend/services/queue.js
```diff
@@ -163,9 +166,9 @@ class ImportQueue {
 	importFile(corID, filePath, orgFileName, databaseName, modelName, userName, copy, tag, desc, importAnimations = true) {
 		const jsonFilename = `${this.sharedSpacePath}/${corID}.json`;
```
Charence, line 167: What is the difference between `this.sharedSpacePath` and `sharedSpacePH` and why is `sharedSpacePH` not used here?

carmenfan: In short `this.sharedSpacePath` points to the directory where the sharedSpace (common folder between 3drepo.io and 3drepobouncer nodes) is placed.

`sharedSpacePH` is a placeholder tag for communications to the queue, where the worker nodes should replace it with whereever they've put the sharedSpace.

e.g. let's say the file is called `a.json`  and my sharedSpacePath is `/efs/queue`

`jsonFilename` would be `/efs/queue/a.json`

if we use `sharedSpacePH` (which is a placeholder label for sharedSpace path), it'd be `$SHARED_SPACE/a.json`.

we use `jsonFileName` to write the file in the sharedSpace so it needs to have the real path to sharedSpace. `sharedSpacePH` is what sent to the queue, so bouncer worker can replace it with whereever the bouncer node has the sharedSpace.

In the past we assume the bouncer worker and repo.io must have the sharedspace in the identical space (and with some loopy workaround if it's a linux box to windows), so sending a placeholder tag through the queue removes this constraint.

Charence (approved):

PR #2403, opened 2021-01-19
ISSUE #2326 - Fix updating of mitigation status prop by kanban board

### LINKED ISSUE #2326: Risk becomes inaccessible after updating its risk in Kanban Board
### Description
When I use Kanban Board, it's missing the Unmitigated Column. If there is a risk already, I can drag other across to update their status to Unmitigated. Unfortunately, they disappear and I can't find them. If I find them again, I can't open them, they give me a white screen in both Kanban Board & Model Viewer. As a result, I can no longer access that risk or change anything attributed to it.

### Steps to replicate
- Open the Kanban Board and change it to show: **Risks** Group by: **Mitigation Status**
- Update a Risk card to Unmitigated
- Open the recently updated Risk
- White screen appears

**Problem 1**
Unmitigated status doesn't exist as a column (whilst other empty columns are present)
image

**Problem 2**
Changing the Risk status to unmitigated makes it disappear
Update Unmitigated Risk

**Problem 3**
After updating the risk's status to Unmitigated I can no longer open it & change it back. _White screen_
Risk Open fail
Open Risk Fails

**Problem 4**
Mitigation Status contains a key-value "_mitigation_status_", should be "_unmitigated_"
image

#### Migration script:
`fixMitigationStatus.py <mongoURL> <mongoPort> <userName> <password>`

m1chalkubiak:
This fixes #2326
#### Description
we need the following:

- A db migration script to check if there is any risk with `mitigation_status` set to `mitigation_status`, then set it to empty string.
- if we can also add a check to on POST/PATCH request to see if the status is something we recognise, that would be useful too.

### DIFF
--- a/frontend/routes/board/board.component.tsx
+++ b/frontend/routes/board/board.component.tsx
@@ -266,13 +266,25 @@ export function Board(props: IProps) {
 		handleOpenDialog();
 	};
 
+	const getUpdatedProps = ({ filterProp, toLaneId }) => {
+		if (filterProp === ISSUE_FILTER_PROPS.assigned_roles.value) {
+			return [toLaneId];
+		}
+
+		if (filterProp === RISK_FILTER_PROPS.mitigation_status.value && filterProp === toLaneId ) {
+			return '';
+		}
+
+		return toLaneId;
+	};
+
 	const handleCardMove = (fromLaneId, toLaneId, cardId) => {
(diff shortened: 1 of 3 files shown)


Charence:
@m1chalkubiak @carmenfan DB migration script added

carmenfan (approved):

PR #2407, opened 2021-01-20
Issue 2391 - Removed mongoose from user, account permissions and template permissions

### LINKED ISSUE #2391: Remove Mongoose - User
### Description

Remove Mongoose backend dependency from User

### Goals

- [x] Update tests
- [x] Remove Mongoose
- [x] Clean up

sanmont3drepo:
This fixes #2391


backend/models/accountPermissions.js
```diff
@@ -0,0 +1,102 @@
...
+};
+
+const checkValidUpdate = async (teamspace, teamMember, permissions) => {
+	const User = require("./user");
+	const userToCheck = await User.findByUserName(teamMember);
+	if(!userToCheck) {
+		throw (responseCodes.USER_NOT_FOUND);
+	}
+
+	if(!User.isMemberOfTeamspace(userToCheck, teamspace.user)) {
```
carmenfan, line 37: I don't think we need to do L32 anymore, 37 should just take in a username?

It's only coded this way because it was a mongoose method i believe

The `isMemberOfTeamspace` method should do L32-37 within the function

backend/models/accountPermissions.js
```diff
@@ -0,0 +1,102 @@
...
+};
+
+AccountPermissions.remove = async function(teamspace, userToRemove) {
+	const updatedPermissions = this.get(teamspace).filter(perm => perm.user !== userToRemove);
+
+	if (updatedPermissions.length >=  this.get(teamspace).length) {
+		throw responseCodes.ACCOUNT_PERM_NOT_FOUND;
+	}
+
+	const projects = await Project.find({ account: teamspace.user },{ "permissions.user": userToRemove});
```
carmenfan, line 92: There should probably be a method in Project to remove a user from the project, instead of account Permissions depicting everything

tagging @Charence  since he's doing project so he knows on his branch!

backend/models/invitations.js
```diff
@@ -121,7 +121,7 @@ invitations.create = async (email, teamspace, job, username, permissions = {}) =
 		throw responseCodes.INVALID_MODEL_ID;
 	}
 
-	if (await teamspaceObject.hasReachedLicenceLimit()) {
+	if (await User.hasReachedLicenceLimit(teamspaceObject)) {
```
carmenfan, line 124: This should just take the teamspace name and leave User to fetch the teamspace object?

backend/models/notification.js
```diff
@@ -401,7 +401,7 @@ module.exports = {
 	insertUserReferencedNotification: async function (referrer, teamspace, modelId, type, _id, referee) {
 		try {
 			const user = await User.findByUserName(referee);
-			if (await user.isMemberOfTeamspace(teamspace)) {
+			if (User.isMemberOfTeamspace(user, teamspace)) {
```
carmenfan, line 404: similar sentiment here, notification.js shouldn't need to get the whole user data out just to do `isMemberOfTeamspace`.

`isMemberOfTeamspace` can fetch that data out with projections to save IO and also less coupling

carmenfan (changes requested):

backend/models/accountPermissions.js
sanmont3drepo: isMemberOfTeamspace takes the user object, it checks for the roles from that user.

backend/models/invitations.js
sanmont3drepo: the problem with that is that you are fetching several times the same user when you can do it just onces and pass it through to check whatever

backend/models/accountPermissions.js
carmenfan: that's what i mean, the only reason why this code needs to get the user object is to pass it in for isMemberOfTeamspace to use. It should instead take the teamspace as a string and does it itself

backend/models/invitations.js
carmenfan: apart from that's not really the use case here, once again we're fetching the user object just to call this function.

backend/models/notification.js
sanmont3drepo: As above isMemberOfTeamspace is taking a user in order to reuse the fetch instead of fetching the user multiple times. In this case user is only being use once. I could add isMemberOfTeamspaceByUsername.

carmenfan: Once again, here the code is only fetching User just to call that function.

Considering User consist of a lot of data (including an image for avatar), I think it's a lot smarter to let isMemberOfTeamspace fetch it with a projection.

I agree if it has to reuse it somewhere, but so far none of it is the case. So I think we don't need the original function, just replace it with the one that takes it as a string

backend/models/invitations.js
carmenfan: Can we do something similar to the `isMemberOfTeamspace ` with `hasReachedLicenceLimit`? Once again it's getting the object just to call this function here, and the only other place that his is used is within `user.js`

backend/models/user.js
```diff
@@ -1068,155 +928,98 @@ function _createAccounts(roles, userName) {
...
 
-			promises.push(foundProjects.map(project =>
-				project.updateAttrs({ permissions: project.permissions.filter(p => p.user !== username) })));
+	return await Role.revokeTeamSpaceRoleFromUser(userToRemove, teamspace.user);
+};
 
-			promises.push(removeUserFromAnyJob(this.user, username));
-			return Promise.all(promises);
-		}
+User.addTeamMember = async function(teamspace, userToAdd, job, permissions) {
```
carmenfan, line 988: Can we also make this pass in a teamspace name instead of object?

This is used in invitations.js and routes/, both of which are currently fetching the user object just to call this function

carmenfan (changes requested):

carmenfan (approved):

PR #2410, opened 2021-01-21
ISSUE #2383 - Removed mongoose from model/role

### LINKED ISSUE #2383: Remove Mongoose - Role
### Description

Remove Mongoose backend dependency from Role

### Goals

- [x] Update tests
- [x] Remove Mongoose
- [x] Clean up

sanmont3drepo:
This fixes #2383

### DIFF
--- a/backend/models/role.js
+++ b/backend/models/role.js
@@ -20,12 +20,16 @@
 const C = require("../constants");
 const DB = require("../handler/db");
 
+const findByRoleID = async function(id) {
+	return await DB.findOne("admin", "system.roles", { _id: id});
+};
+
 const Role = {};
 
 Role.createTeamSpaceRole = async function (account) {
 	const roleId = `${account}.${C.DEFAULT_MEMBER_ROLE}`;
 
-	const roleFound = await Role.findByRoleID(roleId);
+	const roleFound = await findByRoleID(roleId);
 
 	if(roleFound) {
 		return ;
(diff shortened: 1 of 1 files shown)


backend/models/role.js
```diff
@@ -16,95 +16,72 @@
...
 
-	schema.statics.grantTeamSpaceRoleToUser = function (username, account) {
+	const roleFound = await Role.findByRoleID(roleId);
 
-		const grantRoleCmd = {
-			grantRolesToUser: username,
-			roles: [{role: C.DEFAULT_MEMBER_ROLE, db: account}]
-		};
+	if(roleFound) {
+		return { role: roleFound.role, db: roleFound.db};
```
carmenfan, line 31: i don't think this function actually needs to return anything, it's never used (and makes sense, as it's just to make sure the db is setup.)

image

backend/models/role.js
```diff
@@ -16,95 +16,72 @@
...
+		"roles": []
 	};
 
-	schema.statics.findByRoleID = function(id) {
-		return this.findOne({ account: "admin"}, { _id: id});
+	await DB.runCommand(account, createRoleCmd);
+	return {role: roleName, db: account};
+};
+
+Role.dropTeamSpaceRole = async function (account) {
```
carmenfan, line 51: we can remove this, this is never used.

We'd never want to remove this role unless we're removing the whole teamspace, which is not possible to do.

backend/models/role.js
```diff
@@ -16,95 +16,72 @@
...
-		"Role",
-		schema,
-		() => {
-			return "system.roles";
-		}
-	);
+	return await DB.runCommand("admin", grantRoleCmd);
+};
+
+Role.findByRoleID = async function(id) {
```
carmenfan, line 74: This should be a private function. It's only used in this file. (and in context of 3D Repo it'll never be used anywhere else)

carmenfan (changes requested):

carmenfan (approved):

PR #2419, opened 2021-01-22
ISSUE 2345 - Remove Mongoose from Project

### LINKED ISSUE #2345: Remove Mongoose - Project
### Description

Remove Mongoose backend dependency from Project

### Goals

- [x] Update tests
- [x] Remove Mongoose
- [x] Clean up

Charence:
This fixes #2345


backend/models/project.js
```diff
@@ -18,266 +18,204 @@
...
 
-	schema.statics.findByIds = function(account, _ids) {
-		return Project.find({account}, { _id: { $in: _ids.map(utils.stringToUUID) } });
+		return projects;
 	};
 
-	schema.statics.populateUsers = function(userList, project) {
-		userList.forEach(user => {
-			const userFound = project.permissions.find(perm => perm.user === user);
+	Project.findProjectAndPopulateUsers = async function(teamspace, projectName) {
```
carmenfan, line 174: This needs better naming. The naming is really misleading.

After looking around, this is only used by the endpoint, which looks like it's to show the permissions page on the frontend. So it needs to know all the users within teamspace and whether they have any permissions.

PopulateUsers suggests we're doing a write. This is more like `getAllUsersProjectPermissions`  or something (i'm bad at naming, but hopefully you see the direction i'm trying to go towards)

backend/models/project.js
```diff
@@ -18,266 +18,204 @@
...
-		return this.permissions.find(perm => perm.user === username);
-	};
+	Project.findProjectsById = async function(teamspace, ids) {
+		const foundProjects = await Project.listProjects(teamspace, { _id: { $in: ids.map(utils.stringToUUID) } });
 
-	schema.statics.isProjectAdmin = async function(account, model, user) {
-		const projection = { "permissions": { "$elemMatch": { user: user } }};
-		const project = await Project.findOne({account}, {models: model}, projection);
-		const hasProjectPermissions = project && project.permissions.length > 0;
+		foundProjects.forEach((project) => {
```
carmenfan, line 189: there should be a clean function to do this? so it's inline with every other file?

backend/models/project.js
```diff
@@ -18,266 +18,204 @@
...
+		return projects;
 	};
 
-	schema.statics.populateUsers = function(userList, project) {
-		userList.forEach(user => {
-			const userFound = project.permissions.find(perm => perm.user === user);
+	Project.findProjectAndPopulateUsers = async function(teamspace, projectName) {
+		const User = require("./user");
+		const userList = await User.getAllUsersInTeamspace(teamspace);
+		const project = await Project.findOneProject(teamspace, {name: projectName});
```
carmenfan, line 177: can do these 2 queries simultaneously?

backend/models/project.js
```diff
@@ -18,266 +18,204 @@
...
+
+		return prepareProject(foundProject);
 	};
 
-	schema.statics.setUserAsProjectAdminById = async function(teamspace, project, user) {
-		const projectObj = await Project.findOne({ account: teamspace }, {_id: utils.stringToUUID(project)});
-		const projectPermission = { user, permissions: ["admin_project"]};
-		return await projectObj.updateAttrs({ permissions: projectObj.permissions.concat(projectPermission) });
+	Project.findProjectPermsByUser = async function(teamspace, model, username) {
+		const project = await Project.findOneProject(teamspace, {name: model});
```
carmenfan, line 203: add projection `{permissions:1 }` since we don't use anything else

backend/models/project.js
```diff
@@ -326,46 +264,134 @@
...
-		const project = await this.findOne({ account }, { name: projectName });
+		return hasProjectPermissions && project.permissions[0].permissions.includes(C.PERM_PROJECT_ADMIN);
+	};
+
+	Project.removeProjectModel = async function(teamspace, model) {
+		return db.update(teamspace, PROJECTS_COLLECTION_NAME, { models: model }, { "$pull" : { "models": model}}, {"multi": true});
+	};
+
+	Project.setUserAsProjectAdmin = async function(teamspace, project, user) {
+		const projectObj = await Project.findOneProject(teamspace, {name: project});
```
carmenfan, line 280: just do a update?

carmenfan: do we check if the user is in the teamspace? (do we have test for this?)

backend/models/project.js
```diff
@@ -326,46 +264,134 @@
...
+		const projectPermission = { user, permissions: ["admin_project"]};
+
+		return await Project.updateAttrs(teamspace, project, { permissions: projectObj.permissions.concat(projectPermission) });
+	};
+
+	Project.setUserAsProjectAdminById = async function(teamspace, project, user) {
```
carmenfan, line 286: I think we should refactor this and 279 to call a private function `setUsersAsProjectByQuery` which both calls with the query, as that's the only difference between them

backend/models/user.js
```diff
@@ -736,12 +736,14 @@ function _addProjects(account, username, models) {
 		query = { models: { $in: models } };
 	}
 
-	return Project.find({ account: account.account }, query).then(projects => {
+	return listProjects(account.account, query).then(projects => {
```
carmenfan, line 739: Looking at the way _addProjects is called, `models` is always undefined. so we're just getting all the projects all the time.

backend/models/user.js
```diff
@@ -736,12 +736,14 @@ function _addProjects(account, username, models) {
...
 
 		projects.forEach((project, i) => {
 
-			project = project.toObject();
+			project._id = utils.uuidToString(project._id);
 
+			// FIXME
```
carmenfan, line 745: remove? (there's more of these in the file)

backend/models/user.js
```diff
@@ -915,8 +917,7 @@ function _createAccounts(roles, userName) {
 				let account = null;
 				const query = { "permissions": { "$elemMatch": { user: userName } } };
 				const projection = { "permissions": { "$elemMatch": { user: userName } }, "models": 1, "name": 1 };
-				return Project.find({ account: user.user }, query, projection).then(projects => {
-
+				return listProjects(user.user, query, projection).then(projects => {
```
carmenfan, line 920: Move this logic into project. There should be a function in project to get all projects the user has access to. this is very high coupling 😆

backend/models/user.js
```diff
@@ -1135,8 +1147,7 @@ schema.methods.removeTeamMember = function (username, cascadeRemove) {
 
 	const teamspacePerm = this.customData.permissions.findByUser(username);
 	// check if they have any permissions assigned
-	return Project.find({ account: this.user }, { "permissions.user": username }).then(projects => {
-
+	return listProjects(this.user, { "permissions.user": username }).then(projects => {
```
carmenfan, line 1150: there should be a function in project to get projects the user has access to.

carmenfan (changes requested):
@Charence  can you address this here: https://github.com/3drepo/3drepo.io/pull/2407#discussion_r562561582

backend/models/project.js
```diff
@@ -18,266 +18,204 @@
...
+			}];
 		}
 
-		return check.then(() => {
-			Object.keys(data).forEach(key => {
-				if(whitelist.indexOf(key) !== -1) {
-					this[key] = data[key];
-				}
-			});
+		await checkInvalidName(name);
```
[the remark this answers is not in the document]
Charence: It seems like `checkDupName` still needs await, so I've kept that there

PR #2424, opened 2021-01-26
ISSUE #2374 rename Navis to 3D Repo plugin

### LINKED ISSUE #2374: rename .bim file format and related error messages
3drepo.io is stil referencing .bim as navis plugin, it should be renamed to 3drepo plugins on things like the revision list, and make sure we check error messages too:
image

carmenfan:
This fixes #2374
#### Description
.bim export should now show as "3D REPO PLUGIN" instead of "NAVIS".


Charence (changes requested):
@carmenfan I think we also want to change this? https://github.com/3drepo/3drepo.io/blob/8734d7c13d77343beb69f0342904e19a55a20773/backend/response_codes.js#L93

carmenfan:
@Charence  removed navisworks from that message

Charence (approved):

PR #2459, opened 2021-02-09
ISSUE 2453 - Password checks on backend

### LINKED ISSUE #2453: Pentest M5 - Enforce Password Policy on Server side
### Description

Currently UI is enforcing password policies which are not enforced via the APIs. This needs to be aligned to enforce the same policies on the back end

Charence:
This fixes #2453

#### Test cases
- SIgn up with password < 8 chars
- SIgn up with simple password


carmenfan (changes requested):
We also need to add this when/if user changes their password. Whether through the forgot password motion or changing it in their profile

carmenfan (changes requested):

PR #2464, opened 2021-02-10
Issue 2431 - ability to add a legend, change the name of a sequence

### LINKED ISSUE #2431: Add API support to rename a sequence, and sequence legend
### Description
This is a supporting issue for the front end work done in #2401.
Splitting it up as we are about to do more backend work in synchro (which will build on this), so it might be better to have this separate to avoid issue dependencies.

### Goals
- [x] Be able to change the name of a sequence
- [x] Be able to add a legend for the sequence
- [x] Be able to fetch a legend for the sequence
- [x] Be able to alter the legend for the sequence
- [x] Be able to delete the legend for the sequence
- [x] Be able to set a legend as default

### Tasks
- [x] Add route to allow renaming a sequence
- [x] ensure character limit is respected
- [x] ensure permission restriction is respected
- [x] Endpoint to add/remove/get legend
- [x] model settings endpoint update to take in a default legend
- [x] add tests
    - [x] getting a legend where a sequence does not have one should return an empty legend (no default legend)
    - [x] getting a legend where a sequence has one should succeed
    - [x] setting a default legend should succeed
    - [x] getting a legend where the sequence does not have one should return the default legend
    - [x] removing the default legend should succeed
    - [x] getting a legend with an invalid sequence id should fail
    - [x] permissions check on read/write
- [x] API docs

carmenfan:
This fixes #2431

#### Description
Backend support for:
1. ability to update the name of a sequence
2. Ability to add a legend
3. Set a default legend
4. fetch legend for a particular sequence
5. delete a legend for the sequence

#### Test cases
These endpoints should work as specified in travis test cases


backend/models/sequence.js
```diff
@@ -20,101 +20,173 @@ const db = require("../handler/db");
...
-		const dbCol = await db.getCollection(account, model + ".sequences");
+	return sequences;
+};
 
-		const sequences = await (dbCol.find({"rev_id": history._id}).toArray());
-		sequences.forEach((sequence) => {
-			sequence.teamspace = account;
-			sequence.model = model;
+Sequence.updateSequence = async (account, model, sequenceId, data) => {
+	if (data && data.name && utils.isString(data.name) && data.name !== "" && data.name.length < 30) {
```
sanmont3drepo, line 147: I think this is a little clearer
```suggestion
	if (!data || !data.name || !utils.isString(data.name) || data.name == ""  || data.name.length >= 30) {
		throw responseCodes.INVALID_ARGUMENTS;
	}
```

and get rid of the else part

sanmont3drepo (changes requested):

PR #2471, opened 2021-02-15
ISSUE 2451 - Account lockout

### LINKED ISSUE #2451: Pentest M3- Account Lockout policy
### Description

Currently there's is no password lockout policy and this means passwords can be brute forced. ALthough we can monitor this there is no failsafe if someone brute forces in the middle of the night

The following goals were set based off Microsoft guidelines [Available Here](https://docs.microsoft.com/en-gb/archive/blogs/secguide/configuring-account-lockout)

### Goals

- [x] After 10 bad attempts at password entry a users account is locked (this should be a variable in case we want to change it later)
- [x] The account is locked for 15 minutes (as above, may need to be changed later)
- [x] If no incorrect password is entered for a period of 15 mins the counter is reset (eg. if I enter it wrong 8 times, then do nothing for 15 mins, the counter is set back to 0)
- [x] Account locking triggers a flag which can be monitored on either Elastic or Intercom (or perhaps both)
- [x] Password reset only works when account is unlocked
- [x] If the number of attempts remaining is <=5 the Login screen will explain "You have x attempts remaining"
  - [x] backend
  - [x] frontend

Charence:
This fixes #2451


backend/models/user.js
```diff
@@ -48,10 +49,26 @@ const FileRef = require("./fileRef");
 const PermissionTemplates = require("./permissionTemplates");
 const { get } = require("lodash");
 
+const appendRemainingLoginsInfo = function (resCode, remaining) {
```
carmenfan, line 52: Instead of doing this? why don't we just have a function that spits out the object? And only call this function when you're certain you want to throw this error?

Currently it feels very hacky and trying to do too much. It should only call it when it needs to

```suggestion
return { 
    ...responseCodes.INCORRECT_USERNAME_OR_PASSWORD,
    message: `${responseCodes.INCORRECT_USERNAME_OR_PASSWORD.message}  (Remaining attempts: ${remaining})`
};
    
}
```

backend/models/user.js
```diff
@@ -109,14 +171,18 @@ User.authenticate =  async function (logger, username, password) {
...
-
 		if (!user)  {
 			user = await User.findByUserName(username);
 		}
 
+		if (hasReachedFailedLoginLimit(user)) {
+			throw ({ resCode: responseCodes.TOO_MANY_LOGIN_ATTEMPTS });
+		}
+
+		authDB = await DB.getAuthDB();
```
carmenfan, line 182: We should move this below L188. There's no point authenticating the user if it's inactive.

backend/models/user.js
```diff
@@ -109,14 +171,18 @@ User.authenticate =  async function (logger, username, password) {
...
 
-		authDB = await DB.getAuthDB();
-		await authDB.authenticate(username, password);
-		authDB.close();
-
 		if (!user)  {
 			user = await User.findByUserName(username);
 		}
 
+		if (hasReachedFailedLoginLimit(user)) {
```
carmenfan, line 178: I think this is better described as `isAccountLocked`

Coz you can have reached the failed login count, but the lockout time has expired.

backend/models/user.js
```diff
@@ -74,12 +91,57 @@ const findOne = async function (query, projection) {
...
+		const elapsedTime = user.customData.lastFailedLoginAt ?
+			currentTime - user.customData.lastFailedLoginAt : undefined;
+
+		const failedLoginCount = user.customData.failedLoginCount && elapsedTime && elapsedTime < config.lockoutDuration ?
+			user.customData.failedLoginCount + 1 : 1;
+
+		const loginLocked = failedLoginCount >= config.maxUnsuccessfulLoginAttempts ? true : undefined;
+
+		await User.update(username, {
+			"customData.lastFailedLoginAt": currentTime,
```
carmenfan, line 108: Can we also group these 3 things into a single embedded object within customData?

backend/models/user.js
```diff
@@ -74,12 +91,57 @@ const findOne = async function (query, projection) {
...
+			"customData.failedLoginCount": failedLoginCount,
+			"customData.loginLocked": loginLocked
```
carmenfan, line 110: I don't think we should be storing this? we should always deduce it at runtime?

Also doesn't look like we use it

backend/models/user.js
```diff
@@ -74,12 +91,57 @@ const findOne = async function (query, projection) {
...
+		});
+
+		return Math.max(config.maxUnsuccessfulLoginAttempts - failedLoginCount, 0);
+	} catch(err) {
+		// suppress update failure
```
carmenfan, line 115: What kind of failures are we suppressing here?

backend/models/user.js
```diff
@@ -74,12 +91,57 @@ const findOne = async function (query, projection) {
...
+		const user = await User.findByUserName(username);
+		const elapsedTime = user.customData.lastFailedLoginAt ?
+			currentTime - user.customData.lastFailedLoginAt : undefined;
+
+		const failedLoginCount = user.customData.failedLoginCount && elapsedTime && elapsedTime < config.lockoutDuration ?
+			user.customData.failedLoginCount + 1 : 1;
+
+		const loginLocked = failedLoginCount >= config.maxUnsuccessfulLoginAttempts ? true : undefined;
+
+		await User.update(username, {
```
carmenfan, line 107: I think we should stop having a generate update function and just go ahead and call $set here.

backend/models/user.js
```diff
@@ -74,12 +91,57 @@ const findOne = async function (query, projection) {
 	return await DB.findOne("admin", COLL_NAME, query, projection);
 };
 
+const handleAuthenticateFail = async function (username) {
+	const currentTime = Date.now();
+
+	try {
+		const user = await User.findByUserName(username);
```
carmenfan, line 98: I don't think we need to do any of that here. All we need to do is increment the failedLoginCount and update the last time we tried it? I don't think you need to do a read at all prior to the update.

You can use `$inc` to increment something https://docs.mongodb.com/manual/reference/operator/update/inc/

And do a `findAndUpdate` to update the document and get the counter back out all in one query

backend/models/user.js
```diff
@@ -74,12 +91,57 @@ const findOne = async function (query, projection) {
...
+			"customData.loginLocked": loginLocked
+		});
+
+		return Math.max(config.maxUnsuccessfulLoginAttempts - failedLoginCount, 0);
+	} catch(err) {
+		// suppress update failure
+	}
+};
+
 const COLL_NAME = "system.users";
```
carmenfan, line 119: This should be at the top of the file

backend/models/user.js
```diff
@@ -74,12 +91,57 @@ const findOne = async function (query, projection) {
...
 
 const User = {};
 
 User.update = async function (username, data) {
-	return DB.update("admin", COLL_NAME, {user: username}, {$set: data});
+	const toUpdate = {};
```
carmenfan, line 124: we should really be ditching this function instead of expanding on it, 

In this case it doesn't even do type checking, so it's quite useless

carmenfan (changes requested):

backend/models/user.js
Charence: We don't use it. It is always deduced at runtime. But this was put in to satisfy: 
> Account locking triggers a flag which can be monitored on either Elastic or Intercom (or perhaps both)

carmenfan: let's remove it, considering the states can't be guaranteed it's pointless to store it. It's not like Elastic/Intercom has the ability to probe the database anyway, what we actually want to do is to send a signal to them to notify them this account has been locked out when it happens.

@mosment3DRepo if you want this to be sent to intercom you need to tell us how you want it done, do we update the user data?

@keifgwinn can you advise charence on what he needs to send?

backend/models/user.js
Charence: @carmenfan If the user doesn't exist it goes through here as well. I guess I should change that and maybe pass in the already retrieve `user` as well.

backend/models/user.js
mosment3DRepo: @carmenfan can we post to the events service [Intercom API docs here](https://developers.intercom.com/intercom-api-reference/reference#submitting-events)

Endpoint: https://api.intercom.io/events

JSON would be
```
{
  "event_name": "Password Lockout",
  "created_at": 1613668686,
  "user_id": "example_user"
}
```

backend/models/user.js
Charence: @carmenfan This actually changes how this function works. At the moment, login/user tests expect an incorrect password to be returned before an error is thrown about an inactive account.

From a user's point of view, there's a difference, but I'm not sure if it matters - if they get their password wrong and their account is also inactive, it'll tell them their account is inactive instead of letting them know they got their password wrong. We can change it to this, but we'll also have to change the tests to match.

backend/models/user.js
Charence: @mosment3DRepo Done! Event name is `password-lockout`.

backend/constants.js
```diff
@@ -476,6 +476,15 @@
 	define("USERNAME_REGEXP",  /^[a-zA-Z][\w]{1,19}$/);
 	define("EMAIL_REGEXP", /^([a-zA-Z0-9_\-.]+)@([a-zA-Z0-9_\-.]+)\.([a-zA-Z]{2,5})$/);
 
+	// -----------------------------------------------------------------------------
+	//
+	// Password requirements
+	//
+	// -----------------------------------------------------------------------------
+
+	define("MIN_PASSWORD_LENGTH", 8);
```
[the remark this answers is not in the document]
Charence: Yes, I branched this off the password one to minimise conflicts as I thought it would be touching the same files. Could we merge the password one in first when it's ready?

backend/models/user.js
carmenfan: yea I don't think it matters, if it's inactive it's inactive. It'll probably help support as well

backend/config.js
```diff
@@ -293,4 +293,12 @@ config.liveChatLicense = config.liveChatLicense || 0;
 
 config.cachePolicy = config.cachePolicy || {maxAge: 60 * 60 * 24 * 30};
 
+// login lockout policy
+if (!config.loginPolicy) {
```
carmenfan, line 297:
```suggestion
 config.loginPolicy = config.loginPolicy || {};
```

backend/models/user.js
```diff
@@ -121,14 +157,18 @@ User.authenticate =  async function (logger, username, password) {
...
-
 		if (!user)  {
 			user = await User.findByUserName(username);
 		}
 
+		if (isAccountLocked(user)) {
+			throw ({ resCode: responseCodes.TOO_MANY_LOGIN_ATTEMPTS });
+		}
+
+		authDB = await db.getAuthDB();
```
carmenfan, line 168: can we refactor this to go go in to the dbhandler? 

so this file should only need to call `db.authenticate()`; and shouldn't need to deal with the db connections

backend/models/user.js
```diff
@@ -139,17 +179,28 @@ User.authenticate =  async function (logger, username, password) {
...
 		user.customData.lastLoginAt = new Date();
 
-		await User.update(username, {"customData.lastLoginAt": user.customData.lastLoginAt});
+		await db.update("admin", COLL_NAME, {user: username}, {$set: {"customData.lastLoginAt": user.customData.lastLoginAt}});
 
 		logger.logInfo("User has logged in", {username});
 
 		return user;
 	} catch(err) {
+		let resCodeError = err.resCode ? err : { resCode: utils.mongoErrorToResCode(err) };
```
carmenfan, line 188: can we move the try/catch?

We're catching exceptions that we are throwing here unnecessarily.

Effectively the catch should only be around the authentication part. Then you wouldn't have to probe the resCode value just to check if the user failed to authenticate itself, or or check if user exists etc.

i.e. this whole blurb can simplify if (err.resCode) { //deal with account lockout code } otherwise, throw mongo error

carmenfan (changes requested):

backend/models/user.js
```diff
@@ -146,62 +146,56 @@ User.authenticate =  async function (logger, username, password) {
...
-		logger.logInfo("User has logged in", {username});
+	if (!user)  {
+		user = await User.findByUserName(username);
+	}
 
-		return user;
-	} catch(err) {
-		let resCodeError = err.resCode ? err : { resCode: utils.mongoErrorToResCode(err) };
+	if (isAccountLocked(user)) {
+		throw ({ resCode: responseCodes.TOO_MANY_LOGIN_ATTEMPTS });
```
carmenfan, line 164: this should just be throw responseCodes.TOO_MANY_LOGIN_ATTEMPTS ?

backend/models/user.js
```diff
@@ -146,62 +146,56 @@ User.authenticate =  async function (logger, username, password) {
...
+		const resCode = utils.mongoErrorToResCode(err);
 
 		if (user) {
 			const remainingLoginAttempts = await handleAuthenticateFail(user, username);
 
-			if (resCodeError.resCode.value === responseCodes.INCORRECT_USERNAME_OR_PASSWORD.value &&
+			if (resCode.value === responseCodes.INCORRECT_USERNAME_OR_PASSWORD.value &&
 				remainingLoginAttempts <= config.loginPolicy.remainingLoginAttemptsPromptThreshold) {
-				resCodeError = { resCode: appendRemainingLoginsInfo(resCodeError.resCode, remainingLoginAttempts) };
+				throw { resCode: appendRemainingLoginsInfo(resCode, remainingLoginAttempts) };
```
carmenfan, line 177: this as well, should just be throw appendRemainingLoginsInfo(resCode, remainingLoginAttempts)

backend/models/user.js
```diff
@@ -146,62 +146,56 @@ User.authenticate =  async function (logger, username, password) {
...
 			}
 		}
 
-		throw resCodeError;
+		throw { resCode };
 	}
+
+	if (user.customData && user.customData.inactive) {
+		throw ({ resCode: responseCodes.USER_NOT_VERIFIED });
```
carmenfan, line 185: and here

backend/models/user.js
```diff
@@ -146,62 +146,56 @@ User.authenticate =  async function (logger, username, password) {
...
+	}
+
+	if (!user.customData) {
+		user.customData = {};
+	}
+
+	user.customData.lastLoginAt = new Date();
```
carmenfan, line 192: this is redundant as we no longer doing it via mongoose?

backend/models/user.js
```diff
@@ -146,62 +146,56 @@ User.authenticate =  async function (logger, username, password) {
...
+
+	await db.update("admin", COLL_NAME, {user: username}, {$set: {"customData.lastLoginAt": user.customData.lastLoginAt}});
+
+	logger.logInfo("User has logged in", {username});
+
+	return user;
```
carmenfan, line 198: can you check if this return value is used anywhere?

i'm sure we don't return the user data back to the user like that at login

Charence: @carmenfan `login` uses it when checking if the user has read the latest T&Cs in `createSession`. Instead of returning `user`, it would be sufficient to return `lastLoginAt` instead, as long as it makes sense why it is just returning a timestamp.

backend/models/user.js
Charence: This is used to check if the user has read the latest copy of the T&Cs

carmenfan (changes requested):
as discussed:

- let's store lastFailedLoginAt as a ISO Date instead of timestamp so it's consistent with other fields in the user bson
- reset login failed attempts when the user successfully logged in

carmenfan (approved):

carmenfan (approved):

PR #2473, opened 2021-02-16
ISSUE #2401 - Synchro v3 - Front end enhancements

### LINKED ISSUE #2401: Synchro v3 - Front end enhancements
### Description
A series of (mostly) front end changes on the Synchro feature.

### Goals
Click on the links to see the details:
- [x] [Play sequence in frame intervals](https://github.com/3drepo/3drepo.io/issues/2401#issuecomment-762338273)
- [x] [Move the loading icon to somewhere else](https://github.com/3drepo/3drepo.io/issues/2401#issuecomment-762340709)
- [x] [Ability to rename a sequence](https://github.com/3drepo/3drepo.io/issues/2401#issuecomment-762379088)
- [x] [Modify Date Picker defaults](https://github.com/3drepo/3drepo.io/issues/2401#issuecomment-762389501)
- [x] [Cache state data in indexeddb](https://github.com/3drepo/3drepo.io/issues/2401#issuecomment-762387035)
- [x] [Add a legend](https://github.com/3drepo/3drepo.io/issues/2401#issuecomment-762748315)

### Tasks
- [x] Use new @material-ui/pickers instead of deprecated material-ui-pickers
### Related Resources
Coda doc: https://coda.io/d/3D-Repo-Features_d5c0Qp25r6f/4D-v3_suIVb#Synchro-4D-v3_tud5j

m1chalkubiak:
This fixes #2401


m1chalkubiak:
Hi @carmenfan 
Currently, I still have a strange problem with the library responsible for the resizable and draggable functionality of the new panel. The functionality itself works fine, but in conjunction with the additional panel menu and the color picker, it creates a conflict that it is trying to resolve for now.

There are some situations where I am not sure how the Legend panel should behave correctly, I would be grateful for specifying these cases:

Because the entire legend is saved as an object in the
```
{
 name: color,
}
```
This makes it impossible to add two elements with the same name, how should the panel behave when trying to enter a name that already exists in the list?
I think blocking saving and showing error message would be consistent with the rest of the application, but maybe you have another idea.

Since the new legends panel is draggable, it is possible to move it mostly outside the monitor screen, should we somehow block those kinds of situations?

mosment3DRepo:
> Hi @carmenfan
> Currently, I still have a strange problem with the library responsible for the resizable and draggable functionality of the new panel. The functionality itself works fine, but in conjunction with the additional panel menu and the color picker, it creates a conflict that it is trying to resolve for now.
> 
> There are some situations where I am not sure how the Legend panel should behave correctly, I would be grateful for specifying these cases:
> 
> Because the entire legend is saved as an object in the
> 
> ```
> {
>  name: color,
> }
> ```
> 
> This makes it impossible to add two elements with the same name, how should the panel behave when trying to enter a name that already exists in the list?
> I think blocking saving and showing error message would be consistent with the rest of the application, but maybe you have another idea.
> 
> Since the new legends panel is draggable, it is possible to move it mostly outside the monitor screen, should we somehow block those kinds of situations?

@m1chalkubiak I agree, blocking the save icon and having an error message in red, similar to the model revision name length error:

image

it should say "This name is already in use"

In terms of dragging, can we prevent any of the panel from going off the screen, so the moment the panel hits the edge it stops moving. Alternatively if that's not possible, if the panel is dragged off the screen, pop it back to the nearest edge when the user lets go.

m1chalkubiak:
> @m1chalkubiak I agree, blocking the save icon and having an error message in red, similar to the model revision name length error:
> 
> image
> 

@carmenfan Of course, I will add it right away, should I also add any kind of name length limit?

> it should say "This name is already in use"
> 
> In terms of dragging, can we prevent any of the panel from going off the screen, so the moment the panel hits the edge it stops moving. Alternatively if that's not possible, if the panel is dragged off the screen, pop it back to the nearest edge when the user lets go.

Sure, I will investigate it and try to make it work in the best possible way. Hope that blocking from dragging off the screen will be possible.

keifgwinn:
Whilst you’re here could you have the automatic name suggestion cut to 19 not 20 so that it doesn’t go into an error state immediately if you use a file with a long name 

--
Keif

carmenfan:
> @carmenfan Of course, I will add it right away, should I also add any kind of name length limit?

 @m1chalkubiak  @mosment3DRepo says lets let's go with 20!

carmenfan:
> Whilst you’re here could you have the automatic name suggestion cut to 19 not 20 so that it doesn’t go into an error state immediately if you use a file with a long name
> […](#)
> -- Keif
> On 16 Feb 2021, at 13:32, Michał Kubiak ***@***.***> wrote: ﻿ @m1chalkubiak I agree, blocking the save icon and having an error message in red, similar to the model revision name length error: @carmenfan Of course, I will add it right away, should I also add any kind of name length limit? it should say "This name is already in use" In terms of dragging, can we prevent any of the panel from going off the screen, so the moment the panel hits the edge it stops moving. Alternatively if that's not possible, if the panel is dragged off the screen, pop it back to the nearest edge when the user lets go. Sure, I will investigate it and try to make it work in the best possible way. Hope that blocking from dragging off the screen will be possible. — You are receiving this because you are subscribed to this thread. Reply to this email directly, view it on GitHub, or unsubscribe.

Not really in scope for this issue. Make a separate issue?

m1chalkubiak:
Hi @carmenfan,
Apologies for the delays with this task, several times when I was already convinced that I would finish it in a moment, I was finding a new edge case that I had not thought of before, however, I think to know all should work fine.
Let me know If you see anything that should be changed.

Are we going to merge https://github.com/3drepo/3drepo.io/pull/2480 into staging, to merge it to this PR? Since currently, I think the environment won't build properly.

carmenfan:
@m1chalkubiak  

> Are we going to merge #2480 into staging, to merge it to this PR? Since currently, I think the environment won't build properly.

Merged just now! It was waiting on travis and I forgot about it 😆

m1chalkubiak:
@carmenfan I merged staging and found something new. Can You please check whether setting default Legend is working properly? I'm sending an object with defaultLegend and sequence ID as a value, get 200, but after fetching legend again I'm getting empty object. Am I doing something wrong here?

carmenfan:
> @carmenfan I merged staging and found something new. Can You please check whether setting default Legend is working properly? I'm sending an object with defaultLegend and sequence ID as a value, get 200, but after fetching legend again I'm getting empty object. Am I doing something wrong here?

@m1chalkubiak 
If it worked before and no longer works now then I've probably broke something whilst merging the code, i'll take a look.

carmenfan:
@m1chalkubiak  Can you explain exactly what you doing? I just had a quick play and the backend logic seems to be intact.

I'm having trouble however for the front end, sometimes I can see that the legend is blank, but it's not doing a HTTP request to fetch it. This to happens if it's not the first time i'm entering the viewer (i.e. the legend loads fine if I do a refresh)

I'm also seems to be stuck in loading frame forever when I enter a sequence: 
image

And the legend panel should toggle off if I exit the sequence, as the legend info is specific to that particular sequence, i.e. this shouldn't happen: 
image

Can we also reduce the size of the padding here:
image

m1chalkubiak:
> @m1chalkubiak Can you explain exactly what you doing? I just had a quick play and the backend logic seems to be intact.

Of, course @carmenfan  I will try to describe it as detail as possible.

After creating some sample legend with items, fetching of them (GET on)

```
/api/${teamspace}/${modelId}/revision/${revision}/sequences/${sequenceId}/legend
```
work properly, and as a response, I get (for example):

```
{
  Home: "#ff7f00"
  Item 3: "#a6cee3"
  Test: "#ffff99"
}
```

However, if I try to set this as a default (PUT)

```
/api/${teamspace}/${modelId}/settings
```
with body as below:

```
{
  "defaultLegend": ${sequenceId}
}
```

And then try to reset legend to default:

Call DELETE on 
```
/api/${teamspace}/${modelId}/revision/${revision}/sequences/${sequenceId}/legend
```
as described here: https://github.com/3drepo/3drepo.io/issues/2401#issuecomment-777345327

Then GET on the same endpoint to get the previously saved default legend, but got an empty object as response.

So either I don't understand how the default legend should work, or I missing something.

> I'm having trouble however for the front end, sometimes I can see that the legend is blank, but it's not doing a HTTP request to fetch it. This seems to happens if it's not the first time i'm entering the viewer (i.e. the legend loads fine if I do a refresh)

> And the legend panel should toggle off if I exit the sequence, as the legend info is specific to that particular sequence, i.e. this shouldn't happen:
> image

Both above fixed, thanks for catching this.

> I'm also seems to be stuck in loading frame forever when I enter a sequence:
> image
> 

On my "friday" account with the only model having sequence (from my knowledge), I did not run into such a problem.
I don't think I am changing anything in the logic behind the "loading frame" status. Maybe this is problem with specific model, are You able to give me access to the one that You are referring to?

> Can we also reduce the size of the padding here:
> image

I would have to reduce the clickable area of the arrows in the player, because currently there is no available space between them, does this solution suit you?
Zrzut ekranu 2021-02-26 o 06 56 54

carmenfan:
> So either I don't understand how the default legend should work, or I missing something.

@m1chalkubiak You're deleting the content of the legend that is being used as the default legend. When we set the default legend we only peg it against a specific sequence's legend instead of storing a copy of it. This should work if you deleted a legend off a sequence that isn't being used for default. I've tested that and it seems to work ok.

This does bring up a UI improvement - If the legend is the default legend, can we disable `set as default` and `reset to default`?

> > I'm also seems to be stuck in loading frame forever when I enter a sequence:
> > image
> 
> On my "friday" account with the only model having sequence (from my knowledge), I did not run into such a problem.
> I don't think I am changing anything in the logic behind the "loading frame" status. Maybe this is problem with specific model, are You able to give me access to the one that You are referring to?
> 

I tried to upload the same model to your database and it's not doing it... 😆 We can leave it for now. I'll let you know if I have a reliable way of reproducing it!

> > Can we also reduce the size of the padding here:
> > image
> 
> I would have to reduce the clickable area of the arrows in the player, because currently there is no available space between them, does this solution suit you?
> Zrzut ekranu 2021-02-26 o 06 56 54

Sure , looks better I think!

A few more things:
- The max length of the legend only really occupy half of the tab, can we reduce the default width so there's less padding here? 
image
- Can we add more padding to the left so it's more balanced? 
image
- When the user press the + button, can we make it so it doesn't auto save the new entry? instead, just have it in edit mode and only save it if the user click save (A bit like the views):
fds

mosment3DRepo:
@carmenfan this seems to suffer the same bug as before where the final result flashes in and out during play through, Is it a bug or did this miss some of the previous bug fixes?

4D Flash

carmenfan:
> @carmenfan this seems to suffer the same bug as before where the final result flashes in and out during play through, Is it a bug or did this miss some of the previous bug fixes?
> 
> 4D Flash

@mosment3DRepo this has latest of staging in so it has the fix.

@m1chalkubiak can you sanity check that when the sequence card is waiting on the next frame, it's keeping the state of the current frame instead of resetting all the states?

frontend/modules/model/model.sagas.ts
```diff
@@ -58,11 +58,13 @@ export function* updateSettings({ modelData: { teamspace, project, modelId }, se
 
 		yield API.editModelSettings(teamspace, modelId, modifiedSettings);
 
-		yield put(
-			TeamspacesActions.updateModelSuccess(
-				teamspace, modelId, { project, model: modelId, name: modifiedSettings.name, code: modifiedSettings.code }
-			)
-		);
+		if (modifiedSettings.name && modifiedSettings.code) {
```
carmenfan, line 61: does this make sense? not all model settings have a code.  Did you run into some errors?

m1chalkubiak: Good catch, I had an issue during setting default legend but I checked how this was implemented in case of default view and I realized that I overcomplicate it, nothing needs to be changed here, I'm pushing update right now.

m1chalkubiak:
> A few more things:
> 
> * The max length of the legend only really occupy half of the tab, can we reduce the default width so there's less padding here?
> * Can we add more padding to the left so it's more balanced?
> * When the user press the + button, can we make it so it doesn't auto save the new entry? instead, just have it in edit mode and only save it if the user click save (A bit like the views):

@carmenfan All above are addressed.

> > @carmenfan this seems to suffer the same bug as before where the final result flashes in and out during play through, Is it a bug or did this miss some of the previous bug fixes?
> > 4D Flash
> 
> @mosment3DRepo this has latest of staging in so it has the fix.
> 
> @m1chalkubiak can you sanity check that when the sequence card is waiting on the next frame, it's keeping the state of the current frame instead of resetting all the states?

I've unfortunately a problem with finding the cause of the issue reported by Matthew.
Honestly, again I have also a problem with reproducing this issue, on the model available to me it is very difficult to notice whether this behavior is taking place.
But the main change affecting the part responsible for loading the frame was the addition of indexedDB cache, so for now it seems to me that this could be some source of the issue.

@mosment3DRepo Are u using Model Caching option in the case above?
Additionally, is this issue is appearing only when we are using frames as steps, or it also occurs in other types of steps?

mosment3DRepo:
@m1chalkubiak just checked this out:

- I didn't have model caching on
- It was occurring for both frames and days
- When I loaded it in incognito mode it ran fine
- I cleared my browser cache and loaded it back in the regular browser
- The issue stopped happening

Although I didn't have model caching on I may have had it on in the past, so perhaps my cache had something in that was causing this

m1chalkubiak:
> @m1chalkubiak just checked this out:
> 
> * I didn't have model caching on
> * It was occurring for both frames and days
> * When I loaded it in incognito mode it ran fine
> * I cleared my browser cache and loaded it back in the regular browser
> * The issue stopped happening
> 
> Although I didn't have model caching on I may have had it on in the past, so perhaps my cache had something in that was causing this

@mosment3DRepo The only thing that comes to mind is to try to reproduce this bug with model caching on. If the bug does not reproduce then we don't know the correct steps to reproduce it.

m1chalkubiak:
> @m1chalkubiak found a bug with the legend:
> 
> fds
> 
> If I try to alter the colour prior to saving the entry, it disappears.
> 
> Also the random colour it's generating when you add a new entry, can we set alpha to 1 by default? As you can see on the gif here, sometimes it looks like opacity can be 0

@carmenfan Nice catch! Both cases fixed.

PR #2477, opened 2021-02-17
ISSUE #2449 - Turn off Cross-Site Requests

### LINKED ISSUE #2449: Pentest M2 - Turn off Cross-Site Requests
### Description

Cross site requests expose 3D Repo to malicious attacks and is often rendered incompatible by  changes to Chrome (edge/Firefox etc.)

In order to solve both problems we should turn off Cross-Site Requests and restrict all usage of the embedded viewer and APIs to API Key only

If possible this change should be reversible in case we find a client using cross site for a critical tool, in which case we will turn it back on as we assist the client in transitioning to an API key

We have clients attaching via iframe, we want to make sure they are unaffected from this change

Charence:
This fixes #2449

### Tests
- Requests using API key should work
- .io embedded into iframe should work
- Requests from server B to server A should not work e.g. https://issue-2485.dev.3drepo.io/


carmenfan:
@Charence  you can mark this as draft to indicate it's not ready :
image

Charence:
Note: "Strict" doesn't work with iframe; user is unable to proceed from login page.

Charence:
Note: "Lax" also has same behaviour as "Strict" for iframe.

backend/middlewares/sessionManager.js
```diff
@@ -27,6 +27,12 @@ module.exports = async (req, res, next) => {
 		return;
 	}
 
+	if ((req.headers.referer && !req.headers.referer.match(`^${config.public_protocol}://${config.cookie_domain}`)) ||
+		(!req.headers.referer && !config.cookie.skipMissingReferer)) {
+		req[C.REQ_REPO].logger.logError(`express-session internal error: referer (${req.headers.referer}) and cookie (${config.cookie_domain}) domain mismatch`);
```
carmenfan, line 32: This isn't an error, i don't think we should be logging this way.

backend/middlewares/sessionManager.js
```diff
@@ -27,6 +27,12 @@ module.exports = async (req, res, next) => {
 		return;
 	}
 
+	if ((req.headers.referer && !req.headers.referer.match(`^${config.public_protocol}://${config.cookie_domain}`)) ||
+		(!req.headers.referer && !config.cookie.skipMissingReferer)) {
```
carmenfan, line 31: We can remove skipMissingReferer.

We need to always accept missing referrer or plugins won't work

backend/middlewares/sessionManager.js
```diff
@@ -27,6 +27,12 @@ module.exports = async (req, res, next) => {
...
+		req[C.REQ_REPO].logger.logError(`express-session internal error: referer (${req.headers.referer}) and cookie (${config.cookie_domain}) domain mismatch`);
+		return;
```
carmenfan, line 33: If we're not going next(), we need to respond to the request here. at the moment if it enters this condition the request never returns.

carmenfan (changes requested):
Can we also add some tests? I think you can set headers on the agent

https://stackoverflow.com/questions/23725643/with-supertest-can-i-create-an-alternative-request-with-some-headers-set-by-def

Charence:
@carmenfan Suggested changes for storing referrer applied

backend/middlewares/sessionManager.js
```diff
@@ -28,6 +30,11 @@ module.exports = async (req, res, next) => {
 	}
 
 	session(req, res, function(err) {
+		if (req.headers.referer && req.session.user && req.session.user.referer && !req.headers.referer.match(req.session.user.referer)) {
```
carmenfan, line 33: should this check just go here? https://github.com/3drepo/3drepo.io/blob/master/backend/middlewares/checkPermissions.js#L74

backend/services/session.js
```diff
@@ -63,8 +63,14 @@ module.exports.regenerateAuthSession = (req, config, user) => {
 						some(browserType => ua[browserType]); // If any of these browser types matches then is a websession
 				}
 
+				if (req.headers.referer) {
+					const refererDomain = req.headers.referer.match(/^(\w)*:\/\/.*?\//);
```
carmenfan, line 67: Not immediately clear why we have to do this?

Probably needs a comment

carmenfan (changes requested):

carmenfan (approved):

PR #2537, opened 2021-03-18
ISSUE 2440 - Custom sequences support

### LINKED ISSUE #2440: Custom sequences via the API (1) - Create Sequences
## Description
Introduce the ability to add a sequence based on viewpoint schema via API. These sequences will be custom sequences that will be different to the Synchro sequences and are editable/removable. Instead of having states, they will utilise the viewpoint schema and animation transformations will not be supported.

## Goals
- [x] The user will be able to create a custom sequence for via the API.
- [x] The user will be able to remove a custom sequence via the API
- [x] The user will be able to edit a custom sequence via the API.
- [x] The user will be able to view sequences appropriate for the revision specified
- [x] Synchro sequences stay read only - the user should not be able to edit/remove a synchro sequence (other than changing the  sequence name, supported from #2431
- [x] The front end should be showing and interacting with customs the same way as synchro sequences

## Specification

<details>

### Backend
We need to add the following endpoints:
- HTTP POST to add a sequence
- HTTP DELETE to remove a sequence
- HTTP GET to fetch a specific sequence

We need to modify an existing endpoint:
- HTTP PATCH to edit a sequence
- HTTP GET to get a list of sequences

##### Make revision association optional
Unlike synchro sequences, these sequences may not necessary be pegged against a particular revision. (esp the ones who are associated with smart groups.) We want to make sure that revision Id is only stored if the user specified it in their data.

We should move all sequence endpoints to be ` /:teamspace/:modelId/sequences` instead of  `/:teamspace/:modelId/revision/[master/head|:revisionId]/sequences`.

In the GET request, allow the user to provide a query string. ie.:

`HTTP GET /:teamspace/:modelId/sequences?revId=:revId`

This should return all the timeless sequences, and any sequences that matches `revId`.

If no query string is provided, it should return every single sequence regardless of revision.

##### Make sequence list API return a summary instead
At the moment, the sequence list api `/:teamspace/:modelId/revision/[master/head|:revisionId]/sequences` returns the full document of the sequence. As it is likely that we will get more than 1 sequence per revision with this change. Similar to risk/issues, we want to make the sequence list api return a summary (everything apart from the frame details), and create a separate API to return the sequence details given a sequence Id

#### Add a sequence
Proposed route: `HTTP POST /:teamspace/:modelId/sequences`
payload:
````js
{
    name: string
    revId: string //optional
    frames: [
        {
             dateTime: number //timestamp
             viewpoint: {} or viewpointId: string //takes in a viewpoint like issues/risks, or an existing viewpointId.
        }
    ]
}
````
- Viewpoint data should be processed just like issues/risks, where highlighted objects etc are moved into a group instead.
- if the user submits a `viewpointId`, take a copy of the viewpoint data instead of taking the link (so if the viewpoint got removed, the sequence is intact.
- `INVALID_ARGUMENTS` if any of the viewpoints contain a transformation.
- `_id` should be automatically generated by the backend
- `rev_id` should only be stored if it's provided, if it is provided, ensure it is a valid revision
- `customSequence` set to true for all sequences created this way.

#### Editing a sequence
- There is already an API done in #2431 to support editing a sequence. Currently it only supports a `name` change, we should change it so that it also support updating the `frames` property.
- `SEQUENCE_READ_ONLY` if the user is trying to update a sequence where `customSequence` is not set to true.
- frames data should be processed in a similar manner to POST request.

#### Deleting a sequence
Proposed route: `HTTP DELETE /:teamspace/:modelId/sequences/sequenceId`
- this should clean up any group Ids that are only associated with the sequence

### Frontend
- For these sequences, viewpoint is replacing states, so the front end needs to be modified to understand it and processing them accordingly.

</details> 

## Tasks

- [x] Create sequence
  - [x] tests
- [x] Remove sequence
  - [x] tests
- [x] Edit sequence
  - [x] tests
- [x] Get sequence(s)
  - [x] tests
  - [x] backend
    - [x] deprecate old endpoints that require revision
  - [x] frontend
    - [x] use new GET endpoint
- [x] List sequences to retrieve summary only
  - [x] update tests
- [x] Synchro sequences read only
  - [x] tests
  - [x] backend
  - [x] frontend
- [x] Check frontend sequence interaction

## DB migration
See [PR](https://github.com/3drepo/3drepo.io/pull/2537#issue-595790494)

## Related Resources
https://coda.io/d/3D-Repo-Features_d5c0Qp25r6f/4D-v3_suIVb#_luYkt

### PR DESCRIPTION
**Pull request by Charence**
This fixes #2440 

**Description** 
- Added various endpoints specified in the issue
- Front end changes to support viewing a sequence with viewpoints
- migration scripts to update existing synchro sequences to the latest schema. Run in following order:
  1) run `addSequenceStartEndDates.py` to add start/end dates to sequences
  2) run `fixSequencesActivities.py` to migrate activities
  3) run `removeOldActivitiesFiles.py` to remove migrated activities

**Test Case**
- [x] Creating a sequence
    - [x] The user should be able to post a new sequence via API
    - [x] The user should be able to post a new sequence pegged against a specific revision via API
- [x] Editing a sequence
    - [x] The user should be able to edit a custom sequence
    - [x] The user should be able to remove revision ID from a custom sequence
    - [x] The user should only be able to change the name on a synchro sequence via API
    - [x] The user should be able to add/modify a revision ID of a sequence
- [x] Sequence listing: 
    - [x]  Should show all sequences if revision is not specified
    - [x] Should show revision-specific sequences if specified
- [x] Deleting a sequence
    - [x] The user should be able to delete an existing custom sequence via API
    - [x] The user should not be able to delete a synchro sequence via API
- [x] Sequence viewing
    - [x] Custom sequences should be viewable as expected on the frontend
    - [x] Synchro sequences should be viewable as expected
- [x] Legends
    - [x] APIs should still work
    - [x] Legend UI should still work
- [x] Activities should display as expected
    - [x] Synchro sequence activities show up correctly
    - [x] Custom sequence with no activities shouldn't break
    - [x] custom sequence activities should show up correctly
- [x] Activities API should function as expected

### Issues
- [x] [Sequence Upload doesn't create groups properly](https://github.com/3drepo/3drepo.io/pull/2539#issuecomment-815907224)
- [x] Ensure sample GIF, sequence groups, and sequence playback all match
- [x] [Fix activities migration script](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-821872684) (Santiago)
- [x] Check/fix playback for old Synchro sequences (Santiago)
- [x] Pre-fetch groups (NOTE: will need #2562 for prefetch fix on stepping by frames)(Santiago)
- [x] Wait for viewpoints to load (Charence)
- [x] [First frame colour overrides flashes when the sequence is played](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-822411613) (<del>Charence</del> Carmen)
- [x] [Playback by frames gets stuck](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-822415973) (carmen)
- [x] Test updated bouncer (carmen)
         - tested, worked on a small model, bigger model got white screened
         - need retesting after `Check/fix playback for old Synchro sequences ` is fixed
- [x] Test go to sequence date from issues/risks/activity list is not broken.
    - [x] White screened when clicking on sequence start date from an issue
- [x] [Activity stuck on loading state](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-816663852) (carmen)
- [x] [Improvement on activity backend](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-825621487)
- [x] Missing instructions for db migration scripts (add them on the issue please!)
- [x] [Sequences must have at least 1 frame](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-825738998)
- [x] [Incorrect documentation on sequence POST](https://github.com/3drepo/3drepo.io/pull/2537#pullrequestreview-643482622) (Charence)
- [x] [Synchro sequence stuck on loading frame](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-825799305) (Charence)
- [x] [Legend spacing tweak](https://github.com/3drepo/3drepo.io/issues/2401#issuecomment-826646182) (Santiago)
- [x] [Edit/Create sequence has inconsistent schema](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-826675791) (Santiago)
- [x] [Don't take in `revId` and store `rev_id`](https://github.com/3drepo/3drepo.io/pull/2537#pullrequestreview-644463758)
- [x] [Create Custom Sequence with unknown rev tag is being allowed - kind of](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-826682415)(Santiago)
- [x] [Cannot remove a revision ID from a sequence](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-826854492) (Charence)
- [x] [screenshot ref should not be saved as part of the sequence viewpoint](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-826883447) (Charence)
- [x] [Toggling activity list is resulting in white screen if the custom sequence has no activities](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-827512883) (Santiago)
- [x] [activity list ordering should be sorted by startTime](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-827627546) (Charence)
- [x] [Sequence front end should be be getting sequences that are only relevant to the revision](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-827658045)
- [x] [Issues found in code review](https://github.com/3drepo/3drepo.io/pull/2537#pullrequestreview-645915726)
- [x] [Reset button on Sequence doesn't restart from first frame](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-832196540) (Santiago)
- [x] [groups in sequences are being stored as plain string](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-832685251) (Charence)
- [x] [Problems with custom sequence](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-832702309) (Santiago)
- [x] [UI not responsive when scrubbing](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-833325979) (Santiago)
- [x] [Changing the date of the first frame on a PATCH causes a failure](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-833359744)(Carmen)

### CODE REVIEW

File: backend/models/sequence.js [hunk 1]
```diff
@@ -116,24 +224,36 @@ Sequence.getSequenceState = async (account, model, stateId) => {
...
+	const sequences = await db.find(account, sequenceCol(model), sequencesQuery);
 
-	const sequences = await db.find(account, sequenceCol(model), {"rev_id": history._id});
 	sequences.forEach((sequence) => {
-		sequence.teamspace = account;
-		sequence.model = model;
+		sequence.minDate = new Date((sequence.frames[0] || {}).dateTime).getTime();
+		sequence.maxDate = new Date((sequence.frames[sequence.frames.length - 1] || {}).dateTime).getTime();
+
+		delete sequence.frames;
```
carmenfan, line 253: do a exclusion (i.e. `frames: 0`) projection instead doing this.

no point dragging the data out of the database just to ditch it.

backend/models/sequence.js
```diff
@@ -116,24 +224,36 @@ Sequence.getSequenceState = async (account, model, stateId) => {
...
-	const submodelSequencesPromises = Promise.all(submodels.map((submodel) => Sequence.getList(account, submodel, "master", null, cleanResponse)));
+	const submodelSequencesPromises = Promise.all(submodels.map((submodel) => Sequence.getList(account, submodel, submodelBranch, undefined, cleanResponse)));
+
+	const sequences = await db.find(account, sequenceCol(model), sequencesQuery);
 
-	const sequences = await db.find(account, sequenceCol(model), {"rev_id": history._id});
 	sequences.forEach((sequence) => {
-		sequence.teamspace = account;
-		sequence.model = model;
+		sequence.minDate = new Date((sequence.frames[0] || {}).dateTime).getTime();
```
carmenfan, line 250: as discussed, we should just write it into the db if we need it.

The whole purpose of ditching the frames was to reduce the data, so if we're fetching it to the backend it kind of defeats the point

backend/models/sequence.js
```diff
@@ -144,12 +264,43 @@ Sequence.getList = async (account, model, branch, revision, cleanResponse = fals
...
+	if (data.name) {
+		if (!utils.isString(data.name) || data.name === "" || data.name.length >= 30) {
+			throw responseCodes.INVALID_ARGUMENTS;
+		}
+
+		toSet.name = data.name;
+	}
+
+	if (data.revId) {
+		const history = await History.getHistory(account, model, undefined, data.revId);
```
carmenfan, line 282: another spot for revisionExists 😆

backend/models/sequence.js
```diff
@@ -144,12 +264,43 @@ Sequence.getList = async (account, model, branch, revision, cleanResponse = fals
...
 
+	if (data.name) {
+		if (!utils.isString(data.name) || data.name === "" || data.name.length >= 30) {
+			throw responseCodes.INVALID_ARGUMENTS;
+		}
+
+		toSet.name = data.name;
+	}
+
+	if (data.revId) {
```
carmenfan, line 281: might want to check for null explicitly, to allow the option for user to unset a revision should they wish

backend/models/sequence.js
```diff
@@ -144,12 +264,43 @@ Sequence.getList = async (account, model, branch, revision, cleanResponse = fals
...
+		if (!customSequence) {
+			throw responseCodes.SEQUENCE_READ_ONLY;
+		}
+
+		toSet.frames = await handleFrames(account, model, sequenceId, data.frames);
+	}
+
 	const { result } = await db.update(account, sequenceCol(model),
-		{_id: utils.stringToUUID(sequenceId)}, {$set: {name: data.name}});
+		{_id: utils.stringToUUID(sequenceId)}, {$set: toSet});
```
carmenfan, line 303: we should check if sequence exist on the top, otherwise (esp with the frames stuff that would also potentially write things into the database), we could end up with a lot of zombie data.

backend/models/viewpoint.js
```diff
@@ -252,51 +276,27 @@ const createViewpoint = async (account, model, collName, routePrefix, hostId, vp
...
+			await Viewpoint.setExternalScreenshotRef(viewpoint, account, model, collName);
+		}
 	}
 
-	return clean(routePrefix, viewpoint, false);
+	return Viewpoint.cleanViewpoint(routePrefix, viewpoint, false);
 };
 
-const setExternalScreenshotRef = async function(viewpoint, account, model, collName) {
+Viewpoint.setExternalScreenshotRef = async function(viewpoint, account, model, collName) {
```
carmenfan, line 294: is this used outside of this file now? 

how come we made it public?

backend/routes/sequence.js
```diff
@@ -1,5 +1,5 @@
 /**
- *  Copyright (C) 2019 3D Repo Ltd
+ *  Copyright (C) 2021 3D Repo Ltd
```
carmenfan, line 2: shouldn't really be updating copy right notices?

i think if there's update it should say 2019-2021 if you want to pedantic about it 😆

backend/routes/sequence.js
```diff
@@ -29,14 +29,14 @@ const Sequence = require("../models/sequence");
  *
  * @apiParam {String} teamspace Name of teamspace
  * @apiParam {String} model Model ID
- * @apiParam {String} [revId] Revision unique ID
  */
 
 /**
  * @api {get} /:teamspace/:model/revision(/master/head/|/:revId)/sequences/:sequenceId/activities/:activityId Get activity
- * @apiName getSequenceActivityDetail
+ * @apiName getSequenceActivityDetailOld
```
carmenfan, line 36: don't modify anything regarding activities, it's already updated by santiago on his branch

backend/routes/sequence.js
```diff
@@ -156,12 +197,88 @@ router.get("/revision/:revId/sequences/:sequenceId/activities/:activityId", midd
...
+ * 		}
+ * 	]
+ * }
+ */
+router.get("/sequences/:sequenceId/activities", middlewares.issue.canView, getSequenceActivities);
 
 /**
  * @api {get} /:teamspace/:model/revision(/master/head/|/:revId)/sequences/:sequenceId/state/:stateId Get state
- * @apiName getSequenceState
+ * @apiName getSequenceStateOld
```
carmenfan, line 278: just ditch these APIs, we're not providing backwards support since we don't think there's any users who're using the APIs for this atm.

backend/routes/sequence.js
```diff
@@ -298,15 +555,37 @@ router.patch("/revision/:revId/sequences/:sequenceId", middlewares.hasUploadAcce
...
+ * @apiSuccessExample {json} Success-Response
+ * HTTP/1.1 200 OK
+ * {}
+ */
+router.put("/sequences/:sequenceId/legend", middlewares.hasUploadAccessToModel, updateLegend);
 
 /**
  * @api {get} /:teamspace/:model/revision(/master/head/|/:revId)/sequences/:sequenceID/legend get the legend
- * @apiName getLegend
+ * @apiName getLegendOld
```
carmenfan, line 585: same with legends, just remove the old endpoints.

frontend/modules/sequences/sequences.selectors.ts
```diff
@@ -24,9 +24,9 @@ import { getDateByStep, getSelectedFrame } from './sequences.helper';
 
 export const selectSequencesDomain = (state) => (state.sequences);
 
-const getMinMaxDates = ({frames}) => ({
-	minDate: (frames[0] || {}).dateTime,
-	maxDate: (frames[frames.length - 1] || {}).dateTime
+const getMinMaxDates = ({frames, minDate, maxDate}) => ({
```
carmenfan, line 27: with the change they should always have a min/max dates now?

Does this function need to exist?

it's very odd to be passing in the result just to return it.

frontend/routes/viewerGui/components/sequences/components/sequencePlayer/sequencePlayer.component.tsx
```diff
@@ -1,346 +1,346 @@
...
-							label="Full Activity List"
-						/>
-					</FormGroup>
-					{this.renderLoader(this.props.loadingFrame && !this.props.isActivitiesPending)}
-				</SequencePlayerColumn>
-			</SequencePlayerContainer>
-		);
-	}
-}
+/**
```
carmenfan, line 1: very hard to read what's changed here... why does git thinks the whole file is modified? 😟

frontend/modules/sequences/sequences.selectors.ts
carmenfan: also can we rename this to endDate and startDate?

backend/models/sequence.js
carmenfan: Added it on the frontend side but just putting here for consistency too... can we rename to startDate and endDate?

frontend/routes/viewerGui/components/sequences/components/sequencePlayer/sequencePlayer.component.tsx
Charence: @carmenfan I'm not sure. A few (not all) `tsx` files seemed a little odd, they'd load without any syntax highlighting and freeze my computer; this was one of them.

carmenfan: can we try fixing it?

backend/models/viewpoint.js
Charence: @carmenfan I noticed it's being used by `models/bcf`. Its usage wasn't added in this branch, but I thought I'd fix it since I noticed it.

See here: https://github.com/3drepo/3drepo.io/blob/master/backend/models/bcf.js#L928

frontend/routes/viewerGui/components/sequences/components/sequencePlayer/sequencePlayer.component.tsx
Charence: @carmenfan It seems to be related to line endings. I thought git should take care of this, but it seems some `tsx` files have DOS-style line endings.

carmenfan: ahh tsx needs to be added here https://github.com/3drepo/3drepo.io/blob/master/.gitattributes

carmenfan: infact i don't know why it's not auto lf.

everywhere I've set has auto lf and do exclusion if necessary
https://github.com/3drepo/3drepobouncer/blob/master/.gitattributes

backend/models/sequence.js
```diff
@@ -97,6 +151,60 @@ const getDefaultLegend = async (account, model) => {
...
+	if (!data ||
+		!data.name || !utils.typeMatch(data.name, fieldTypes["name"]) ||
+		!data.frames || !utils.typeMatch(data.frames, fieldTypes["frames"])) {
+		throw responseCodes.INVALID_ARGUMENTS;
+	} else {
+		newSequence.name = data.name;
+	}
+
+	if (data.revId) {
+		const history = await History.getHistory(account, model, undefined, data.revId);
```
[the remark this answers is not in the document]
Charence: @carmenfan I've added the projection but not sure if it needs a different `revisionExists` function. `getHistory()` seems to now check for an empty history, so the previous checks after the call are no longer necessary.

backend/models/sequence.js
```diff
@@ -116,24 +224,36 @@ Sequence.getSequenceState = async (account, model, stateId) => {
 };
 
 Sequence.getList = async (account, model, branch, revision, cleanResponse = false) => {
-	const history = await History.getHistory(account, model, branch, revision);
+	let submodelBranch;
+	let sequencesQuery = {};
+
+	if (branch || revision) {
+		const history = await History.getHistory(account, model, branch, revision);
```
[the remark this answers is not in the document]
Charence: @carmenfan See previous comment

backend/models/sequence.js
Charence: @carmenfan See previous comment

backend/models/sequence.js
```diff
@@ -144,16 +307,59 @@ Sequence.getList = async (account, model, branch, revision, cleanResponse = fals
...
+
+	if (data.name) {
+		if (!utils.isString(data.name) || data.name === "" || data.name.length >= 30) {
+			throw responseCodes.INVALID_ARGUMENTS;
+		}
+
+		toSet.name = data.name;
+	}
+
+	if (data.revId) {
```
carmenfan, line 328: you shouldn' tbe able to alter revId either if it's a synchro sequence, so maybe move the custom sequence check out to wrap around this and the frames?

carmenfan (changes requested):
as discussed, also need a db migration script to add startDate endDate to sequences

mosment3DRepo:
### Description

Sequence Upload doesn't create groups properly

### Steps to replicate
- Open this model: [Simple Building](https://issue-2440.dev.3drepo.io/viewer/mosment3ddev/10ff30c0-9874-11eb-8e5c-c9ff1aec95c5)
- The Sequence was uploaded via API
- When you query the sequence you get the following [Sequence Query](https://api3.issue-2440.dev.3drepo.io/api/mosment3ddev/10ff30c0-9874-11eb-8e5c-c9ff1aec95c5/sequences/b5c8df30-9878-11eb-a8eb-2951c087846a)
- Notice that all of the Group IDs are duplicated, and if you query this group it is not found in the model
- The sequence does not play back correctly

<details>
<summary> This is the sequence payload that way uploaded </summary>

````
{
    "name":"Example Sequence",
    "frames":[
        {
            "dateTime":1616400000000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                    "0h79Q0rcfC1gOPK50yoFCv",
                                    "0K5o7g755EZw2RjNI7HcYK",
                                    "0yuGDtpaPCSBT7QB7wvN5I",
                                    "2HBVtaIWv07ud53r01WB6q"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ],
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "2S2omCydz5b9jSgrcLLblk",
                                "0_U7q0Dzj6DfPp4VzMmTUt",
                                "0iMv$JxRL67v6DoyA3RRwz",
                                "1W4yiIKW92qAUdezi70DTY",
                                "00ojKm$5f7luRCAjta0hsu",
                                "0d2LnELub06glJ9mZh2up$",
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
        {
            "dateTime":1616572800000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                    "2S2omCydz5b9jSgrcLLblk",
                                    "0_U7q0Dzj6DfPp4VzMmTUt",
                                    "0iMv$JxRL67v6DoyA3RRwz",
                                    "1W4yiIKW92qAUdezi70DTY"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ],
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "00ojKm$5f7luRCAjta0hsu",
                                "0d2LnELub06glJ9mZh2up$",
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
        {
            "dateTime":1616695200000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                    "2S2omCydz5b9jSgrcLLblk",
                                    "0_U7q0Dzj6DfPp4VzMmTUt",
                                    "0iMv$JxRL67v6DoyA3RRwz",
                                    "1W4yiIKW92qAUdezi70DTY"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ],
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "00ojKm$5f7luRCAjta0hsu",
                                "0d2LnELub06glJ9mZh2up$",
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
        {
            "dateTime":1616781600000,
            "viewpoint":{
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "00ojKm$5f7luRCAjta0hsu",
                                "0d2LnELub06glJ9mZh2up$",
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
        {
            "dateTime":1617091200000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                    "0d2LnELub06glJ9mZh2up$"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ],
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "00ojKm$5f7luRCAjta0hsu",
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
       {
            "dateTime":1617213600000,
            "viewpoint":{
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "00ojKm$5f7luRCAjta0hsu",
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
        {
            "dateTime":1617264000000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ],
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "00ojKm$5f7luRCAjta0hsu"

                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
        {
            "dateTime":1617696000000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi",
                                "00ojKm$5f7luRCAjta0hsu"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ]
            }
        },
        {
            "dateTime":1617732000000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                "00ojKm$5f7luRCAjta0hsu"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ]
            }
        },
        {
            "dateTime":1617818400000,
            "viewpoint":{}
            
        }          
    ]
}
````


</details>

### Current Behaviour
- Sequence does not play back possibly because the groups are not being displayed correctly
### Expected Behaviour
- The sequence frames should load according to the JSON request

**Comment by Charence** [comment]
@mosment3DRepo Groups and playback of viewpoints should now be fixed. Please give the build about 15 mins to become available.

I noticed towards the end of "Example Sequence 2" there's a "Loading activity list" message that doesn't go away; not sure if this is expected.

Also, when loading a viewpoint for each frame, there's an annoying flicker on my computer where you see the whole model before the new groups of the new viewpoint are applied.

**Comment by carmenfan** [comment]
> Also, when loading a viewpoint for each frame, there's an annoying flicker on my computer where you see the whole model before the new groups of the new viewpoint are applied.

@Charence that also happened in states, we basically had to keep the last state until we've fetched everything. Probably worth delving in the viewpoints code and see whether we apply stuff before fetching groups

**Comment by mosment3DRepo** [comment]
@Charence the playback doesn't appear to be correct, looking at the groups it appears they are OK and have captured all the right elements,

The play back should look like the images in this folder:
https://www.dropbox.com/home/3D%20Repo%20Team%20Folder/Integrations/PowerProject/Screenshots

4D Demo

**Comment by Charence** [comment]
> @Charence the playback doesn't appear to be correct, looking at the groups it appears they are OK and have captured all the right elements,

I'm getting this: https://www.dropbox.com/s/u0eiih3isj52xkw/Screen%20Recording%202021-04-09%20at%2016.47.19.mov?dl=0

The last 2 frames don't seem to match up to your GIF, is that right?

**Comment by mosment3DRepo** [comment]
The last 2 and also the first 4 don't match

**Comment by Charence** [comment]
> The last 2 and also the first 4 don't match

This should be fixed now

**Comment by mosment3DRepo** [comment]
Frames correct now, but the flashing seems to be very visible.

This is only happening the first time I run through frames, seems like once they are cached it stops happening:

4D Flash

**Comment by Charence** [comment]
@mosment3DRepo The custom sequence playback flicker should be fixed now. I have run the addSequenceStartEndDates DB script on my teamspace and Carmen's but old Synchro sequences seem to have a problem. Trying to find when this problem was introduced.

**Comment by mosment3DRepo** [comment]
@Charence I think the incompatability is related to a schema change. We need to migrate old sequences over when we go to prod

**Comment by carmenfan** [comment]
> @mosment3DRepo The custom sequence playback flicker should be fixed now. I have run the addSequenceStartEndDates DB script on my teamspace and Carmen's but old Synchro sequences seem to have a problem. Trying to find when this problem was introduced.

you'd also need to run the migration for sequence activity

**Comment by Charence** [comment]
@sanmont3drepo I'm having a problem with the fixSequencesActivities migration script

Screenshot from 2021-04-17 20-08-05

**Comment by mosment3DRepo** [comment]
### First frame colour overrides flashes when the sequence is played

### Steps to replicate
- Open this [model](https://issue-2440.dev.3drepo.io/viewer/mosment3ddev/10ff30c0-9874-11eb-8e5c-c9ff1aec95c5)
- Open Sequence 3
- Play the sequence back by days
- See that the piles start as green then flash to uncoloured before going back to green
- This should not happen (see previous GIF)

### Current Behaviour
- Colour override seems to flash on then off for the initial frames

### Expected Behaviour
- See previous GIF for playback examle

4D Flashv2

**Comment by mosment3DRepo** [comment]
### Playback by frames gets stuck

### Steps to replicate
- Open this [model](https://issue-2440.dev.3drepo.io/viewer/mosment3ddev/10ff30c0-9874-11eb-8e5c-c9ff1aec95c5)
- Open Sequence 3
- Change playback to "frames"
- Play sequence
- Notice after 2 or 3 frames the sequence stops playing back

### Current Behaviour
- The frame playback freezes

### Expected Behaviour
- The playback should continue until the end of the sequence

4D Frame

**Comment by Charence** [comment]
@mosment3DRepo This frame doesn't appear to exist when I manually apply the groups: https://www.dropbox.com/s/a22363e9x4z93ep/1616572800.png?dl=0

It looks like this: 
Screenshot from 2021-04-19 10-48-18

**Comment by Charence** [comment]
@mosment3DRepo I've also noticed that when the override is applied there's a slight delay on my computer. When playing back by days, it looks like one of the earlier frames is skipped, but this seems to be because the colour hasn't been applied yet before the next frame (which doesn't have colour) is loaded.

Screenshot from 2021-04-19 16-27-36

Also, towards the end, it always seems to skip to colouring the top and not doing the sides as well, but in this case there are 2 frames falling on the same day. If the playback is by day, then there's a chance it may miss the earlier frame in the day because the playback doesn't request that one. When playing back by hours or by manually dragging the slider, it appears.

Screenshot from 2021-04-19 16-27-58

Screenshot from 2021-04-19 16-28-14

So the frames seem to match what is stored in the groups.

**Comment by mosment3DRepo** [comment]
@Charence I think the problem is I raised might be separate to what you're talking about. If you play back the sequence by days the 4 column piles start as green, which is correct according to the first frame. The moment you hit play the colour override disappears before coming back the following day.

According to the sequence data those piles should be green at 8am on the 22nd (start of the sequence) and the colour override is removed by the next frame at 8am on the 24th. There isn't an intermediate frame where the colour override is removed.

### Current Behaviour

4D flash

### Expected Behaviour

4D flash2

**Comment by Charence** [comment]
@mosment3DRepo Yes, we're talking about different problems. My last 2 posts were about the playback matching the GIF. In one of frames of the GIF the 4 colours and the square blocks on top of them are all green, but manually apply the groups the override doesn't colour both the columns and square blocks at the same time.

**Comment by mosment3DRepo** [comment]
@Charence I've amended the Sequence so it should match the GIF now

**Comment by mosment3DRepo** [comment]
> ### Playback by frames gets stuck
> ### Steps to replicate
> * Open this [model](https://issue-2440.dev.3drepo.io/viewer/mosment3ddev/10ff30c0-9874-11eb-8e5c-c9ff1aec95c5)
> * Open Sequence 3
> * Change playback to "frames"
> * Play sequence
> * Notice after 2 or 3 frames the sequence stops playing back
> 
> ### Current Behaviour
> * The frame playback freezes
> 
> ### Expected Behaviour
> * The playback should continue until the end of the sequence
> 
> ### Pictures/ Gifs of the error
> 4D Frame

In the following GIF you can see that the UI stops sending network requests when the playback gets stopped. Dragging the slider can force it to retrieve the relevant frames, but when playback is started from the beginning it still gets stuck in the same place, even though the frames are cached

SequenceNetwork

**Comment by Charence** [comment]
> ### First frame colour overrides flashes when the sequence is played
> ### Steps to replicate
> * Open this [model](https://issue-2440.dev.3drepo.io/viewer/mosment3ddev/10ff30c0-9874-11eb-8e5c-c9ff1aec95c5)
> * Open Sequence 3
> * Play the sequence back by days
> * See that the piles start as green then flash to uncoloured before going back to green
> * This should not happen (see previous GIF)

@mosment3DRepo After some debugging, I've found that this flashing occurs as the state of the colour overrides goes from`

```
{
  2ab1b00f-cd3a-4275-9378-f9b1bb36ad8b: "#00FF00"
  6b994a71-d7c4-49b1-8ce9-1eba8cf1baa5: "#00FF00"
  7523ec62-f9ab-4a22-9779-24b79f51f85b: "#00FF00"
  d7e7d28a-3512-4d50-a2cc-bd4cadea85c5: "#00FF00"
}
```

to `{}`, and then it goes from `{}` back to the state above.

Playback works this way currently in order to reset any changes the user may have applied to the model (e.g. additional colour overrides) while the playback is paused; it happens on the first frame that is played back and shouldn't be different to the existing behaviour on prod. The question is "**can we live with this?**"

**Comment by carmenfan** [comment]
> > > ### First frame colour overrides flashes when the sequence is played
> > > ### Steps to replicate
> > > 
> > > * Open this [model](https://issue-2440.dev.3drepo.io/viewer/mosment3ddev/10ff30c0-9874-11eb-8e5c-c9ff1aec95c5)
> > > * Open Sequence 3
> > > * Play the sequence back by days
> > > * See that the piles start as green then flash to uncoloured before going back to green
> > > * This should not happen (see previous GIF)
> > 
> > 
> > @mosment3DRepo After some debugging, I've found that this flashing occurs as the state of the colour overrides goes from`
> > ```
> > {
> >   2ab1b00f-cd3a-4275-9378-f9b1bb36ad8b: "#00FF00"
> >   6b994a71-d7c4-49b1-8ce9-1eba8cf1baa5: "#00FF00"
> >   7523ec62-f9ab-4a22-9779-24b79f51f85b: "#00FF00"
> >   d7e7d28a-3512-4d50-a2cc-bd4cadea85c5: "#00FF00"
> > }
> > ```
> > 
> > 
> > to `{}`, and then it goes from `{}` back to the state above.
> > Playback works this way currently in order to reset any changes the user may have applied to the model (e.g. additional colour overrides) while the playback is paused; it happens on the first frame that is played back and shouldn't be different to the existing behaviour on prod. The question is "**can we live with this?**"
> 
> Ah ok, I see why it's necessary, but can we have it do that cycle without moving on to the second day. If you playback the same sequence by 'frames' it does the below sequence, making it look like frame 1 has no colour overrides
> 
> ### Current Behaviour
> sequenceflashv1
> 
> ### Potential Fix
> sequenceflashv2

@Charence  I think we can achieve this by just adding the interval wait before we start the interval loop https://github.com/3drepo/3drepo.io/blob/master/frontend/routes/viewerGui/components/sequences/components/sequencePlayer/sequencePlayer.component.tsx#L190

So it'll always wait 1s before going to the next step on the first frame

**Comment by mosment3DRepo** [comment]
@Charence @carmenfan issue relating to activites not loading was due to the epoch times on the activities being incorrect.

Is there anyway for the UI to not show the loading icon if there are no relevant activities?

**Comment by carmenfan** [comment]
> @Charence @carmenfan issue relating to activites not loading was due to the epoch times on the activities being incorrect.
> 
> Is there anyway for the UI to not show the loading icon if there are no relevant activities?

@mosment3DRepo  yup I'm pretty sure it's just an oversight.

Also I just triggered a redeployment, the overrides on the first frame, custom sequence getting stuck on frame-stepping should now be fixed.

**Comment by carmenfan** [comment]
**Activities timestamps are being stored as doubles**
Can we make sure it is stored as a int 64? 
image

**Sanity check Activity timestamps**
We should add checks on startDate endDate to ensure they are valid timestamps and if possible, are in ms (matt had some trouble to invalid epoch values with activites, this will help reduce user errors):
https://github.com/3drepo/3drepo.io/pull/2537/files

**Comment by carmenfan** [comment]
**You can edit a sequence to have no frames** 
You can't create a sequence with no frames, but you can make a patch to an existing sequence to set frames to [].

This is inconsistent, and also breaks UI 
image

We should stop user to be able to edit frames to be empty (probably the easiest solution?)

File: backend/routes/sequence.js [hunk 16]
```diff
@@ -298,70 +235,109 @@ router.patch("/revision/:revId/sequences/:sequenceId", middlewares.hasUploadAcce
...
+ * @apiUse Sequences
+ *
+ * @apiExample {post} Example usage
+ * POST /acme/00000000-0000-0000-0000-000000000000/sequences HTTP/1.1
+ * {
+ * 	"name":"Custom Sequence 1",
+ * 	"frames":[
+ * 		{
+ * 			"dateTime":1244246400000,
+ * 			"state":"00000000-0000-0000-0001-000000000002"
```
carmenfan, line 309: This is synchro state, we should be showing example with viewpoint

carmenfan:
Synchro sequences are stuck on loading frames @Charence , it was still working at the beginning of the day, so it must be something to do with the viewpoint load stuff?
image

backend/routes/sequence.js
carmenfan: @Charence  can we make a better example ? the user who go into this document would probably be more interested in things like colour overrides, hiding objects, let's make sure the example includes that

carmenfan:
**Edit/Create sequence has inconsistent schema**

one takes in `revId` and the other `rev_id`
image

backend/models/sequence.js
```diff
@@ -144,25 +286,70 @@ Sequence.getList = async (account, model, branch, revision, cleanResponse = fals
...
+			{_id: utils.stringToUUID(sequenceId), customSequence: true});
+
+		if (!customSequence) {
+			throw responseCodes.SEQUENCE_READ_ONLY;
+		}
+
+		if (data.revId) {
+			const history = await History.getHistory(account, model, undefined, data.revId, {_id: 1});
+
+			toSet.rev_id = history._id;
```
carmenfan, line 318: Can we not do this and just take in `rev_id` instead of revId. it's just add confusion.

carmenfan:
**Create Custom Sequence with unknown rev tag is being allowed - kind of**

Taking in consideration the yup schema, i dd this and it passed even though the rev tag is invalid

image

But this seems to be because it expects revId instead of rev_id (related to [this](https://github.com/3drepo/3drepo.io/pull/2537#pullrequestreview-644463758))

If i send this in as `revId` instead, it throws an error as you'd expect. However, since it's not on the sequence schema and the yup schema has "noUnknown()" on it, why does it even go through?

carmenfan:
Cannot remove a revision ID from a sequence

The user should be able to unassign a custom sequence from a revision.

However, if I do a PATCH request with revId (NOTE: using `revId` at the moment instead of `rev_id` due to [this](https://github.com/3drepo/3drepo.io/pull/2537/files/97500c20c4117ac7f4c3b28e04941cf98f922686#diff-ff1db998ebcbb44e8f3c26c0b6d2360652f0d489a23999bd26637ea1e41c93cb)  set to null, it's giving me INVALID_ARGUMENTS as a response

image

backend/models/sequence.js
carmenfan: Also here:
https://github.com/3drepo/3drepo.io/pull/2537/files#diff-ffb4e106db66bb79d9156f73ad54a3a278b6e7f9c41032b5037a2d790432f890R374

We're taking in `revId` as query string here and also POST/PATCH request . but on GET we're returning `rev_id`.

Just make everything consistent.

carmenfan:
screenshot ref should not be saved as part of the sequence viewpoint

if we created the view through giving it a viewId, it's saving screenshort_ref as null, this is also being passed back through the API endpoint. 

if screenshot isn't defined, it shouldn't be saving a reference

image

carmenfan:
**Toggling activity list is resulting in white screen if the custom sequence has no activities**

- Create a custom sequence 
- view that custom sequence
- toggle on `Full Activity List`

fds

It's complaining about frames being null here: 
image

But it doesn't make sense, as the sequence player was viewing it fine, seems like toggling the activity list is causing this selector to re-evaluate with a null frame somehow?

carmenfan:
activity list ordering should be sorted by startTime

I created an activity with 2 sub activities, phase 1 a and phase 1 b:
image


When I see it on 3drepo.io, the ordering is swapped , even though phase 1a starts before phase 1 b.
image

The sibling activities should probably be ordered by start time

carmenfan:
Sequence front end should be be getting sequences that are only relevant to the revision

We need to impose the `rev_id` when we get Sequences, the sequence circled is pegged against a different revision but it's still showing up

image

frontend/modules/sequences/sequences.selectors.ts
```diff
@@ -189,13 +184,22 @@ export const selectSelectedStateId = createSelector(
...
+
 export const selectLastSelectedStateId = createSelector(
 	selectLastSelectedFrame, (frame) =>  (frame || {}).state
 );
 
-export const selectIsLoadingFrame = createSelector(
-	selectSelectedStateId, selectStateDefinitions,
-	(stateId, stateDefinitions) => !(stateDefinitions || {}).hasOwnProperty(stateId)
+export const selectIsViewpointFrame = createSelector(
+	selectSelectedFrame, (frame) => (frame || {}).hasOwnProperty('viewpoint')
```
carmenfan, line 196: Is it better to just piggy back on L187? i.e.

```suggestion
	selectSelectedFrameViewpoint, (viewpoint) => !!viewpoint
```

frontend/modules/viewpoints/viewpoints.sagas.ts
```diff
@@ -291,15 +291,20 @@ export function * deselectViewsAndLeaveClipping() {
...
 	try  {
 		const revision = yield select(selectCurrentRevisionId);
 		const viewpointsGroups = yield select(selectViewpointsGroups);
 
-		if (!viewpoint) {
+		if (!view.viewpoint) {
 			return;
 		}
 
+		// console.log("fetching groups for viewpoint:");
```
carmenfan, line 303: remove commented out console logs

frontend/modules/viewpoints/viewpoints.selectors.ts
```diff
@@ -18,10 +18,16 @@ import { isEmpty, orderBy, values } from 'lodash';
 import { createSelector } from 'reselect';
 import { addToGroupDictionary } from '../../helpers/colorOverrides';
 import { getTransparency, hasTransparency } from '../../helpers/colors';
-import { selectActiveIssue } from '../issues';
+import { selectActiveIssue } from '../issues'; // TODO: this should be refactored out (?)
 import { selectDefaultView } from '../model';
-import { selectActiveRisk } from '../risks';
+import { selectActiveRisk } from '../risks'; // TODO: this should be refactored out (?)
+// TODO: views/viewpoints are 2 different things
```
carmenfan, line 24: remove the todo, no one ever sees it at the right time.

Just create a maintenance issue, at least that's in a kanban to consider when we do maintenance stuff.

If it already has one, then these stuff should just be description in there.

frontend/services/api/sequences.ts
```diff
@@ -18,41 +18,48 @@
...
  * Get sequences list
  *
  * @param teamspace
  * @returns {*|promise}
  */
-export const getSequences = (teamspace, modelId, revision): Promise<any> => {
-	return api.get(`${teamspace}/${modelId}/revision/${revision || HEAD}/sequences`);
+export const getSequenceList = (teamspace, modelId, revision?): Promise<any> => {
+	if (revision) {
+		return api.get(`${teamspace}/${modelId}/sequences/?revId=${revision}`);
```
carmenfan, line 33: ah that's why [this](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-827658045) isn't working! it needs to be rev_id

carmenfan (changes requested):

keifgwinn:
/azp run

--
Keif

backend/models/sequence.js
```diff
@@ -60,23 +79,98 @@ const cleanSequenceList = (toClean) => {
...
+				}
+			}
+		}
+
+		if (!viewpoint) {
+			// frame missing viewpoint
+			throw responseCodes.INVALID_ARGUMENTS;
+		}
+
+		if (viewpoint.transformation_group_ids || viewpoint.transformation_groups) {
```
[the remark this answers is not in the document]
sanmont3drepo: Now is being checked with yup before everything!

backend/models/sequence.js
```diff
@@ -97,45 +191,92 @@ const getDefaultLegend = async (account, model) => {
...
-	const history = await History.getHistory(account, model, branch, revision);
+	let submodelBranch;
+	let sequencesQuery = {};
 
-	if(!history) {
-		throw responseCodes.INVALID_TAG_NAME;
+	if (branch || revision) {
+		const history = await History.getHistory(account, model, branch, revision, {_id: 1});
+
+		submodelBranch = "master";
```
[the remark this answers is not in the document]
Charence: @carmenfan @mosment3DRepo There is a slight difference and it depends what data the user should be able to query. At the moment, we're allowing users to get ALL federation and sub-model sequences.

If we make sub-model revision branch to always be "master" then they will only ever get the sequences for the latest revision + sequences with unspecified revision in the sub-models. For models and the federation, we retrieve all sequences if no revision is specified.

With this change it won't retrieve all sub-model sequences.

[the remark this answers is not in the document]
Charence: @carmenfan @mosment3DRepo In the case where the user does specify a specific revision, for federations it will retrieve sequences from that revision + sequences with unspecified revision and the latest "master" sequences from sub-models.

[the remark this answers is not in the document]
carmenfan: in that case let's leave it as is.

backend/models/sequence.js
```diff
@@ -19,16 +19,102 @@
...
+			max = frame.dateTime;
+		}
+	});
+
+	return {min, max};
+};
+
+const byDateTime = (frameA, frameB) => frameA.dateTime - frameB.dateTime;
+
+const notEmpty = (obj) => Object.keys(obj).length > 0;
```
carmenfan, line 53: I think this is better placed in utils as it's a very generic function

backend/models/sequence.js
```diff
@@ -19,16 +19,102 @@
...
+// Viewpoint checks
+const viewpointSchema = yup.object().test((viewpoint) => {
+	if (!viewpoint) {
+		return true;
+	}
+
+	const {transformation_group_ids, transformation_groups } = viewpoint;
+
+	// The viewpoints for sequences CANT have transformations groups
+	if (transformation_group_ids || transformation_groups) {
```
carmenfan, line 64: Can be simplified
```suggestion
	return !(transformation_group_ids || transformation_groups);
```

backend/models/sequence.js
```diff
@@ -19,16 +19,102 @@
...
+
+const sequenceSchema = yup.object().shape({
+	name: nameSchema.required(),
+	startDate: yup.number(),
+	endDate: yup.number(),
+	rev_id: yup.string(),
+	frames: yup.array().of(frameSchema).min(1).required()
+}).noUnknown().test(({startDate, endDate, frames}) => {
+	if ((startDate !== undefined || endDate !== undefined) && frames) {
+		const {min, max} = findMinMaxDate(frames);
```
carmenfan, line 90: 2 questions here:

- we're looping through this to find min/max dates, later on in the code we have to sort the dates by order. Is it possible for yup to sort it here and find the minMax after that?
- If we can't do the above, maybe it's better to just leave this logic as a check manually?

In essence I want to avoid the double looping if possible

backend/models/sequence.js
```diff
@@ -19,16 +19,102 @@
...
+		}
+
+		// If the max date found in frames is larger than the passed end date
+		// then the object is inconsistent
+		if(endDate !== undefined && max < endDate) {
+			return false;
+		}
+	}
+
+	if (startDate !== undefined &&  endDate !== undefined && startDate > endDate) {
```
carmenfan, line 105: @jozefdobos  will scream if you do if (x) then true else false.

```suggestion
	return ! (startDate !== undefined &&  endDate !== undefined && startDate > endDate)
```

backend/models/sequence.js
```diff
@@ -19,16 +19,102 @@
...
+const nameSchema = yup.string().min(1).max(30);
+
+const sequenceSchema = yup.object().shape({
+	name: nameSchema.required(),
+	startDate: yup.number(),
+	endDate: yup.number(),
+	rev_id: yup.string(),
+	frames: yup.array().of(frameSchema).min(1).required()
+}).noUnknown().test(({startDate, endDate, frames}) => {
+	if ((startDate !== undefined || endDate !== undefined) && frames) {
```
carmenfan, line 89: Can frames be undefined? does yup still run this test if it failed the `min(1).required()`?

I'm thinking that because L94 onwards could otherwise be reduced to this:
```suggestion
	return !(startDate !== undefined && min > startDate ||endDate !== undefined && max < endDate)
```

backend/models/sequence.js
```diff
@@ -19,16 +19,102 @@
...
+		return false;
+	}
+
+	return true;
+});
+
+const sequenceEditSchema = sequenceSchema.shape({
+	name: nameSchema,
+	rev_id: yup.string().nullable(),
+	frames: yup.array().of(frameSchema).min(1)
```
carmenfan, line 115: if we let user put in the startDate endDate then we should let the user edit the start date and end date?

Also if the frames are updated we may want to update the startDate/endDate?

backend/models/sequence.js
```diff
@@ -60,23 +146,88 @@ const cleanSequenceList = (toClean) => {
...
-const sequenceExists = async (account, model, sequenceId) => {
-	if(!(await getSequenceById(account, model, utils.stringToUUID(sequenceId), {_id: 1}))) {
-		throw responseCodes.SEQUENCE_NOT_FOUND;
+		processedFrames.push({
+			dateTime,
+			viewpoint
+		});
 	}
+
+	processedFrames.sort(byDateTime);
```
carmenfan, line 228: byDateTime is only used here and it's so specific to this cause, I think we should just embed the function in here
```suggestion
	processedFrames.sort((frameA, frameB) => frameA.dateTime - frameB.dateTime);
```

carmenfan (changes requested):

backend/models/sequence.js
sanmont3drepo: We can do the first thing. I'll do the second one.

backend/models/sequence.js
sanmont3drepo: It could be undefined with sequenceEditSchema, It "inherited" from sequenceSchema. But I changed to  yup.object() so is no longer possible.

TBH I think  the change would be a little hard to understand at first glance. I would get rid only of the frames thing.

sanmont3drepo: The frames can be undefined when is checking from  sequenceEditSchema

carmenfan (approved):

mosment3DRepo:
### Reset button on Sequence doesn't restart from first frame

Following on from the earlier issue with the first frame which I think was resolved [here](https://github.com/3drepo/3drepo.io/pull/2537#issuecomment-824135333)

When a sequence is finished and the user hits the "restart" button, the first frame does not load when the playback is set to "by frames". If the playback is set to a time such as "days" it takes 1 day before the first frame loads.

### Steps to replicate:
1. Open [This model](https://issue-2440.dev.3drepo.io/viewer/mosment3ddev/df98b770-a379-11eb-a1ba-8f9871e95712)
2. Select "Example Sequence"
3. Play back by frames
4. Notice first frame is only the cars coloured red
5. Once the playback is finished, hit the reset button
6. Notice the view does not change until the second frame, skipping the first one entirely

sequence

carmenfan:
**groups in sequences are being stored as plain string:**

they should be LUUIDs
image

carmenfan:
**Problems with custom sequence**

1. It's still not waiting on group fetch (you can see on the network tab that the groups are still pending and the frames are already moving
2. Either we're not caching groups that are already fetched, or we keep pre-fetching when we have the group data. (when I replay the sequence, it's still trying to fetch the groups...)

mosment3DRepo:
### UI not responsive when scrubbing

Undesirable but not release preventing

If you have a particularly large model/sequence loaded, when you scrub the time and date, the UI doesn't respond until the frame is loaded. This makes it very hard to actually know what you're scrubbing too. Perhaps we could allow the UI to work and have a 0.5s delay before we load the relevant frame. This would make the UI feel more snappy.


### Steps to replicate:
1. Open [This model](https://issue-2440.dev.3drepo.io/viewer/mosment3ddev/df98b770-a379-11eb-a1ba-8f9871e95712)
2. Select "Sequence v2"
3. Scrub to a date
4. Notice the delay in UI response

sequence2

mosment3DRepo:
### Changing the date of the first frame on a PATCH causes a failure

The Sequence doesn't seem to be able to cope if the first date of a sequence is changes with a PATCH

### Steps to replicate:
1. Open [This model](https://issue-2440.dev.3drepo.io/viewer/mosment3ddev/10ff30c0-9874-11eb-8e5c-c9ff1aec95c5)
2. POST a new sequence using the below code
3. Create a new PATCH request to the newly created sequence with the same payload but changing the first date to be later
4. Open the sequence in 3D Repo and notice it has broken (the sequence start date doesn't match the first date of the frames)
```
curl --location --request POST 'https://issue-2440.dev.3drepo.io/api/mosment3ddev/10ff30c0-9874-11eb-8e5c-c9ff1aec95c5/sequences?key=835b7e2ec2a47c812230493762497b6c' \
--header 'Content-Type: application/json' \
--data-raw '{
    "name":"Example Sequence date change",
    "frames":[
        {
            "dateTime":1626400000000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                    "0h79Q0rcfC1gOPK50yoFCv"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ]
            }
        },
        {
            "dateTime":1616572800000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
									"0h79Q0rcfC1gOPK50yoFCv"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ]
            }
        }
    ]
}'
```


Open the "Broken Date Change" sequence for an example

image

carmenfan:
> ### Changing the date of the first frame on a PATCH causes a failure
> The Sequence doesn't seem to be able to cope if the first date of a sequence is changes with a PATCH
> 
> ### Steps to replicate:
> 1. Open [This model](https://issue-2440.dev.3drepo.io/viewer/mosment3ddev/10ff30c0-9874-11eb-8e5c-c9ff1aec95c5)
> 2. POST a new sequence using the below code
> 3. Create a new PATCH request to the newly created sequence with the same payload but changing the first date to be later
> 4. Open the sequence in 3D Repo and notice it has broken (the sequence start date doesn't match the first date of the frames)
> 
> ```
> curl --location --request POST 'https://issue-2440.dev.3drepo.io/api/mosment3ddev/10ff30c0-9874-11eb-8e5c-c9ff1aec95c5/sequences?key={apikey}' \
> --header 'Content-Type: application/json' \
> --data-raw '{
>     "name":"Example Sequence date change",
>     "frames":[
>         {
>             "dateTime":1626400000000,
>             "viewpoint":{
>                 "override_groups":[
>                     {
>                         "color":[
>                             0,
>                             255,
>                             0
>                         ],
>                         "rules":[
>                             {
>                                 "field":"IFC GUID",
>                                 "operator":"IS",
>                                 "values":[
>                                     "0h79Q0rcfC1gOPK50yoFCv"
>                                 ]
>                             }
>                         ],
>                         "account":"mosment3d",
>                         "model":"29460f20-8d41-11eb-894b-072add228ef4"
>                     }
>                 ]
>             }
>         },
>         {
>             "dateTime":1616572800000,
>             "viewpoint":{
>                 "override_groups":[
>                     {
>                         "color":[
>                             0,
>                             255,
>                             0
>                         ],
>                         "rules":[
>                             {
>                                 "field":"IFC GUID",
>                                 "operator":"IS",
>                                 "values":[
> 									"0h79Q0rcfC1gOPK50yoFCv"
>                                 ]
>                             }
>                         ],
>                         "account":"mosment3d",
>                         "model":"29460f20-8d41-11eb-894b-072add228ef4"
>                     }
>                 ]
>             }
>         }
>     ]
> }'
> ```
> 
> Open the "Broken Date Change" sequence for an example
> 
> image

@mosment3DRepo I've just commited a fix for this (not deployed yet, just for future reference). But  this won't fix your existing sequence as the fix is on the PATCH request.

PR #2539, opened 2021-03-19
ISSUE 2441 - Custom sequences activities API

### LINKED ISSUE #2441: Custom sequences via the API (2) - Create Activities
## Description
Following on from #2440 , once a custom sequence exists, the user should be able to add/edit activities within this sequence

## Goals
- [x] Be able to add activities to a sequence
- [x] Be able to edit an existing activity
- [x] Be able to remove an activity

## Specifications
<details>

### Activity Schema
The current activity schema is designed around the assumption that it's write once, read only after. Which may be difficult to replicate for this purpose. It consists of 2 different dataset:
1. Activity list - a json file stored as a blob that serves as a cache to the front end to display the activity list as a tree structure (note: this example does not have sub tasks, but it is possible)
image 
2. Activities - Each activity is a separate entry within the database, consist of metadata, and resources it the activity is associated with: 
image

keeping up with 2. is simple, as it is just a write to the database. But as 1 is just a blob file, writing it every time there's been an edit to any activity can be exhausting (especially the usually will most likely make multiple edits before a activity list read request is made.) So my proposal is that:
- When we create/update activity in a sequence, we remove the existing blob file for the activity list without generating a new one
- When there is a request to get the activity list, 
    - if blob file exists for this sequence, return the activity list
    - otherwise, generate it on the fly base on the activities within the database, store the newly generated data as a blob file

### Other things to note:
- activities should be removed if there is a request to remove the whole sequence
- Sequence imported from synchro (i.e. `customSequence` is not set to true) should remain read only and the user should not be able to add/edit/remove activities within them

#### To add an activity:
Proposed route: `HTTP POST /:teamspace/:modelId/:sequences/:sequenceId/activities`
Payload:
````js
{
    name: string
    startDate: timestamp
    endDate: timestamp
    resources: object //see "Activities Schema change for details
    parents; []  //array of parent activity ids. can be undefined. I think in our use case this will always be 1 or nothing.
    data: [
        //other metadata for this activity
       {
            key: string
            value: <dataType>
       }
    ]
}
````

- `startDate` and `endDate` doesn't exist in the current schema, but we should store it as we need to generate the activityList on the fly. Convert to ISO Date.
- `_id` should be generated by the backend (uuid)
- add a `sequence_id` property when stored in the database, so we can remove these activities if the user wishes to remove the sequence.
- User should not be allowed to do this action if the associated sequence does not have `customSequence` set to true.

#### To edit an activity
Proposed route: `HTTP PUT  /:teamspace/:modelId/:sequences/:sequenceId/activities/:activityId`
- See POST request, essentially the same.
- should throw error if `activityId` not found.

#### To delete an activity
Proposed route: `HTTP DELETE /:teamspace/:modelId/:sequences/:sequenceId/activities/:activityId`

##### Metadata key/value change
Notice the schema of activities has been changed from `data : { key: value }` to `data: [ {key: keyString, value: value}]`. This makes the keys indexable to future proof for any queries we may want to do. 

##### Resource array change
We want resource array to work more like groups. so instead of taking in an array of shared Ids, we'll allow rules in replacement of `shared_ids`. i.e. the resource array should either be

storing a list of shared_ids:
````js
{
  resources: {
    shared_ids: [array]
  }
  //...
}
````
or rules:
````js
{
  resources: {
    rules: [...] //Same schema as group
  }
}
````

This will require work on the following:
1. A db migration script to modify existing activities
2. bouncer changes so future synchro imports also follows the same schema
3. front end modification to understand this schema

</details>

## Tasks
- [x] Create activity in model
- [x] Add a create activity test 
- [x] Edit activity in model
- [x] Edit activity test 
- [x] Remove activity  in model
- [x] Remove activity test
- [x] Generate file from activities table
- [x] Modify get activities to generate a file and save it if none is there or to return the already available file
- [x] Add tests on get activities taking in consideration the deletion of the activities file when updated
- [x] Create migration script for the new schema 
- [x] Fill in apidocs

## Related Resources
https://coda.io/d/3D-Repo-Features_d5c0Qp25r6f/4D-v3_suIVb#_luYkt

### PR DESCRIPTION
**Pull request by sanmont3drepo**
This fixes #2441

### CODE REVIEW

File: backend/models/sequenceActivities.js [hunk 1]
```diff
@@ -0,0 +1,298 @@
...
+	await Promise.all([
+		db.insert(account,  activityCol(model), activity),
+		FileRef.removeFile(account, model, "activities", sequenceId)
+	]);
+
+	return {...activity, _id, sequenceId: sequenceId};
+};
+
+SequenceActivities.edit = async (account, model, sequenceId, activityId, activity) => {
+	const sequence = await db.findOne(account, model + ".sequences", { _id: utils.stringToUUID(sequenceId)});
```
carmenfan, line 202: call sequenceExists()

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	const query =  {_id: utils.stringToUUID(activityId), sequenceId: utils.stringToUUID(sequenceId)};
+	const {result} = await db.update(account, activityCol(model), query, {$set: activity});
+
+	if (!result.n) {
+		throw responseCodes.ACTIVITY_NOT_FOUND;
+	}
+
+	await FileRef.removeFile(account, model, "activities", sequenceId);
```
carmenfan, line 223: might be good to make this a separate function so it's clearer, like `clearActivityListCache`  or something

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,298 @@
...
+};
+
+SequenceActivities.remove = async (account, model, sequenceId, activityId) => {
+	const sequence = await db.findOne(account, model + ".sequences", { _id: utils.stringToUUID(sequenceId)});
```
carmenfan, line 227: call sequenceExists()

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,298 @@
...
+
+	if (!sequence) {
+		throw responseCodes.SEQUENCE_NOT_FOUND;
+	}
+	const activities = await db.find(account,activityCol(model), {sequenceId: utils.stringToUUID(sequenceId), parent:{$exists: true}} , {parent:1, _id:1});
+
+	const idsToDelete = getDescendantsIds(activities, activityId);
+	idsToDelete.push(utils.stringToUUID(activityId));
+
+	// console.log(idsToDelete.map(utils.uuidToString));
```
carmenfan, line 237: remove?

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,298 @@
...
+
+SequenceActivities.remove = async (account, model, sequenceId, activityId) => {
+	const sequence = await db.findOne(account, model + ".sequences", { _id: utils.stringToUUID(sequenceId)});
+
+	if (!sequence) {
+		throw responseCodes.SEQUENCE_NOT_FOUND;
+	}
+	const activities = await db.find(account,activityCol(model), {sequenceId: utils.stringToUUID(sequenceId), parent:{$exists: true}} , {parent:1, _id:1});
+
+	const idsToDelete = getDescendantsIds(activities, activityId);
```
carmenfan, line 234: I guess this may change anyway since we're going to store children instead of parents.

But I think it's better to just let the db to do the query (even if it means we have to do it multiple times with different parents)

i.e. just do `parent: activityId`

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,298 @@
...
+		throw responseCodes.ACTIVITY_NOT_FOUND;
+	}
+
+	await FileRef.removeFile(account, model, "activities", sequenceId);
+
+	return { _id: activityId};
+};
+
+SequenceActivities.get = async (account, model, sequenceId) => {
+	const sequence = await db.findOne(account, model + ".sequences", { _id: utils.stringToUUID(sequenceId)});
```
carmenfan, line 252: call sequenceExists()

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,298 @@
...
+	if (!sequence) {
+		throw responseCodes.SEQUENCE_NOT_FOUND;
+	}
+
+	let activities = {};
+	try {
+		activities = await FileRef.getSequenceActivitiesFile(account, model, utils.uuidToString(sequenceId));
+	} catch(e) {
+		activities = await createActivitiesTree(account, model, sequenceId);
+		await FileRef.storeFile(account, activityCol(model) + ".ref", account, utils.uuidToString(nodeuuid()), JSON.stringify(activities),  {"_id": sequenceId});
```
carmenfan, line 263: maybe we shouldn't await this one?

whether the cache is stored successfully into the fileshare or not doesn't affect the output of the network request. And we'd be waiting extra time when the user could use the data straight away

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	return activities;
+};
+
+SequenceActivities.createActivities = async (account, model, sequenceId, activitiesTree, overwrite) => {
+	const sequence = await db.findOne(account, model + ".sequences", { _id: utils.stringToUUID(sequenceId)});
```
carmenfan, line 270: sequenceExists

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,298 @@
...
+
+	if (!sequence) {
+		throw responseCodes.SEQUENCE_NOT_FOUND;
+	}
+
+	if (!activityTreeSchema.isValidSync(activitiesTree, { strict: true })) {
+		throw responseCodes.INVALID_ARGUMENTS;
+	}
+
+	fillIds(activitiesTree);
```
carmenfan, line 280: The activityTreeSchema check at L276 and this line means we're traversing the tree twice.

I think it's better if we just rewrite it to traverse the tree once, and do both the schema check and append IDs

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,298 @@
...
+
+	const activitiesList = unrollActivities(activitiesTree, utils.stringToUUID(sequenceId));
```
carmenfan, line 282: this is another traversal, could we do this along with the other 2 functions?

backend/routes/sequence.js
```diff
@@ -388,27 +305,241 @@ router.delete("/revision/:revId/sequences/:sequenceId/legend", middlewares.hasUp
...
+ *    {"key":"Estimated Rate","value":0},
+ *    {"key":"Planned Physical Volume","value":6.6},
+ *    {"key":"Actual Physical Volume","value":0.9},
+ *    {"key":"Remaining Physical Volume","value":5.7},
+ *    {"key":"Budgeted Cost","value":30},
+ *    {"key":"Actual Cost","value":9999.99}
+ *  ]
+ * }
+ */
+router.get("/sequences/:sequenceId/activities/:activityId", middlewares.issue.canView, getSequenceActivityDetail);
```
carmenfan, line 351:
```suggestion
router.get("/sequences/:sequenceId/activities/:activityId", middlewares.hasReadAccessToModel, getSequenceActivityDetail);
```

scripts/db_migrations/v4.16/fixSequencesActivities.py
```diff
@@ -0,0 +1,153 @@
...
+  print("dryRun = False: Commiting to database")
+else:
+  print("dryRun = True: Trial run")
+
+connString = "mongodb://"+ userName + ":" + password +"@"+mongoURL + ":" + mongoPort + "/"
+
+
+def updateActivitiesSchema(db, modelId, sequenceId):
+  activitiesCollName = modelId + ".activities"
+  fileContents = getFile(db, fsDirectory, activitiesCollName, str(sequenceId))
```
carmenfan, line 128: instead of reading the gridFS list, isn't it better to just loop through the activities database?

From what I can tell, we're reading the cache file, then base on that take the ID and get the file and alter the schema.

And we ultimately ditch the activitiyList cache, so we're reading the file only to get the IDs?

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,298 @@
...
+
+		if (activity.subActivities) {
+			fillIds(activity.subActivities, sequenceId);
+		}
+	});
+};
+
+const unrollActivities = (activityTrees, sequenceId, parentId, plainActivities = []) => {
+	activityTrees.forEach((activity) => {
+		const plainActivity = pick(activity,"_id", "name", "startDate", "endDate", "data", "resources");
```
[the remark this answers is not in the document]
sanmont3drepo: Nope, that would just  extract  the fields, the equivalent would be :

```
const {_id, name, startDate, /* etc... */ } = activity;
const plainActivity` = {_id, name, startDate, /* etc... */ };
```

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,298 @@
...
+		}
+	}).map(activity => {
+		const subActivities = getSubActivities(utils.uuidToString(activity._id), activities);
+		return  simplifyActivity(activity, subActivities);
+	});
+};
+
+const createActivitiesTree = async(account, model, sequenceId) => {
+	let activities = {};
+	const foundActivities = await db.find(account, model + ".activities",{sequenceId: utils.stringToUUID(sequenceId)}); // filter by sequenceId
```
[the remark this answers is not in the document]
sanmont3drepo: I have to check which actual fields are being use (for example parent is being use but get rid in simplify activiy). But some sort of simplify activity should be in order because it transforms the plain activity to a tree like activity (using parent and no subactivities).
Maybe I should change the name, because is does not reflect exactly what it does.

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,298 @@
...
+		plainActivity.sequenceId = sequenceId;
+
+		if (parentId) {
+			plainActivity.parent = parentId;
+		}
+
+		plainActivities.push(plainActivity);
+
+		if (activity.subActivities) {
+			unrollActivities(activity.subActivities, sequenceId, activity._id, plainActivities);
```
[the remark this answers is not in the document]
sanmont3drepo: I will try to do so!

backend/models/sequenceActivities.js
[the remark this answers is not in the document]
carmenfan: but I think we def don't need the metadata part right? If that's not used then I think we should at least omit that as that's the bulky part

you can do projection with exclusion instead of inclusion, i.e. `{metadata: 0}`. That should return everything but the fields you specified.

backend/models/sequenceActivities.js
sanmont3drepo: What happens if the time between requests is less than the time it takes to store the file?

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,337 @@
...
+
+			idsDictionary[id] = true;
+			activity._id =  id;
+		}
+
+		const plainActivity = pick(activity,"_id", "name", "startDate", "endDate", "data", "resources");
+		plainActivity.sequenceId = sequenceIdUUID;
+
+		if (parent) {
+			plainActivity.parent = parent._id;
```
carmenfan, line 212: we can just pick it from the activity if we do the rewrite in `traverseActivities`

backend/models/sequenceActivities.js
```diff
@@ -0,0 +1,337 @@
...
+	const activity = await db.findOne(account, activityCol(model), {"_id": utils.stringToUUID(activityId)});
+
+	if (!activity) {
+		throw responseCodes.ACTIVITY_NOT_FOUND;
+	}
+
+	return cleanActivityDetail(activity);
+};
+
+SequenceActivities.create = async (account, model, sequenceId, activity) => {
```
carmenfan, line 237: I don't think this function is used anymore?

carmenfan (changes requested):

carmenfan (approved):

carmenfan:
This is ready to merge, but we need the Sequence side of the work to test the whole feature.

@Charence  please merge this branch into your branch and we will use your branch as the main PR

mosment3DRepo:
### Description

Sequence Upload doesn't create groups properly

### Steps to replicate
- Open this model: [Simple Building](https://issue-2440.dev.3drepo.io/viewer/mosment3ddev/10ff30c0-9874-11eb-8e5c-c9ff1aec95c5)
- The Sequence was uploaded via API
- When you query the sequence you get the following [Sequence Query](https://api3.issue-2440.dev.3drepo.io/api/mosment3ddev/10ff30c0-9874-11eb-8e5c-c9ff1aec95c5/sequences/b5c8df30-9878-11eb-a8eb-2951c087846a)
- Notice that all of the Group IDs are duplicated, and if you query this group it is not found in the model
- The sequence does not play back correctly

<details>
<summary> This is the sequence payload that way uploaded </summary>

````
{
    "name":"Example Sequence",
    "frames":[
        {
            "dateTime":1616400000000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                    "0h79Q0rcfC1gOPK50yoFCv",
                                    "0K5o7g755EZw2RjNI7HcYK",
                                    "0yuGDtpaPCSBT7QB7wvN5I",
                                    "2HBVtaIWv07ud53r01WB6q"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ],
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "2S2omCydz5b9jSgrcLLblk",
                                "0_U7q0Dzj6DfPp4VzMmTUt",
                                "0iMv$JxRL67v6DoyA3RRwz",
                                "1W4yiIKW92qAUdezi70DTY",
                                "00ojKm$5f7luRCAjta0hsu",
                                "0d2LnELub06glJ9mZh2up$",
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
        {
            "dateTime":1616572800000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                    "2S2omCydz5b9jSgrcLLblk",
                                    "0_U7q0Dzj6DfPp4VzMmTUt",
                                    "0iMv$JxRL67v6DoyA3RRwz",
                                    "1W4yiIKW92qAUdezi70DTY"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ],
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "00ojKm$5f7luRCAjta0hsu",
                                "0d2LnELub06glJ9mZh2up$",
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
        {
            "dateTime":1616695200000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                    "2S2omCydz5b9jSgrcLLblk",
                                    "0_U7q0Dzj6DfPp4VzMmTUt",
                                    "0iMv$JxRL67v6DoyA3RRwz",
                                    "1W4yiIKW92qAUdezi70DTY"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ],
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "00ojKm$5f7luRCAjta0hsu",
                                "0d2LnELub06glJ9mZh2up$",
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
        {
            "dateTime":1616781600000,
            "viewpoint":{
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "00ojKm$5f7luRCAjta0hsu",
                                "0d2LnELub06glJ9mZh2up$",
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
        {
            "dateTime":1617091200000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                    "0d2LnELub06glJ9mZh2up$"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ],
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "00ojKm$5f7luRCAjta0hsu",
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
       {
            "dateTime":1617213600000,
            "viewpoint":{
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "00ojKm$5f7luRCAjta0hsu",
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
        {
            "dateTime":1617264000000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ],
                "hidden_group":{
                    "rules":[
                        {
                            "field":"IFC GUID",
                            "operator":"IS",
                            "values":[
                                "00ojKm$5f7luRCAjta0hsu"

                            ]
                        }
                    ],
                    "account":"mosment3d",
                    "model":"29460f20-8d41-11eb-894b-072add228ef4"
                }
            }
        },
        {
            "dateTime":1617696000000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                "37gui3POjDQgmIadjhr$ek",
                                "3XAjSwznb6PfZG9t_wAFXi",
                                "00ojKm$5f7luRCAjta0hsu"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ]
            }
        },
        {
            "dateTime":1617732000000,
            "viewpoint":{
                "override_groups":[
                    {
                        "color":[
                            0,
                            255,
                            0
                        ],
                        "rules":[
                            {
                                "field":"IFC GUID",
                                "operator":"IS",
                                "values":[
                                "00ojKm$5f7luRCAjta0hsu"
                                ]
                            }
                        ],
                        "account":"mosment3d",
                        "model":"29460f20-8d41-11eb-894b-072add228ef4"
                    }
                ]
            }
        },
        {
            "dateTime":1617818400000,
            "viewpoint":{}
            
        }          
    ]
}
````


</details>

### Current Behaviour
- Sequence does not play back possibly because the groups are not being displayed correctly
### Expected Behaviour
- The sequence frames should load according to the JSON request

carmenfan:
closing this as it's been merged into #2440

PR #2544, opened 2021-03-19
Issue 2541 fix reporting

### LINKED ISSUE #2541: genStats is running out of memory
### Description
GenStats used to run with the heap size set to ~8GB.
image

With the increase in usage, it is now struggling and even at 12GB it's failing.

carmenfan:
This fixes #2541
#### Description
Skip any issues that has a timestamp that is clearly invalid.


keifgwinn:
Need to port this to statsGenElastic as this has also been failing since November 

--
Keif

Charence (approved):

PR #2555, opened 2021-04-08
Issue 2532

### LINKED ISSUE #2532: enforce character limit in the field "residual risk" on the backend
### Description
The front end only allows 220 characters on `residual risk`.  However the backend does not enforce this.

This has lead to problems where users upload their risks via another service (e.g. Excel VBA script) with `residual risk` longer than 220 characters. When they view it on 3D repo they're struggling to update the risk as the form is in an error state.

We should increase this character limit to accommodate the existing risks that users has uploaded, and add character length check on any new risks to avoid the problem from appearing again.

### Steps to replicate
- Create a risk somewhere
- through the API (or just overwrite it in the db), update residual risk to character length > 220
- go to the risk on 3drepo, try to update any of the properties

### Current Behaviour
- Nothing happens, the risk does not update

### Expected Behaviour
- The risk should update.

Charence:
This fixes #2532


frontend/routes/viewerGui/components/risks/components/riskDetails/riskDetailsForm.component.tsx
```diff
@@ -80,9 +80,9 @@ interface IState {
 }
 
 export const RiskSchema = Yup.object().shape({
-	description: Yup.string().max(220, VALIDATIONS_MESSAGES.TOO_LONG_STRING),
-	mitigation_desc: Yup.string().max(220, VALIDATIONS_MESSAGES.TOO_LONG_STRING),
-	residual_risk: Yup.string().max(220, VALIDATIONS_MESSAGES.TOO_LONG_STRING)
+	description: Yup.string().max(660, VALIDATIONS_MESSAGES.TOO_LONG_STRING),
```
carmenfan, line 83: Can we use a variable constant here?

carmenfan: Also it looks like we're missing the change on issues?

carmenfan (changes requested):

carmenfan (approved):

PR #2564, opened 2021-04-14
ISSUE #1979 - Add copy to clipboard functionality

### LINKED ISSUE #1979: One click copy value to clipboard
We have a number of data fields available in the UI that could be enhanced with a copy to clipboard button for simple mouse over and click, rather than triple-click and keyboard shortcuts. This would make usage easier for people who need to access the GUIDs or metadata in models for use elsewhere.

Here is an example of the implementation in Azure portal.
[image](https://user-images.githubusercontent.com/23369849/79563236-4ba02600-80a4-11ea-8278-ed90cf59c977.png)

### Goals

- [ ] Add one click copy to clipboard

### Tasks

- [ ] Identify commonly copied text fields
- [ ] Add one click copy to clipboard

m1chalkubiak:
This fixes #1979

#### Description
Functionality was added to generic textField component used in many places in the application.
It can be enabled by adding `withCopyButton` property. However, it is worth mentioning that this was only tested for non-editable fields. Probably in the case of editable text fields, there would have to do some additional style adjustment.

It can be checked on modelID and apiKey (on user profile) fields.

### DIFF
--- a/frontend/routes/components/textField/textField.component.tsx
+++ b/frontend/routes/components/textField/textField.component.tsx
@@ -54,6 +54,7 @@ interface IProps extends StandardTextFieldProps {
 	withCopyButton?: boolean;
 	onCancel?: () => void;
 	showSnackbar?: (text: string) => void;
+	value?: string;
 }
 
 interface IState {
@@ -107,7 +108,7 @@ export class TextField extends React.PureComponent<IProps, IState> {
 	}
 
 	get fieldValue() {
-		return this.props.requiredConfirm ? this.state.currentValue : this.props.value as string;
+		return this.props.requiredConfirm ? this.state.currentValue : this.props.value;
 	}
 
 	private checkIfGotLongContent = () => {
--- a/frontend/routes/components/textField/textField.component.tsx
+++ b/frontend/routes/components/textField/textField.component.tsx
@@ -21,14 +21,17 @@ import { StandardTextFieldProps } from '@material-ui/core/TextField';
 import CancelIcon from '@material-ui/icons/Cancel';
 import EditIcon from '@material-ui/icons/Edit';
 import SaveIcon from '@material-ui/icons/Save';
+import copy from 'copy-to-clipboard';
 import { Field, Formik } from 'formik';
 
+import CopyIcon from '@material-ui/icons/FileCopy';
 import { ENTER_KEY } from '../../../constants/keys';
 import { renderWhenTrue } from '../../../helpers/rendering';
 import { ExpandAction } from '../../viewerGui/components/risks/components/riskDetails/riskDetails.styles';
 import {
 	ActionsLine,
 	Container,
+	CopyButton,
 	FieldLabel,
 	FieldWrapper,
 	MutableActionsLine,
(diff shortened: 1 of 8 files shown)


frontend/routes/components/textField/textField.component.tsx
```diff
@@ -102,7 +107,7 @@ export class TextField extends React.PureComponent<IProps, IState> {
 	}
 
 	get fieldValue() {
-		return this.props.requiredConfirm ? this.state.currentValue : this.props.value;
+		return this.props.requiredConfirm ? this.state.currentValue : this.props.value as string;
```
carmenfan, line 110: Is the string cast this necessary? I'm a bit dubious since we never had it previously.

frontend/routes/components/textField/textField.styles.ts
```diff
@@ -15,14 +15,28 @@
...
+
+const containerNonEditModeStyles = css`
+	align-items: flex-end;
+`;
+
 export const Container = styled.div`
 	position: relative;
-` as any;
+	display: flex;
+	align-items: baseline;
```
carmenfan, line 37: This is always overwritten by the next line so i think it's redundant?

frontend/routes/components/textField/textField.styles.ts
```diff
@@ -15,14 +15,28 @@
...
+import { ContainedButton } from '../../viewerGui/components/containedButton/containedButton.component';
 import { LinkableField } from '../linkableField/linkableField.component';
 import { MarkdownField } from '../markdownField/markdownField.component';
 
+const containerEditModeStyles = css`
+	align-items: baseline;
+`;
+
+const containerNonEditModeStyles = css`
+	align-items: flex-end;
```
carmenfan, line 31: do we need the flex-end at all? I played with this property on a few text fields and I'm not really seeing the significance.. Admittedly I'm also not too familiar with CSS 😆

carmenfan (commented):
@m1chalkubiak no request for change at the moment, just a few questions that maybe just need answering to clear up (or may lead to code change). Let me know your thoughts

frontend/routes/components/textField/textField.component.tsx
m1chalkubiak: This is because the `copy` function from` copy-to-clipboard` expects this value to be a string, but it is not certain because props `value` declared in BaseTextFieldProps interface is an unknown type.
However, we can redeclare it in our own IProps interface as an optional string, and it seems to do the trick, I don't think we should expect anything else than string here.

frontend/routes/components/textField/textField.styles.ts
m1chalkubiak: This is mainly about this case with the new copy button, and the way that buttons are disabled... because some times button are disabled by passing `disabled` property, and some times by using:
```js
inputProps={{
 readOnly: true
}}
```
One affects some style and display changes in general, the second one is just setting `readonly` attribute, which affects only in a functional way to the rendered input.

### disabled case
With `flex-end`
Zrzut ekranu 2021-04-20 o 12 21 43

With `baseline;`
Zrzut ekranu 2021-04-20 o 12 22 51

### readonly case
With `flex-end`
Zrzut ekranu 2021-04-20 o 12 52 29

With `baseline`
Zrzut ekranu 2021-04-20 o 12 51 40

carmenfan (approved):

PR #2569, opened 2021-04-21
ISSUE #2549 - Fix kanban board styling

### LINKED ISSUE #2549: Kanban Board control sliders hide behind the UI when you use Filters.
### Description

Currently, you are able to scroll up/down and left/right in the Kanban Board. Frequently our users get stuck when they start filtering the Kanban Board. Even though the are still able to slide across, it takes time for them to realise that it is the case and they need to simply scroll all the way down. It's not a User friend or intuative UI. These control sliders should be fixed regardless of whether I am using filters or not. Other features should not put me in a difficult situation where I need to spend additional time figuring out what's happening.

### Goals

- [x] Fix the Kanban Board sliders

### Related Resources

Kanban Board

abyszewski:
This fixes #2549

#### Description
Fix board height when filters, search are opened


.idea/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Default ignored files
```
m1chalkubiak: ?

frontend/routes/board/board.component.tsx
```diff
@@ -34,7 +34,8 @@ import { filtersValuesMap as issuesFilters, getHeaderMenuItems as getIssueMenuIt
 import { renderWhenTrue } from '../../helpers/rendering';
 import { filtersValuesMap as risksFilters, getHeaderMenuItems as getRisksMenuItems  } from '../../helpers/risks';
 import {
-	ISSUE_FILTER_PROPS, ISSUE_FILTER_VALUES, RISK_FILTER_PROPS, RISK_FILTER_VALUES
+	FILTERS_COMPONENT_HEIGHT,
```
m1chalkubiak, line 37: Please fix formatting

frontend/routes/board/board.styles.ts
```diff
@@ -32,8 +32,8 @@ export const Container = styled.div`
 	flex-direction: column;
 `;
 
-export const BoardContainer = styled.div`
-	height: calc(100% - 50px);
+export const BoardContainer = styled.div<{offset: number}>`
+	height: ${({offset}) => offset ? `calc(100% - ${50 + offset}px)` : 'calc(100% - 50px)'};
```
m1chalkubiak, line 36: Please move those additional styles to separate variables

frontend/routes/components/filterPanel/filterPanel.component.tsx
```diff
@@ -102,6 +102,7 @@ interface IProps {
 	left?: boolean;
 	onChange: (selectedFilters) => void;
 	onDataTypeChange?: (selectedDataTypes) => void;
+	setFiltersOpen: Dispatch<SetStateAction<boolean>>;
```
m1chalkubiak, line 105: Did You check all places were this component is used ?

frontend/modules/board/board.constants.ts
```diff
@@ -88,3 +88,6 @@ export const RISK_FILTER_PROPS = {
 
 export const ISSUE_FILTER_VALUES = values(ISSUE_FILTER_PROPS);
 export const RISK_FILTER_VALUES = values(RISK_FILTER_PROPS);
+
+export const SEARCH_INPUT_HEIGHT = 50;
+export const FILTERS_COMPONENT_HEIGHT = 50;
```
m1chalkubiak, line 93: If fix need knowing of element height, we should depend on real values

m1chalkubiak (changes requested):

carmenfan:
image


@abyszewski  @m1chalkubiak  can we make sure we're not doing force pushes? It's probably fine for now since no one else is working on this branch at the moment, but It's pretty bad practice 😆

m1chalkubiak:
Sure, @carmenfan, it's Antek's first task in this project and We still agree on the rules of work here. Of course, also I suggested internally to Antek to no making force pushes.

carmenfan:
@keifgwinn it's struggling to build https://github.com/3drepo/3drepo.io/runs/2420264425

Seems to be related to the change you made yday. Any ideas why/what needs to be done??

BDavydov:
@mosment3DRepo Looks good to me, does what expected.
Sometimes you get an odd jitter or two, scrolling Up & Down quickly, but it's barely noticeable.

sanmont3drepo:
I think the approach ussing just css in this branch is simpler:
https://github.com/3drepo/3drepo.io/commits/ISSUE_2549b

m1chalkubiak:
> I think the approach using only css in this branch is simpler:
> https://github.com/3drepo/3drepo.io/commits/ISSUE_2549b

Hi @sanmont3drepo I can't argue with fact that it is simpler, but it changes diametrically what and how is scrollable on board.
I'm not saying it's not working or looking bad but it's just different from the current implementation.

Can you please build it independently for your team internally and compare it with the current version on staging?

carmenfan:
@sanmont3drepo  can you make PR for that branch and trigger a build?

carmenfan:
@m1chalkubiak 

> Can you please build it independently for your team internally and compare it with the current version on staging?

Are there situations where you suspect it may not work as before? whilst we can try to thoroughly test it , it would be good to make sure we cover any suspected case that are likely to break

sanmont3drepo:
> > I think the approach using only css in this branch is simpler:
> > https://github.com/3drepo/3drepo.io/commits/ISSUE_2549b
> 
> Hi @sanmont3drepo I can't argue with fact that it is simpler, but it changes diametrically what and how is scrollable on board.
> I'm not saying it's not working or looking bad but it's just different from the current implementation.
> 
> Can you please build it independently for your team internally and compare it with the current version on staging?

https://github.com/3drepo/3drepo.io/pull/2592

m1chalkubiak:
> @m1chalkubiak
> 
> > Can you please build it independently for your team internally and compare it with the current version on staging?
> 
> Are there situations where you suspect it may not work as before? whilst we can try to thoroughly test it , it would be good to make sure we cover any suspected case that are likely to break (i.e. know what we're testing!)

@carmenfan 
I'm attaching screen motion captures, as this probably be the best way to compare those two solutions.

The current way the scroll works inside the kanban board
CPT2105051334-995x508

The way the scroll works with the changes proposed by Santiago:
CPT2105051415-995x508

I will repeat it again to make it clear, I am not saying that one solution is good and the other is bad. They are just different, it is up to you to decide which way this board works for you.
Of course, it shows the extreme situation on a low height monitor and with a large number of selected filters only to illustrate what the difference is.

sanmont3drepo:
> > @m1chalkubiak
> > > Can you please build it independently for your team internally and compare it with the current version on staging?
> > 
> > 
> > Are there situations where you suspect it may not work as before? whilst we can try to thoroughly test it , it would be good to make sure we cover any suspected case that are likely to break (i.e. know what we're testing!)
> 
> @carmenfan
> I'm attaching screen motion captures, as this probably be the best way to compare those two solutions.
> 
> The current way the scroll works inside the kanban board
> CPT2105051334-995x508
> 
> The way the scroll works with the changes proposed by Santiago:
> CPT2105051415-995x508
> 
> I will repeat it again to make it clear, I am not saying that one solution is good and the other is bad. They are just different, it is up to you to decide which way this board works for you.
> Of course, it shows the extreme situation on a low height monitor and with a large number of selected filters only to illustrate what the difference is.

Hi Michal. I think that you compared 2549b (or 2549) to staging unwillingly. I just copied the behaviour of what you guys did but in css.
This is the same test in 2549
branch2549

m1chalkubiak:
> 
> Hi Michal. I think that you compared 2549b (or 2549) to staging unwillingly. I just copied the behaviour of what you guys did but in css.
> This is the same test with your code from branch ISSUE_2549
> branch2549

Ok, I see it now, apologize Santiago for this confusion, there were some missing commits on my local that I didn't pull. Now I see both solutions change this behavior in the same way. I was referring to commit 80ea6aa which was working better (the same way like on staging). 

I think the main question is whether this change is ok for all of You, if so then we should do it according to Your suggestion.

carmenfan:
closing in favour of #2592

PR #2570, opened 2021-04-21
Issue 2562 frame stepping issues on sequences

### LINKED ISSUE #2562: 4D Sequence playback based on frames can crash
### Description

When playing back a 4D sequence based on frames it can suddenly stop displaying the model. Also loading times between frames is slow (no prefetching?)

- This requires a hotfix to production: no (fix for next release) 

### Steps to replicate
- Open [this model](https://staging.dev.3drepo.io/viewer/andrew/e11a4930-9c38-11eb-86ad-4bacbc759634) (you may need to clear your cache if you've already opened the model before)
- Open the sequence
- Select playback via frames
- Play through the sequence
- Notice the first few frames don't load (this may be a separate issue)
- Around 27/9/2021 the model disappears (see video below)
- Going to the same date in day playback does not produce the error

Sometimes the following icon appears in the top left after the model disappears
sad face

### Current Behaviour
- The sequence playback seems to fail at some point

### Expected Behaviour
- Playback should play smoothly

[video on dropbox](https://www.dropbox.com/home/3D%20Repo%20Team%20Folder/Support%20Team/GitHub%20Issues/Images?preview=2021-04-14+11-23-40.mp4)

carmenfan:
This fixes #2562

#### Description
- Fixed frames prefetching: 
    - Changed getSelectedFrame to return an index as well for easy access to the next frame
- Fixed indexeddb caching on states
   - database is created by DataCacheService itself instead of viewerGui (that seemed very high coupling and breaks MVC)
   - There's a problem with getValues not returning even when the promise has been resolved, so had to rewrite a yield into a promise.

#### Test cases
- the test case mentioned in the issue should now be resolved

### DIFF
--- a/frontend/modules/sequences/sequences.helper.ts
+++ b/frontend/modules/sequences/sequences.helper.ts
@@ -22,7 +22,7 @@ import { selectGetMeshesByIds, selectGetNodesIdsFromSharedIds } from '../tree';
 
 export const getSelectedFrame = (frames, endingDate) => {
 	if (!frames.length) {
-		return null;
+		return {};
 	}
 
 	let leftMargin = 0;
@@ -39,14 +39,14 @@ export const getSelectedFrame = (frames, endingDate) => {
 	}
 
 	if (frames[rightMargin].dateTime <= endingDate) {
-		return frames[rightMargin];
+		return { frame: frames[rightMargin], index: rightMargin };
 	}
 
 	if ( frames[leftMargin].dateTime <= endingDate) {
-		return frames[leftMargin];
+		return { frame: frames[leftMargin], index: leftMargin };
 	}
 
-	return null;
+	return {};
 };
 
 export const getDateByStep = (date, stepScale, step) => {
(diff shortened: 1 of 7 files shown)


sanmont3drepo (approved):

frontend/modules/sequences/sequences.helper.ts
```diff
@@ -39,14 +39,14 @@ export const getSelectedFrame = (frames, endingDate) => {
 	}
 
 	if (frames[rightMargin].dateTime <= endingDate) {
-		return frames[rightMargin];
+		return { frame: frames[rightMargin], index: rightMargin };
```
sanmont3drepo, line 42: What if we rename this function to getSelectedFrameIndex?

replace this one with
``export const getSelectedFrame = (frames, endingDate) => frames[getSelectedFrameIndex(frames, endingDate)] ;``

And use ``getSelectedFrameIndex(....)`` in ``selectNextKeyFramesDates`` ?

PR #2572, opened 2021-04-21
ISSUE #2557 - Split requests if the list of groups is too long to proceed in deleting of groups

### LINKED ISSUE #2557: Can't delete a long list of groups. Throws a Network Error message
### Description

I have a long list of groups in my model. I have decided to clean it up and utilise the Delete all functionality. When I press on the Delete all button, I get a Network error message every single time.

- This is happening in: prod

### Steps to replicate
- Open this model: [structure](https://www.3drepo.io/viewer/Bogdan_D/6dba85c6-58d5-4407-9c48-0a7ed744a1eb)
- Open the Groups card
- Click on the Delete all button

### Current Behaviour
- Fails. Network Error 

### Expected Behaviour
- Delete all of the Groups

image

m1chalkubiak:
This fixes #2557

### DIFF
--- a/frontend/constants/requests.ts
+++ b/frontend/constants/requests.ts
@@ -1,4 +1,4 @@
-export const MAX_URL_LENGTH = 140;
+export const MAX_URL_LENGTH = 2000;
 
 const normalizeString = (text): string => text.replace(/^,/, '');
 
--- a/frontend/constants/requests.ts
+++ b/frontend/constants/requests.ts
@@ -1,15 +1,28 @@
-export const MAX_URL_LENGTH = 2000;
+export const MAX_URL_LENGTH = 240;
+export const IDS_ARGUMENT_INDICATION = '?ids=';
 
 const normalizeString = (text): string => text.replace(/^,/, '');
 
-export const splitString = (text: string, by: string, limit: number) => {
-	text = normalizeString(text);
+const getPartLength = (text: string, by: string = ',') => text.split(by)[0]?.length || 0;
+
+const getSplitPoint = (text: string, limit: number, by: string = ',') => {
 	const splitPoint = text.indexOf(by, limit);
+
+	if (splitPoint > -1) { return splitPoint; }
+
+	const partLength = getPartLength(text);
+
+	return partLength || getSplitPoint(text, limit - partLength, by);
+};
+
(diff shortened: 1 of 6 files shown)


carmenfan (changes requested):
@m1chalkubiak  I don't think this is the right approach..

This is a general problem of API requests and not specific to groups (we've ran into this in issues/risks before). So to add all the logic on the group saga seems to be over complicating the group saga with a requirement of HTTP requests.

I think we can just simply fix it on the API service, impose a max group per request, if it's bigger, send another request.

m1chalkubiak:
@carmenfan 
As suggested, I changed my approach to solving this issue and transferred the auxiliary logic to the API service. I have tried to analyze the requests that occur in the application and it seems to me that the only ones that have this problem are ones responsible for deletion, where we provide resource references as the value of a specific attribute. From what I checked, this occurs in the case of groups and risks removal, however, regarding the second one probably not being used anywhere in the application.

I tried to test it extensively before pushing it out, but I would feel a lot safer with this solution if we had tests written for it. Is there an addition of unit tests on the frontend part of the application somewhere in the nearest app development plans? I would be very happy if I could help with it then and start adding tests to, for example, this type of auxiliary function that from time to time.

carmenfan:
@m1chalkubiak  I still think you over complicated it 😆 

I can see you're attempting to generalise it so it's useful for every request, but on the other hand it's making a lot of assumptions.

- The usage of `IDS_ARGUMENT_INDICATION` means you're assuming the request the request will always start with `ids`
- The way the text is split you're assuming there's no other query string after ?id=
- You're assuming every value within id is the same length as the first one

Whilst i'm sure it works right now, as soon as we add anything that can vary, it will break and probably become one of those bugs that is not even obvious why it's breaking until the person delve very deep. If you're going to generalise it, I think you need to make it super generalised (i.e. make none of the assumptions above.)

Whilst it's probably this is probably the best approach that if you've removed all the assumptions we will never run into this problem, I'm happy with just a simple solution where you just check the size of group [here ](https://github.com/3drepo/3drepo.io/blob/master/frontend/services/api/groups.ts#L83)and call api.delete multiple times if need be... Because this is the place where you can make those assumptions that you have made, as we know group ID is of type UUID and they will always be fix length.

carmenfan (approved):

Issue #2573, opened 2021-04-22
Azure Pipeline Values Override on a Branch


keifgwinn:
### Description

Occasionally we want to override the default values in the Azure pipelines in order to support custom bouncers or other such features. We want to be able to add additional -set commands to the helm package based on the branch.

### Goals

- [x] Defaults can be overridden on a per branch basis.
- [x] Some way to ensure that these don't always make their way up to staging, could require a review of pull requests. https://github.com/3drepo/3drepo.io/pull/2576

### Tasks

- [x] Work out how to add commands
- [x] Test
- [x] Discussion around pull requests

keifgwinn:
https://github.com/3drepo/3drepo.io/pull/2576 covers ensuring it doesn't make changes further up the tree

carmenfan:
@keifgwinn  this can close?

keifgwinn:
Yes

--
Keif

PR #2574, opened 2021-04-22
ISSUE #2573 - Testing custom helm attributes per branch

keifgwinn:
This tests #2573

#### Description
<!-- Add a high level description of the changes made (possibly quite similar to task list if it was a feature) 
e.g.
- I've added a variable to be able to add additional overrides to the helm chart so that we can fix certain values.

#### Test cases

Ensure that the task completes, and the new values are sticky.


carmenfan:
closing as it doesn't need to be merged in

PR #2577, opened 2021-04-26
Issue 2571 - Endpoint to request textures by node UID

### LINKED ISSUE #2571: Add API to return Texture Maps
### Description
Add an API to return a texture by UID to support [textured SRCs for Unreal](https://github.com/3drepo/3drepoUnreal/issues/31)

### Goals
- [x] A new API (e.g. :model/textures/:uid) will return a texture map.
- [x] The API will support HTTP, returning a binary stream with the correct mime type
- [x] The API will work with the Unreal WebRequestManager.

### Tasks
- [x] Add a new route for the API
- [x] Add a new helper class similar to mesh
- [x] Ensure different texture types are supported correctly

### Related Resources
The API should return textures for any textured model. Minimal working textured models for development are available here: https://www.dropbox.com/sh/m3206ivyefukuz9/AADXiuDz8D3gOGyGzBwKduH6a?dl=0

sebjf:
This fixes #2571

#### Description
This PR adds a route/API to request textures from model (repo scene) texture nodes by UID. It is intended to support https://github.com/3drepo/3drepoUnreal/issues/31, though is independent of it. The API returns an HTTP response with the content set to the data member of the texture node, and the mime type set based on the extension member of the node.

#### Test cases
There is no automated test; the API should return compliant HTTP responses for any texture in any model.


backend/VERSION.json
```diff
@@ -1,4 +1,4 @@
-{ "VERSION" : "4.15.0",
+{ "VERSION" : "4.14.5",
```
carmenfan, line 1: ? it doesn't look like you based off master, but the version number is altered?

carmenfan (changes requested):

backend/VERSION.json
sebjf: Hmm I don't know what to make of that! My last commit integrated changes from the branch holding the 4.14.5 tag made at the end of March. I always pull the latest version before making the PR, so the only thing I can think of is I accidentally merged master instead of staging!? I will revert whatever it was either way and fix it...

sebjf: @carmenfan OK it should be fixed now.

carmenfan (approved):
Looks fine, waiting on verification that it works as expected from the unreal PR

PR #2579, opened 2021-04-27
ISSUE #2548 - Handle unauthorized model access

### LINKED ISSUE #2548: return user to the teamspace view screen if they do not have permissions to access a model
### Description
If the user tries to enter a model where they have no permissions, there should be a single warning message indicating that they do not have access rights, then redirect them back to the teamspace view.

(this used to work, once upon a time...)

### Steps to replicate
- open a link to a model which you do not have access

### Current Behaviour
- you will get spammed by loads of network failures, eventually seeing a blank viewer window

### Expected Behaviour
- there should be 1 single message to tell you you have insufficient permission to access the model, and upon clicking on confirm, redirect you to the teamspace listing.

abyszewski:
This fixes #2548

#### Description
Show single dialog and redirect the user to team space after accessing not permitted model


m1chalkubiak:
@carmenfan In this case, we would need your decision as to whether it is better to use the generic errors dialogue for this case or should we create a new custom one about the lack of access to the model with a button saying "Back to teamspace"? In my option, this generic one does not fit in this case, especially regarding what will happen after clicking the button.
Zrzut ekranu 2021-04-27 o 09 05 03

frontend/routes/app/app.component.tsx
```diff
@@ -23,6 +23,7 @@ import { PUBLIC_ROUTES, ROUTES } from '../../constants/routes';
 import { getCookie, setCookie } from '../../helpers/cookies';
 import { renderWhenTrue } from '../../helpers/rendering';
 import { WebGLChecker } from '../../helpers/webglChecker';
+import {history} from '../../modules/store';
```
m1chalkubiak, line 26: Please fix formatting

frontend/modules/dialog/dialog.redux.ts
```diff
@@ -27,6 +27,7 @@ export interface IDialogConfig {
 	template?: () => JSX.Element;
 	content?: string;
 	data?: any;
+	permissionDenied?: boolean;
```
m1chalkubiak, line 30: Do we need this `prop` ?

m1chalkubiak:
> @carmenfan In this case, we would need your decision as to whether it is better to use the generic errors dialogue for this case or should we create a new custom one about the lack of access to the model with a button saying "Back to teamspace"? In my option, this generic one does not fit in this case, especially regarding what will happen after clicking the button.
> Zrzut ekranu 2021-04-27 o 09 05 03

@carmenfan We would be grateful if you would be able to take a look at this and decide on what would be the best approach in Your opinion here.

carmenfan:
> > @carmenfan In this case, we would need your decision as to whether it is better to use the generic errors dialogue for this case or should we create a new custom one about the lack of access to the model with a button saying "Back to teamspace"? In my option, this generic one does not fit in this case, especially regarding what will happen after clicking the button.
> > Zrzut ekranu 2021-04-27 o 09 05 03
> 
> @carmenfan We would be grateful if you would be able to take a look at this and decide on what would be the best approach in Your opinion here.

Yea, create a new one that does what it need.

Note that it shouldn't be used for every Not authorized network request errors, it should really only be used for this very special case.

frontend/modules/viewerGui/viewerGui.sagas.ts
```diff
@@ -51,6 +51,19 @@ import {
 } from './viewerGui.selectors';
 
 function* fetchData({ teamspace, model }) {
+	try {
```
carmenfan, line 54: use fetchSettings. Fetching model settings as the first action makes more sense than metadata keys.

frontend/modules/viewerGui/viewerGui.sagas.ts
```diff
@@ -51,6 +51,19 @@ import {
...
+		const {data: metaKeys} = yield API.getMetaKeys(teamspace, model);
+		yield put(ModelActions.fetchMetaKeysSuccess(metaKeys));
+	} catch (e) {
+		if (e?.response.status === 401) {
+			yield put(DialogActions.showEndpointErrorDialog('access', 'model', e));
```
carmenfan, line 59: Actually if model settings failed, there's pretty much no point carrying on, so i don't think there's a need to reroute.

If we can make the dialog to be generic to just say "There's a problem accessing the model" , then show the network request error, and a button to redirect to teamspace list, that should do.

carmenfan (changes requested):

frontend/modules/viewerGui/viewerGui.sagas.ts
m1chalkubiak: Fetching model settings is now used instead of metadata

frontend/modules/viewerGui/viewerGui.sagas.ts
m1chalkubiak: This comment is marked as `Outdated` as I change code here, altho the general logic stays the same, this is why I'm not sure if it fits right now with Your comment or not. In general, if this error condition is met, no further action is continued.

carmenfan: the comment is still applicable. I'll mark this as resolve and pin point it in the new version

frontend/modules/viewerGui/viewerGui.sagas.ts
```diff
@@ -51,6 +51,20 @@ import {
...
 
 function* fetchData({ teamspace, model }) {
+	try {
+		yield put(ModelActions.setPendingState(true));
+		const { data: settings } = yield API.getModelSettings(teamspace, model);
+
+		yield put(ModelActions.fetchSettingsSuccess(settings));
+		yield put(ModelActions.setPendingState(false));
+	} catch (e) {
+			if (e.response?.status === 401) {
```
carmenfan, line 61: I think if there's been an error fetching model settings, we want to show a dialog to redirect to teamspace regardless.

Can we make it so that it's not a `showUnauthorizedModelAccessDialog`, but a `redirectToTeamspaceDialog`

and regardless of what type of error it is here, we show that dialog.

We can pretty much keep the format of endpoint error dialog but with a small change, something like this:

image

frontend/modules/dialog/dialog.redux.ts
```diff
@@ -161,6 +164,18 @@ const showLoggedOutDialog = (state = INITIAL_STATE, action) => {
 	return showDialog(state, { config });
 };
 
+const showUnauthorizedModelAccessDialog = (state = INITIAL_STATE, action) => {
+	const config = {
+		title: 'Unauthorized Access',
+		content: 'Insufficient permission to access the model',
+		template: Dialogs.UnauthorizedAccessDialog,
+		onCancel: () => push(ROUTES.TEAMSPACES),
+		onConfirm: () => {}
```
carmenfan, line 173: should it not also do push(ROUTES.TEAMSPACES) here? so we don't have the redirect logic on both this place and the component?

carmenfan (changes requested):

frontend/modules/dialog/dialog.redux.ts
m1chalkubiak: Yes, You are right, it makes a lot more sense, thanks for the suggestion.

frontend/modules/viewerGui/viewerGui.sagas.ts
m1chalkubiak: Done

carmenfan (approved):

PR #2587, opened 2021-04-29
ISSUE 2586 - E-mail validation regex

### LINKED ISSUE #2586: Change email regex to support TLD > 5 chars
### Description

A user's email address xxxxxxxx@yyyy.london was rejected at the login screen, further investigation show that the regex is currently limited to 5 character TLDs

### Current Behaviour
- TLD of .london is rejected using the regex: `define("EMAIL_REGEXP", /^([a-zA-Z0-9_\-.]+)@([a-zA-Z0-9_\-.]+)\.([a-zA-Z]{2,5})$/);`

### Expected Behaviour
- Extended TLDs should be supported as they are becoming more common
- https://emailregex.com/ offers a solution which seems to resolve this issue

Charence:
This fixes #2586

#### Description
Adopted yup e-mail validation on backend for login authentication. This is for consistency with frontend and also to accept a wider range of valid e-mail addresses.

#### Test cases
SIgn up and login using e-mail (in particular, with uncommon TLDs such as .london)


backend/models/user.js
```diff
@@ -153,7 +153,7 @@ User.authenticate =  async function (logger, username, password) {
 
 	let user = null;
 
-	if (C.EMAIL_REGEXP.test(username)) { // if the submited username is the email
+	if (await utils.isEmail(username)) { // if the submited username is the email
```
carmenfan, line 156: if we're using yup, it's better to just have the schema validation of the request instead of using  yup as part of the utils.

carmenfan (changes requested):

carmenfan (changes requested):
looks fine, but auto test broke

carmenfan (approved):

PR #2591, opened 2021-04-30
ISSUE #2463 - Handle deselecting items on issues, risks & groups panel

### LINKED ISSUE #2463: Clicking white space in the cards to deselect Issues/Risks/Groups
### Description

At the moment is is tricky to deselect an Issue/Risk/Group which means the pins stay visible even when all pins are turned off, or it is not the correct time in the sequence.

When the user clicks the whitespace it should deselect the current selection

### Goals

- [ ] Allow users to deselect the currently selected Issue/Risk/View

### Related Resources
issues area crop

m1chalkubiak:
This fixes #2463

### DIFF
--- a/frontend/modules/issues/issues.redux.ts
+++ b/frontend/modules/issues/issues.redux.ts
@@ -38,6 +38,7 @@ export const { Types: IssuesTypes, Creators: IssuesActions } = createActions({
 	showDetails: ['revision', 'issueId'],
 	closeDetails: [],
 	setActiveIssue: ['issue', 'revision', 'ignoreViewer'],
+	resetComponentState: [],
 	togglePendingState: ['isPending'],
 	toggleDetailsPendingState: ['isPending'],
 	togglePostCommentPendingState: ['isPending'],
@@ -105,6 +106,10 @@ const updateIssueProps = (issuesMap, issueId, props = {}) => {
 			...props
 		}
 	};
+};
+
+const resetComponentState = (state = INITIAL_STATE) => {
+	return { ...state, componentState: INITIAL_STATE.componentState };
 };
 
 export const togglePendingState = (state = INITIAL_STATE, { isPending }) => ({ ...state, isPending });
@@ -281,6 +286,7 @@ export const reducer = createReducer(INITIAL_STATE, {
 	[IssuesTypes.FETCH_ISSUE_SUCCESS]: fetchIssueSuccess,
 	[IssuesTypes.FETCH_ISSUE_FAILURE]: fetchIssueFailure,
 	[IssuesTypes.SET_COMPONENT_STATE]: setComponentState,
+	[IssuesTypes.RESET_COMPONENT_STATE]: resetComponentState,
 	[IssuesTypes.SAVE_ISSUE_SUCCESS]: saveIssueSuccess,
 	[IssuesTypes.TOGGLE_PENDING_STATE]: togglePendingState,
 	[IssuesTypes.TOGGLE_DETAILS_PENDING_STATE]: toggleDetailsPendingState,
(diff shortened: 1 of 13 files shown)


frontend/modules/issues/issues.redux.ts
```diff
@@ -107,6 +108,10 @@ const updateIssueProps = (issuesMap, issueId, props = {}) => {
 	};
 };
 
+const resetComponentState = (state = INITIAL_STATE) => {
+	return { ...state, componentState: INITIAL_STATE.componentState };
```
sanmont3drepo, line 112: This clears everything in the state in the component. That means that the filters , showpins  or any criteria changed regarding order are cleared.
If this is what we want is ok, otherwise we should use `setActiveIssue()/setActiveRisk()`  and fix it to accept null issues and deselect the issue/risk.

frontend/routes/viewerGui/components/groups/groups.component.tsx
```diff
@@ -135,6 +135,17 @@ export class Groups extends React.PureComponent<IProps, IState> {
...
+
+	public handleClickOutside = ({ target }) => {
+		const button = this.addItemButtonRef.current;
+		const listContainer = this.groupsContainerRef.current;
+		const containsButton = target === button || button.contains(target);
+
+		if (!listContainer.contains(target) && !containsButton) {
+			this.resetActiveGroup();
+		}
+	}
```
sanmont3drepo, line 148: I think it might be better to handle the propagation in the click handlers  instead of creating refs  unless there is another reason to use the reference to the other components.

frontend/routes/viewerGui/components/groups/groups.component.tsx
```diff
@@ -182,16 +193,17 @@ export class Groups extends React.PureComponent<IProps, IState> {
 
 	public renderListView = renderWhenTrue(() => (
 		<>
-			<ViewerPanelContent>
+			<ViewerPanelContent onClick={this.handleClickOutside}>
 				{this.renderEmptyState(!this.props.searchEnabled && !this.state.filteredGroups.length)}
 				{this.renderNotFound(this.props.searchEnabled && !this.state.filteredGroups.length)}
 				{this.renderGroupsList(this.state.filteredGroups.length)}
```
sanmont3drepo, line 199: Instead of referencing each individual component, the propagation of the click could be handled here by wrapping the components and handling the click.
```suggestion
<div onClick={(e:MouseEvent) => e.stopPropagation()}>
				{this.renderEmptyState(!this.props.searchEnabled && !this.state.filteredGroups.length)}
				{this.renderNotFound(this.props.searchEnabled && !this.state.filteredGroups.length)}
				{this.renderGroupsList(this.state.filteredGroups.length)}
</div>
```
Same goes with reportedItems.

sanmont3drepo (changes requested):

frontend/modules/issues/issues.redux.ts
m1chalkubiak: You're right, the previous implementation was too little thought out. This has now been corrected.

m1chalkubiak:
In the last commit, apart from the suggested refactoring, I also added the removal of the highlighted overrides elements, which seems to be missing in the previous implementation but was included in the requirements for this task.

frontend/modules/issues/issues.redux.ts
sanmont3drepo: Lets remove this changes because now is just dead code.

frontend/routes/components/submitButton/submitButton.component.tsx
```diff
@@ -29,7 +29,8 @@ interface IProps {
 	variant?: string;
 }
 
-export const SubmitButton = ({ pending, disabled, children, ...props}: IProps) => {
+export const SubmitButton = React.forwardRef(
+	({ pending, disabled, children, ...props}: IProps, ref: React.Ref<HTMLElement>) => {
```
sanmont3drepo, line 33: Im guessing this ref is not being used anymore? If so, remove them.

sanmont3drepo (changes requested):

PR #2601, opened 2021-05-11
ISSUE #2334 - Preventing overwritten your own new card when create ticket event arrives

### LINKED ISSUE #2334: When multiple users create issues/risks simultaneously, whoever submits first will reset everyone's card.
### Description
I am working on a model and start raising an Issue/Risk. Simultaneously, another user is creating an Issue/Risk. If another user submits an issue/risk before mine it will reset my Issue/Risk card. As a result, I've lost any information that I have written in. 
This behaviour is happening in both the model viewer & kanban board.

### Steps to replicate
Requires two accounts ie. User A & User B
- User A opens a model in 3drepo.io
- User A starts creating an issue (do not submit yet)
- Simultaneously, User B opens the same model in 3drepo.io
- User B starts creating an issue as well
- User B submits his/her issue (works fine) 
- User A hasn't submitted his/her issue yet
- User A suddenly looses all of the information input (resets the Issue Card)

Problem is that users can't work in 3D Repo simultaneously & unable raising issues/risks independent of each other. 

Simultaneous SafetiBase

m1chalkubiak:
This fixes #2334


frontend/modules/issues/issues.redux.ts
```diff
@@ -147,10 +147,17 @@ export const saveIssueSuccess = (state = INITIAL_STATE, { issue }) => {
 		issuesMap[state.componentState.activeIssue].position = oldPosition;
 	}
 
+	if (resetComponentState) {
```
carmenfan, line 150: Can we reduce duplication here? Do something like this?

```suggestion
	const newComponentState = { ...state.componentState };
	if (resetComponentState) {
	      newComponentState.newIssue = {};
	}
	return {
			...state,
			issuesMap,
			componentState: newComponentState
		};
```

frontend/modules/risks/risks.redux.ts
```diff
@@ -165,10 +165,17 @@ export const saveRiskSuccess = (state = INITIAL_STATE, { risk }) => {
 		risksMap[state.componentState.activeRisk].position = oldPosition;
 	}
 
+	if (resetComponentState) {
```
carmenfan, line 168: here as well

carmenfan (changes requested):

frontend/modules/risks/risks.redux.ts
m1chalkubiak: done

frontend/modules/issues/issues.redux.ts
m1chalkubiak: Of course, it looks better now, thanks for spotting this.

carmenfan (changes requested):
The card stays up now, but the data within the card goes missing.

notice all the drop down menu goes blank here:

m1chalkubiak:
> The card stays up now, but the data within the card goes missing.
> 
> notice all the drop down menu goes blank here:
> 
> 

@carmenfan This shouldn't be happening and I can't reproduce it locally, can you please check it's the newest ISSUE_2334?

carmenfan:
> > The card stays up now, but the data within the card goes missing.
> > notice all the drop down menu goes blank here:
> > 
> 
> @carmenfan This shouldn't be happening and I can't reproduce it locally, can you please check it's the newest ISSUE_2334?

Hmmm maybe i didn't refresh my browser!

I can no longer reproduce it when I create a new issue, but if I update an existing issue (e.g. dragging something across the kanban, it's still happening.

m1chalkubiak:
@carmenfan Just a quick update - I see right now this case with changing issue status on kanban board, the other card is in fact resetting. I'm checking why.

m1chalkubiak:
Found it, fortunately, it was easy to fix.
Thanks for paying attention to the kanban board, I completely missed it.

carmenfan (approved):

PR #2604, opened 2021-05-13
Issue 2599 - make metadata all endpoint stream the response

### LINKED ISSUE #2599: get all metadata endpoint is crashing the server if the data set is too large
### Description
The metadata set can get is too large for the server to handle. At this point the server seems to crash.

We should prevent that from happening, even if it means limiting the amount of data we serve. Couple of ideas (probably should do both):

- add pagination to this endpoint, limit to x amount of documents at a time
-<del> add query support where the user can filter the metadata that they're interested in.</del> this is effectively [rules endpoint](https://3drepo.github.io/3drepo.io/#api-Meta-queryMetadataByRules)

### Steps to replicate
- run get all metadata endpoint on this model : www.3drepo.io/viewer/Demo_3D_Repo/19ce3220-6e89-11ea-9f08-1387f50c0299

### Current Behaviour
- the server crashes and you get an nginx error 

### Expected Behaviour
- The endpoint should run, even if it means we need to have concessions.

carmenfan:
This fixes #2599
#### Description
- small refactor of Metadata file to get rid of the class and be more inline with what we're doing in other files
- introduced a function in ref.js `getSubModels` which will return you an array with all the sub models with the write branch/revision info (so we don't have to do check if `_rid== C.MASTER_BRANCH` everywhere....). This also has a callback function to reduce looping.
- Rewrote getAllMetadata to:
    - return a stream buffer instead of a json
    - use `findNodesByType` for a much more efficient query (as types are indexed)
    - loop through the models synchronously and pipe the resulting json into a stream (to reduce memory footprint)

#### Test cases
- The test case identified in the issue should now work without crashing
- everything else should be working as before


backend/models/meta.js
```diff
@@ -40,391 +40,410 @@ function clean(metadataToClean) {
...
-		}
+	const subMeta = await Promise.all(getMeta);
 
-		// Check for submodel references
-		const refs = await getRefNodes(account, model, branch, rev);
+	const obj = await findNodesByField(account, model, branch, rev, fullFieldName);
 
-		const getMeta = [];
+	if (!obj) {
+		return Promise.reject(responseCodes.METADATA_NOT_FOUND);
```
Charence, line 308: We can `throw` here?

backend/models/meta.js
```diff
@@ -40,391 +40,410 @@ function clean(metadataToClean) {
...
-							model: ref.project
-						};
-					})
-					.catch(() => {
-						// Just because a sub model fails doesn't mean everything failed - do nothing
-					})
-			);
-		});
+	if (!settings) {
+		return Promise.reject(responseCodes.MODEL_NOT_FOUND);
```
Charence, line 329: `throw`

backend/models/meta.js
```diff
@@ -40,391 +40,410 @@ function clean(metadataToClean) {
...
+	}
 
-		const subMeta = await Promise.all(getMeta);
+	if (!settings.fourDSequenceTag) {
+		return Promise.reject(responseCodes.SEQ_TAG_NOT_FOUND);
```
Charence, line 333: `throw`

backend/models/meta.js
```diff
@@ -40,391 +40,410 @@ function clean(metadataToClean) {
...
+	});
+};
+
+Meta.getIfcGuids = async (account, model) => {
+	return db.find(account, getSceneCollectionName(model), { type: "meta" }, { "metadata.IFC GUID": 1 });
+};
+
+Meta.ifcGuidsToUUIDs = async (account, model, branch, revId, ifcGuids) => {
+	if (!ifcGuids || ifcGuids.length === 0) {
+		return Promise.resolve([]);
```
Charence, line 191: Is the `Promise.resolve` bit needed?

Charence:
Is there a copy of http://www.3drepo.io/viewer/Demo_3D_Repo/19ce3220-6e89-11ea-9f08-1387f50c0299 on dev?

backend/models/meta.js
carmenfan: sure, i didn't code this (or anything else you pointed out)

All I did was remove the class 😆

carmenfan:
> @carmenfan @3drepo/product-team Is there a copy of http://www.3drepo.io/viewer/Demo_3D_Repo/19ce3220-6e89-11ea-9f08-1387f50c0299 on dev?

@BDavydov

Charence (approved):

mosment3DRepo (approved):
Tested on Postman, Logic Apps, PowerBI and Excel.

All managed to get the data from the endpoint

PR #2614, opened 2021-05-18
ISSUE 2489 - MS Logic Apps support

### LINKED ISSUE #2489: Add 'Chunking' to 3D Repo model upload
### Description

Currently 3D Repo does not appear to support file 'chunking', This means that on integration platforms such as Microsoft Power Automate (also know as Logic Apps on Azure) any files >50MB fail as the platform uploads files in 50MB chunks. With chunking files up to 1GB can be supported on these platforms

Currently when chunking is attempted the response from the API is "The multipart content body cannot be null."

### Goals

- [x] Allow files to be uploaded >50MB on Microsoft Power Automate
- [ ] Better connect with Projectwise and Asite, who also have connectors on these platforms

### Tasks

Currently envisage 3 API endpoints to be required to handle this MS Logic App specific uploading.

1. We tell the server that we plan on starting an upload and we pass through the information that a regular 3drepo.io upload needs such as tag and description. We get a `corID` back in the response and we'll use this to keep track of which upload it is for the POST/PATCH requests from the logic app.
2. This is the upload initiation POST request mentioned in the Logic App documentation. A POST is sent with information about the size of the upload in the header. In response, you'll get a `Location`, which is the API for uploading the subsequent file chunks, and a desired chunk size. **NB: I believe Logic Apps should be handling this next chunk upload step automatically**
3. This last step should hopefully happen automatically. It is a PATCH for uploading individual model chunks. The response of each PATCH confirms the byte range uploaded and the desired chunk size for the next upload.

1) **`/:teamspace/:model/upload/ms-chunking`**
- [x] Initiate upload session with server and obtain Correlation ID (corID) to keep track of upload
- [x] Tests

Example:
```
POST /:teamspace/:model/upload/ms-chunking

body: {
  filename,
  tag,
  desc,
  importAnimation
}

response: {
  corID: <correlation ID for upload>
}

```

2) **`/:teamspace/:model/upload/ms-chunking/:corID`**
- [x] Upload request POST endpoint with an empty message in body. Request header includes: `x-ms-transfer-mode` and `x-ms-content-length`
  - [x] verify that `x-ms-transfer-mode` is `chunked`; we're not expecting anything other than `chunked` so it will throw an `INVALID_ARGUMENTS` error if this header is not as expected
  - [x] verify that `x-ms-content-length` is in the header and is a number, otherwise throw `INVALID_ARGUMENTS`
  - ~we (maybe) need to store this expected content length so we can recognise when we have received all expected chunks~
  - [x] need to calculate upload chunk size and know max chunk size for Logic Apps
  - [x] Respond with URL location for PATCH messages
  - [x] Tests

Example:
```
POST /:teamspace/:model/upload/ms-chunking/:corID

header: {
  x-ms-transfer-mode: "chunked",
  x-ms-content-length: 4096
}

response: {
  x-ms-chunk-size: 1024,
  Location: "https://<domain>/api/<teamspace>/<model>/upload/ms-chunking/<corID>"
}
```

3) **`/:teamspace/:model/upload/ms-chunking/:corID`**
- [x] Chunk file upload PATCH endpoint
  - [x] Saving of uploaded chunk ~(similar to existing multer file upload?)~
    - current implementation saves to `cn_queue.upload_dir`, which is the same as the existing multipart file upload
    - alternatively, it could be held in memory, but I think we might not want to do this due to the file sizes, especially if there's more than 1 upload
  - [x] Concatenate uploaded chunks
    - if we assume chunks are sequential (as stated in docs) then we can join files as they are uploaded
  - [x] Record/save the `Content-Range` uploaded
    - `Content-Range` is in the format `bytes=<start byte>-<end_byte>/<total bytes>`, e.g. `bytes=0-1023/10100`
    - so potentially if we assume chunks are uploaded sequentially (according to documentation it is), we could maybe get away with just checking if the current chunk is the last one based on the `end_byte + 1 = total_bytes`
  - [x] When all expected chunks received, continue with same post file upload processing as normal `upload` endpoint
  - [x] Tests

Example:
```
PATCH /:teamspace/:model/upload/ms-chunking/:corID

header: {
  file: <file chunk>,
  Content-Range: "bytes=0-1023/4096",
  Content-Type: <file type>,
  Content-Length: 1024
}

response: {
  Range: "bytes=0-1023",
  x-ms-chunk-size: 1024
}
```

### Related Resources

[Azure documentation on chunking](https://docs.microsoft.com/en-us/azure/logic-apps/logic-apps-handle-large-messages)

[HTTP specification relating to chunking](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.6.1)

Charence:
This fixes #2489 

#### Description
- Support for MS Logic Apps chunking uploads
- New POST endpoint for requesting chunking upload that returns a `corID`
- New POST endpoint for initiating upload with file information
- New PATCH endpoint for uploading file chunks

#### Test cases
- Upload of large files via MS Logic Apps should work


backend/models/upload.js
```diff
@@ -0,0 +1,301 @@
...
+		req.on("end", () => {
+			writeStream.end();
+			return resolve();
+		});
+	});
+};
+
+const stitchChunks = (corID, newFilename) => {
+	const sharedSpacePath = importQueue.getSharedSpacePath();
+	const filePath = sharedSpacePath + "/" + corID + "/chunks";
```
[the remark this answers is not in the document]
carmenfan: I think `getSharedSpacePath` should be private for the queue - it's a bit high coupling otherwise.

Can we have a function in the queue service to return you a the path to corId? Something like
```suggestion
	const filePath = `${importQueue.getTaskFolder(corID)}/chunks`
```

backend/routes/model.js
```diff
@@ -1572,6 +1573,109 @@ router.get("/:model/revision/:revId/subModelRevisions", middlewares.hasReadAcces
  */
 router.delete("/:model", middlewares.hasDeleteAccessToModel, deleteModel);
 
+/**
+ * @api {post} /:teamspace/:model/upload/ms-chunking Upload model request
```
carmenfan, line 1577:
```suggestion
 * @api {post} /:teamspace/:model/upload/ms-chunking Initialise a chunking request
```

backend/routes/model.js
```diff
@@ -1572,6 +1573,109 @@ router.get("/:model/revision/:revId/subModelRevisions", middlewares.hasReadAcces
...
+ * @apiName uploadModelRequest
```
carmenfan, line 1578:
```suggestion
 * @apiName InitialiseModelUploadChunking
```

backend/routes/model.js
```diff
@@ -1572,6 +1573,109 @@ router.get("/:model/revision/:revId/subModelRevisions", middlewares.hasReadAcces
...
+ * 	"desc": "Revision 2"
+ * }
+ *
+ * @apiSuccessExample {json} Success-Response:
+ * HTTP/1.1 200 OK
+ * {
+ * 	"corID": "00000000-0000-1111-2222-333333333333"
+ * }
+ */
+router.post("/:model/upload/ms-chunking", middlewares.hasUploadAccessToModel, uploadModelRequest);
```
carmenfan, line 1603:
```suggestion
router.post("/:model/upload/ms-chunking", middlewares.hasUploadAccessToModel, initChunkRequest);
```

backend/routes/model.js
```diff
@@ -1939,6 +2043,90 @@ function downloadLatest(req, res, next) {
...
+	try {
+		const chunkingHeader = await Upload.uploadChunk(account, model, corID, req);
+		res.writeHead(200, chunkingHeader);
+		res.end();
+	} catch(err) {
+		const errMsg = err.resCode ? err.resCode : err;
+		responseCodes.respond(responsePlace, req, res, next, errMsg, errMsg);
+	}
+
+	/*
```
carmenfan, line 2095: remove if not needed?

backend/routes/model.js
```diff
@@ -1939,6 +2043,90 @@ function downloadLatest(req, res, next) {
...
+		res.end();
+	} catch(err) {
+		const errMsg = err.resCode ? err.resCode : err;
+		responseCodes.respond(responsePlace, req, res, next, errMsg, errMsg);
+	}
+}
+
+async function uploadChunk(req, res, next) {
+	const responsePlace = utils.APIInfo(req);
+	const account = req.params.account;
```
carmenfan, line 2081: use deconstructed assignment ?
```suggestion
	const { account, model, corId}  = req.params;
```

backend/routes/model.js
```diff
@@ -1939,6 +2043,90 @@ function downloadLatest(req, res, next) {
...
+		responseCodes.respond(responsePlace, req, res, next, responseCodes.OK, corID);
+	} catch(err) {
+		const errMsg = err.resCode ? err.resCode : err;
+		responseCodes.respond(responsePlace, req, res, next, errMsg, errMsg);
+	}
+}
+
+async function initUploadChunks(req, res, next) {
+	const responsePlace = utils.APIInfo(req);
+	const account = req.params.account;
```
carmenfan, line 2063: use deconstructed assignment ?

backend/test/integrated/upload.js
```diff
@@ -219,4 +247,264 @@ describe("Uploading a model", function () {
...
+						"tag": "bad tag!"
+					})
+					.expect(400, function(err, res) {
+						expect(res.body.value).to.equal(responseCodes.INVALID_TAG_NAME.value);
+						done(err);
+					});
+			});
+
+			/*
+			it("duplicate tag should fail", function(done) {
```
carmenfan, line 298: ?

backend/test/integrated/upload.js
```diff
@@ -219,4 +247,264 @@ describe("Uploading a model", function () {
...
+					.set("x-ms-content-length", "118832273")
+					.expect(200, function(err, res) {
+						expect(parseInt(res.headers["x-ms-chunk-size"])).to.equal(C.MS_CHUNK_BYTES_LIMIT);
+						expect(res.headers["location"]).to.exist;
+						done(err);
+					});
+			});
+		});
+
+		describe("Handle MS chunked uploads", function() {
```
carmenfan, line 420: should check invalid corID as well?

carmenfan (changes requested):

backend/models/upload.js
```diff
@@ -0,0 +1,301 @@
...
+
+	const contentMax = contentRangeValue.split("-")[1];
+
+	const sizeRemaining = contentSize - contentMax - 1;
+	const chunkSize = Math.min(C.MS_CHUNK_BYTES_LIMIT, sizeRemaining);
+
+	const sharedSpacePath = importQueue.getSharedSpacePath();
+	const timestamp = Date.now();
+
+	await handleChunkStream(req, `${sharedSpacePath}/${corID}/chunks/${timestamp}`);
```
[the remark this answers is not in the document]
Charence: The documentation says:

> A content chunk based on x-ms-chunk-size or some internally calculated size until all the content totaling x-ms-content-length is sequentially uploaded

https://docs.microsoft.com/en-us/azure/logic-apps/logic-apps-handle-large-messages#upload-content-in-chunks

backend/models/upload.js
```diff
@@ -0,0 +1,301 @@
...
+
+	systemLogger.logInfo(`CONTENT-RANGE=${req.headers["content-range"]}`);
+	systemLogger.logInfo(`CHUNKSIZE=${chunkSize}`);
+	if (chunkSize === 0) {
+		modelStatusChanged(null, teamspace, model, { status: "uploaded" });
+		await stitchChunks(corID, "upload");
```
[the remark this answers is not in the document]
Charence: @carmenfan The workflow for normal uploads always wants to move a file to this `corID/filename` location. The `move` function seems to not like it if source and destination are the same and I think some more changes to distinguish between a normal upload and chunked upload would have to be made if we don't want to move the chunked file.

backend/models/upload.js
```diff
@@ -0,0 +1,301 @@
...
+		const { filename } = JSON.parse(fs.readFileSync(`${sharedSpacePath}/${corID}.json`, "utf8"));
+		importQueue.importFile(corID, `${sharedSpacePath}/${corID}/upload`, filename, null);
```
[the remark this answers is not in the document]
Charence: @carmenfan Yes, this would be necessary if we don't temporarily store it as `corID/upload`.

backend/models/upload.js
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	return {
+		"Range": `bytes=0-${contentMax}`,
```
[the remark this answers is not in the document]
Charence: @carmenfan it was complaining about the `x-ms-...` ones so I ended up putting it on this too. Can be removed for `Range` but not for content length, etc.

backend/models/upload.js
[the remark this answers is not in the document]
carmenfan: I think we want to do a check and not bother with the mv if the file is already in the right place. There's not a lot of point naming something something else just because you need to do a move that doesn't need to happen

backend/models/upload.js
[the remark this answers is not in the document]
carmenfan: ok we can keep it then if the doc guarantees it 👍

backend/models/upload.js
```diff
@@ -0,0 +1,301 @@
...
+
+/** *****************************************************************************
+ * @param {corID} corID - correlation ID of upload
+ * @param {databaseName} databaseName - name of database to commit to
+ * @param {modelName} modelName - name of model to commit to
+ * @param {userName} userName - name of user
+ * @param {tag} tag - revision tag
+ * @param {desc} desc - revison description
+ *******************************************************************************/
+Upload.writeImportData = async (corID, databaseName, modelName, userName, newFileName, tag, desc, importAnimations = true) => {
```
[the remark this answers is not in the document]
Charence: @carmenfan I noticed there's `importSuccess` and `importFail` in `model/helper/model` that uses sharedSpacePath too. Should that stay or be moved as well?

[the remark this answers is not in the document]
carmenfan: let's leave it for now, there's an inevitable refactoring on those functions eventually

backend/models/history.js
```diff
@@ -167,14 +167,14 @@ History.clean = function(histories, branch) { // or history
 };
 
 History.isValidTag = async function(account, model, tag) {
-	if(!tag) {
-		return null;
+	if (!tag) {
+		throw responseCodes.INVALID_TAG_NAME;
```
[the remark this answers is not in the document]
Charence: Returning `null` was supposed to be a positive outcome. I've changed this function to not return `null`.

Charence:
@mosment3DRepo This is ready for some more testing

mosment3DRepo (approved):
Tested in principal, further testing to be carried out when Microsoft answer outstanding queries. Any changes can be applied in 4.20

backend/models/upload.js
```diff
@@ -0,0 +1,240 @@
...
+const stitchChunks = (corID, newFilename) => {
+	const filePath = `${importQueue.getTaskPath(corID)}/chunks`;
+
+	const files = fs.readdirSync(filePath);
+
+	// sort timestamps into ascending order
+	files.sort((a, b) => a - b);
+
+	files.forEach((file) => {
+		// for debugging
```
carmenfan, line 69: remove debugging

backend/models/upload.js
```diff
@@ -0,0 +1,240 @@
...
+
+	if (!modelSetting) {
+		throw responseCodes.MODEL_NOT_FOUND;
+	}
+
+	const newFileName = data.filename.replace(C.FILENAME_REGEXP, "_");
+
+	await isValidTag(teamspace, model, data.tag);
+
+	const corID = nodeuuid();
```
carmenfan, line 100: use utils.generateUUID({string: true})

backend/models/upload.js
```diff
@@ -0,0 +1,240 @@
...
+		data.desc,
+		data.importAnimations
+	);
+
+	return { corID };
+};
+
+Upload.uploadFile = async (req) => {
+	if (!config.cn_queue) {
+		return Promise.reject(responseCodes.QUEUE_NO_CONFIG);
```
carmenfan, line 118: throw?

backend/models/upload.js
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	const { account, model } = req.params;
```
carmenfan, line 121: could we refactor this so anything request/multer related stays in route folder. Where possible we want to avoid request data bleeding into the models folder. route handles requests, model handles internal logic.

I think majority of the logic here should just be in `routes`

backend/models/upload.js
```diff
@@ -0,0 +1,240 @@
...
+	});
+
+	// req.body.tag wont be defined after the file has been uploaded
+	await isValidTag(account, model, req.body.tag);
+
+	return uploadedFile;
+};
+
+Upload.uploadChunksStart = async (teamspace, model, corID, username, headers) => {
+	if (!headers["x-ms-transfer-mode"] ||
```
carmenfan, line 164: probably good to make these strings constants somewhere as you reference it more than once.

backend/models/upload.js
```diff
@@ -0,0 +1,240 @@
...
+		headers["x-ms-transfer-mode"] !== "chunked" ||
+		!headers["x-ms-content-length"] ||
+		isNaN(headers["x-ms-content-length"])) {
+		// for debugging
```
carmenfan, line 168: remove debugging

backend/models/upload.js
```diff
@@ -0,0 +1,240 @@
...
+Upload.uploadChunk = async (teamspace, model, corID, req) => {
+	if (!fs.existsSync(`${importQueue.getTaskPath(corID)}.json`)) {
+		throw responseCodes.CORRELATION_ID_NOT_FOUND;
+	}
+
+	if (!config.cn_queue) {
+		return Promise.reject(responseCodes.QUEUE_NO_CONFIG);
+	}
+
+	if (!req.headers["content-range"]) {
```
carmenfan, line 199: constant

backend/models/upload.js
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	const [contentRange, totalContentSize] = req.headers["content-range"].split("/");
+	const [sizeUnit, contentRangeValue] = contentRange.split(" ");
+
+	if (sizeUnit !== "bytes") {
+		throw responseCodes.INVALID_ARGUMENTS;
+	}
+
+	await middlewares.checkSufficientSpace(teamspace, parseInt(totalContentSize));
```
carmenfan, line 210: does this mean you're checking it everytime there's a chunk coming in?

I think we only need to check it on the last chunk. it's a bit excessive to do it at every chunk.

backend/response_codes.js
```diff
@@ -485,6 +486,13 @@
 		// next();
 	};
 
+	responseCodes.writeHead =  function (place, req, res, next, customHeaders) {
```
carmenfan, line 489: this is confusing, we have a function to write a standard response, and we have a function to write a streamed response.

You're essentially replying with a standard response, without any body, and some extra customheaders. don't write a new function for this, just expand the existing function to take in custom headers.

carmenfan (changes requested):

carmenfan (approved):

PR #2620, opened 2021-05-21
Issue 2551 - 2D Sheet navigation mode

### LINKED ISSUE #2551: 2D sheet viewing mode
### Description
With the support for 2D imports, there will be models coming in that are basically sheets. The current viewer is purely designed to navigate/interact in 3 dimensions, thus does not work well for these kind of situations

### Goals
- Allow the user to flag something as `2D Drawing` which initiates 2D mode when they enter the viewer
- Add option to change background colour to ensure the drawings doesn't blend into the background

#### Turning on 2D Sheet Mode
Unity will have a function to toggle the viewer between 2D and 3D. The differences are:
      - 2D has it's own navigation, turnable/helicopter mode will not be available
      - Default home view is top down, looking at the sheet

If the model type of the model is of `2D Drawing`, React should call a function in UnityUtils to turn the viewer into 2D mode. It should also change the background to white.

In 2D Sheet Mode there is a special 2D navigation, so UI should reflect this accordingly.

#### Turning off 2D Sheet mode
2D sheet mode should reset when the user exits the model. It should also exit 2D sheet mode if the user change the 2D navigation back to turnable/helicopter mode

### Tasks

### Related Resources
Coda doc: https://coda.io/d/3D-Repo-Features_d5c0Qp25r6f/2D-Document-Viewing_su4ro#_luC7c

carmenfan:
This fixes #2551

#### Description
- Added a new model type `2D Sheet`
- Added background settings into viewer setting (background colour, background skybox)
- moved model type constants into the `constants` folder instead of part of teamspace component
- Add logic to enable/disable 2D Sheet mode.

#### Test cases
- The user should be able to create a model of 2D Sheet type
- When viewing 2D Sheet type model, the navigation defaults to 2D Sheet
-  If the user goes to a 3D Model that's not 2D Sheet type, this should reset to turntable.
- If the user turns off skybox, the viewer should change the background to whatever background colour presented
- If the user turns on skybox, the viewer should reinstate the skybox
- if skybox is enabled, colour configuration is disabled
- Which setting (2d or 3d) used should match whether the nav mode is on 2D Sheet
- orthographic/perspective cam setting should not appear if the viewer is in 2D sheet mode

#### Missing features 
- If the model is set to be of type 2D sheet and it's not actually a sheet, the nav mode should auto revert to turntable (waiting on unity)
- The ability to realign the model (waiting on decision on approach)
- Ability to flip the sheet (waiting on the line above, and UI design)


keifgwinn:
I've deleted this as it was interferring with the callbackq messages on staging. when it's ready to be tested again run /azp *run* again to recreate and it shouldn't break it now.

carmenfan:
this is parked. revive when appropriate.

PR #2623, opened 2021-05-25
ISSUE #2479 - Reset autosugestion list on blur in groups card

### LINKED ISSUE #2479: Smart group auto suggest should reset if the field value changes
see https://github.com/3drepo/3drepo.io/issues/2474#issuecomment-780716326

m1chalkubiak:
This fixes #2479


carmenfan (changes requested):
- Not sure if this one is introduced by this branch or not, but it's probably useful to fix - clicking on the auto-suggestion doesn't always load the text into the input box:
aa

- Can we also make it so that if it's re-focused the auto-suggest prompt comes back?

m1chalkubiak:
> * Not sure if this one is introduced by this branch or not, but it's probably useful to fix - clicking on the auto-suggestion doesn't always load the text into the input box:
> * Can we also make it so that if it's re-focused the auto-suggest prompt comes back?

@carmenfan Of course, I'm already looking at both things.

m1chalkubiak:
@carmenfan Currently, everything works fine in my opinion, but I think it is worth mentioning what I realized yesterday.
There are currently 4 different implementations of how fields with autosuggestion work in the project. They all use the same external library, but their configuration and thus their behavior differs. Currently, to fix how this field works in the case of groups card, I checked how it is implemented elsewhere in the application, and mostly I copied the implementations from there. Therefore, I think we could go a step further and get rid of different implementations of this field, as this may cause duplicate problems in the future (inconsistencies in how the auto-suggest field behaves in different parts of the application).

I am not sure if it has such a high priority at the moment, I leave the decisions to You. 
Anyway, it is definitely worth including such a task in the backlog.

carmenfan:
@m1chalkubiak 

Good to know. considering a rewrite is imminent, i think we'll leave for now.

sanmont3drepo (approved):

PR #2627, opened 2021-05-27
ISSUE #2456 - Activate save button when at least one object is selected

### LINKED ISSUE #2456: save button is disabled when you try to highlight objects before going into the create group tab
### Description

If you highlight the items you wish to create a group with prior to going into the create group panel, the save button is disabled

- This is happening in: both

### Steps to replicate
- Open any model
- Highlight some objects
- click into create group

### Current Behaviour
- save button is disabled, even though you have items highlighted
image

### Expected Behaviour
- Save button should be enabled as long as you have >0 objects currently highlighted

m1chalkubiak:
This fixes #2456

I added the missing logic to check if anything is actually selected when creating a group. I'm not sure if it works properly for smart groups. It would be helpful if you could check it and let me know.

### DIFF
--- a/frontend/routes/components/submitButton/submitButton.component.tsx
+++ b/frontend/routes/components/submitButton/submitButton.component.tsx
@@ -29,7 +29,8 @@ interface IProps {
 	variant?: string;
 }
 
-export const SubmitButton = ({ pending, disabled, children, ...props }: IProps) => {
+export const SubmitButton = React.forwardRef(
+		({ pending, disabled, children, ...props}: IProps, ref: React.Ref<HTMLElement>) => {
 	const additionalProps = merge({
 		type: 'submit',
 		variant: 'contained',
@@ -46,6 +47,7 @@ export const SubmitButton = ({ pending, disabled, children, ...props }: IProps) 
 
 	return (
 		<Button
+			ref={ref}
 			disabled={pending || disabled}
 			{...additionalProps}
 		>
@@ -57,4 +59,4 @@ export const SubmitButton = ({ pending, disabled, children, ...props }: IProps) 
 			)}
 		</Button>
 	);
-};
+});
--- a/frontend/routes/viewerGui/components/groups/components/groupDetails/groupDetailsForm.component.tsx
+++ b/frontend/routes/viewerGui/components/groups/components/groupDetails/groupDetailsForm.component.tsx
@@ -49,6 +49,10 @@ export class GroupDetailsForm extends React.PureComponent<IProps, any> {
 
 	public componentDidMount() {
 		this.props.setIsFormValid(this.isNewGroup);
+
+		if (this.props.selectedNodes.length) {
+			this.props.setIsFormDirty(this.isNewGroup);
+		}
 	}
 
 	public componentDidUpdate(prevProps) {
@@ -58,9 +62,10 @@ export class GroupDetailsForm extends React.PureComponent<IProps, any> {
 		const groupChanged = !isEqual(this.props.group, prevProps.group);
 		const isNormalGroup = this.props.group.type === GROUPS_TYPES.NORMAL;
 		const sharedIdsChanged = isNormalGroup ? this.areSharedIdsChanged(this.props.selectedNodes, objects) : false;
+		const isAnyNodeSelected = !!this.props.selectedNodes.length;
 
 		const isFormDirtyAndValid = (!isEqual(initialValues, currentValues) && groupChanged) || sharedIdsChanged;
-		this.props.setIsFormValid(isFormDirtyAndValid);
+		this.props.setIsFormValid(isFormDirtyAndValid && isAnyNodeSelected);
 		this.props.setIsFormDirty(isFormDirtyAndValid);
 	}
 
--- a/frontend/routes/viewerGui/components/groups/groups.component.tsx
+++ b/frontend/routes/viewerGui/components/groups/groups.component.tsx
@@ -134,7 +134,6 @@ export class Groups extends React.PureComponent<IProps, IState> {
 		filteredGroups: []
 	};
 
-	public addGroupButtonRef = React.createRef<any>();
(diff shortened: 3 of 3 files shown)


sanmont3drepo:
I look a little into this and I think the component was reacting properly, the only problem that I've seen is that it was deselecting the meshes when clicking the create group button. I dont know if Im missing something but I created an alternative branch
https://github.com/3drepo/3drepo.io/tree/ISSUE_2456b

m1chalkubiak:
> I look a little into this and I think the component was reacting properly, the only problem that I've seen is that it was deselecting the meshes when clicking the create group button. I dont know if Im missing something but I created an alternative branch
> #2631

Hi  @sanmont3drepo thanks for checking PR, I saw the issue with deselecting nodes after clicking on creating group button, I think this is something that we missed during https://github.com/3drepo/3drepo.io/pull/2591.
To prevent it I've added additional logic to `resetActiveGroup` method, and it seemed to work fine.



https://user-images.githubusercontent.com/24941866/120305278-db1c8e80-c2d0-11eb-87d2-b83a1fa7f3e9.mp4

carmenfan:
> I think this is something that we missed during #2591.

@m1chalkubiak  actually this problem pre-dates the work done for that issue. We've had this issue for the past year... 😆 

@sanmont3drepo's branch seems to work as expected and seems to be a much simplier solution, I'd be more inclined to merge that one in as a fix if react is "reacting" correctly as it is - we shouldn't require to do extra unnecessary checks . Do you mind taking a look at his changes and see if there's anything that might be missing?

m1chalkubiak:
@carmenfan Sure, I've removed redundant checking from my PR, after merging Santiago's one, it should work properly.

sanmont3drepo:
> @carmenfan Sure, I've removed redundant checking from my PR, after merging Santiago's one, it should work properly.

I think is  already working correctly in the other branch. The logic for enabling the button is here
https://github.com/3drepo/3drepo.io/blob/a32a86bb4e557f3300d96ffbf048e18cb7e8c90d/frontend/routes/viewerGui/components/groups/components/groupDetails/groupDetailsForm.component.tsx#L60

Is only being disabled when you just add the group because you need to enter a name.

m1chalkubiak:
> > @carmenfan Sure, I've removed redundant checking from my PR, after merging Santiago's one, it should work properly.
> 
> I think is already working correctly in the other branch. The logic for enabling the button is here
> https://github.com/3drepo/3drepo.io/blob/a32a86bb4e557f3300d96ffbf048e18cb7e8c90d/frontend/routes/viewerGui/components/groups/components/groupDetails/groupDetailsForm.component.tsx#L60
> 
> Is only being disabled when you just add the group because you need to enter a name.

@sanmont3drepo which branch are you referring to?

sanmont3drepo:
> > > @carmenfan Sure, I've removed redundant checking from my PR, after merging Santiago's one, it should work properly.
> > 
> > 
> > I think is already working correctly in the other branch. The logic for enabling the button is here
> > https://github.com/3drepo/3drepo.io/blob/a32a86bb4e557f3300d96ffbf048e18cb7e8c90d/frontend/routes/viewerGui/components/groups/components/groupDetails/groupDetailsForm.component.tsx#L60
> > 
> > Is only being disabled when you just add the group because you need to enter a name.
> 
> @sanmont3drepo which branch are you referring to?
It was in staging all along.

m1chalkubiak:
> > > > @carmenfan Sure, I've removed redundant checking from my PR, after merging Santiago's one, it should work properly.
> > > 
> > > 
> > > I think is already working correctly in the other branch. The logic for enabling the button is here
> > > https://github.com/3drepo/3drepo.io/blob/a32a86bb4e557f3300d96ffbf048e18cb7e8c90d/frontend/routes/viewerGui/components/groups/components/groupDetails/groupDetailsForm.component.tsx#L60
> > > 
> > > Is only being disabled when you just add the group because you need to enter a name.
> > 
> > 
> > @sanmont3drepo which branch are you referring to?
> 
> It was in staging all along. The only problem was that is was clearing up the selected meshes the logic for enabling the button was there.

@sanmont3drepo But You can still create a group without any selected node, right? And it's not a problem?

carmenfan:
> @sanmont3drepo But You can still create a group without any selected node, right? And it's not a problem?

@m1chalkubiak  no - you should only be able to create a group if it has a name and is either:
 - a smart group with at least 1 rule
 - a normal group with at least 1 object

m1chalkubiak:
> > @sanmont3drepo But You can still create a group without any selected node, right? And it's not a problem?
> 
> @m1chalkubiak no - you should only be able to create a group if it has a name and is either:
> 
> * a smart group with at least 1 rule
> * a normal group with at least 1 object

@carmenfan  This is the current state on staging, my patch fixes it.

https://user-images.githubusercontent.com/24941866/120337539-b2f25700-c2f3-11eb-9a04-59e9a8750b24.mp4

sanmont3drepo:
> > > @sanmont3drepo But You can still create a group without any selected node, right? And it's not a problem?
> > 
> > 
> > @m1chalkubiak no - you should only be able to create a group if it has a name and is either:
> > 
> > * a smart group with at least 1 rule
> > * a normal group with at least 1 object
> 
> @carmenfan This is the current state on staging, my patch fixes it.
> 
>  ezgif.com-gif-maker.1.mp4

The patch is just forcing an enabled button when one mesh is selected and its creating the situations where the rest of the form is not ok (like the name not entered) and enabling the button:
image

m1chalkubiak:
@sanmont3drepo Ok, thanks for pointing this case

@carmenfan So we still have problems that we have to handle regarding this form, do I have to deal with this on this branch or on a different issue?

sanmont3drepo:
I created a new issue to address the problem of  updating a normal group to 0 meshes:
https://github.com/3drepo/3drepo.io/issues/2634

PR #2633, opened 2021-06-02
ISSUE #2618 - Add linter rule to ensure license header exists on all files

### LINKED ISSUE #2618: some files are missing license headers
### Description
some files (for e..g `frontend/herlps/filtering.ts`) is missing the licensing header. The licensing header should appear on every file that contains that we coded.

We should address this. Also it would be good to add a linter rule to error out if the file is missing a license header.

### Tasks:
-  [x] Add linter rule to ensure license header exists
-  [x] apply license header on any file that does not have it

m1chalkubiak:
This fixes #2618


carmenfan (changes requested):
@m1chalkubiak 

Copy right terms that only display 1 number for year should be displaying the year that it the copy right is valid from, not the current year.

Could we make it so that we didn't change all the license header years?


image

m1chalkubiak:
> @m1chalkubiak
> 
> Copy right terms that only display 1 number for year should be displaying the year that it the copy right is valid from, not the current year.

@carmenfan I'm not sure what do You mean by that exactly, should this be the year of creating a specific file? 

> Could we make it so that we didn't change all the license header years?
> 
> image

Actually, the rule isn't so sophisticated to update specifically years, it only checks if there is a license and if its content is valid with the original one. If no, fix is adding license at the beginning of the file, but doesn't remove old one. In the situation that You are referring to, the problem was line 15, a new license block has added by lint fix and I removed manually the whole old one. I corrected the year to match the original one right now.

carmenfan:
> @carmenfan I'm not sure what do You mean by that exactly, should this be the year of creating a specific file?

Yes - if we created the file last year it should say 2020 and stays that way.

Obviously anything with a missing license file it'll be difficult to figure out - putting 2021 on those one it's fine.

The main point here is we shouldn't be updating the year of the files.

> Actually, the rule isn't so sophisticated to update specifically years, it only checks if there is a license and if its content is valid with the original one.

I'm confused. the linter shouldn't be updating years. it should either check there is a license on it, or not, regardless of what the year is printed on it.

i..e if we have a file that has a 2020 license, it should detect it has a licence and thus it should not impose a new one on it

m1chalkubiak:
> 
> > Actually, the rule isn't so sophisticated to update specifically years, it only checks if there is a license and if its content is valid with the original one.
> 
> I'm confused. the linter shouldn't be updating years. it should either check there is a license on it, or not, regardless of what the year is printed on it.
> 
> i..e if we have a file that has a 2020 license, it should detect it has a licence and thus it should not impose a new one on it

@m1chalkubiak As I wrote earlier, the problem was not the year but line no. 15 which differed from what the original license looks like (no extra space).

It was me who "changed the year", by removing the previous "wrong" license after adding a new one by lint fix.

m1chalkubiak:
@carmenfan The main problem was that some licenses were written differently, or even in some cases had their content changed. For the rule to work properly, I had to standardize its content in all files.

carmenfan:
> @carmenfan The main problem was that some licenses were written differently, or even in some cases had their content changed. For the rule to work properly, I had to standardize its content in all files. 


@m1chalkubiak  that's fine. I just wanted ot make sure the linter understands the old ones and doesn't auto update the year/add an additional license on them 😆

m1chalkubiak:
@carmenfan Are You happy with the way that rule is working currently. If You think that rule should work differently just let me know, I will do my best to fit it to Your expectations.

carmenfan (changes requested):
- [ ] Can we also add this on the eslint on the backend?
- [ ] I can't run the linter on the anymore, it's giving me an error: 
image
image
Just checked staging and it runs fine for me, but staging doesn't try to run with bin.js

m1chalkubiak:
@carmenfan Let me know what do You think about this, but it might be better to publish this rule as a normal eslint rule package on npm, thus we could apply this rule directly in FE and BE part of the application without having to repeat this implementation. As well as easily add it to other projects if necessary.

carmenfan:
> @carmenfan Let me know what do You think about this, but it might be better to publish this rule as a normal eslint rule package on npm, thus we could apply this rule directly in FE and BE part of the application without having to repeat this implementation. As well as easily add it to other projects if necessary.

@m1chalkubiak  I think first and foremost, we need to resolve the fact that it doesn't work on windows.

m1chalkubiak:
> > @carmenfan Let me know what do You think about this, but it might be better to publish this rule as a normal eslint rule package on npm, thus we could apply this rule directly in FE and BE part of the application without having to repeat this implementation. As well as easily add it to other projects if necessary.
> 
> @m1chalkubiak I think first and foremost, we need to resolve the fact that it doesn't work on windows.

Unfontently, at this moment I don't have easy access to any computer with Windows to debug it, however, it is related to the fact that for using custom rule written in typescript I needed to use ts-node instead of node. This would not be needed if the rule would be available externally (it would already be compiled).

Alternatively, we could compile the file to JS then ts-node will not be needed. However, this is related to the need to keep a file in the repository that does not allow easy manual editing.

carmenfan:
@m1chalkubiak  it's a bit chicken and egg atm. I can't tell you if I'm happy with the rule without testing it, and i can't test it without running... 😆

m1chalkubiak:
> @m1chalkubiak it's a bit chicken and egg atm. I can't tell you if I'm happy with the rule without testing it, and i can't test it without running... 😆

@carmenfan I totally get it, you are right. I added rule compiled to JS. Now everything will work as before, using node. Let's determine if it works properly, and then we will consider what to do next.

m1chalkubiak:
@carmenfan Apologize, this morning I realized that I deleted liveChat component on the wrong branch.

carmenfan:
> @carmenfan Apologize, this morning I realized that I deleted liveChat component on the wrong branch.

I noticed, so I cherry picked the commit to the GTM branch, GTM branch is now merged in and i merged staging into this branch, so it was correct, you didn't need to revert it. 😛 

I just reverted that commit so it doesn't' add the component back in. Should've wrote a message here about it, my bad!

carmenfan:
@m1chalkubiak  i think this is working fine

> but it might be better to publish this rule as a normal eslint rule package on npm, thus we could apply this rule directly in FE and BE part of the application without having to repeat this implementation.

considering the linter rule has a 3D Repo explicit license, does it make much sense to do it? It feels weird to be publishing something on npm that is so bespoke.

m1chalkubiak:
@carmenfan I've added an option to pass license owner to rule, more to show that we can adapt this solution so that publishing it is less "weird". At the same time, I want to emphasize that I do not insist that we have to solve it this way, more that currently it will require repeating the same code in at least two places (FE and BE).

carmenfan:
> @carmenfan I've added an option to pass license owner to rule, more to show that we can adapt this solution so that publishing it is less "weird". At the same time, I want to emphasize that I do not insist that we have to solve it this way, more that currently it will require repeating the same code in at least two places (FE and BE).

@m1chalkubiak   I think that would work, since at that point it's just a generic AGPL3 license header. I think we can publish it that way

carmenfan:
@m1chalkubiak in preparation for the revamp kick off, @sanmont3drepo  wants to do some folder restructuring. This is going to cause a lot of conflicts with any PR changes on the front end so we're keen to merge as many of these in as possible.

just wondering if you can make this a priority to finish off?

m1chalkubiak:
Hoi @carmenfan 
As we established in this task I've made migration from tslint to eslint, which in the end turned out to be a larger and more time-consuming task than I initially assumed.

I approached this migration several times, in the end sticking to the option that was most often suggested on the web. So I used the https://github.com/typescript-eslint/tslint-to-eslint-config tool to convert the current tslint configuration to the closest possible in eslint.
Even though it sounds very cool in theory, after the migration I received about 40,000 new errors, then the quite tedious part of analyzing individual rules, removing or silencing them began.

I've made also a few corrections in cases where the suggestion of eslint seemed justified to me, most of these changes can be found on this commit: https://github.com/3drepo/3drepo.io/pull/2633/commits/e12b72a976f973cd6836aea6df328d9a59ffc884

In case of the rules for which I changed the warning level from error to warn, I did it most often due to the fact that the rule itself is valuable in my opinion, but the number of places where corrections should be made was too large - I think that it would significantly increase the size of this PR - which is already very large due to the number of license-header fixes in many files.

I am waiting for your feedback as I think there is quite a lot of room for discussion here.

I was surprised by the build errors that appeared yesterday and were directly related to typing errors in typescript: https://github.com/3drepo/3drepo.io/pull/2633/commits/c86f62ab5c50c4ca789d525bba7ec402c3e56c0e

The first thing that jumps in this commit is the change from `[string]` to `string[]`, unless I'm not mistaken, the first one means arrays with one element that is a string, and I think type-check rightly noticed here that it is not this is the correct type for this variable, I just don't know why it just came out now.

P.S. There are more entries of this type `[string]` in the project and it seems to me that in most cases it may be the wrong type.

carmenfan:
@m1chalkubiak ok thanks!

It probably doesn't matter too much, for the revamp our plan is to rewrite everything and use a linter that has rules that are pretty close to the 3D send one. so even if we're keeping everything as is for certain files we'd have a manual screening and change anything appriopriate.

carmenfan (approved):

Issue #2635, opened 2021-06-03
allow unity to recover processing if it was forced to die 


carmenfan:
supporting issue for https://github.com/3drepo/3drepounity/issues/384

PR #2636, opened 2021-06-03
ISSUE #2635 allow unity to rerun if it was terminated mid operation

carmenfan:
See https://github.com/3drepo/3drepounity/pull/387 for details


.azure/build-and-deploy.yaml
```diff
@@ -31,7 +31,7 @@ variables:
   branchName: master
   group: tests-group
   customHelmOverride: |
-    config.CUSTOM=example
+    config.APP_QUEUE_DNS=issue-2635-rabbitmq
```
Charence, line 34: is this a temporary config for testing or we want to keep it as this value?

carmenfan: this file is under some special setting where any changes are ignored when we merge i believe. @keifgwinn ?

keifgwinn: https://github.com/3drepo/3drepo.io/blob/1d8de93c45a426bcd8c84008cef4240bb5f4abe7/.gitattributes#L8 is supposed to ignore changes to this file when doing the merge, however I noticed it didn't work last time.

carmenfan: sounds like this is only triggered if there's a conflict https://stackoverflow.com/a/27234774

so it wouldn't work if the staging version was not touched as well.

keifgwinn: regarding the original merge, this file can be abandoned it's only for moving it to a dedicated queue.

carmenfan: @Charence reverted

Charence (approved):

PR #2648, opened 2021-06-14
Issue 2611 - Successful logins are logged in the database

### LINKED ISSUE #2611: Record user logins
### Description
We want to start logging more information regarding the user logging in. The main aim is as follows:

- In the future, we may want to show the user a log of their session activity; identifying where they logged in from and with what kind of device.
- We want to start gathering statistics about which plugin version with which CAD software the user is using, so we can make better decisions on deprecating support. (this will require https://github.com/3drepo/3drepoPlugin/issues/181)

### Goals
- [ ] Every time a successful login request has been processed, we should gather data about the user origin and record an entry in the database.
- [ ] Send a copy of the record to elastic

### Specification
Upon successful login, we should extrapolate the user origin, type of device etc  from the request and store it as an entry. 

We should store it like notifications, where we have a database called `loginRecords`, where each user has its own collection. (NOTE: we need to blacklist the database to prevent users from creating a user account with the same name)

The schema should (probably) look like this:
````js
{
   _id: <string> //session ID so we can refer to it if necessary (to check if the session is still active?)
   loginTime: <MongoDB date>, //date time when the user logged in
   referrer: <string>, //the domain the request came from (e.g. https://www.3drepo.io or somewhere else which embedded our viewer),
   ipAddr: <string>, //ip address of the request
   location: {
        country: <string>, //country code
        city: <string>
    },
   application: {
       name: <string>, //If this came from a plugin, this should be Revit/Civils/Autocad/Navis, if it's browser it should be "chrome", "firefox" etc.
       version: <string>,
       type: <string> //"browser" or "plugin" or "unknown" (unknown used if there's no user agent information)
   },
   engine: {
       name: <string>, //If this came from a plugin, this should say "3drepoplugin"
       version: <string> //If this came from a plugin, this should be the version of the plugin
   },
   os: {
      name: <string>,
      version: <string>
   },
   device: <string> //type of device, e.g. mobile, computer etc
         
}
````

#### Considerations
- express seems to store the [ip ](http://expressjs.com/en/api.html#req.ip) as part of the request object. so we should be able to get the ip from just doing `req.ip`
- We utilise a nginx box in front of the web io boxes for traffic forwarding... so to get the original ip in these situatiosn, we need to check [req.ips](http://expressjs.com/en/api.html#req.ips) - make sure we cater for both cases.
- location info we should be able to extract from the IP address by utilising a nodejs library
- the other info should be extractable from the user agent (again, with a 3rd party nodejs library)
- We will need to come up with a format for plugin user agent and parse it for the info accordingly.

### Tasks
to be filled out by the developer

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #2611

Successful login requests should be logged in the database. The login records should include the following:

{
   _id: <string> //session ID so we can refer to it if necessary (to check if the session is still active?)
   loginTime: <MongoDB date>, //date time when the user logged in
   referrer: <string>, //the domain the request came from (e.g. https://www.3drepo.io or somewhere else which embedded our viewer),
   ipAddr: <string>, //ip address of the request
   location: {
        country: <string>, //country code
        city: <string>
    },
   application: {
       name: <string>, //If this came from a plugin, this should be Revit/Civils/Autocad/Navis, if it's browser it should be "chrome", "firefox" etc.
       version: <string>,
       type: <string> //"browser" or "plugin" or "unknown" (unknown used if there's no user agent information)
   },
   engine: {
       name: <string>, //If this came from a plugin, this should say "3drepoplugin"
       version: <string> //If this came from a plugin, this should be the version of the plugin
   },
   os: {
      name: <string>,
      version: <string>
   },
   device: <string> //type of device, e.g. mobile, computer etc        
}

### CODE REVIEW

File: backend/models/loginRecord.js [hunk 1]
```diff
@@ -0,0 +1,127 @@
+/**
+ *	Copyright (C) 2014 3D Repo Ltd
```
keifgwinn, line 2: 2021?

carmenfan: yes this should be 2021.

@sanmont3drepo  also did the same on another PR... Is there a vscode template that auto inserting this with the wrong date?

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+ *
+ *	You should have received a copy of the GNU Affero General Public License
+ *	along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+ "use strict";
+ const db = require("../handler/db");
+ 
+ //detects edge as browser but not device
+ var uaParserJs = require('ua-parser-js');
```
carmenfan, line 22: use `const`.

Always `const` unless you have to reassign them.

Var is discouraged for ES6+ javascript. You shouldn't really find yourself needing to use var anymore. It has some weird scoping definitions which you want to avoid https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+ var device = require('device');
+ var ip2location = require('ip-to-location');
+
+ const LoginRecord = {};
+ 
+ LoginRecord.saveLoginRecord = async (req) => {
+        var loginRecord = {
```
carmenfan, line 29: const

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+                loginTime : new Date(),
+                ipAddr: req.ips[0] || req.ip                    
+        }
+
+        const referrer = req.header('Referer');
+        if(referrer != null){
+                loginRecord.referrer = referrer;
+        }
+
+        const location = await getLocationFromIPAddress(loginRecord.ipAddr);
```
carmenfan, line 40: It's a bit here or there, but you could use destructuring assignments here (and people seem to prefer this in modern javascript programming)
```suggestion
        const {country_name, city} = await getLocationFromIPAddress(loginRecord.ipAddr);   
        loginRecord.location = {
                country: country_name,
                city
        };
        
```

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        const location = await getLocationFromIPAddress(loginRecord.ipAddr);   
+        loginRecord.location = {
+                country: location.country_name,
+                city: location.city
+        };    
+       
+         const userAgentString = req.headers['user-agent'];
+         var uaInfo = {};
```
carmenfan, line 47:
```suggestion
         const uaInfo = isUserAgentFromPlugin(userAgentString) ? getUserAgentInfoFromPlugin(userAgentString) : getUserAgentInfoFromBrowser(userAgentString);
```

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+         else{
+                uaInfo = getUserAgentInfoFromBrowser(userAgentString);
+         }
+
+         loginRecord.application = uaInfo.application;
+         loginRecord.engine = uaInfo.engine;
+         loginRecord.os = uaInfo.os;
+         loginRecord.device = uaInfo.device;          
+
+         const col = await db.getCollection("loginRecords", req.body.username);
```
carmenfan, line 61: avoid getting your own collection object to do the insert, call [this ](https://github.com/3drepo/3drepo.io/blob/master/backend/handler/db.js#L184) in db.js

The idea is that if we abstract ourselves away from the db driver then it'd make our life easier should they change their API in future upgrades, also less of these double awaits everywhere

so you should be able to just do this
```suggestion
    await db.insert("loginRecords", req.body.username, loginRecord);
```

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+         await col.insertOne(loginRecord);
+ }
+ 
+ //Format: 
+ //PLUGIN: {OS Name}/{OS Version} {Host Software Name}/{Host Software Version} {Plugin Type}/{Plugin Version}
+ //Example:
+ //PLUGIN: Windows/10.0.19042.0 REVIT/2021.1 PUBLISH/4.15.0
+ function getUserAgentInfoFromPlugin(userAgentString){
```
carmenfan, line 69: declare functions as arrow functions
```suggestion
const getUserAgentInfoFromPlugin => (userAgentString) {
```

There's some nuance difference between this and classic function definition, main benefit is the binding of `this` - the `this` keyword is super confusing in the JavaScript world...

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+        const userAgentInfo = {};
```
carmenfan, line 70: unlike classic programming languages there isn't much need or point to pre-declare your variable like this.
it's less code and arugably clearer to just merge this declaration with the assignments below:

```suggestion
        const userAgentInfo = {
            application: {
                name: userAgentComponents[1].split('/')[0],
                /*...*/
            },
            engine: {
              /*...*/
            }
            /*...*/
        };        
```

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+         await col.insertOne(loginRecord);
+ }
+ 
+ //Format: 
+ //PLUGIN: {OS Name}/{OS Version} {Host Software Name}/{Host Software Version} {Plugin Type}/{Plugin Version}
+ //Example:
+ //PLUGIN: Windows/10.0.19042.0 REVIT/2021.1 PUBLISH/4.15.0
+ function getUserAgentInfoFromPlugin(userAgentString){
+        const userAgentInfo = {};        
+        const userAgentComponents = userAgentString.replace("PLUGIN: ", "").split(' ');
```
carmenfan, line 71: Here is a place where i think it'll benefit with destructuring assignements
```suggestion
        const [osInfo, appInfo, engineInfo]= userAgentString.replace("PLUGIN: ", "").split(' ');                  
```

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+ //Format: 
+ //PLUGIN: {OS Name}/{OS Version} {Host Software Name}/{Host Software Version} {Plugin Type}/{Plugin Version}
+ //Example:
+ //PLUGIN: Windows/10.0.19042.0 REVIT/2021.1 PUBLISH/4.15.0
+ function getUserAgentInfoFromPlugin(userAgentString){
+        const userAgentInfo = {};        
+        const userAgentComponents = userAgentString.replace("PLUGIN: ", "").split(' ');                  
+
+        userAgentInfo.application = {
+                name: userAgentComponents[1].split('/')[0],
```
carmenfan, line 74: Probably best to avoid keep calling split on every line, store it in a variable

```suggestion
                const appData = userAgentComponents[1].split('/');
                userAgentInfo.application = {
                    name: appData[0],
```

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+                osVersion: userAgentComponents[0].split('/')[1]
+        }
+
+        userAgentInfo.device = "desktop";
+
+        return userAgentInfo;
+ }
+
+ function getUserAgentInfoFromBrowser(userAgentString){
+        const userAgentInfo = {};
```
carmenfan, line 95: Same here, just declare this as you assign it below

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+        userAgentInfo.device = "desktop";
+
+        return userAgentInfo;
+ }
+
+ function getUserAgentInfoFromBrowser(userAgentString){
+        const userAgentInfo = {};        
+        const userAgentObject = uaParserJs(userAgentString);       
+
+        userAgentInfo.application = {
```
carmenfan, line 98: you could use the spread operator to copy the object from browser:

```suggestion
        userAgentInfo.application = {
            ...userAgentObject.browser,
            type: "browser"
        }
            
```

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+                name: userAgentObject.browser.name,
+                version: userAgentObject.browser.version,
+                type: "browser"        
+        }
+        
+        userAgentInfo.engine = {
```
carmenfan, line 104:
```suggestion
        userAgentInfo.engine = userAgentObject.engine
```

carmenfan: Infact you could just do this:

````js
const {browser, engine, os} = uaParserJs(userAgentString);
const userAgentInfo = {
    application : {...browser, type: "browser"},
    engine,
    os,
    device: device(userAgentString).type
}

````

File: backend/models/loginRecord.js [hunk 14]
```diff
@@ -0,0 +1,127 @@
...
+                name: userAgentObject.engine.name,
+                version: userAgentObject.engine.version
+        }
+
+        userAgentInfo.os = {
+                osName: userAgentObject.os.name,
+                osVersion: userAgentObject.os.version
+        }
+
+        userAgentInfo.device = device(userAgentString).type;
```
carmenfan, line 114: What happens if User agent is null? would this break the app?

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+ }
+
+ function getUserAgentInfoFromBrowser(userAgentString){
+        const userAgentInfo = {};        
+        const userAgentObject = uaParserJs(userAgentString);       
+
+        userAgentInfo.application = {
+                name: userAgentObject.browser.name,
+                version: userAgentObject.browser.version,
+                type: "browser"
```
carmenfan, line 101: what if UA is empty? would this be recording it as browser?

backend/models/loginRecord.js
```diff
@@ -0,0 +1,127 @@
...
+
+        return userAgentInfo;
+}
+
+async function getLocationFromIPAddress(ipAddress){
+        return await ip2location.fetch(ipAddress);
+}
+
+function isUserAgentFromPlugin(userAgent){        
+        return userAgent.split(" ")[0] == "PLUGIN:";
```
carmenfan, line 124: You almost never use `==` in javascript - there's a very small amount of situations where you find yourself needing to use `==` instead of `===` legitimately

```suggestion
        return userAgent.split(" ")[0] === "PLUGIN:";
```

backend/models/loginRecord.js
carmenfan, line 22: Can we make the first letter to be in caps for requires?

backend/routes/auth.js
```diff
@@ -570,6 +572,8 @@ function createSession(place, req, res, next, user) {
 			return removeSessions(ids);
 		}).then(() => {
 			responseCodes.respond(place, req, res, next, responseCodes.OK, user);
+		}).then(()=>{
```
carmenfan, line 575: This would imply that you would only want to save the record after you respond. We don't need this to be done synchronously.

I would probably put it on L556

```` js 
.then(() => { 
   LoginRecord.saveLoginRecord(req);
   return getSessionsByUsername(user.username));
}
````

**Review by carmenfan — CHANGES_REQUESTED** [review]
@ChristosTsiotsias  Pretty good first attempt at javascript 👍

File: backend/models/loginRecord.js (hunk 2 above)
Comment (carmenfan) [reply backend/models/loginRecord.js:22]: Also you haven't updated package.json. Meaning anyone who's running this won't have the library installed.

When you introduce a new package into the repo you should run `yarn add <package name>`

This will add it into package.json so we can pull it automatically when we run `yarn install` (A bit like nuget)

File: backend/models/loginRecord.js (hunk 1 above)
Comment (ChristosTsiotsias) [reply backend/models/loginRecord.js:2]: Woops .. I actually copied and pasted it from another file and forgot to change the year. Wasn't sure if we needed this comment block on top or not so I thought Im going to include it anyway

File: backend/models/loginRecord.js [hunk 18]
```diff
@@ -0,0 +1,127 @@
...
+ 
+ LoginRecord.saveLoginRecord = async (req) => {
+        var loginRecord = {
+                _id : req.sessionID,
+                loginTime : new Date(),
+                ipAddr: req.ips[0] || req.ip                    
+        }
+
+        const referrer = req.header('Referer');
+        if(referrer != null){
```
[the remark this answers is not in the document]
ChristosTsiotsias: I want to check whether this is not null though.. would that be 

if(referrer){
}

backend/models/loginRecord.js
ChristosTsiotsias: browser has both major and full version so in that case we might not want to do that ?

image

backend/models/loginRecord.js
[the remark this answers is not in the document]
carmenfan: yes well spotted 😆

backend/models/loginRecord.js
carmenfan: actually we may as well record both if it gives us both.

the major version is probably more useful for stats purposes but we may want to check specific versions too

backend/models/loginRecord.js
```diff
@@ -15,113 +15,97 @@
...
+	};
+
+	return userAgentInfo;
+};
+
+const getUserAgentInfoFromBrowser = (userAgentString) => {
+
+	const { browser, engine, os } = UaParserJs(userAgentString);
+	const userAgentInfo = {
+		application: { ...browser, type: browser === null ? null : "browser"},
```
carmenfan, line 94: doesn't look like this happens, i just tried it with `undefined` and an empty string, these are the results:

image


So the library always return the object, just the names might be undefined.

so I think what we want to do here is this:
```suggestion
		application: browser.name ? {...browser, type: "browser"} : {type: "unknown"},
```

backend/package.json
```diff
@@ -31,6 +31,7 @@
     "connect-mongodb-session": "2.1.0",
     "cors": "2.8.5",
     "csv-parse": "4.8.5",
+    "device": "^0.3.12",
```
carmenfan, line 34: lock the versioning (applies to the other lines you added too)

^ means minimum version, we often end up with compilation problems when the libraries bump their major version and change the apis. so we try to lock down the version and manually upgrade when needed.
```suggestion
    "device": "0.3.12",

```

carmenfan (changes requested):

carmenfan:
@keifgwinn you want this in elastic, can you instruct what to do here? 

at [this point](https://github.com/3drepo/3drepo.io/pull/2648/files#diff-f35478d063da884b8182ec0bf363c1b7c01490c80fe4288b8c40d1a0e3145e97R55) We'd have the login record as per the schema from the issue

backend/models/loginRecord.js
```diff
@@ -15,113 +15,97 @@
...
+	loginRecord.location = {
+		country: country_name,
+		city
+	};
+
+	const userAgentString = req.headers["user-agent"];
+	const uaInfo = isUserAgentFromPlugin(userAgentString) ?
+		getUserAgentInfoFromPlugin(userAgentString) : getUserAgentInfoFromBrowser(userAgentString);
+
+	loginRecord.application = uaInfo.application;
```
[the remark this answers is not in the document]
ChristosTsiotsias: Thats complaining I think because loginRecord is in both sides (?) but I did something very simillar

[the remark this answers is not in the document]
carmenfan: did you change it to `let` instead of `const`? that might be why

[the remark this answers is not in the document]
ChristosTsiotsias: that was it 😆😆

backend/models/loginRecord.js
```diff
@@ -0,0 +1,134 @@
...
+
+// detects edge as browser but not device
+const UaParserJs = require("ua-parser-js");
+const Device = require("device");
+const Ip2location = require("ip-to-location");
+const Elastic = require("../handler/elastic");
+
+const LoginRecord = {};
+
+LoginRecord.saveLoginRecord = async (req) => {
```
carmenfan, line 29: Actually can we pass in the user agent and ip here only?

If we can help it I want to avoid req being accessible anywhere within models folder.

config/config_sample.js
```diff
@@ -129,6 +129,10 @@ module.exports = {
 		secretKey: "",  // This is for validating the user identity in intercom
 		accessToken: "" // This is for the api request to intercom in the backend
 	},
-
+    elastic: {
+        enabled: false,
```
carmenfan, line 133: remove, it's not even used.

carmenfan (changes requested):
image
We need to use v6 of the api until we upgrade nodejs.

backend/models/loginRecord.js
ChristosTsiotsias: we use some other properties of req like sessionId, referrer and body.username. Shall I pass all the ones we use as params ?

carmenfan: yep

backend/handler/elastic.js
```diff
@@ -0,0 +1,141 @@
...
+		auth: {
+			username: elasticCredentials[0],
+			password: elasticCredentials[1]
+		},
+		reload_connections: elasticConfig.reload_connections,
+		maxRetries: elasticConfig.maxRetries,
+		request_timeout: elasticConfig.request_timeout
+	};
+
+	const client = new Client(config);
```
carmenfan, line 69: as discussed, we can just leave the burden of structuring the config to the user and assume elastic config is how elastic wants it.

so we can just do this 
```suggestion
	const client = new Client(elasticConfig);
```

backend/handler/elastic.js
```diff
@@ -0,0 +1,141 @@
...
+	try {
+		await client.cluster.health();
+		systemLogger.logInfo(`Succesfully connected to ${elasticConfig.cloudId.trim()}`);
+		await establishIndices(client);
+	} catch (err) {
+		systemLogger.logError("Health check failed on elastic connection, please check settings.");
+		return;
```
carmenfan, line 76: we do want to fail this As it means the user did put in elastic config and have an intention to connect to it.

```suggestion
		throw "Health check failed on elastic connection, please check settings.";
```

backend/handler/elastic.js
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	return client;
+};
+
+const establishIndices = async (elasticClient)=>{
+	indicesMappings.forEach(async (indexMapping) => {
```
carmenfan, line 83: The parameter in forEach is essentially a callback function. so unless you know what you're doing it's typically not great to use async, as that promise is not waited on. i.e. as it stands, `establishIndices` will return before the indices are established.

what you want to do is actually  capture the returned promise from each iteration and wait on it, so the best thing to do here is probably `map`

```suggestion
	return Promise.all(indicesMappings.map(async (indexMapping) => {
```

carmenfan: if you do this then you won't have to keep doing `indexMapping.` in the function

```suggestion
	indicesMappings.forEach(async ({index, mapping}) => {
```

backend/handler/elastic.js
```diff
@@ -0,0 +1,141 @@
...
+const elasticClientPromise = createElasticClient();
+
+const createElasticRecord = async (elasticIndex, elasticBody, id) => {
+	try {
+		const elasticClient = await elasticClientPromise;
+
+		if (elasticBody) {
+			await elasticClient.index({
+				index: elasticIndex,
+				id: id,
```
carmenfan, line 108:
```suggestion
				id,
```

backend/handler/elastic.js
```diff
@@ -0,0 +1,141 @@
...
+
+const elasticClientPromise = createElasticClient();
+
+const createElasticRecord = async (elasticIndex, elasticBody, id) => {
+	try {
+		const elasticClient = await elasticClientPromise;
+
+		if (elasticBody) {
+			await elasticClient.index({
+				index: elasticIndex,
```
carmenfan, line 107: if you rename `elasticIndex` to `index` then this can be simplified

```suggestion
				index,
```

backend/handler/elastic.js
```diff
@@ -0,0 +1,141 @@
...
+				id: id,
+				refresh: true,
+				body: elasticBody
```
carmenfan, line 110: same here

backend/handler/elastic.js
```diff
@@ -0,0 +1,141 @@
...
+			});
+			systemLogger.logInfo(`created doc ${elasticIndex} ${Object.values(elasticBody).toString()}`);
```
carmenfan, line 112: probably makes more sense to just do JSON.stringify here? 

```suggestion
			systemLogger.logInfo(`created doc ${elasticIndex} ${JSON.stringify(elasticBody)}`);
```

carmenfan (changes requested):

carmenfan (dismissed):
Looks fine now. I added a commit so we force the initialisation to happen synchronously otherwise the application won't exit on error.

@keifgwinn  can you verify that elastic is getting the data you expect?

carmenfan:
> if this is the case we should update the example conf file with a complete copy of what's required for elastic.

I've added a commit to change it to just point to elastic documentation. I think that's more useful.

keifgwinn (dismissed):
Tested on dev cluster, looks good to me.

PR #2649, opened 2021-06-16
Issue #2637 - Colour override bugs fixes

### LINKED ISSUE #2637: Element Colour Override not consistent across Views and Issues
<details>
<summary> Expand to see original post </summary>

### Description
Transparency and colour overrides with Smart Groups do not come through consistently when clicking on Issues and Views that contain these overrides. The camera position updates correctly but the overrides fail to save. This is happening in prod.io. Does it require a hotfix? @mosment3DRepo 

### Steps to replicate
- Open this model: [https://www.3drepo.io/viewer/PowerBIVisual/1099fb90-b329-11eb-87e0-f722fb2ce12b](url) 
- Create a view with colour override (you can use the Smart groups that are already created) and save it
- Create an issue that has a view with colour override and save it with the view as a screenshot
- Refresh the browser page and toggle between different views and issues in the list to see how colour and transparency overrides behave

### Current Behaviour
- **Issue no 1:** Clicking anywhere in space on the model viewer resets view and removes overrides 
- **Issue no 2:** Toggle between different views in the Views list: camera position updates but some coloured elements come through while others don't, 
- **Issue no 3:** Click on different issues in the Issues list and open them: camera position updates but some elements overrides are missing
- **Issue no 4:** Click on different views with colour overrides. Some elements are highlighted in yellow even if view doesn't contain a specific selection. 

### Expected Behaviour
1. Clicking anywhere in space should retain whatever colour/transparency overrides is set for the opened view and not make them disappear
2. Clicking on any Issue or View in the lists should retain all colour/transparency overrides
3. Clicking on any View with overrides should not select and highlight elements in yellow

[https://www.dropbox.com/preview/3D%20Repo%20Team%20Folder/Support%20Team/GitHub%20Issues/Videos/Element%20colour%20overrides%201.mp4?role=work](url)
[https://www.dropbox.com/home/3D%20Repo%20Team%20Folder/Support%20Team/GitHub%20Issues/Videos?preview=Element+colour+overrides+2.mp4](url)

</details>

**Bug list:**
- [ ] [Colour override's alpha values are cleared if you toggle between 2 views](https://github.com/3drepo/3drepo.io/issues/2637#issuecomment-856785906)
- [ ] [When you click on an issue, it's fetching the groups information but not applying it](https://github.com/3drepo/3drepo.io/issues/2637#issuecomment-856829804)
- [ ] [Similar colour override problems in views](https://github.com/3drepo/3drepo.io/issues/2637#issuecomment-856839773)
- [ ] [Colour overrides doesn't go away when you navigate outside of the issue](https://github.com/3drepo/3drepo.io/issues/2637#issuecomment-856843967)
- [ ] [Only one of the colour overrides are being applied when I have 3](https://github.com/3drepo/3drepo.io/issues/2637#issuecomment-856847853)
- [ ] [We may want to add a spinning wheel icon when the groups are loading?](https://github.com/3drepo/3drepo.io/issues/2637#issuecomment-856854767)

sanmont3drepo:
This fixes #2637 

#### Description
e.g.
- FIxed bad iteration for the viewpoint groups (it wasnt fetching all the groups for a viewpoint)
- Updated transparencies colour override to work also with gl colours
- Deseleting an issue/risk deselects its colour also

#### Test cases
see the test cases in #2637


carmenfan (approved):

carmenfan (changes requested):
- Colour overrides/highlights shouldn't reset if i try to create a new issue:
aa

> Deseleting an issue/risk deselects its colour also

I don't think it should be doing that, and I think this is the problem causing the issue above... The context of resetting the view is very different to deselecting an issue in my opinion, hence the previous conversation we had about there shouldn't an idea of a "selected view", as it could be an aggregate of many actions (including mouse clicks).  It's a bit half baked as if we are really treating view to be a state then we should also be resetting the camera viewpoint and reset hidden objects. But I don't think it's intended to be used this way.

mosment3DRepo (approved):
Reviewed and confirmed to work by @miadibe

carmenfan (approved):

PR #2651, opened 2021-06-21
Issue #2634 - Fixed creating normal groups with 0 meshes and refactor

### LINKED ISSUE #2634: Normal groups update can have 0 meshes
### Description
- This is happening in: both 

### Steps to replicate
- Open any model
- Create a normal group
- Deselect all meshes
- Update the group

### Expected Behaviour
- **Normal** groups should have the save button disabled when deselecting all meshes

sanmont3drepo:
This fixes #2634 

#### Description
- Refactored the groups edition component to be have a simpler  method to determine if the form is dirty or valid
- Added the more than 0 meshes selected and normal group type as criteria to prove that the form is valid

#### Test cases
- Try to create or update a normal group to have 0 meshes selected, now the button is disabled all the time.

### DIFF
--- a/frontend/modules/groups/groups.sagas.ts
+++ b/frontend/modules/groups/groups.sagas.ts
@@ -407,8 +407,6 @@ function* resetToSavedSelection({ groupId }) {
 
 	yield all([
 		put(GroupsActions.selectGroup(activeGroup))
-		// TODO: has to restore highlighted groups
-		// put(GroupsActions.setComponentState({ newGroup: activeGroup }))
 	]);
 }
 
--- a/frontend/modules/groups/groups.sagas.ts
+++ b/frontend/modules/groups/groups.sagas.ts
@@ -15,6 +15,7 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { cloneDeep } from 'lodash';
 import { all, put, select, takeLatest } from 'redux-saga/effects';
 
 import { CHAT_CHANNELS } from '../../constants/chat';
@@ -228,7 +229,7 @@ function* showDetails({ group, revision }) {
 		yield put(GroupsActions.highlightGroup(group));
 		yield put(GroupsActions.setComponentState({
 			showDetails: true,
-			newGroup: {...group},
+			newGroup: cloneDeep(group),
 			criteriaFieldState: INITIAL_CRITERIA_FIELD_STATE
 		}));
 	} catch (error) {
@@ -407,8 +408,9 @@ function* resetToSavedSelection({ groupId }) {
 	activeGroup.rules = (activeGroup || { rules: [] }).rules;
 
 	yield all([
-		put(GroupsActions.selectGroup(activeGroup)),
-		put(GroupsActions.setComponentState({ newGroup: activeGroup }))
+		put(GroupsActions.selectGroup(activeGroup))
+		// TODO: has to restore highlighted groups
+		// put(GroupsActions.setComponentState({ newGroup: activeGroup }))
 	]);
 }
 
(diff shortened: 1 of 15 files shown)


frontend/routes/viewerGui/components/groups/components/groupDetails/groupDetails.component.tsx
```diff
@@ -108,82 +112,93 @@ export class GroupDetails extends React.PureComponent<IProps, IState> {
...
 		/>
 	));
 
+	public areSharedIdsChanged = (selectedNodes = [], groupObjects = []) => {
+		const toFullIdsDict = (dict, val) => {
+			val.shared_ids.forEach((e) => dict[val.account + '.' + val.model + '.' + e] = true);
+			return dict;
+		};
+
+		selectedNodes = selectedNodes.reduce(toFullIdsDict, {});
```
carmenfan, line 137: My worry is when the number of objects get huge (and it can be if someone has a massive federation and decided to select all the objects somehow...), this will grind to a halt.

I think we need some early rejects, like if the lengths doesn't compare (both the account/model level and the number of shared Ids within each account/model entry)

And I think we can assume the order as well (unity keep these as a list and just spit it out anyway so the order can be assumed) - so I don't think we need the reduction. It's much better to act quicker but potentially false positives than to grind to a halt every time.


This is definitely neater than what I'm proposing, but speed matters here i think!

carmenfan (changes requested):

frontend/modules/groups/groups.sagas.ts
```diff
@@ -403,12 +402,13 @@ function* unsubscribeFromChanges({ teamspace, modelId }) {
...
 function* resetToSavedSelection({ groupId }) {
-	const activeGroup = yield select(selectUnalteredActiveGroupDetails);
+	const activeGroup = yield select(selectActiveGroupDetails);
 	activeGroup.rules = (activeGroup || { rules: [] }).rules;
 
 	yield all([
-		put(GroupsActions.selectGroup(activeGroup)),
-		put(GroupsActions.setComponentState({ newGroup: activeGroup }))
+		put(GroupsActions.selectGroup(activeGroup))
+		// TODO: has to restore highlighted groups
```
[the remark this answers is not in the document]
sanmont3drepo: I dont know to be honest, it was a quick note after testing. Right now is not doing that in prod or dev. 
This is how it is working:

group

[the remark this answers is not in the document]
carmenfan: I think the refresh button should disable if you're under new group... as it's kind of meaningless.

frontend/routes/viewerGui/components/groups/components/groupDetails/groupDetails.component.tsx
sanmont3drepo: Unfortunately we cant assume order, as it depends on the order which the meshes were selected.

carmenfan: Yea, what I mean is if the order isn't right even if the same set of mesh we just return and say something has changed, as that would only be caused by user has done some interaction with the viewer.

carmenfan: it's a false positive, but it's better than going through with the whole process... I think.

carmenfan (changes requested):
- [x] Cancelling highlight shouldn't trigger the save button to be enabled if the group is a smart group - the rules were not changed:
aa

- [x] removing the rule then re-pasting it back in is causing for the save button to be enabled even though I have made no change:
aa

carmenfan (approved):

PR #2661, opened 2021-07-01
ISSUE #2561 - Return project id in proper format after creating it

### LINKED ISSUE #2561: New Project Invitations don't work until you do a hard refresh Ctrl + R
### Description

When I create a new project and head over to User Management to send the invite, I get an error message. The error message disappears and the invitation works after the hard refresh Ctrl + R. This issue happens regardless of whether you have models in that project or not.
 
- This is happening in: prod

### Steps to replicate
- Log in to your 3D Repo account
- Create a new project
- Head over to UM
- Create a new user invitation
- Select the new project
- Send the Invite

### Current Behaviour
- Error message appears **Invalid Project ID**

### Expected Behaviour
- Send the invite as you would after Ctrl + R

[BUG Video](https://www.dropbox.com/preview/3D%20Repo%20Team%20Folder/Support%20Team/GitHub%20Issues/Videos/User%20Management_Invitation.mp4)

image

m1chalkubiak:
This fixes #2561

### DIFF
--- a/backend/models/project.js
+++ b/backend/models/project.js
@@ -128,6 +128,7 @@
 
 		await db.insert(teamspace, PROJECTS_COLLECTION_NAME, project);
 
+		project._id = utils.uuidToString(project._id);
 		project.permissions = C.IMPLIED_PERM[C.PERM_PROJECT_ADMIN].project;
 
 		return project;


m1chalkubiak:
Hi @carmenfan
I gave this task yesterday to Krzysztof (@klesnyatapptension), who is just getting started with this project. Task required getting acquainted with several functionalities in the application, which seemed to be a good way to get to know it. However, as it turned out, the problem appears because in response to the request to create a new project we received an incorrect project id.

I looked at it more closely and it turned out that when returning data about the newly created project, there was no UUID to string conversion. And because in this situation we update the state of the application ourselves, we had a project with an incorrect id. However, when refreshing the page, the entire list of projects is loaded from the endpoint where the ids are already in the correct format.

After adding a small fix in BE it now seems to be working fine.

carmenfan (approved):

PR #2667, opened 2021-07-02
ISSUE_1914 - Email Verification token reduced in size and reusable method for ID added

### LINKED ISSUE #1914: Register verify token is extremely long
It's 128 characters for not a lot of reason:
image

ChristosTsiotsias:
This fixes #1914

#### Description
Email verification token Id reduced from 128 chars to 32 chars and a method for generating random ids is added to utils to be reused by other places

### DIFF
--- a/backend/utils.js
+++ b/backend/utils.js
@@ -97,8 +97,8 @@ function Utils() {
 		} , {});
 	};
 
-	this.getRandomBytesToHexString = (length = 16) => {
-		return crypto.randomBytes(length).toString("hex");
+	this.generateHashString = (length = 32) => {
+		return crypto.randomBytes(length / 2).toString("hex");
 	};
 
 	/** *****************************************************************************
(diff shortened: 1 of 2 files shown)


backend/utils.js
```diff
@@ -96,6 +97,10 @@ function Utils() {
 		} , {});
 	};
 
+	this.getRandomBytesToHexString = (length = 16) => {
```
carmenfan, line 100: 2 things:

1. Can we rename this to `generateHashString`? as this is esssentially the purpose of this function
2. It's confusing for the caller to pass in a length but get a length *2 string length as a return. Can we make it so it takes in 2x length so it matches what the string length would be ultimately? i.e
```suggestion
	this.getRandomBytesToHexString = (length = 32) => {
	    return crypto.randomBytes(length/2).toString("hex");
```

carmenfan (changes requested):
just 2 tiny things

carmenfan (approved):

PR #2677, opened 2021-07-07
ISSUE #2662 Support for PixelStreaming Viewer

### LINKED ISSUE #2662: Add PixelStreaming Viewer Support to UnityUtil
### Description
The UnityUtil class is used to initialise a Unity viewer. This should be updated to allow initialising an Unreal PixelStreaming viewer instead.

The implementation of the PixelStreaming viewer will be in a separate file, delivered from the signalling server that is used to make the connection to a PixelStreaming process. It is assumed that this source will be loaded before the command to UnityUtil is given.

### Goals
- [x] The existing UnityUtil API is supported
- [x] When the viewer is set to Unreal, a PixelStreaming instance is established that operates as Unity does

### Tasks
- [x] Implement a method to initialise a PixelStreaming [viewer](https://github.com/3drepo/3drepoUnreal/issues/68)

### Related Resources
Unreal Issue: https://github.com/3drepo/3drepoUnreal/issues/40
Viewer Issue: https://github.com/3drepo/3drepoUnreal/issues/68
PlanBase Issue: https://github.com/3drepo/PlanBase/issues/9

sebjf:
This fixes #2662

#### Description
This PR updates the UnityUtil class so that a pixel streaming based viewer can transparently be loaded in place of the Unity WebGL based one.

To support this, a new method, `loadUnreal` has been added that can be called instead of `loadUnity`. When this is done a PixelStreaming viewer is loaded but the UnityUtil class & APIs otherwise function identically.

The implementation of the pixel streaming viewer is split across three services: the actual viewer process that runs remotely, a matchmaking server that makes the connection between the browser & the process, and the browser side functionality to make a connection to the matchmaker & set up the DOM.

Before calling `loadUnreal` a page must load the PixelStreaming dependencies. These are four static files, three of which are currently hosted by the matchmaking server.

For example, the [PlanBase](https://github.com/3drepo/PlanBase/issues/9) `head` contains the following snippet:

```
<!-- These links and scripts import the Unreal Pixel Streaming browser side dependencies. Once loaded a PixelStreaming instance can be created through UnityUtil.loadUnreal() -->
<script type="text/javascript" src="https://webrtc.github.io/adapter/adapter-latest.js"></script>
<link type="text/css" rel="stylesheet" href="http://localhost/player.css">
<script type="text/javascript" src="http://localhost/scripts/webRtcPlayer.js"></script>
<script type="text/javascript" src="http://localhost/scripts/app.js"></script>
```

To load Unreal instead of Unity, an application would replace the call, e.g.

```
UnityUtil.loadUnity('unity', PREFIX + '/unity/Build/unity.json', 2130706432 / 10);
```

With, e.g.

```
UnityUtil.loadUnreal(div, {serverUri: "ws://localhost/", autoConnect: false});
```

#### Test cases
Testing is performed in the dependent issue https://github.com/3drepo/3drepoUnreal/issues/68

### DIFF
--- a/frontend/globals/unity-util.ts
+++ b/frontend/globals/unity-util.ts
@@ -208,8 +208,8 @@ export class UnityUtil {
 	}
 
 	/**
-	 * Load an Unreal PixelStreaming Viewer into the provided Canvas. The Pixel Streaming Viewer is a js class
-	 * that manages a WebRtc PeerConnection to a remote Unreal instance via a video stream & data channel
+	 * Load an Unreal PixelStreaming Viewer into the provided Div. The PixelStreaming Viewer is a class
+	 * that manages a WebRtc PeerConnection to a remote Unreal instance via a video stream & data channel.
  	 * @category Configurations
 	 * @param div - the html dom element that the viewer should be created under
 	 * @param config - an object describing the configuration of the viewer
@@ -218,7 +218,9 @@ export class UnityUtil {
 	 */
 	public static loadUnreal(div: any, config: UnrealConfig) : Promise<void> {
 		// The client must load the PixelStreaming dependencies before beginning if they wish to use PixelStreaming
-		// todo: test whether the pixel streaming functionality is available here/update this to load the pixel streaming functionality
+		if(typeof createPixelStreamingInstance != "function"){
+			console.error("An application must load the Unreal PixelStreaming dependencies before attemping to load the viewer");
+		}
 
 		if (Object.prototype.toString.call(div) === '[object String]') {
 			// tslint:disable-next-line
--- a/frontend/globals/unity-util.ts
+++ b/frontend/globals/unity-util.ts
@@ -14,10 +14,19 @@
 * You should have received a copy of the GNU Affero General Public License
 * along with this program. If not, see <http://www.gnu.org/licenses/>.
 */
-declare var SendMessage;
-declare var createUnityInstance;
 
-import { IS_FIREFOX } from '../helpers/browser';
+declare var createUnityInstance;
+declare var createPixelStreamingInstance;
+
+/**
+ * The required properties of the Pixel Streaming configuration for the loadUnreal method.
+ */
+interface UnrealConfig {
+	/** The fully qualified address of the matchmaking server */
+	serverUri: string,
+	/** Whether to attempt to connect as soon as createUnrealInstance is called; be aware that many browsers will not allow the session to start without explicit interaction from the user. */
+	autoConnect: boolean
+}
 
 export class UnityUtil {
 	/** @hidden */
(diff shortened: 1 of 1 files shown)


Charence (changes requested):
At the moment this fails lint as there's a class for `UnrealConfig` and another for `UnityUtil` when lint prefers to have 1 (we could change this).

Instead, it might be better to split this into 3 files [`ViewerUtil`, `UnityUtil`, `UnrealUtil`]? 

Typescript supports inheritance so we could make use of that. The common functions can be kept in `ViewerUtil` while the Unity/Unreal specific functions can be kept in their respective files?

This will involve quite a bit of refactoring/renaming but hopefully they will make things clearer. Functions dependent on the existing `unity-util` should be unaffected as that should remain in the same place with the same functions.

frontend/globals/unity-util.ts
```diff
@@ -196,6 +196,37 @@ export class UnityUtil {
...
+	 * Load an Unreal PixelStreaming Viewer into the provided Div. The PixelStreaming Viewer is a class
+	 * that manages a WebRtc PeerConnection to a remote Unreal instance via a video stream & data channel.
+ 	 * @category Configurations
+	 * @param div - the html dom element that the viewer should be created under
+	 * @param config - an object describing the configuration of the viewer
+	 * @return returns a promise which resolves when the game is loaded.
+	 */
+	public static loadUnreal(div: any, config: any): Promise<void> {
+		// The client must load the PixelStreaming dependencies before beginning if they wish to use PixelStreaming
+		if (typeof createPixelStreamingInstance !== 'function') {
```
carmenfan, line 209: @sebjf if this is just a function, can we get users to pass it in as part of `loadUnreal`?

Unlike sendMessage - we can't control as unity generated code depicts that. We want to avoid random variables that the users need to know to declare before calling stuff.

frontend/globals/unity-util.ts
```diff
@@ -196,6 +196,37 @@ export class UnityUtil {
...
 
+	/**
+	 * Load an Unreal PixelStreaming Viewer into the provided Div. The PixelStreaming Viewer is a class
+	 * that manages a WebRtc PeerConnection to a remote Unreal instance via a video stream & data channel.
+ 	 * @category Configurations
+	 * @param div - the html dom element that the viewer should be created under
+	 * @param config - an object describing the configuration of the viewer
+	 * @return returns a promise which resolves when the game is loaded.
+	 */
+	public static loadUnreal(div: any, config: any): Promise<void> {
```
carmenfan, line 207: instead of removing the class declaration and make config: any, i think we can use interface?

carmenfan (changes requested):

keifgwinn:
removed branch as it was interfering, re-run azp to recreate when it's needed

PR #2699, opened 2021-07-16
ISSUE 2412 - Upgrade mongodb driver

### LINKED ISSUE #2412: add functional tests to handler/db.js
It'll be good to have some functional test in the test suite for the db handler.

Mongo API seems to have some peculiarities, and the API calls may change without throwing exceptions (just ignore the entry quietly)

so It would be good if we have a set of functional tests so we know if that's broken (esp for testing driver upgrades), instead of trying to figure it out from the the endpoint tests.

Charence:
This fixes #2412 

#### Description
- Added tests for mongoDB handler
- Update no longer supported function calls
- Upgrade mongo driver version from 2.2.21 -> 3.6.10

#### Test cases
- All automated tests should pass
- All database operation should work as intended


backend/handler/db.js
```diff
@@ -268,6 +275,12 @@
 		return collection.update(query, data, options);
 	};
 
+	Handler.updateMany = async function (database, colName, query, data, upsert = false) {
```
carmenfan, line 278: you added an `updateMany` when `update` function is still here. 

If we want to stop using update then we should outright remove the function so it doesn't get confusing.

backend/handler/db.js
```diff
@@ -100,41 +142,20 @@
...
-				return db;
+				return db.db(database);
 			});
 		}
 	};
 
 	Handler.getAuthDB = function () {
 		return MongoClient.connect(getURL("admin"), connConfig).then(_db => {
-			return _db;
+			return _db.db("admin");
```
carmenfan, line 158: I don't know why we reconnect for authDB, it's a bit weird..

we should just return Handler.getDB("admin") ?

backend/models/user.js
```diff
@@ -169,7 +169,7 @@ User.authenticate =  async function (username, password) {
 	try {
 		await db.authenticate(user.user, password);
 	} catch (err) {
-		const resCode = utils.mongoErrorToResCode(err);
+		const resCode = err;
```
carmenfan, line 172: i don't think we need to do this declaration anymore? either rename `err` to resCode` or just use err throughout?

backend/test/integrated/impliedPermission.js
```diff
@@ -160,7 +160,11 @@ describe("Implied permission::", function () {
 		it("can delete project", function(done) {
 			agent
 				.delete(`/${sharedTeamspace}/projects/${project}`)
-				.expect(200, done);
+				.expect(200, function(err, res) {
+					console.log(err);
```
carmenfan, line 164: remove console log?

backend/handler/db.js
```diff
@@ -27,21 +27,61 @@
 		autoReconnect: true
 	};
 
+	function getGridFSBucket(database, collection, chunksize = null) {
+		return Handler.getDB(database).then(dbConn => {
```
carmenfan, line 31: I know you just moved functions around, but while we're here can we rewrite this to async await?

carmenfan (changes requested):

backend/handler/db.js
```diff
@@ -81,11 +121,13 @@
 		const collection = await Handler.getCollection(database, colName);
 		// NOTE: documentation states it should be { projection, sort } so when we upgrade we may have to change.
 		//       Also: projection stops working if you pass in a sort with empty obj.
-		if(sort) {
-			projection.sort = sort;
+		const options = { projection };
```
carmenfan, line 124: remove comment above as it's addressed?

carmenfan: also in `find`,  could we potentially rewrite it to be similar to findOne?

I had to write it that way because of the old API didn't work otherwise. be good if the way we access their api is consistent.

carmenfan (changes requested):

carmenfan:
@Charence  also i dont' know why merging staging in broke the tests... can you take a look 😆

carmenfan (approved):

PR #2724, opened 2021-08-03
ISSUE #2688 - New framework for v5

### LINKED ISSUE #2688: Revamping - New framework
### Description
The goal of the new framework is to provide a more developer friendly way of coding the frontend. This issue depends on #2686. 

### Goals
- [x] Have a new framework for integrating react and redux using hooks in a maner that is easier to use
- [x] Have to separate lint options for v4 and v5
- [x] Have a new testing framework for sagas
- [x] Have a new route /v5 for the new dashboard with a dummy page

### Tasks
- [x] Add service files for wrapping selectors to provide an easier redux integration
- [x] Add service files for wrapping actions to provide an easier redux integration
- [x] Add sagas testing  package
- [x] Update webpack config to support the new saga tests. Because we are testing the typescript source code, we need to write the tests in typescript and transpile them
- [x] Modify package.json to add a test sagas script
- [x] Add a saga testing example
- [x] Add a dummy layout page for v5
- [x] Add a route that shows the dummy layout
- [x] Add a command to package json to run linter for v4 and v5

### Related Resources
Dashboard scoping:
https://docs.google.com/document/d/1Kla4KIoo6Naw092LuZz-AKSz5kmKU1j9Zi3uY8bQBBQ/edit?usp=sharing

sanmont3drepo:
This fixes #2688 

#### Description
- Added wrappers to create selectors hooks directly in frontend
- Added wrappers to create actions directly from frontend (with autocomplete and typing)
- Added two different eslint files for v5 and v4
- Added a dummy route /v5 with a react page using the new framework
- Added an example of a saga test


carmenfan:
@m1chalkubiak  TODO:
1. add example of component
2. change mocha to jest (or use mocha)

m1chalkubiak:
@sanmont3drepo In general I'm not convinced of this approach with importing selectors, actions, etc from `v4`, at some point we will need to get rid of the v4 folder, then we will be forced to copy all the files used in this way to the right place? Wouldn't it be better to transfer what we really need and at the same time verify what we don't need, just during the implementation of specific pages/features?

sanmont3drepo:
> @sanmont3drepo In general I'm not convinced of this approach with importing selectors, actions, etc from `v4`, at some point we will need to get rid of the v4 folder, then we will be forced to copy all the files used in this way to the right place? Wouldn't it be better to transfer what we really need and at the same time verify what we don't need, just during the implementation of specific pages/features?

The idea is to exactly transfer / or migrate if necessary what we need when they are needed, to have control of what is happening and  not blinding migrate/copy everything without being tested.
Eventually v4  everything will be migrated to v5 and the v4  folder will be deleted.
The example that I used is just a mockup to show how the wrappers work and how they would be use in  a new v5 file.

m1chalkubiak:
> The idea is to exactly transfer / or migrate if necessary what we need when they are needed, to have control of what is happening and not blinding migrate/copy everything without being tested.
> Eventually v4 everything will be migrated to v5 and the v4 folder will be deleted.
> The example that I used is just a mockup to show how the wrappers work and how they would be use in a new v5 file.
> 
> The same as the saga test example

@sanmont3drepo Just to make sure, so for `currentUser.selector`, when we will come to moment when we want to actual use it, this example file should looks something like this:

```js
import { createSelector } from 'reselect';
import { wrapSelectors } from './selectors.helper';

const selectCurrentUserDomain = (state) => state.currentUser;

const selectCurrentUser = createSelector(
	selectCurrentUserDomain, (state) => state.currentUser || {},
);
const selectUsername = createSelector(
	selectCurrentUser, (state) => state.username || '',
);

export const CurrentUserSelectors = wrapSelectors({
	selectUsername,
});
```

For this example I'm exporting only `selectUsername` as this is the only selector used in v5 right now.

Is this make sense for You?

frontend/src/v4/modules/currentUser/currentUser.selectors.ts
```diff
@@ -17,6 +17,7 @@
 
 import { isNil } from 'lodash';
 import { createSelector } from 'reselect';
+import { boolean } from 'yup/lib/locale';
```
m1chalkubiak, line 20: it is probably added by mistake

frontend/src/main.tsx
```diff
@@ -25,24 +26,32 @@ import '@/v4/services/fontAwesome';
 import 'simplebar/dist/simplebar.min.css';
 
 import 'simplebar';
```
m1chalkubiak, line 28: I checked that this is for custom scrollbars, but we should use `simplebar-react`. Currently, both packages are installed, but in fact, neither is used. Do we want to leave it like that?

m1chalkubiak: @sanmont3drepo ignore this comment, I realize we are using one webpack configuration and one output file. 
But this can cause us problems in the future. Did you consider moving the new version of the application completely to the subfolder, along with the configuration of the webpack, package.json etc.?
I think this would be much safer. Inadvertently, it will be easy to break something in the old version of the application and not even notice it.

m1chalkubiak:
Do we want to completely abandon naming components like `COMPONENT_NAME.component.tsx`? If so, should files with styled components still be named `COMPONENT_NAME.styles.ts` or are we assuming some other convention?

sanmont3drepo:
> > The idea is to exactly transfer / or migrate if necessary what we need when they are needed, to have control of what is happening and not blinding migrate/copy everything without being tested.
> > Eventually v4 everything will be migrated to v5 and the v4 folder will be deleted.
> > The example that I used is just a mockup to show how the wrappers work and how they would be use in a new v5 file.
> > The same as the saga test example
> 
> @sanmont3drepo Just to make sure, so for `currentUser.selector`, when we will come to moment when we want to actual use it, this example file should looks something like this:
> 
> ```js
> import { createSelector } from 'reselect';
> import { wrapSelectors } from './selectors.helper';
> 
> const selectCurrentUserDomain = (state) => state.currentUser;
> 
> const selectCurrentUser = createSelector(
> 	selectCurrentUserDomain, (state) => state.currentUser || {},
> );
> const selectUsername = createSelector(
> 	selectCurrentUser, (state) => state.username || '',
> );
> 
> export const CurrentUserSelectors = wrapSelectors({
> 	selectUsername,
> });
> ```
> 
> For this example I'm exporting only `selectUsername` as this is the only selector used in v5 right now.
> 
> Is this make sense for You?

I renamed and moved the hooks generators wrappers and action dispatchers to a different directory. I hope this makes it clearer what they actually are. Im still not sure if that's the correct place or naming, but I feel that is a little improvement.

> Do we want to completely abandon naming components like `COMPONENT_NAME.component.tsx`? If so, should files with styled components still be named `COMPONENT_NAME.styles.ts` or are we assuming some other convention?

I think we can keep that naming convention, I dont see a problem with it.

carmenfan:
@m1chalkubiak if you can take another look here before the end of the day and make sure to make sure you're happy with things, then we can merge this and get everything in one place! :)

carmenfan:
@sanmont3drepo  since we're moving to jest on the backend, it might be good to use jest here too

m1chalkubiak (approved):

carmenfan (approved):
tried it, seems to be working ok. Merging!

PR #2725, opened 2021-08-04
ISSUE #2629 - Add category support in BIM panel

### LINKED ISSUE #2629: Category support in the BIM card
### Description
At the moment we store the category information as a prefix to the metadata label, with `::` as the delimiter.

E.g. 
    `Classification::Source` is a metadata value for `Source`, that belongs in the `Classification category`

This results in metadata looking like a mess and hard for users to make sense of it. e.g. Bim card of a certain object in 3drepo
image

When the same data presented in a different software:
image

### Goals
- [ ] **Introduce category grouping in the BIM data card.**
Key value needs to be looped through and split by the delimiter `::`, any keys that has the same categories should be grouped together for view. Note that the keys could nested categories, i.e. `1::2::3::key_a` should be placed inside category `1`, with a subcategory `2`, which has a sub category `3`. This would not be in the same grouping as `3::key_b`

The end result should look a bit like the measuring card:
image

Note:
- we should collapse all categories by default
- Add a button on the `...` button to allow collapse all

- [ ] **Ensure filtering still works**
Filtering behaviour should be similar to filtering activity list:
    - if it matches a category, the entire category (plus children) should show
    - if it matches a sub category, the parent category + the sub category should show (but not the siblings unless they are also  matches)
    - If a metadata label matches, it should show the parent ancestor categories, but not the sibling metadata (unless they are also matches)
- [ ] **Ensure favourites still work**
    - If I favourited a metadata label of a certain category, it should show it on the favourite tab.
    - It should not show on the favourite tab if it matches the label , but under a different category.
- [ ] **Ensure Copy group filter button still work**
    - The copy group filter button should copy exactly what we get right now (i.e. with the original string value with categories.)
- [ ] **Select Element... button still work**
    - Select element with the same value should trigger a API request that sends the same payload as right now (i.e. with original string value with categories)

### Tasks
To be filled out by the developer

m1chalkubiak:
This fixes #2629

Updates:
- [x] [Colouring in sub items](https://github.com/3drepo/3drepo.io/pull/2725#issuecomment-892813688)
- [x] [Incorrect Parameter name](https://github.com/3drepo/3drepo.io/pull/2725#issuecomment-892813792)
- [x] [Text overflow not working on Parameter name](https://github.com/3drepo/3drepo.io/pull/2725#issuecomment-892813871)
- [x] [Copy Group & Select Elements not working](https://github.com/3drepo/3drepo.io/pull/2725#issuecomment-892813990)
- [x] [Reorder List](https://github.com/3drepo/3drepo.io/pull/2725#issuecomment-892814082)


m1chalkubiak:
@carmenfan Please, check out the current BIM panel that supports grouping by categories. I was not able to find a project that would contain metadata with this data naming, so I was doing it based on the mocked data, assuming that if there are '::' characters in the name, then we should handle grouping. I tried to use the elements that were in the measurements panel indicated by you, but I do not know if I correctly interpreted the issues of styling this data. The main difference is the lack of different coloring for every odd line, but I think it would look weird when combined with nested categories.

carmenfan:
@mosment3DRepo can you take a look?

mosment3DRepo:
@m1chalkubiak I have added you as an admin to a project (mosment3ddev/IFC Parameter Testing) there are models in there with additional parameters.

I have a few comments:

mosment3DRepo:
### Colouring in sub items

As you mentioned the alternating row colouring doesn't work in this situation but I think it would be good to show which values exist on the child level by colouring them as per the below mock up

image

mosment3DRepo:
### Incorrect Parameter name

The child parameter name is currently just repeating the parent, where it should be the second part of the original string (after the :: delimeter):

image

mosment3DRepo:
### Text overflow not working on Parameter name

On long parameter names the text is no longer overflowing onto the next line, instead it is cropped. This causes an issue if 2 parameters start with the same text but are cut off at the end

image

mosment3DRepo:
### Copy Group & Select Elements not working

The 2 hover over buttons only appear to take one half of the parameter string on the child level, so they do not select anything, and the group logic is incorrect

image

mosment3DRepo:
### Reorder List

In order to make it more user friendly can we reorder the list of metadata so that it puts all the parent items at the bottom of the list and sorts all the orphans alphabetically.

At the moment the parents showing up alongside the other values is a bit jarring

image

mosment3DRepo (changes requested):
Added feedback and additional changes

m1chalkubiak:
> @m1chalkubiak I have added you as an admin to a project (mosment3ddev/IFC Parameter Testing) there are models in there with additional parameters.
> 
> I have a few comments:

@mosment3DRepo can you specify what account you have added to this project?
Most of the time in dev environment I'm using `TestDev` account created by a previous developer (dleszczynski@apptension.com)

carmenfan:
I think Matt added you on the staging database. I'll import it to
apptensions dB so you can test it easier

carmenfan:
@m1chalkubiak  you should able to find the same model in apptension db under Sergi/20b5fce0-f5dc-11eb-bb5f-bb77d6a1eae7

Sergi's password is still password 😆

m1chalkubiak:
@mosment3DRepo After the recently implemented changes, is there any feedback about this functionality?

carmenfan (changes requested):
@m1chalkubiak it's treating 0 in a strange way:
image

This is happening on any fields that has a value `0`.

Also, the text is overlapping if the key is too long:
image

m1chalkubiak:
@mosment3DRepo Could you please check this functionality? I made the last two fixes to the bugs reported by Carmen, and I think that is now ready for merge.

carmenfan:
@m1chalkubiak 

The alignments are still a bit weird, I have entries that look like this, where the value is intruding into the key column:
image

But there's also items where it wraps like you'd expect: 
image

BDavydov:
**Description**
Parameter filter functionality is a little less user-friendly for sub-items. Here is the deal. It filters the main list of parameters, but any sublists continuously collapse when you click on another model element. Without a doubt, users will find it irritating, should they choose to check 3 or more model elements.

Would it be possible to collapse the list once and keep it that way?

BIM Card Fold

BDavydov:
**Description**
It is not a direct issue with the BIM card. However, a comment on the overall User Experience.

**Search Filter in BIM Card**
You can find specific parameters using the keywords and in no particular order. So if the parameter is **_ArchiCAD::BIM Data_**, you will find it using the search phrase **_BIM Data_**.

**Search Filter in Smart Groups**
You can only find parameters if you spell them out as they're written. So if the parameter is **_ArchiCAD::BIM Data_** you won't find it using the search phrase **_BIM Data_**. Instead, your search must start with **_ArchiCAD::_** minimum.

Now, the above logic becomes a little tricky for Novice users to implement with the new BIM Card layout. Whilst I may find the parameter I need by simply providing a keyphrase **_elevation to_**, I will struggle to find that parameter in Smart Groups unless I remember to include the Parent value **_ArchiCADQuantities_**.
I fear some users will forget about that and get confused over the fact that some parameters appear, but those from the sublists don't.

**Filter the BIM Card**
BIM Card Filter

**Filter the Smart Groups Parameters**
Smart Group Filter

**Filter the Smart Groups Parameters (properly)**
Smart Groups FIlter Proper

m1chalkubiak:
> **Description**
> Parameter filter functionality is a little less user-friendly for sub-items. Here is the deal. It filters the main list of parameters, but any sublists continuously collapse when you click on another model element. Without a doubt, users will find it irritating, should they choose to check 3 or more model elements.
> 
> Would it be possible to collapse the list once and keep it that way?

@carmenfan If I understand correctly, Bogdan would like that in case we expand a category and chose another element with the same category, the previously expanded category should stay expanded.

Currently, it works that every time you click on an element, the whole list is rendered and according to the requirement described in this task categories are collapsed:
>  - we should collapse all categories by default

But I understand that the scenario described by Bogdan is not very user-friendly for the user.

If we would like to change it, I would have to move the information about which categories are expanded to the state of the application (currently it is the individual state of the component) when rendering a new list, check if there are categories that were previously marked as expanded and if yes, render them expanded.

Do we want to make such a change?

carmenfan:
@m1chalkubiak just had a chat with @BDavydov , and I think the use case he's talking about is more in a situation where the user has a particular field they're looking for, and identified as such via starred or search functionality. Essentially we want the search/starred to work in such a way that it always expand enough to show the entry that satisfy the search field. So I think if we do the following, it will be a happy medium:

- if you're looking at the starred tab - always expanded (as favourites are on the individual entries not categories)

- if you're in a search - if the search field matches the parameter name (i.e. searching apple in cat1::apples), expand the category so you see the "apples" automatically

- if you're in a search - if the search field matches the category name (i.e. searching cat1 in cat1::apples), expand the category.
	
- otherwise (i.e. not search, not starred tab) - collapsed by default like it is currently.

m1chalkubiak:
> @m1chalkubiak just had a chat with @BDavydov , and I think the use case he's talking about is more in a situation where the user has a particular field they're looking for, and identified as such via starred or search functionality. Essentially we want the search/starred to work in such a way that it always expand enough to show the entry that satisfy the search field. So I think if we do the following, it will be a happy medium:
> 
> * if you're looking at the starred tab - always expanded (as favourites are on the individual entries not categories)
> * if you're in a search - if the search field matches the parameter name (i.e. searching apple in cat1::apples), expand the category so you see the "apples" automatically
> * if you're in a search - if the search field matches the category name (i.e. searching cat1 in cat1::apples), expand the category.
> * otherwise (i.e. not search, not starred tab) - collapsed by default like it is currently.

Of course @carmenfan , I understand, I will start to make appropriate modifications to ensure this functionality works as described.

carmenfan:
@m1chalkubiak thanks!

Another thing I just noticed - we no longer have clickable links on the BIM tab:

ISSUE_2629:
image


staging:
image

m1chalkubiak:
> **Description**
> It is not a direct issue with the BIM card. However, a comment on the overall User Experience.
> 
> **Search Filter in BIM Card**
> You can find specific parameters using the keywords and in no particular order. So if the parameter is **_ArchiCAD::BIM Data_**, you will find it using the search phrase **_BIM Data_**.
> 
> **Search Filter in Smart Groups**
> You can only find parameters if you spell them out as they're written. So if the parameter is **_ArchiCAD::BIM Data_** you won't find it using the search phrase **_BIM Data_**. Instead, your search must start with **_ArchiCAD::_** minimum.
> 
> Now, the above logic becomes a little tricky for Novice users to implement with the new BIM Card layout. Whilst I may find the parameter I need by simply providing a keyphrase **_elevation to_**, I will struggle to find that parameter in Smart Groups unless I remember to include the Parent value **_ArchiCADQuantities_**.
> I fear some users will forget about that and get confused over the fact that some parameters appear, but those from the sublists don't.
> 

@carmenfan I think that we should do this in a separate issue,  do you agree? Basically, it's about improving the search in the groups tab, so it's not really related to this task. Nevertheless, its a good catch @BDavydov definitely worth improving.

carmenfan:
@m1chalkubiak  yes, agreed

m1chalkubiak:
@carmenfan @BDavydov Everything above seems to be addressed, can You check if it works as expected?

carmenfan (approved):

carmenfan:
@BDavydov approved. merging

PR #2728, opened 2021-08-05
ISSUE 2017 - Alluxio support

### LINKED ISSUE #2017: Add sufficient support to test Alluxio performance
### Description
@keifgwinn  is looking into replacing the use of fileShare/S3 with Alluxio, which will have us the flexibility of using one interface but plugging in different type of file storage solutions, with the added benefit of hot cache (as far as I understand it.

Keif has already setup an Alluxio server with the existing fileshare as the storage solution. The aim at this stage is do the minimal amount of work to connect to it to read existing model files, so we can test the reliability and the performance of this setup, also to see if there are any undiscovered caveats to this approach.

### Goals
To be able to successfully load an existing model through Alluxio API. 

### Tasks
- Add configurations needed to connect to alluxio
- Add alluxio as a new storage solution via external services 
- Add read support via alluxio (don't worry about the writes yet unless it gets in your way of making this work)

Charence:
This fixes #2017

#### Description
- Alluxio file handler
- Use alluxio as default storage in tests

#### Test cases
- read/write using alluxio as default storage should work
- read/write for existing sources should continue to work


backend/src/v4/handler/alluxio.js
```diff
@@ -0,0 +1,90 @@
...
+	constructor() {
+		if (config.alluxio) {
+			const {hostname, port, levels } = config.alluxio;
+			this.client = new AlluxioClient(`${hostname}:${port}`);
+			this.levels = levels;
+		}
+	}
+
+	async storeFile(data) {
+		const _id = nodeuuid();
```
carmenfan, line 54: use utils

backend/src/v4/handler/alluxio.js
```diff
@@ -0,0 +1,90 @@
...
+		const folderNames = generateFoldernames(_id, this.levels);
+		const link = path.join(folderNames, _id);
+		await this.client.uploadFile(this.getAlluxioPathFormat(link), data);
+		return ({_id, link, size:data.length, type: "alluxio"});
+	}
+
+	async getFileStream(key) {
+		try {
+			return await this.client.downloadFileStream(this.getAlluxioPathFormat(key));
```
carmenfan, line 63: return is not necessary here

backend/src/v4/handler/alluxio.js
```diff
@@ -0,0 +1,90 @@
...
+
+const config = require("../config.js");
+const path = require("path");
+const ResponseCodes = require("../response_codes");
+const nodeuuid = require("uuid").v1;
+const farmhash = require("farmhash");
+const AlluxioClient = require("../models/alluxioClient");
+const slash = require("slash");
+
+const generateFoldernames = (fileName, dirLevels) => {
```
carmenfan, line 28: Is this identical to the one we use in fs? if so we may want to move this out to a common file (in utils or something)

backend/src/v4/handler/alluxio.js
```diff
@@ -0,0 +1,90 @@
...
+		folders.push(fileNameHash & 255);
+	}
+	return folders.join("/");
+};
+
+class AlluxioHandler {
+	constructor() {
+		if (config.alluxio) {
+			const {hostname, port, levels } = config.alluxio;
+			this.client = new AlluxioClient(`${hostname}:${port}`);
```
carmenfan, line 48: can we test the connection and break it if it fails? We do this for every service we launch

backend/src/v4/handler/alluxio.js
```diff
@@ -0,0 +1,90 @@
...
+	async removeFile(key) {
+		return await this.client.delete(this.getAlluxioPathFormat(key));
+	}
+
+	async removeFiles(keys) {
+		return await Promise.all(keys.map(this.removeFile, this));
+	}
+
+	getAlluxioPathFormat(link) {
+		return "/" + slash(link);
```
carmenfan, line 86:
```suggestion
		return `/${slash(link)}`;
```

backend/src/v4/models/alluxioClient.js
```diff
@@ -0,0 +1,160 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+"use strict";
+const axios = require("axios");
```
carmenfan, line 19: this should live in handlers. It has nothing to do with 3drepo data models

backend/src/v4/models/alluxioClient.js
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	/**
+	 * Creates a directory
+	 *
+	 * @param {string} path - The absolute directory path to be created.
+	 * @param {string} opts - The absolute directory path to be created.
+	 */
+	async createDirectory(path, opts) {
+		return await this.postToPathRoute(path, opts, "create-directory");
```
carmenfan, line 67: either these don't need to be async, or it doesn't need to have return.

i.e. 
```suggestion
async createDirectory(path, opts) {
    await this.postToPathRoute(path, opts, "create-directory");
}
```

or

```suggestion
createDirectory(path, opts) {
   return this.postToPathRoute(path, opts, "create-directory");
}
````

File: backend/src/v4/models/alluxioClient.js [hunk 8]
```diff
@@ -0,0 +1,160 @@
...
+
+		try {
+			await this.closeFile(id);
+		} catch(e) {
+			throw new AlluxioError(`Couldn't close file "${path}": unknown error.`, e);
+		}
+
+	}
+
+	async write(id, data) {
```
carmenfan, line 147: this doesn't even have await in the function 😆

backend/src/v4/models/sequenceActivities.js
```diff
@@ -241,7 +241,7 @@ SequenceActivities.get = async (account, model, sequenceId) => {
 		activities = await FileRef.getSequenceActivitiesFile(account, model, utils.uuidToString(sequenceId));
 	} catch(e) {
 		activities = await createActivitiesTree(account, model, sequenceId);
-		FileRef.storeFile(account, activityCol(model) + ".ref", account, utils.generateUUID({string: true}), JSON.stringify(activities),  {"_id": sequenceId});
+		await FileRef.storeFile(account, activityCol(model) + ".ref", account, utils.generateUUID({string: true}), JSON.stringify(activities),  {"_id": sequenceId});
```
carmenfan, line 244: did you need await here for a test?

The whole point is the user doesn't need to wait for it to be cached, even if it failed, we don't care. that's why we're not awaiting here.

carmenfan (changes requested):
Like the database PR, can we write a bunch of unit tests for the alluxio client?

backend/src/v4/models/sequenceActivities.js
Charence: @carmenfan This is what was causing the sequence activities tests to fail. I'll see what else can be done here.

PR #2731, opened 2021-08-09
Issue 2722 - v5 backend initialisation

### LINKED ISSUE #2722: backend v5 foundation
### Description
For the v5 revamp (front end). we want to take the opportunity to also restructure the backend - at the very least, new endpoints should be added in a more desirable way so we are not adding extra technical debt for every extra logic going forward.

The aim is to have a place to code new features where we have a structure/frameworks that we wish to use going forward, whilst slowly migrate everything else during maintenance.

### Tasks
- [x] separate v4 and v5 similar to #2686
- [x] do one endpoint as an example
- [x] setup launch logic to tap into both so we run one server with endpoints supported from both versions
- [x] setup testing framework (properly)
- [x] setup coverall (again?)
- [x] setup better linting check
- [ ] swagger documentation ?
- [x] typescript - try it and see. 
- [x] allow launch from yarn run
- [x] response code revamp -> be able to send default error codes with customised message
- [ ] event listening to deal with things like emails and notifications (try it out)
- [ ] plugin js -> remove and see if it works?

### Proposed folder structure
- backend
  - tests
    - v4
    - v5
  - src
    - v4
    - v5
     - 3drepo.js
        - utils
          - config
          - logger
          - permissions (access rights check)          
          - helpers (essentially what was in utils but split up better, common helper functions etc)
        - middleware
          - auth
          - validator (input/output processing, so we stop doign things like UUID-to-string, ensure things are right type everywhere and can make assumptions)
        - routes
           - teamspace
             - project
               - containers
               - federation
               - common
                 - tickets
                   - issues
                   - risks
           - user
        - model (1 to 1 mapping to data models)
        - processors 
        - handlers (database, gridfs, fileshare, alluxio etc)
        - services (api server, sock.io server etc)

carmenfan:
This fixes #2722

#### Description
- added an example endpoint (list teamspaces)
- setup jest as testing framework
- added swagger documentation


keifgwinn:
looks like the permissions have been changed?

❯ k logs issue-2722-54c57b8c65-8lc2x
error: exec: "./run/run_app": permission denied

carmenfan:
> looks like the permissions have been changed?
> 
> ❯ k logs issue-2722-54c57b8c65-8lc2x
> error: exec: "./run/run_app": permission denied

not intentionally, I modified it on windows...

Charence (changes requested):

backend/src/v5/utils/helper/strings.js
```diff
@@ -0,0 +1,33 @@
...
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const { camelCase, snakeCase } = require('lodash');
+
+const StringHelper = {};
+
+// Turns thisIsUs to THIS_IS_US
+StringHelper.toSnakeCase = (str) => snakeCase(str).toUpperCase();
```
sanmont3drepo, line 23: I would change the naming from `.toSnakeCase` to `.toConstantCase` (That's how people are naming it  ¯\_(ツ)_/¯) to avoid the confusion

backend/src/v5/utils/permissions/permissions.constants.js
```diff
@@ -0,0 +1,22 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const PermConst = {};
+
+PermConst.teamspaceAdmin = 'teamspace_admin';
```
sanmont3drepo, line 20: ***Just a suggestion:**
change it to `PERM` or `PERMISSIONS`  or `PERMISSION_CONST` with the constant casing?

I dont know why but this looks a bit more like constants definition, the other way seems more like an object with behaviour. 

```suggestion
const PERMISSIONS_CONST = {
	teamspaceAdmin: 'teamspace_admin',
};
```

backend/src/v5/utils/responseCodes.js
```diff
@@ -0,0 +1,93 @@
...
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+const { toCamelCase, toSnakeCase } = require('./helper/strings');
+
+const ResponseCodes = {};
+
+ResponseCodes.template = {
```
sanmont3drepo, line 21: I think  `.template` to `.templates` would reflect a bit better this because in the rest of the system is being used as a collection of templates or response messages.

backend/src/v5/utils/responseCodes.js
```diff
@@ -0,0 +1,93 @@
...
+		};
+	});
+
+	return responses;
+};
+
+ResponseCodes.codeExists = (code) => !!ResponseCodes.template[toCamelCase(code)];
+
+ResponseCodes.createResponseCode = (errCode, message) => {
+	const res = ResponseCodes.codeExists(errCode?.code) ? errCode : ResponseCodes.template.unknown;
```
sanmont3drepo, line 89: Is this as intended or this should be
`const res = ResponseCodes.codeExists(errCode?.code) ? ResponseCodes.templates[errCode?.code]: ResponseCodes.template.unknown;`
?

sanmont3drepo (changes requested):

backend/src/v5/utils/responseCodes.js
carmenfan: yep should be your version, it's more robust

backend/src/v5/utils/permissions/permissions.constants.js
carmenfan: as discussed, changing `teamspaceAdmin` to `TEAMSPACE_ADMIN` as that's the variable that gets exported.

backend/src/v5/utils/helper/strings.js
carmenfan: fair enough 😆

backend/tests/v4/integrated/signup.js
```diff
@@ -0,0 +1,624 @@
...
+
+	const username = 'signup_helloworld';
+	const uppercase_username = 'Signup_helloworld';
+	const password = 'Str0ngPassword!';
+	const email = 'test3drepo_signup@mailinator.com';
+	const firstName = 'Hello';
+	const lastName = 'World';
+	const countryCode = 'GB';
+	const company = 'company';
+	// const jobTitle = "someJobTitle";
```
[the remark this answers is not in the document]
carmenfan: Yes, it's temporarily removed because we couldn't get the legal docs out in time. so we don't want to get rid of this as it'll be revived in the next release.

run/run_app
```diff
@@ -9,5 +9,8 @@ else
             echo "WARNING: If running locally then you may require sudo due to port issues "
             echo ""
     fi
-    NODE_ENV=$1 NODE_CONFIG_DIR='./config' cd backend && yarn start
+    echo "running the bash stuff"
```
Charence, line 12: debugging message?

Charence, line 12: debugging message?

Charence:
@carmenfan How do I create the swagger docs?

carmenfan:
> @carmenfan How do I create the swagger docs?

you don't. It's auto generated as part of the front end: https://issue-2722.dev.3drepo.io/docs/

backend/tests/v5/unit/utils/helper/strings.test.js
```diff
@@ -0,0 +1,74 @@
...
+		test(`with ${source} should result in ${target}`, () => {
+			matchHelper(StringHelper.toConstantCase, source, target);
+		});
+	});
+};
+
+describe('utils/helper/strings', () => {
+	testGetURLDomain();
+	testToCamelCase();
+	testToSnakeCase();
```
sanmont3drepo, line 73: rename the test testToConstantsCase .

sanmont3drepo (changes requested):

sanmont3drepo (approved):

PR #2739, opened 2021-08-16
#2738 - add NODE_OPTIONS

### LINKED ISSUE #2738: Add ability to send NODE_OPTIONS to /run/run_app
### Description

./run/run_app supports NODE_OPTIONS via $2, but the docker init script doesn't support it yet.

### Goals

- [ ] Ability to send NODE_OPTIONS to start up script

### Tasks

- [ ] Add variable to init script.

keifgwinn:
This fixes #2738 

#### Description
<!-- Add a high level description of the changes made (possibly quite similar to task list if it was a feature) 
e.g.
- Added NODE_OPTIONS variable

#### Test cases
Can send in additional params like '--max_old_space_size=8192'

### DIFF
--- a/.azure/Docker/src/init.sh
+++ b/.azure/Docker/src/init.sh
@@ -17,7 +17,7 @@ if [ "$1" = 'web' ]; then
 
     umask 2
     cd /home/node/3drepo.io
-    exec gosu $NODE_USERNAME ./run/run_app $NODE_ENV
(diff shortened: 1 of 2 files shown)


carmenfan (changes requested):
it's not going to work when #2731 get merged in.

it's probably better to just set it as an env var 
image

PR #2752, opened 2021-08-23
ISSUE #2721 - Added lingui and minimal example

### LINKED ISSUE #2721: Have support for multiple languages in v5

sanmont3drepo:
This fixes #2721

#### Description
- Added lingui without the macros

### DIFF
--- a/frontend/.babelrc
+++ b/frontend/.babelrc
@@ -1,0 +1,3 @@
+{
+	"presets": ["@babel/env", "@babel/react", "@lingui/babel-preset-react"]
+}
--- a/frontend/.linguirc
+++ b/frontend/.linguirc
@@ -1,0 +1,13 @@
+{
+   "locales": ["en", "es"],
+   "sourceLocale": "en",
+   "catalogs": [{
+      "path": "src/locales/{locale}/messages",
+      "include": ["src"]
+   }],
+   "format": "minimal",
+   "compileNamespace": "ts",
+   "extractBabelOptions": {
+    "presets": ["@babel/preset-typescript"]
+   }
+}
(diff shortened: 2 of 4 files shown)


carmenfan:
@m1chalkubiak  can you take a look and see if you think this will work?

m1chalkubiak:
@sanmont3drepo 
The scenario that I think we should definitely check is whether in this solution we are able to embed html and actually render it?
Here I can give as an example, a text with a link to a page in a given language. Depending on the language, the link will be in a different part of the sentence and will lead to a different page (in a given language), this is a scenario that often happens in projects.

A similar scenario would be to pass some React component as a variable to translation, it could be some icon, but it could just be some data wrapped in a container that somehow stylizes it specifically.

It would be also good to have an example with some plural.
Something similar to what we had in 3d send:
```js
defaultMessage: 'Expires {days, plural, =0 {today} one {in one day} other {in {days} days}}',
```

Also, did thought already,  how we will be adding new translations? I am most afraid of a scenario in which we will have one large file and if we introduce changes in several PRs at the same time, conflicts and erroneous deletions may occur.

sanmont3drepo:
> @sanmont3drepo
> The scenario that I think we should definitely check is whether in this solution we are able to embed html and actually render it?
> Here I can give as an example, a text with a link to a page in a given language. Depending on the language, the link will be in a different part of the sentence and will lead to a different page (in a given language), this is a scenario that often happens in projects.

Can you show or point to an example of the problem and how is solved in 3dsend?

> 
> A similar scenario would be to pass some React component as a variable to translation, it could be some icon, but it could just be some data wrapped in a container that somehow stylizes it specifically.

I dont understand this one in particular

> 
> It would be also good to have an example with some plural.
> Something similar to what we had in 3d send:
> 
> ```js
> defaultMessage: 'Expires {days, plural, =0 {today} one {in one day} other {in {days} days}}',
> ```
> 
> Also, did thought already, how we will be adding new translations? I am most afraid of a scenario in which we will have one large file and if we introduce changes in several PRs at the same time, conflicts and erroneous deletions may occur.

There are two commands: `extract` and `compile` it generates files with the translations in .ts . The code in this example is only for showing how would the <Trans /> look.

PR #2761, opened 2021-09-01
Issue 2706 - Added routing and skeleton for dashboard

### LINKED ISSUE #2706: Revamping - routing
### Description
Create routes  for v5 pages. Routing should be as decoupled as possible to make it easy to tweak and see what's right. 

### Goals
- [ ] Create some sort of decoupled routing system
- [ ] Set containers/federations routes

### Tasks
- [ ] Create a service file that returns the route taking teamspace/project/ federationid, etc (depending on the route) as parameters 
- [ ] Group the functions in a logical maner (maybe separate the ones in dashboard and viewer)
- [ ] Use this service files also in the route paths when defining the paths in react router to avoid having outdated routes

sanmont3drepo:
This fixes #2706 

#### Description
- Added support for routing 
- Fully functional navigation for dashboard

#### Test cases
- Check that changing the selection boxes changes the url correctly.
- Check that selecting a tab changes the url correctly.
- Check that it doesnt reload the whole page when navigating.
- Check that refeshing matches the state of the selection boxes
- Check that foward and back button from the browser works correctly


m1chalkubiak:
Hey @sanmont3drepo I hope you feel better now, I have checked at your routing proposal and found some minor bugs it would be great if you could have a look at them.

When we enter the teamspace settings page and click on settings link again, we are transfered to the settings/settings page. Of course, we could add disable to the link if the last element in path is settings, but I don't know if that's what we want to fix this problem.

Zrzut ekranu 2021-09-7 o 10 51 34


When we are on the 'containers' page, we can still see that the content of the project is rendered, do you have any specific idea how we would like to avoid it in the final implementation? The same thing happens for federations and project settings.

Zrzut ekranu 2021-09-7 o 11 05 48

m1chalkubiak (approved):

PR #2775, opened 2021-09-07
ISSUE 2774 - Scripts for getting face count

### LINKED ISSUE #2774: Scripts for counting mesh faces
### Description

Scripts for counting faces in meshes/models/federations

### Goals

- [x] Output faces_count for models/federations

### Tasks

- [x] Output faces_count for models
- [x] Output faces_count for federations

Charence:
This fixes #2774

#### Description
Python scripts for getting face count of meshes from DB

#### Test cases
- Run script with valid args should produce output
- Run script with invalid args should exit


carmenfan:
@Charence  what is the context of this usage? why do we need these scripts?

Charence:
> @Charence what is the context of this usage? why do we need these scripts?

Getting model stats for product team's debugging purposes

carmenfan:
closing this is getting a bit crazy and the query doesn't work. I think we got what we needed right now and we can look into putting it in github when we are less busy.

I have a script that will give you the right numbers for federation count. Talk to me if you need any numbers.

PR #2776, opened 2021-09-07
ISSUE 2773 - remove dep on farmhash

### LINKED ISSUE #2773: Replace farmhash dependency due to deployment issue
### Description

Farmhash, which we are using to generate random filenames, is crashing on a private deployment. This affects files saved through IO such as screenshots and other file attachments to a model. We are only using this library to generate a random filename, alternatives can be used.

- This is happening in: private prod deployment 

### Steps to replicate
- Use 3drepo on private deployment
- Open a model
- Save a view

Note: This does not happen if you create a issue/risk through the Kanban board without a screenshot

### Current Behaviour
- IO crashes when attaching file

### Expected Behaviour
- IO to accept the file

Charence:
This fixes #2773 

#### Description
Farmhash no longer used to generate a large random number. Math.random is being used instead. Not aware of any particular disadvantages of just relying on Math.random for folder names.

#### Test cases
- Should work on private deployment
- Should distribute `fs` files across different folders randomly as before

### DIFF
--- a/backend/handler/fs.js
+++ b/backend/handler/fs.js
@@ -24,19 +24,16 @@ const ResponseCodes = require("../response_codes");
 const systemLogger = require("../logger").systemLogger;
 const utils = require("../utils");
 
-const generateFoldernames = (fileName, dirLevels) => {
+const generateFoldernames = (dirLevels) => {
 	if (dirLevels < 1) {
 		return "";
 	}
 	const folders = [];
-	const minChunkLen = 4;
-	const nameChunkLen = Math.max(fileName.length / dirLevels, minChunkLen);
 
 	for(let i = 0 ; i < dirLevels; i++) {
-		const chunkStart = (i * nameChunkLen) % fileName.length;
 		// Generate a relatively large random number for the file directory name
-		const fileNameHash = Math.random() * 100000;
-		folders.push(fileNameHash & 255);
+		const randomNumber = Math.random() * 100000;
+		folders.push(randomNumber & 255);
 	}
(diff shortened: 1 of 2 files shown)


sanmont3drepo (approved):

Charence:
@keifgwinn @carmenfan I do not have permissions to merge into `master`

backend/handler/fs.js
```diff
@@ -22,21 +22,18 @@ const fs = require("fs");
...
 	const folders = [];
-	const minChunkLen = 4;
-	const nameChunkLen = Math.max(fileName.length / dirLevels, minChunkLen);
 
 	for(let i = 0 ; i < dirLevels; i++) {
-		const chunkStart = (i * nameChunkLen) % fileName.length;
-		const fileNameHash = farmhash.fingerprint32(fileName.substr(chunkStart,nameChunkLen) + Math.random());
-		folders.push(fileNameHash & 255);
+		// Generate a relatively large random number for the file directory name
+		const randomNumber = Math.random() * 100000;
```
carmenfan, line 35: Why do we need a massive number? If we just want it between 0-255 shouldn't we just *255?

Charence: @carmenfan Yes, 255 should be sufficient. It's only larger to replicate the number that farmhash was previously generating.

carmenfan:
Ok then let's make that change and I'll merge it

carmenfan (approved):

carmenfan:
@Charence merged

PR #2782, opened 2021-09-10
ISSUE #2696 - Add new popup dialogs system

### LINKED ISSUE #2696: Revamping - New popup dialogs system
### Description
A new popup dialog system that's decoupled from redux is needed in order to have cleaner code. This will also provide the way yo have testable sagas because as it is now it the dependencies generated by having presentation code in redux makes it untestable

### Goals
- [ ] Have an easy to use popup dialog system

### Tasks
- [ ] Create the popup system in v5/ui/controls

m1chalkubiak:
This fixes #2696

#### Description
- Added support for popup dialogs system
- Sample template for info type of popup with hardcoded data.

I imagine the use of this system in such a way that we will be calling a dialogue with a given type. In `dialog.component.tsx` there is a mapped list of types with specific templates, so adding a new dialogue will require adding a new dialogue type and a template for it. Probably in the future there will also be more generic types of dialogs, such as the one regarding errors, for it, like in v4, we will probably provide some information and the content of the dialogue itself will be created according to the scheme, but at the moment we do not have designs for this, so it is difficult to define in how any subsequent templates should be separated.


carmenfan:
@m1chalkubiak can you help resolve conflicts here?

m1chalkubiak:
> @m1chalkubiak can you help resolve conflicts here?

Of course @carmenfan , I'm handling this.

frontend/src/v4/modules/reducers.ts
```diff
@@ -68,6 +69,7 @@ export default function createReducer(history) {
 		currentUser: currentUserReducer,
 		userManagement: userManagementReducer,
 		dialog: dialogReducer,
+		dialogs: dialogsReducer,
```
carmenfan, line 72: `dialog` and `dialogs` is going to be confusing as hell.

Can we name this as`dialogsV5` or something?

frontend/src/v5/ui/routes/dashboard/modalsDemo/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+		const action = DialogsActions.open('alert', { message: 'This is an alert message!' });
+		dispatch(action);
+	};
+
+	const showFormModal = () => {
+		setOpen(true);
+	};
+
+	const onSubmit = () => {
+	// console.log(JSON.stringify(data));
```
carmenfan, line 41: remove?

carmenfan (changes requested):

carmenfan:
Some visual stuff:

@george-hadfield's design has a different colour on the background overlay than the one demonstrated here. Also the visuals are blurred for anything that is in the background, could we make it the same? 
image
image

The error message design has a completely different template (namely there's no `X` button, no title, buttons are placed in different place as well. So I think the alert dialog modal design needs to change.... Having said that, it doesn't look like the design has been approved yet, @george-hadfield @mosment3DRepo can we verify if it is final?
image
image

george-hadfield:
The error message design is final

carmenfan (approved):

carmenfan:
@george-hadfield  ready for review, link here:  https://issue-2696.dev.3drepo.io/v5/modals

george-hadfield:
@m1chalkubiak For the alert modal, please remove the ability for a user to click on the background to close the modal. This behaviour is fine though for the form modal.

george-hadfield:
### Change small text colour in alert modal to #6B778C (base.main)
The text _"<Detail where the user will be taken by clicking ‘OK, close window’>"_ is incorrect and should change to the base.main grey please

### Images 
image

george-hadfield:
### Incorrect font size for alert modal title
The design uses 18px (1.2rem) but the build is using 24px (1.5rem) 

### Images

Build
image

Design
image

george-hadfield:
### Cross should be horizontally aligned with the modal title 
Currently the cross is too low and needs to be brought up to align with the text

### Images

Build
image

Design
image

PR #2784, opened 2021-09-13
ISSUE 2763 - Get federation list

### LINKED ISSUE #2763: Create an endpoint for a list of federations the user can access
#### Federations
so for `GET /teamspaces/:teamspace/projects/:projectId/federations`, It should return the following:

````
{
    federations: [
        {_id: <fed id>, name: "fed name", role: "admin" , isFavourite: true}
        {_id: <fed id>, name: "fed name 2", role: "collaborator" , isFavourite: false}
    ]
}
````
If the user does not have sufficient permission, or the teamspace does not exist, it should return 404 `TEAMSPACE_NOT_FOUND`
If the project does not exist, it should return 404 `PROJECT_NOT_FOUND`

### Goals
- [x] Create an endpoint that will return a list of containers within the project the user has access to

### Tasks
- [x] Create endpoint
- [x] Create swagger docs
- [x] Create e2e tests
- [x] Create unit tests for each method created

### PR DESCRIPTION
**Pull request by Charence**
This fixes #2763 

#### Description
- v5 API for get federation list

#### Test cases
- list of federations should be returned for a valid teamspace and project
- otherwise, error thrown

### CODE REVIEW

File: backend/src/v5/processors/teamspaces/projects/federations/federations.js [hunk 1]
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+const { hasProjectAdminPermissions, isTeamspaceAdmin } = require('../../../../utils/permissions/permissions');
+const { getFavourites } = require('../../../../models/users');
+const { getFederations } = require('../../../../models/modelSettings');
+const { getProjectById } = require('../../../../models/projects');
+
+const Federations = {};
+
+Federations.getFederationList = async (teamspace, project, user) => {
```
carmenfan, line 25: this is going to clash with @ChristosTsiotsias favourites PR as he did the same thing there.

But in general can we avoid duplicated code between fed and containers, any common code should go into a `common.js`?

... or we have a models folder instead, where we have federations.js and containers.js and a models.js or something.

carmenfan (changes requested):
@Charence only thing to change is to reduce duplication between fed/containers.

do you want to change that here and establish the structure, and @ChristosTsiotsias  can just copy that in favourites? (as this one is more likely to be merged in first)

carmenfan (approved):
@ChristosTsiotsias merging this in, note the processing folder changes

PR #2792, opened 2021-09-15
Issue 2756 - Revisions endpoints

### LINKED ISSUE #2756: container Revision endpoint
### Description
Have an endpoint to fetch the revision details of a given container

`GET /teamspaces/:teamspace/projects/:project/containers/:container/revisions?showVoid=true` 
payload:
````js
{
   revisions: [
       {
		"_id":"00000000-0000-0000-0000-000000000001",
		"author":"alice",
		"timestamp":12345567, //timestamp
		"name":"00000000-0000-0000-0000-000000000001",
	}
    ]
}
````
NOTES:
- `showVoid` is optional and should be false by default
- if `showVoid` is set to true, show void revisions as well as active revisions
- if container not found or if the user has no permissiosn to access the container, return `404 CONTAINER_NOT_FOUND`

### Goals
- [ ] The user should be able to get the list of revisions 
   - [ ]  Void revisions should only be visible when the user has specified showVoid to true

### Tasks
- [ ]  Implement endpoint
- [ ]  Add unit tests
- [ ]  Add e2e tests
- [ ]  Add swagger doc

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #2756 

Endpoints for getting all revisions of a container and updating a revision's status are created.
Unit and E2E tests are added.
Swagger documentation is added.

#### Test cases
All non-void revisions should be retrieved successfully if the user has permission to view the container and teamspace/project/container exist.
All revisions including void should be retrieved if ?showVoid=true is included in the query params.
A revisions status should be updated successfully if the user has permission to view the container and teamspace/project/container/revision exist.

### CODE REVIEW

File: backend/src/v5/processors/teamspaces/projects/containers/containers.js [hunk 1]
```diff
@@ -0,0 +1,84 @@
...
+const { getContainerById, getContainers } = require('../../../../models/modelSettings');
+const { getLatestRevision, getRevisionCount, getRevisions, updateRevisionStatus } = require('../../../../models/revisions');
+const { hasProjectAdminPermissions, isTeamspaceAdmin } = require('../../../../utils/permissions/permissions');
+const { UUIDToString } = require('../../../../utils/helper/uuids');
+const { getFavourites } = require('../../../../models/users');
+const { getProjectById } = require('../../../../models/projects');
+
+const Containers = {};
+
+Containers.getContainerList = async (teamspace, project, user) => {
```
carmenfan, line 27: @ChristosTsiotsias  this merge conflict didn't resolve properly.

containers folder no longer exist. talk to @Charence if you're unsure how to go about this.

.gitignore
```diff
@@ -110,3 +110,4 @@ package-lock.json
 .vs/3drepo.io/v16/Browse.VC.opendb
 backend/globalConfig.json
 *.swn
+.vs/3drepo.io/config/applicationhost.config
```
carmenfan, line 113: maybe it's easier to just ignore the whole `.vs` folder?

backend/src/v5/models/revisions.js
```diff
@@ -25,6 +26,27 @@ const excludeIncomplete = { incomplete: { $exists: false } };
 
 const collectionName = (model) => `${model}.history`;
 
+const findRevisionsByModel = async (teamspace, model, query, projection) => {
+	const revisions = await db.find(teamspace, collectionName(model), query, projection, {});
+
+	if (!revisions || revisions.length === 0) {
+		throw templates.containerNotFound;
```
carmenfan, line 33: I don't think we want to throw an error, if there's no revision just return an empty array?

Just thinking the use case of when the user created a model but haven't uploaded a revision yet, you don't expect getting the revision listing for that model to throw you an error, it should just give you no revision.

backend/src/v5/models/revisions.js
```diff
@@ -25,6 +26,27 @@ const excludeIncomplete = { incomplete: { $exists: false } };
...
+	}
+
+	return revisions;
+};
+
+const findOneRevisionByID = async (teamspace, model, revision, projection, sort) => {
+	const revisions = await findRevisionsByModel(teamspace, model, {}, projection, sort);
```
carmenfan, line 40: You should just use `findOneRevisionByQuery`  where `_id: revision`. If you are looking for a single entry you should never draw all the data out and filter here - let the database do the heavy lifting.

revision id should have been converted prior to the data entering the processor - so you shouldn't be doing any UUID to string converasions here.. This was explained on the blurb i sent out before I went on holiday: 
image

backend/src/v5/models/revisions.js
```diff
@@ -45,4 +67,25 @@ Revisions.getRevisionCount = (teamspace, model) => {
 	return db.count(teamspace, collectionName(model), query);
 };
 
+Revisions.getRevisions = (teamspace, model, showVoid, projection = {}) => {
+	const query = { ...excludeIncomplete };
+
+	if (!showVoid) {
+		query.void = { $ne: true };
```
carmenfan, line 74: use the constants up top
```suggestion
		query.void = excludeVoids.void
```

backend/src/v5/models/revisions.js
```diff
@@ -45,4 +67,25 @@ Revisions.getRevisionCount = (teamspace, model) => {
...
+	}
+
+	return findRevisionsByModel(teamspace, model, query, projection);
+};
+
+Revisions.updateRevisionStatus = async (teamspace, model, revision, status) => {
+	const rev = await findOneRevisionByID(teamspace, model, revision, { _id: 1, void: 1 });
```
carmenfan, line 81: I don't see a reason why you need care about the previous status of void.

this function should just do an update.

if void? set it to true, otherwise unset should be sufficient?

backend/src/v5/processors/teamspaces/projects/models/containers.js
```diff
@@ -55,4 +56,18 @@ Containers.getContainerStats = async (teamspace, project, container) => {
...
+		{ _id: 1, author: 1, timestamp: 1, tag: 1, void: 1 });
+
+	return revisions.map((revision) => ({ ...revision, _id: UUIDToString(revision._id) }));
+};
+
+Containers.updateRevisionStatus = async (teamspace, container, revision, status) => {
+	if (typeof status === 'boolean') {
+		await updateRevisionStatus(teamspace, container, revision, status);
+	}
+	return {};
```
carmenfan, line 70: no reason to return {}

backend/src/v5/processors/teamspaces/projects/models/containers.js
```diff
@@ -55,4 +56,18 @@ Containers.getContainerStats = async (teamspace, project, container) => {
...
 
+Containers.getRevisions = async (teamspace, container, showVoid) => {
+	const revisions = await getRevisions(teamspace, container, showVoid,
+		{ _id: 1, author: 1, timestamp: 1, tag: 1, void: 1 });
+
+	return revisions.map((revision) => ({ ...revision, _id: UUIDToString(revision._id) }));
+};
+
+Containers.updateRevisionStatus = async (teamspace, container, revision, status) => {
+	if (typeof status === 'boolean') {
```
carmenfan, line 67: we shouldn't be doing type checking here. Type checks should happen in the middleware.
image

In essence anything entering processor, we assume is type checked and validated.

Also, if the type check failed we should be throwing an error, as it stands you'd be returning a response 200 telling the user you've done the request - which will not be correct

backend/src/v5/processors/teamspaces/projects/models/containers.js
```diff
@@ -55,4 +56,18 @@ Containers.getContainerStats = async (teamspace, project, container) => {
 	};
 };
 
+Containers.getRevisions = async (teamspace, container, showVoid) => {
+	const revisions = await getRevisions(teamspace, container, showVoid,
+		{ _id: 1, author: 1, timestamp: 1, tag: 1, void: 1 });
+
+	return revisions.map((revision) => ({ ...revision, _id: UUIDToString(revision._id) }));
```
carmenfan, line 63: do the conversion in routes.

this is serialisation, we only need it for network requests. serialising/deserialising should only happen in middleware (inputs) or routes(outputs)

backend/src/v5/routes/teamspaces/projects/containers/containers.js
```diff
@@ -43,6 +43,24 @@ const getContainerStats = async (req, res) => {
 	}).catch((err) => respond(req, res, err));
 };
 
+const getRevisions = async (req, res) => {
+	const { teamspace, container } = req.params;
+	const showVoid = req.query.showVoid === 'true';
+
+	Containers.getRevisions(teamspace, container, showVoid).then((revisions) => {
+		respond(req, res, templates.ok, revisions);
```
carmenfan, line 51:
```suggestion
		respond(req, res, templates.ok, { revisions } );
```

It's a better practice to avoid returning arrays as the root level (also makes life a lot easier in the C# world where you need to serialise the json - as arrays aren't recognised as a top level thing...)

So for v5 we want to avoid returning arrays, wrap them in a json

backend/src/v5/routes/teamspaces/projects/containers/containers.js
```diff
@@ -43,6 +43,24 @@ const getContainerStats = async (req, res) => {
...
+	}).catch((err) => respond(req, res, err));
+};
+
+const updateRevisionStatus = async (req, res) => {
+	const { teamspace, container, revision } = req.params;
+	const status = req.body.void;
+
+	Containers.updateRevisionStatus(teamspace, container, revision, status).then((rev) => {
+		respond(req, res, templates.ok, rev);
```
carmenfan, line 60: i don't think we need to return anything here?

backend/src/v5/routes/teamspaces/projects/containers/containers.js
```diff
@@ -184,6 +202,132 @@ const establishRoutes = () => {
...
+	 *               - status
+	 *     responses:
+	 *       401:
+	 *         $ref: "#/components/responses/notLoggedIn"
+	 *       404:
+	 *         $ref: "#/components/responses/teamspaceNotFound"
+	 *       200:
+	 *         description: updates the status of the revision
+	 */
+	router.patch('/:container/revisions/:revision', hasReadAccessToContainer, updateRevisionStatus);
```
carmenfan, line 329: 2 things:
- you need write access to the container to update a revision status, so you need to write a new validator.
- you need to add a middleware to check the payload (e.g. it only has void in it and it's of type boolean)
   - maybe add it in `middleware/dataConverter/revisions` ?
   - use [yup](https://www.npmjs.com/package/yup)

backend/src/v5/models/revisions.js
carmenfan: You can also update a revision status via the tag. I think your query is missing that part?

backend/src/v5/routes/teamspaces/projects/containers/containers.js
```diff
@@ -184,6 +202,132 @@ const establishRoutes = () => {
...
+	 *
+	 *
+	 */
+	router.get('/:container/revisions', hasReadAccessToContainer, getRevisions);
+
+	/**
+	 * @openapi
+	 * /teamspaces/{teamspace}/projects/{project}/containers/{container}/revisions/:revision:
+	 *   patch:
+	 *     description: Updates the void status of a revision
```
carmenfan, line 279:
```suggestion
	 *     description: Update a revision. Currently only the void status can be updated.
```

backend/src/v5/routes/teamspaces/projects/containers/containers.js
```diff
@@ -184,6 +202,132 @@ const establishRoutes = () => {
...
+   	 *       - container:
+	 *         name: container
+	 *         description: Container ID
+	 *         in: path
+	 *         required: true
+	 *         schema:
+	 *         type: string
+	 *       - revision:
+	 *         name: revision
+	 *         description: Revision ID
```
carmenfan, line 306: we should take tag as well here.

backend/tests/v5/e2e/routes/teamspaces/projects/containers/containers.test.js
```diff
@@ -80,7 +88,7 @@ const setupData = async () => {
 		...modelProms,
 		ServiceHelper.db.createUser(nobody),
 		ServiceHelper.db.createProject(teamspace, project.id, project.name, models.map(({ _id }) => _id)),
-		ServiceHelper.db.createRevision(teamspace, modelWithRev._id, revision),
+		ServiceHelper.db.createRevisions(teamspace, modelWithRev._id, revisions),
```
carmenfan, line 91: Can we keep this singular?

```suggestion
		...revisions.map((revision) => ServiceHelper.db.createRevision(teamspace, modelWithRev._id, revision)),
```

backend/tests/v5/e2e/routes/teamspaces/projects/containers/containers.test.js
```diff
@@ -173,6 +182,125 @@ const testGetContainerStats = () => {
...
 };
 
+const formatRevisions = (revs, includeVoids = false) => {
+	let formattedRevisions = revs;
+
+	if (!includeVoids) {
+		formattedRevisions = formattedRevisions.filter((rev) => !rev.void);
+	}
+
+	formattedRevisions = formattedRevisions.map((rev) => ({ ...rev, timestamp: rev.timestamp.toISOString() }))
```
carmenfan, line 192: if you use flatMap then you don't need to do a filter at L189

```suggestion
	formattedRevisions = formattedRevisions.flatMap((rev) => includeVoid || !rev.void ? { ...rev, timestamp: rev.timestamp.toISOString() } : []);
```

backend/tests/v5/e2e/routes/teamspaces/projects/containers/containers.test.js
```diff
@@ -173,6 +182,125 @@ const testGetContainerStats = () => {
...
+
+		test('should fail if the revision doesn\'t exist', async () => {
+			const res = await agent.patch(`/v5/teamspaces/${teamspace}/projects/${project.id}/containers/${modelWithRev._id}/revisions/fdefaxsxsa?key=${users.tsAdmin.apiKey}`)
+				.send({ void: false }).expect(templates.revisionNotFound.status);
+			expect(res.body.code).toEqual(templates.revisionNotFound.code);
+		});
+
+		test('should not update a revision\'s status if the body if the request is not boolean', async () => {
+			await agent.patch(`${route(voidRevision._id)}?key=${users.tsAdmin.apiKey}`)
+				.send({ void: 123 }).expect(templates.ok.status);
```
carmenfan, line 284: this should throw invalidArguments

backend/tests/v5/e2e/routes/teamspaces/projects/containers/containers.test.js
```diff
@@ -173,6 +182,125 @@ const testGetContainerStats = () => {
...
+		test('should update a revision\'s status if the body of the request is boolean', async () => {
+			await agent.patch(`${route(voidRevision._id)}?key=${users.tsAdmin.apiKey}`)
+				.send({ void: false }).expect(templates.ok.status);
+
+			const res = await agent.get(`/v5/teamspaces/${teamspace}/projects/${project.id}/containers/${modelWithRev._id}/revisions?showVoid=true&key=${users.tsAdmin.apiKey}`)
+				.expect(templates.ok.status);
+			const revision = res.body.find((r) => r._id === voidRevision._id);
+			expect(revision.void).toEqual(false);
+		});
+	});
```
carmenfan, line 301: missing tests:
- throw invalidArguments if it has anything extra in the payload (e.g. someone trying to change the tag of the revision)
- user with only viewer/commenter access right should not be able to update the revision

backend/tests/v5/helper/services.js
```diff
@@ -68,31 +68,28 @@ db.createModel = (teamspace, _id, name, props) => {
...
-	const rev = {
-		_id: stringToUUID(_id),
-		author,
-		timestamp,
-		tag,
-	};
-	return DbHandler.insertOne(teamspace, `${modelId}.history`, rev);
+db.createRevisions = (teamspace, modelId, revisions) => {
+	const formattedRevisions = revisions.map((rev) => ({ ...rev, _id: stringToUUID(rev._id) }));
+	DbHandler.insertMany(teamspace, `${modelId}.history`, formattedRevisions);
```
carmenfan, line 73: As above I think we should change this back to singular.

But for reference, you're missing a return here 
```suggestion
	return DbHandler.insertMany(teamspace, `${modelId}.history`, formattedRevisions);
```

insertMany returns a promise, and without returning that upstream, the caller would not know when it finishes.

backend/tests/v5/unit/models/revisions.test.js
```diff
@@ -53,7 +53,75 @@ const testGetLatestRevision = () => {
...
 
+const testGetRevisions = () => {
+	describe('GetRevisions', () => {
+		const expectedData = [
+			{ _id: 1, author: 'someUser', timestamp: new Date() },
+			{ _id: 2, author: 'someUser', timestamp: new Date() },
+			{ _id: 3, author: 'someUser', timestamp: new Date(), void: true },
+		];
+
+		test('Should return non void container revisions', async () => {
```
carmenfan, line 64: this test is pretty pointless, as the logic of void filtering is done by the db query (and in this case, it is done by the mock value logic here). I think you only really the test on L71

this function basically should return what the database query gives you, so as long as the return value matches what db.find returns this is the only test we need to do.

backend/tests/v5/unit/models/revisions.test.js
```diff
@@ -53,7 +53,75 @@ const testGetLatestRevision = () => {
...
+		});
+
+		test('Should return all container revisions', async () => {
+			jest.spyOn(db, 'find').mockResolvedValue(expectedData);
+			const res = await Revisions.getRevisions('someTS', 'someModel', true);
+			expect(res).toEqual(expectedData);
+		});
+
+		test('Should throw CONTAINER_NOT_FOUND if there is no revisions table', async () => {
+			jest.spyOn(db, 'find').mockResolvedValue({ length: 0 });
```
carmenfan, line 78: i think find returns an empty array if it's empty?

backend/tests/v5/unit/models/revisions.test.js
```diff
@@ -53,7 +53,75 @@ const testGetLatestRevision = () => {
...
+			await expect(Revisions.getRevisions('someTS', 'someModel')).rejects.toEqual(templates.containerNotFound);
```
carmenfan, line 79: as discussed, it should return the empty array without throwing.

backend/tests/v5/unit/models/revisions.test.js
```diff
@@ -53,7 +53,75 @@ const testGetLatestRevision = () => {
...
+		const expectedData = [
+			{ _id: 1, author: 'someUser', timestamp: new Date() },
+			{ _id: 2, author: 'someUser', timestamp: new Date() },
+			{ _id: 3, author: 'someUser', timestamp: new Date(), void: true },
+		];
+
+		test('Should update the void status of a revision', async () => {
+			jest.spyOn(db, 'find').mockResolvedValue(expectedData);
+			await Revisions.updateRevisionStatus('someTS', 'someModel', 3, false);
+			expect(expectedData.find((r) => r._id === 3).void).toEqual(false);
```
carmenfan, line 95: similar to the favourites PR, there's no need to check the expected data here.

we expect this function to test that the db query succeeds. so as long as that happened this is fine.

Let the e2e test the integrity of the data.

backend/tests/v5/unit/models/revisions.test.js
```diff
@@ -53,7 +53,75 @@ const testGetLatestRevision = () => {
...
+		});
+
+		test('Should not update the void status of a revision if void is not set and new status is false', async () => {
```
carmenfan, line 98: this test is redundant i think?

After the change were we stop probing what was the last state and just set/unset the void status regardless, it's not really testing anything new to the previous test.

backend/tests/v5/unit/models/revisions.test.js
```diff
@@ -53,7 +53,75 @@ const testGetLatestRevision = () => {
...
+			jest.spyOn(db, 'find').mockResolvedValue(expectedData);
+			await Revisions.updateRevisionStatus('someTS', 'someModel', 1, false);
+			expect(expectedData.find((r) => r._id === 1).void).toEqual(undefined);
+		});
+
+		test('Should not update the void status of a revision if void has the same value as status', async () => {
```
carmenfan, line 104: same here

backend/tests/v5/unit/models/revisions.test.js
```diff
@@ -53,7 +53,75 @@ const testGetLatestRevision = () => {
...
+			jest.spyOn(db, 'find').mockResolvedValue(expectedData);
+			await Revisions.updateRevisionStatus('someTS', 'someModel', 3, true);
+			expect(expectedData.find((r) => r._id === 3).void).toEqual(true);
+		});
+
+		test('Should throw CONTAINER_NOT_FOUND if there is no revisions table', async () => {
```
carmenfan, line 110: this shouldn't happen

backend/tests/v5/unit/models/revisions.test.js
```diff
@@ -53,7 +53,75 @@ const testGetLatestRevision = () => {
...
+			jest.spyOn(db, 'find').mockResolvedValue(undefined);
+			await expect(Revisions.updateRevisionStatus('someTS', 'someModel', 3, true)).rejects.toEqual(templates.containerNotFound);
+		});
+
+		test('Should throw REVISION_NOT_FOUND if it cannot find the revision in the revisions table', async () => {
+			jest.spyOn(db, 'find').mockResolvedValue(expectedData);
```
carmenfan, line 116: this should mock `findOne` to return `undefined`

backend/tests/v5/unit/processors/teamspaces/projects/containers/containers.test.js
```diff
@@ -143,7 +158,37 @@ const testGetContainerStats = () => {
...
+			const res = await Containers.getRevisions('teamspace', 1, true);
+			expect(res).toEqual(model1Revisions);
+		});
+	});
+};
+
+const testUpdateRevisionStatus = () => {
+	const revision3 = model1Revisions.find((r) => r._id === 3);
+	describe('Get container revisions', () => {
+		test('should not update revision if the request body is not boolean', async () => {
```
carmenfan, line 177: not necessary as commented on `containers.js`

carmenfan (changes requested):

backend/src/v5/models/revisions.js
carmenfan: image


I lied about the v4 part... turns out only some v4 revision endpoints support tags and others don't.

going forward we want to support tags regardless.

backend/src/v5/models/revisions.js
ChristosTsiotsias: Regarding the first one I did that just to have an extra check that the revisions table exists cause I thought that by the time you create a model we automatically generate a revisions history table

backend/src/v5/models/revisions.js
ChristosTsiotsias: For the first one I thought that if the status provided is the same as the one in the db we shouldn't make the db call.

backend/src/v5/models/revisions.js
carmenfan: no we don't , nor does it matter

backend/src/v5/models/revisions.js
carmenfan: sure, but there's not a lot of reason to do that. it's more costly to run 2 queries than 1

backend/tests/v5/unit/models/revisions.test.js
ChristosTsiotsias: if I totally remove this I get an 'uncovered line' complain for line 62
image

backend/src/v5/models/revisions.js
```diff
@@ -45,4 +50,21 @@ Revisions.getRevisionCount = (teamspace, model) => {
...
+		query.void = excludeVoids.void;
+	}
+
+	const revisions = await findRevisionsByQuery(teamspace, model, query, projection);
+	return revisions;
+};
+
+Revisions.updateRevisionStatus = async (teamspace, model, revision, status) => {
+	const query = { $or: [{ _id: revision }, { tag: revision }] };
+	await findOneRevisionByQuery(teamspace, model, query, { _id: 1, void: 1 });
```
carmenfan, line 66: you don't need to run this query.

If you need to know if the revision existed, just run the `updateOne` query and [check the counts](https://docs.mongodb.com/manual/reference/method/db.collection.updateOne/#returns)

backend/src/v5/processors/teamspaces/projects/models/containers.js
```diff
@@ -56,6 +56,17 @@ Containers.getContainerStats = async (teamspace, project, container) => {
 	};
 };
 
+Containers.getRevisions = async (teamspace, container, showVoid) => {
```
carmenfan, line 59:
```suggestion
Containers.getRevisions = teamspace, container, showVoid) => getRevisions(teamspace, container, showVoid, { _id: 1, author: 1, timestamp: 1, tag: 1, void: 1 });
```

backend/src/v5/processors/teamspaces/projects/models/containers.js
```diff
@@ -56,6 +56,17 @@ Containers.getContainerStats = async (teamspace, project, container) => {
...
+	const revisions = await getRevisions(teamspace, container, showVoid,
+		{ _id: 1, author: 1, timestamp: 1, tag: 1, void: 1 });
+
+	return revisions;
+};
+
+Containers.updateRevisionStatus = async (teamspace, container, revision, status) => {
```
carmenfan, line 66:
```suggestion
Containers.updateRevisionStatus = updateRevisionStatus;
```

backend/src/v5/routes/teamspaces/projects/containers/containers.js
```diff
@@ -43,6 +43,25 @@ const getContainerStats = async (req, res) => {
...
+		respond(req, res, templates.ok, { revisions });
+	}).catch((err) => respond(req, res, err));
+};
+
+const updateRevisionStatus = async (req, res) => {
+	const { teamspace, container, revision } = req.params;
+	const status = req.body.void;
+
+	Containers.updateRevisionStatus(teamspace, container, revision, status).then(() => {
+		respond(req, res, templates.ok, {});
```
carmenfan, line 61: don't need `{}`

backend/tests/v5/e2e/routes/teamspaces/projects/containers/containers.test.js
```diff
@@ -43,6 +44,7 @@ const models = [
 		_id: ServiceHelper.generateUUIDString(),
 		name: ServiceHelper.generateRandomString(),
 		isFavourite: true,
+		permissions: [{ user: users.limitedPermissions, permission: 'viewer' }, { user: users.limitedPermissions, permission: 'commentator' }],
```
carmenfan, line 47: It's "commenter"

carmenfan: also doesn't make sense for a user to be both viewer and commenter at the same time. Make 2 users

backend/tests/v5/e2e/routes/teamspaces/projects/containers/containers.test.js
```diff
@@ -157,13 +167,14 @@ const testGetContainerStats = () => {
 			expect(res.body.code).toEqual(templates.containerNotFound.code);
 		});
 
-		test('should fail if the container doesn\'t exist', async () => {
-			const res = await agent.get(`${route('jibberish')}?key=${users.tsAdmin.apiKey}`).expect(templates.containerNotFound.status);
-			expect(res.body.code).toEqual(templates.containerNotFound.code);
+		test('should fail if the container does not exist', async () => {
+			const res = await agent.get(`${route('jibberish')}?key=${users.tsAdmin.apiKey}`).expect(templates.modelNotFound.status);
```
carmenfan, line 171: could we keep it as `containerNotFound`?

You can catch the `modelNotFound` error and check the req.params to determine whether we need containerNotFound or federationNotFound

backend/tests/v5/e2e/routes/teamspaces/projects/containers/containers.test.js
```diff
@@ -265,6 +276,120 @@ const testDeleteFavourites = () => {
...
 };
 
+const formatRevisions = (revs, includeVoid = false) => {
+	let formattedRevisions = revs;
+
+	formattedRevisions = formattedRevisions.flatMap((rev) => (includeVoid
+		|| !rev.void ? { ...rev, timestamp: rev.timestamp.toISOString() } : []));
+	formattedRevisions = formattedRevisions.sort((a, b) => b.timestamp - a.timestamp);
+
+	return { revisions: [...formattedRevisions] };
```
carmenfan, line 286: I don't think there's any reason to use the spread operator on this then just to reconstruct it into an array again?

```suggestion
	return { revisions: formattedRevisions };
```

backend/tests/v5/e2e/routes/teamspaces/projects/containers/containers.test.js
```diff
@@ -265,6 +276,120 @@ const testDeleteFavourites = () => {
...
+			expect(res.body.code).toEqual(templates.teamspaceNotFound.code);
+		});
+
+		test('should fail if the project does not exist', async () => {
+			const res = await agent.get(`/v5/teamspaces/${teamspace}/projects/dflkdsjfs/containers/${modelWithRev._id}/revisions?key=${users.tsAdmin.apiKey}`).expect(templates.projectNotFound.status);
+			expect(res.body.code).toEqual(templates.projectNotFound.code);
+		});
+
+		test('should fail if the container does not exist', async () => {
+			const res = await agent.get(`${route('jibberish')}&key=${users.tsAdmin.apiKey}`).expect(templates.modelNotFound.status);
```
carmenfan, line 308: if we can throw `containerNotFound` here that would be good

backend/tests/v5/e2e/routes/teamspaces/projects/containers/containers.test.js
```diff
@@ -265,6 +276,120 @@ const testDeleteFavourites = () => {
...
+			expect(res.body.code).toEqual(templates.teamspaceNotFound.code);
+		});
+
+		test('should fail if the user does not have access to the project', async () => {
+			const res = await agent.patch(`${route(voidRevision._id)}?key=${users.noProjectAccess.apiKey}`)
+				.send({ void: false }).expect(templates.notAuthorized.status);
+			expect(res.body.code).toEqual(templates.notAuthorized.code);
+		});
+
+		test('should fail if the user does not have adequate permissions to edit the model', async () => {
```
carmenfan, line 350:
```suggestion
		test('should fail if the user does not have adequate permissions to edit the container', async () => {
```

backend/tests/v5/e2e/routes/teamspaces/projects/containers/containers.test.js
```diff
@@ -265,6 +276,120 @@ const testDeleteFavourites = () => {
...
+
+		test('should fail if the project does not exist', async () => {
+			const res = await agent.patch(`/v5/teamspaces/${teamspace}/projects/dflkdsjfs/containers/${modelWithRev._id}/revisions/${voidRevision._id}?key=${users.tsAdmin.apiKey}`)
+				.send({ void: false }).expect(templates.projectNotFound.status);
+			expect(res.body.code).toEqual(templates.projectNotFound.code);
+		});
+
+		test('should fail if the container does not exist', async () => {
+			const res = await agent.patch(`/v5/teamspaces/${teamspace}/projects/${project.id}/containers/dfsfaewfc/revisions/${voidRevision._id}?key=${users.tsAdmin.apiKey}`)
+				.send({ void: false }).expect(templates.modelNotFound.status);
```
carmenfan, line 364:
```suggestion
				.send({ void: false }).expect(templates.containerNotFound.status);
```

backend/tests/v5/e2e/routes/teamspaces/projects/containers/containers.test.js
```diff
@@ -265,6 +276,120 @@ const testDeleteFavourites = () => {
...
+			expect(res.body.code).toEqual(templates.modelNotFound.code);
+		});
+
+		test('should fail if the revision does not exist', async () => {
+			const res = await agent.patch(`/v5/teamspaces/${teamspace}/projects/${project.id}/containers/${modelWithRev._id}/revisions/fdefaxsxsa?key=${users.tsAdmin.apiKey}`)
+				.send({ void: false }).expect(templates.revisionNotFound.status);
+			expect(res.body.code).toEqual(templates.revisionNotFound.code);
+		});
+
+		test('should return erorr if the body of the request is not boolean', async () => {
```
carmenfan, line 374:
```suggestion
		test('should return error if the body of the request is not boolean', async () => {
```

backend/tests/v5/e2e/routes/teamspaces/projects/containers/containers.test.js
```diff
@@ -265,6 +276,120 @@ const testDeleteFavourites = () => {
...
+			const res = await agent.patch(`${route(voidRevision._id)}?key=${users.tsAdmin.apiKey}`)
+				.send({ void: 123 }).expect(templates.invalidArguments.status);
+			expect(res.body.code).toEqual(templates.invalidArguments.code);
+		});
+
+		test('should return erorr if the body of the request containes exitra payload', async () => {
```
carmenfan, line 380:
```suggestion
		test('should return error if the body of the request contains extra payload', async () => {
```

backend/tests/v5/e2e/routes/teamspaces/projects/containers/containers.test.js
```diff
@@ -265,6 +276,120 @@ const testDeleteFavourites = () => {
...
+			const res = await agent.patch(`${route(voidRevision._id)}?key=${users.tsAdmin.apiKey}`)
+				.send({ void: true, extra: 123 }).expect(templates.invalidArguments.status);
+			expect(res.body.code).toEqual(templates.invalidArguments.code);
+		});
+
+		test('should update a revision\'s status if the body of the request is boolean', async () => {
+			await agent.patch(`${route(voidRevision._id)}?key=${users.tsAdmin.apiKey}`)
+				.send({ void: false }).expect(templates.ok.status);
```
carmenfan, line 388: can we use the read endpoint to verify it's been updated?

backend/tests/v5/helper/services.js
```diff
@@ -68,31 +68,28 @@ db.createModel = (teamspace, _id, name, props) => {
...
 
-ServiceHelper.generateRevisionEntry = () => ({
+ServiceHelper.generateRevisionEntry = (isVoid = false) => ({
 	_id: ServiceHelper.generateUUIDString(),
 	tag: ServiceHelper.generateRandomString(),
-	timestamp: new Date(),
 	author: ServiceHelper.generateRandomString(),
-
+	timestamp: ServiceHelper.generateRandomDate(new Date(2018, 1, 1), new Date()),
+	void: isVoid,
```
carmenfan, line 92: cast it to boolean incase somsone pass in a string 

```suggestion
	void: !!isVoid,
```

backend/tests/v5/helper/services.js
```diff
@@ -68,31 +68,28 @@ db.createModel = (teamspace, _id, name, props) => {
...
-	};
-	return DbHandler.insertOne(teamspace, `${modelId}.history`, rev);
+db.createRevision = (teamspace, modelId, revision) => {
+	const formattedRevision = { ...revision, _id: stringToUUID(revision._id) };
+	return DbHandler.insertOne(teamspace, `${modelId}.history`, formattedRevision);
 };
 
 ServiceHelper.generateUUIDString = () => uuidToString(generateUUID());
 ServiceHelper.generateRandomString = () => Crypto.randomBytes(15).toString('hex');
+ServiceHelper.generateRandomDate = (start, end) => new Date(start.getTime() + Math.random()
```
carmenfan, line 78: Might be good to give it default params so whoever use it don't have to think about it
```suggestion
ServiceHelper.generateRandomDate = (start = new Date(2018, 1, 1), end = new Date()) => new Date(start.getTime() + Math.random()
```

backend/tests/v5/unit/middleware/dataConverter/revisions.test.js
```diff
@@ -0,0 +1,90 @@
...
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
```
carmenfan, line 18: double license

backend/tests/v5/unit/middleware/dataConverter/revisions.test.js
```diff
@@ -0,0 +1,90 @@
...
+const testHasValidArgsForVoidUpdate = () => {
+	describe('Check if req arguments for void status update are valid', () => {
+		test('should respond with invalidArguments if there are no body arguments', () => {
+			const mockCB = jest.fn(() => {});
+			Revisions.hasValidArgsForVoidUpdate({ body: { } }, {}, mockCB);
+			expect(mockCB.mock.calls.length).toBe(0);
+			expect(Responder.respond.mock.calls.length).toBe(1);
+			expect(Responder.respond.mock.results[0].value).toEqual(templates.invalidArguments);
+		});
+
```
carmenfan, line 55: additional tests:
- body is of a different type (i.e. { body: 1})
- body is undefined (i.e. { })

backend/tests/v5/unit/models/projects.test.js
```diff
@@ -64,7 +64,26 @@ const testGetProjectList = () => {
...
+			const res = await Project.modelExistsInProject('someTS', 'someProject', 'a');
+			expect(res).toEqual(true);
+		});
+
+		test('should return false if a model is not part of a project', async () => {
+			const project = { models: ['a', 'b', 'c'] };
+			jest.spyOn(db, 'findOne').mockResolvedValue(project);
+			const res = await Project.modelExistsInProject('someTS', 'someProject', 'd');
+			expect(res).toEqual(false);
+		});
```
carmenfan, line 81: additional test:

- project doesn't exist should throw projectNotFound

backend/tests/v5/unit/models/revisions.test.js
```diff
@@ -53,7 +53,54 @@ const testGetLatestRevision = () => {
...
+	describe('GetRevisions', () => {
+		const expectedData = [
+			{ _id: 1, author: 'someUser', timestamp: new Date() },
+			{ _id: 2, author: 'someUser', timestamp: new Date() },
+			{ _id: 3, author: 'someUser', timestamp: new Date(), void: true },
+		];
+
+		test('Should return all container revisions', async () => {
+			jest.spyOn(db, 'find').mockResolvedValue(expectedData);
+			const res = await Revisions.getRevisions('someTS', 'someModel', true);
```
carmenfan, line 66: check the querys we're passing through.

just like the ones I did in favourites: https://github.com/3drepo/3drepo.io/blob/ISSUE_2711/backend/tests/v5/unit/models/users.test.js#L126

backend/tests/v5/unit/models/revisions.test.js
```diff
@@ -53,7 +53,54 @@ const testGetLatestRevision = () => {
...
+			expect(res).toEqual(expectedData);
+		});
+
+		test('Should return non void container revisions', async () => {
```
carmenfan, line 70: this test is not checking anything new unless you check the query.

backend/tests/v5/unit/models/revisions.test.js
```diff
@@ -53,7 +53,54 @@ const testGetLatestRevision = () => {
...
+	});
+};
+
+const testUpdateRevisionStatus = () => {
+	describe('UpdateRevisionStatus', () => {
+		const expectedData = { _id: 1, author: 'someUser', timestamp: new Date(), void: true };
+
+		test('Should update the void status of a revision', async () => {
+			jest.spyOn(db, 'findOne').mockResolvedValue(expectedData);
+			await Revisions.updateRevisionStatus('someTS', 'someModel', 3, false);
```
carmenfan, line 91: check the update query

backend/tests/v5/unit/processors/teamspaces/projects/models/containers.test.js
```diff
@@ -205,9 +220,34 @@ const testGetContainerStats = () => {
 	});
 };
 
+const testGetRevisions = () => {
+	describe('Get container revisions', () => {
+		test('should return non-void revisions if the container exists', async () => {
+			const res = await Containers.getRevisions('teamspace', 1, false);
+			expect(res).toEqual(model1Revisions.filter((r) => !r.void));
```
carmenfan, line 227: you're testing your mock implementation here, not the source code.

It's probably better to just mock Revisions.getRevisions and check it's being called with the right parameters.

backend/tests/v5/unit/processors/teamspaces/projects/models/containers.test.js
```diff
@@ -205,9 +220,34 @@ const testGetContainerStats = () => {
...
+		});
+		test('should return all revisions if the container exists', async () => {
+			const res = await Containers.getRevisions('teamspace', 1, true);
+			expect(res).toEqual(model1Revisions);
```
carmenfan, line 231: same here

backend/tests/v5/unit/processors/teamspaces/projects/models/containers.test.js
```diff
@@ -205,9 +220,34 @@ const testGetContainerStats = () => {
...
+		});
+	});
+};
+
+const testUpdateRevisionStatus = () => {
+	const revision3 = model1Revisions.find((r) => r._id === 3);
+	describe('Get container revisions', () => {
+		test('should update revision if the request body is boolean', async () => {
+			await Containers.updateRevisionStatus('teamspace', 1, revision3._id, false);
+			expect(revision3.void).toEqual(false);
```
carmenfan, line 241: again, you're checking your mock implementation 😆 

Just spy on the update call and check it's being called + with right parameters.

backend/tests/v5/unit/utils/permissions/permissions.test.js
```diff
@@ -100,14 +100,50 @@ const testHasReadAccessToModel = () => {
...
+		['b', false, true],
+		['c', false, false],
+		['projAdmin', false, false],
+		['projAdmin', true, true],
+		['tsAdmin', false, false],
+		['tsAdmin', true, true],
+		['tsAdmin', undefined, true],
+		['nobody', false, false],
+		['nobody', true, false],
+	])('Has read access to model', (user, adminCheck, result) => {
```
carmenfan, line 128:
```suggestion
	])('Has write access to model', (user, adminCheck, result) => {
```

backend/tests/v5/unit/utils/permissions/permissions.test.js
```diff
@@ -100,14 +100,50 @@ const testHasReadAccessToModel = () => {
...
+		test(`${user} ${result ? 'have' : 'does not have'} read access (adminCheck: ${adminCheck})`, async () => {
```
carmenfan, line 129:
```suggestion
		test(`${user} ${result ? 'have' : 'does not have'} write access (adminCheck: ${adminCheck})`, async () => {
```

backend/tests/v5/unit/utils/permissions/permissions.test.js
```diff
@@ -100,14 +100,50 @@ const testHasReadAccessToModel = () => {
...
+			Projects.modelExistsInProject.mockImplementation(() => true);
+			expect(await Permissions.hasWriteAccessToModel('teamspace', 'project', 'model', user, adminCheck)).toBe(result);
+		});
+	});
+
+	describe('Container does not belong to the project', () => {
```
carmenfan, line 135:
```suggestion
	describe('Has write access to model (2)', () => {
```

carmenfan (changes requested):

backend/src/v5/middleware/dataConverter/revisions.js
```diff
@@ -0,0 +1,37 @@
...
+
+const { hasWriteAccessToContainer } = require('../permissions/permissions');
+const { respond } = require('../../utils/responder');
+const { templates } = require('../../utils/responseCodes');
+const { validateMany } = require('../common');
+
+const Revisions = {};
+
+Revisions.hasValidArgsForVoidUpdate = (req, res, next) => {
+	const numOfProps = Object.keys(req.body).length;
```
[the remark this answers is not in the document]
ChristosTsiotsias: yup.js helper is not in staging yet, shall I create it in 2756 too ?

carmenfan (approved):

carmenfan (approved):

PR #2799, opened 2021-09-21
ISSUE #2757 - Hook up endpoints with basic layout

### LINKED ISSUE #2757: Hook up endpoints with basic layout
### Description
Add integration for fetching teamspaces and projects with the new endpoints
`/v5/teamspaces` for getting the list of teamspaces that will populate the select box to switch teamspaces 
`/v5/teamspaces/:teamspace/projects/` for getting the projects that will populate the select box to change projects

### Goals
- [ ] Have the selection combo box of teamspaces with data from the endpoint
- [ ] Have the selection combo box of projects  with data from the endpoint, matching the selected teamspace

### Tasks
- [ ] The logged user info should be fetch at the start (login is done in v4) and should be triggered from src\v5\ui\routes\index.tsx for now. When we have a real integration of v4 and v5 it will be have to be in a common root.
- [ ] Add api service for v5
- [ ] Add redux saga/state/selector for fetching teamspaces in v5
- [ ] Add test for saga using mockup data.
- [ ] Use   for 
- [ ] Fetch teamspaces on load in  src\v5\ui\routes\index.tsx  by triggering an action using the 
- [ ] Add redux saga/state/selector for fetching projects in v5
- [ ] Fetch projects based on the url (try to not refetch if the projects for a teamspace has already been fetched)
- [ ] Use createHooksSelectors for creating the hooks for teamspacesSelectors  and projectsSelectos  as is being used in **src\v5\services\selectorsHooks\currentUserSelectors.hooks.ts** 
- [ ] Use those SelectorsHooks in the the components to present the data.
- [ ] use  createActionsDispatchers for the actions as they are used in **src\v5\services\actionsDispatchers\currentUsersActions.dispatchers.ts** and use those to dispatch actions

m1chalkubiak:
This fixes #2757


carmenfan:
@m1chalkubiak  had a quick look as we had a problem with deployments yesterday so I was checking if this deployed properly 😆 

I found a few issues, if you think some of them are better addressed else where then please make a note in the respective issues:

**1. Teamspace drop down isn't routing to the right page**
when I go back to the teamspace list and selected a different teamspace, it's appending the teamspace name in the URL instead of replacing the existing teamspace name

aa

**2. Federation tab should be selected by default**
Federation tab should be selected by default when a project is selected. The top right are tabs as discussed so there should never be a situation where none of them are selected.

**3. Placement of the scroll bar is odd**
The scrollbar placement doesn't really match what it's trying to scroll. when you try to scroll you can see the app bar isn't moving, but the scroll bar is placed for the whole page. This is also causing the app bar to shift slightly when you change pages. I think the scroll bar should start beneath the app bar to get better alignment and represent truly what it's trying to scroll

image

Also as discussed from yesterday the styling of the tabs needs to be changed

m1chalkubiak:
Hey @carmenfan, I forgot to let you know yesterday, all the above comments and the two issues raised by George have been corrected, the branch is ready for your feedback.

 I added George's feedback below for clarity.
 
> ### Selected state for pages in the nav title bar is incorrect
> The build is showing an underline for the selected page in white, but it shouldn't be underlined and should be teal. The page should also include a teal bar underneath the selected nav title.
> 
> ### Steps to replicate
> 1. Click on a page e.g. Federations
> 2. Move your mouse away from the nav title
> 3. Notice it's underlined instead of being not underlined with a teal colour
> 
> ### Pictures & Links
> Build
> image
> 
> Design
> image
> 
> Drag and drop here

> ### Reduce spacing between the end of the project title and the chevron
> It's a little too generous the spacing, it just needs to move to the left a few pixels to match the design. Could we also please increase the size of the chevron by a few pixels to also match the design
> 
> ### Pictures & Links
> Build
> image
> 
> Design
> image
> 
> Drag and drop here

sanmont3drepo:
When I run the tests it throws an error
image

m1chalkubiak:
Hi @sanmont3drepo did You install all new dependencies? Everything seems to work fine on my machine
Zrzut ekranu 2021-09-23 o 15 08 11

sanmont3drepo:
> Hi @sanmont3drepo did You install all new dependencies? Everything seems to work fine on my machine
> Zrzut ekranu 2021-09-23 o 15 08 11

Yes I did, Ill try deleting nod_modules and yarn.lock and reinstalling.

m1chalkubiak:
> 
> Yes I did, Ill try deleting nod_modules and yarn.lock and reinstalling.

It seems to me that the problem may also be the question of how the environment in which you run the tests is interpreted. Earlier, I also had this problem and I used this solution:
https://github.com/babel/babel/issues/10086#issuecomment-501480362
I chose the second option, it must probably be a place where it works wrongly for you.

sanmont3drepo:
> > Yes I did, Ill try deleting nod_modules and yarn.lock and reinstalling.
> 
> It seems to me that the problem may also be the question of how the environment in which you run the tests is interpreted. Earlier, I also had this problem and I used this solution:
> [babel/babel#10086 (comment)](https://github.com/babel/babel/issues/10086#issuecomment-501480362)
> I chose the second option, it must probably be a place where it works wrongly for you.

Yeah, run it again after reinstalling everthing, got the same result.  I think then we have to change the test script (or add a new one?) in package.json so it sets the env and we can run it it locally .

m1chalkubiak:
@carmenfan I think it would be a good idea to add the test run to the checklist in github / Travis. So that we will make sure that the tests will run correctly both locally and on external machines.

sanmont3drepo (changes requested):

carmenfan:
> @carmenfan I think it would be a good idea to add the test run to the checklist in github / Travis. So that we will make sure that the tests will run correctly both locally and on external machines

@m1chalkubiak  

you can do this by editing .travis.yml . just add a new set of test (link shows how the backend v5 test is declared https://github.com/3drepo/3drepo.io/blob/staging/.travis.yml#L71)

m1chalkubiak:
@sanmont3drepo All above done. Only question is do you prefer some structure within the tests folder, or should the ones I just added be in a `store` subfolder, or directly like now?
Zrzut ekranu 2021-09-23 o 16 29 42

m1chalkubiak:
> > @carmenfan I think it would be a good idea to add the test run to the checklist in github / Travis. So that we will make sure that the tests will run correctly both locally and on external machines
> 
> @m1chalkubiak
> 
> you can do this by editing .travis.yml . just add a new set of test (link shows how the backend v5 test is declared https://github.com/3drepo/3drepo.io/blob/staging/.travis.yml#L71)

@carmenfan @sanmont3drepo I added and it looks like everything runs correctly.

carmenfan:
> > > @carmenfan I think it would be a good idea to add the test run to the checklist in github / Travis. So that we will make sure that the tests will run correctly both locally and on external machines
> > 
> > 
> > @m1chalkubiak
> > you can do this by editing .travis.yml . just add a new set of test (link shows how the backend v5 test is declared https://github.com/3drepo/3drepo.io/blob/staging/.travis.yml#L71)
> 
> @carmenfan @sanmont3drepo I added and it looks like everything runs correctly.

@m1chalkubiak  thanks! I just removed the SSL installation on `before_install` - we only needed it for the backend test as we're using the mongo in memory library as the database

m1chalkubiak:
@carmenfan @sanmont3drepo Do You have any feedback regarding this PR?

sanmont3drepo:
> @sanmont3drepo All above done. Only question is do you prefer some structure within the tests folder, or should the ones I just added be in a `store` subfolder, or directly like now?
> Zrzut ekranu 2021-09-23 o 16 29 42

Thanks, I think It might be better to have it into folders "teamspaces" and "projects"

m1chalkubiak:
> > @sanmont3drepo All above done. Only question is do you prefer some structure within the tests folder, or should the ones I just added be in a `store` subfolder, or directly like now?
> > Zrzut ekranu 2021-09-23 o 16 29 42
> 
> Thanks, I think It might be better to have it into folders "teamspaces" and "projects"

@sanmont3drepo Done

sanmont3drepo:
@george-hadfield  take a look at the changes and let us know if everyhing is fine!

carmenfan:
@m1chalkubiak I'm pretty sure we talked about this in a meeting, but can we stick to the routing where there's a `/t/` before the tab selection:

````
/v5/<teamspace>/<projectId>/t/<tab identifier>/federations
````

sanmont3drepo:
@george-hadfield  this is the link after you logged in 
https://issue-2757.dev.3drepo.io/v5/carmen/07c0a840-17c7-11ec-b533-193cec4d7aa3

george-hadfield:
### The tab titles are still showing underlines when clicked
I have to click off somewhere else in order for it to disappear

### Pictures and Links
image

george-hadfield (dismissed):
Approved under premise of having the underline issue fixed in the containers PR

carmenfan (approved):

PR #2806, opened 2021-10-01
Issue 2711- group import/export

### LINKED ISSUE #2711: Groups Post API to accept array of Groups and UI to allow upload
### Description

Currently our Groups card lets you download a JSON of groups (as per the screenshot below), but once you have the file there isn't a simple way to reuse it in 3D Repo. Since the POST groups API endpoint only supports a single object our customers need to understand how to loop through request and it becomes much more complicated,

In order to help our customers with some of their model validation workflows it would be very useful to allow the groups API to accept an array of groups.

### Goals

- [ ] Add an option to download smart groups only (exclude regular groups) as JSON and exclude the "objects" element from the response, as it is unnecessary bloat in this use case
- [ ] Allow an array of groups to be uploaded via the API
- [ ] Where a group exists with the same ID, perform a PUT action to replace the old record with the new one. Where one doesn't exist, add a new group and create a new ID
- [ ] Add support on the UI to "Upload JSON" of an array of groups into a model/federation

#### Front end
- [x] Add 2 filter options (toggle inside hamburger button, default on)
  - show smart groups
  - show normal groups
- [x] 2 extra options (inside hamburger button)
  - Export groups
  - Import groups

#### Backend
- [x] `HTTP POST` endpoint (POST instead of GET to avoid the problem with query string being too long) to export groups
  - payload should contain an array of groups that the user wants to export
  - The group listing will contain the standard information about groups (title, name, createdAt etc), and also
    - if the group is a normal group if will return the `ifc_guids` if possible, otherwise `shared_ids` - never both
    - if the group is a smart group, it will return the rules (not objects)
- [x] `HTTP POST` endpoint to import a group 
  - This will take in the schema exported from the previous endpoint
  - If group id already exists in the fed/model, overwrite the content, otherwise create a new group
  - Ensure chat server is wired up to fire up the notifications
- [x] Adapt the endpoints to v4
- [x] Hook up with notifications
  - [x] Establish the pub/sub service for notifications
  - [x] hook v4 notifications onto it

### Related Resources

image

carmenfan:
This fixes #2711

#### Description
- Frontend:
  - Ability to filter group list by smart/standard groups
  - Ability to export groups
  - Ability to import groups by inserting the expected data
  - Removed all duplicated groups filtering logic and make them all use the one in the selector
  
- Backend:
  - Introduced 2 new v5 routes (import/export)
    - These routes are backported for v4 (so on the front end I don't have to deal with whether it's a container/federation or what's its projectId...). The v4 route is just using the v5 middleware & processors
  - Introduced an event manager in V5.
    - This is a pub/sub service for events
    - Introduced new events: `NEW_GROUPS`, `UPDATE_GROUP`
    - These events are being listened to by v4 ChatEvent to trigger socket.io notifications
  - Added a Yup helper to help store commonly used types
  - Introduced a `schema` folder in DataConvertor
    -  To store schemas that needs to be used for both inputs and outputs
        - the group schema is used for inputs of import to validate the data
        - the group scheme is used for outputs of export to serialise the data.
  - Some refactoring on permissions logic to reduce redundancies
  - 
#### Test cases
- filter should work accordingly
- The user should be able to import/export groups

#### Bugs
- [x] Should give feedback if the import failed
- [x] adapt schema to cast away the argument
image


carmenfan:
@mosment3DRepo ready for QA

BDavydov:
@carmenfan I'm struggling to import groups back in. I can select JSON file, but nothing changes in the Groups card. :/
https://issue-2711.dev.3drepo.io/

carmenfan:
> @carmenfan I'm struggling to import groups back in. I can select JSON file, but nothing changes in the Groups card. :/ https://issue-2711.dev.3drepo.io/

works for me....

aa

Have you tried refreshing? the instant update requires socket.io to be working - which as we know is a bit flakey at best of times...

carmenfan:
@BDavydov  if refresh still shows nothing, let me know your steps & model you used

BDavydov:
@carmenfan Hmm.... Doesn't seem to work, even after Ctrl + R. Here is the model I'm exporting the Groups from.
issue-2711.dev.3drepo.io/viewer/Bogdan_D/c1f49210-24ee-11ec-91df-a52e9012a48b
Here is the model I'm uploading them to:
issue-2711.dev.3drepo.io/viewer/Bogdan_D/e677e1d0-24f0-11ec-90c9-bb535576a7c9

[Structure_04_10_2021_10_50_16_groups.zip](https://github.com/3drepo/3drepo.io/files/7277159/Structure_04_10_2021_10_50_16_groups.zip)

carmenfan:
> @carmenfan Hmm.... Doesn't seem to work, even after Ctrl + R. Here is the model I'm exporting the Groups from. issue-2711.dev.3drepo.io/viewer/Bogdan_D/c1f49210-24ee-11ec-91df-a52e9012a48b Here is the model I'm uploading them to: issue-2711.dev.3drepo.io/viewer/Bogdan_D/e677e1d0-24f0-11ec-90c9-bb535576a7c9
> 
> [Structure_04_10_2021_10_50_16_groups.zip](https://github.com/3drepo/3drepo.io/files/7277159/Structure_04_10_2021_10_50_16_groups.zip)

Ah, it failed with an invalid schema. Let me fix that.

carmenfan:
@BDavydov should be working now, but you will need to export again as the problem was on the export.

BDavydov:
@carmenfan It's working now. Thank you

backend/src/v5/middleware/dataConverter/schemas/groups.js
```diff
@@ -0,0 +1,148 @@
...
+	case 'NOT_EQUALS':
+	case 'GT':
+	case 'GTE':
+	case 'LT':
+	case 'LTE':
+		return values.every(isNumber);
+	default:
+		// range checks
+		return values.length % 2 === 0 && values.every(isNumber);
+	}
```
Charence, line 71: This works and looks good, but we could also do the same with `if..else`. I guess it only becomes a problem if someone edits the order in a wrong way.

carmenfan: I think it's pretty common understanding that case statement falls through unless you returned/ break?

carmenfan: I can change it to an if, if that's more obvious

PR #2808, opened 2021-10-04
Issue 2108 - groups number of object fix

### LINKED ISSUE #2108: Smart Groups Object Count is based on our Object IDs not element GUIDs
### Description
Model Groups provide incorrect object count. 
In this example, I have 12 Windows of Family type "FE - Dachfenster". When I create a smart group it comes up as 48 objects for Revit, Navisworks & IFC Files. 
In Naviswork model's tree structure, you can see that each Window contains up to 4 child elements. When users query Smart Groups results in 3D Repo or using APIs, they get false object count. It would be great to be able to group those objects based on their GUIDs ie. List Unique GUIDs.
If I perform an API call for List All Metadata and filter for these 12 Windows, I get 48 objects for Revit file. However, when I create the same API call for Navisworks or IFC models, I can group these results based on unique GUIDs. Therefore, I am able to get a true count of windows = 12.

### Steps to replicate
- Upload the same model in 3 file formats Revit, NWC & IFC
- Create a Smart Group for each file format
-  Compare the object count results

**Windows Schedule in Revit**
Revit Windows

**FE - Dachfenster Windows Smart Group**
Windows Smart Group

**Each Windows contains 4 Objects**
Window Smart Group
Window Parts

**Windows from Navisworks**
Window Navisworks

**Windows from IFC**
Window IFC

carmenfan:
This fixes #2108

#### Description
- added a new selector in tree that gives you the number of "nodes" given the meshes.
- added a new group helper `prepareGroupWithCount` which is a promise that uses the selector above
- added a new group action so we can call `prepareGroupWithCount` on groups updated via socket.io
- only start preparing the group after the tree has been loaded

#### Test cases
- If you have an object that has multiple meshes (e.g. a window) - the groups count should now count that as 1 object instead of the number of meshes

### DIFF
--- a/frontend/src/v4/helpers/groups.ts
+++ b/frontend/src/v4/helpers/groups.ts
@@ -18,7 +18,7 @@
 import { isEqual, omit, pick } from 'lodash';
 import { select } from 'redux-saga/effects';
 import { GROUP_TYPES_ICONS, GROUPS_TYPES } from '../constants/groups';
-import { selectGetMeshesByIds, selectGetNodesIdsFromSharedIds } from '../modules/tree';
+import { selectGetMeshesByIds, selectGetNodesIdsFromSharedIds, selectGetNumNodesByMeshSharedIdsArray } from '../modules/tree';
 import { COLOR } from '../styles';
 import { hasSameElements } from './arrays';
 import { getGroupHexColor, hexToArray } from './colors';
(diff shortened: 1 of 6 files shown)


carmenfan:
@mosment3DRepo ready for testing (once it's deployed)

carmenfan:
@mosment3DRepo rebasing to master is a bit of a nightmare due to the major folder restructuring for v5. It'll be faster if I make a new branch base of master and hand pick the changes over. So I'll do that and redeploy (which obviously will need to be retested)

@Charence hold off from reviewing the PR until that is done.

carmenfan:
use #2812 instead

PR #2823, opened 2021-10-13
ISSUE #2811 - Render a 404 page for unknown routes

### LINKED ISSUE #2811: render a 404 page for unknown routes
### Description
any routes that are not recognised should render/redirect to the 404 page

### Goals
- Redirect unknown routes to 404 page

### Tasks
<to be filled in by the developer>

### Related Resources
Figma mock up : https://www.figma.com/proto/djgxvRKOaMMyutswn5yywU/Teamspace-(live)?node-id=2260%3A33895&scaling=min-zoom&page-id=1694%3A21238

emil-litwiniec:
This fixes #2811

### Description
- Add NotFound page
- Add fallback for unknown routes

I've had to make path to dashboard more explicit to allow handling unknown routes. So example path to containers page will be: `/v5/dashboard/:teamspace/:projectId/t/containers` instead of `/v5/:teamspace/:projectId/t/containers`

### DIFF
--- a/frontend/src/v5/ui/routes/dashboard/index.tsx
+++ b/frontend/src/v5/ui/routes/dashboard/index.tsx
@@ -74,9 +74,7 @@ export const Dashboard = () => {
 					</DashboardLayout>
 				</Route>
 				<Route path="*">
-					<DashboardLayout
-						hasTopNavigation={false}
-					>
+					<DashboardLayout>
 						<NotFound />
 					</DashboardLayout>
 				</Route>
--- a/frontend/src/v5/ui/components/dashboard/dashboardLayout/dashboardLayout.component.tsx
+++ b/frontend/src/v5/ui/components/dashboard/dashboardLayout/dashboardLayout.component.tsx
@@ -23,12 +23,11 @@ import { MAIN_HEADER_PORTAL_TARGET_ID } from './dashboardLayout.constants';
 
 interface IDashboardLayout {
 	children: ReactNode;
-	hasTopNavigation?: boolean;
 }
 
-export const DashboardLayout = ({ children, hasTopNavigation = true }: IDashboardLayout): JSX.Element => (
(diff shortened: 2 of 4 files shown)


EmilTheSadCat:
@carmenfan The part with house icon, breadcrumbs stays the same. Only component containing tabs/setting button was changed. The proper way to fix it would be when we apply new version of top navigation, but I don't think it should be a part of this issue.

frontend/src/v5/ui/controls/button/button.component.tsx
```diff
@@ -18,20 +18,21 @@
 import React from 'react';
 import MuiButton, { ButtonProps } from '@material-ui/core/Button';
 import { Typography } from '@/v5/ui/controls/typography';
+import { ButtonTypeMap } from '@material-ui/core/Button/Button';
 import { LabelButton } from './button.styles';
 
 type ButtonVariants = ButtonProps['variant'] | 'label' | 'label-outlined';
 
-type IButton = Omit<ButtonProps, 'variant'> & {
+type IButton<T extends React.ElementType = ButtonTypeMap['defaultComponent']> = Omit<ButtonProps<T>, 'variant'> & {
```
sanmont3drepo, line 26: Just a  couple of questions here, not something to change.
- Why has this changed?
- is there any documentation about it?

frontend/src/v5/ui/routes/dashboard/index.tsx
```diff
@@ -46,37 +44,41 @@ export const Dashboard = () => {
...
+								<ProjectContent />
+							</Route>
+						</Switch>
+					</DashboardLayout>
+				</Route>
+				<Route exact path={`${path}/modals/`}>
+					<DashboardLayout>
+						<ModalsDemo />
+					</DashboardLayout>
+				</Route>
```
sanmont3drepo, line 75: We can get rid of this it was only temporary for showing the modals capacity

PR #2826, opened 2021-10-14
ISSUE #2800 - Show error popup for sagas errors

### LINKED ISSUE #2800: Revamping - Error message popup
### Description
When there's an error in the frontend there should be a standard way to show an error message. The idea is to have an easy to trigger  popup message with a customisable error message and the origin of the error.

### Goals
- [x] Show the error popup when something goes wrong

### Tasks
- [x] Add a template popup message that matches the ui design
- [x] Add redux action to trigger the popup with the custom message and origin
- [ ] Add tests for the actions.

### Related Resources
MockUp : https://www.figma.com/file/djgxvRKOaMMyutswn5yywU/Teamspace-(live)?node-id=1694%3A21238

m1chalkubiak:
This fixes #2800

### DIFF
--- a/.azure/build-and-deploy.yaml
+++ b/.azure/build-and-deploy.yaml
@@ -71,6 +71,8 @@ stages:
   displayName: Deploy
   jobs:
     - job: DeployHelmChart
+      pool:
+        vmImage: $(vmImageName)    
       variables:
       - name: DECODE_PERCENTS
         value: false


carmenfan:
> Hi @carmenfan @sanmont3drepo I just wanted to ask if this error message popup implementation is suitable for you? I'm still wondering whether or not to extend it the way it was in v4, namely that we would only give the method and dataType and generate the main message based on it. Currently, it is implemented in a very direct way in relation to designs.

I think we definitely need more details than message from network error.

If it's a network error we need to know about the http status and also the error code that was thrown, otherwise it can be very confusing when the user comes back to us for support.

@george-hadfield  can you enrich the template?

carmenfan:
> > @m1chalkubiak just fyi the azure pipeline is broken right now, @keifgwinn is looking into it
> 
> Ohh, I see, sorry for spamming @carmenfan !
> 
> So I would just like to consult if my current implementation is correct, below is a screenshot of what the popup would look like in the case of 404 when fetching projects Zrzut ekranu 2021-10-19 o 13 06 27

The error message and code looks a bit generic?

3drepo error message typically looks like this:
image

Let's assume this is the error message, then `Request failed with status code 404` should instead be `Teamspace not found`

and `404 - Not found` should be `404 - TEAMSPACE_NOT_FOUND`

Actually it will be useful to put the place info on there as well. maybe the same line as the status/code? @george-hadfield

carmenfan:
@m1chalkubiak  a minor point - can we remove `the` from `fetching the projects` ?

m1chalkubiak:
The title has wrapped up in the second line now, so probably George should take another look at that.

carmenfan:
@george-hadfield  available for QA https://issue-2800.dev.3drepo.io/v5/dslfkjdslf/dslfkjsdlfsd/t/federations

m1chalkubiak:
> @m1chalkubiak I think we need an error message + redirect here. I just tried to go to a valid teamspace but invalid project ID, and it just shows me the teamspace page, with the project URL and no error indication:
> 
> image

Done @carmenfan 
Zrzut ekranu 2021-10-20 o 09 52 57

Altho, maybe you want to change this text, I'm not good at copywriting.

Issue #2827, opened 2021-10-15
Feature testing Mongo 5


keifgwinn:
### Description

Initial tests have shown that we could support up to mongo v5 on the backend, and I would like to see travis have a go.

### Goals

- [ ] Travis passes all tests with a v5 mongo backend.

### Tasks
- [ ] Update travis

PR #2829, opened 2021-10-15
ISSUE #2827 - travis mongo version

keifgwinn:
This tests #2827


.azure/build-and-deploy.yaml
```diff
@@ -20,7 +20,7 @@ resources:
 
 variables:
   # Agent VM image name
-  vmImageName: 'ubuntu-20.04'
+  vmImageName: 'ubuntu-18.04'
```
carmenfan, line 23: @keifgwinn do we want this?

keifgwinn: We should be on 20.04

carmenfan (approved):

PR #2845, opened 2021-10-28
ISSUE #2662 unreal integration

keifgwinn:
#### Description
This is being spun up to review the Unreal pixel streaming for planbase for a temp deployment.

#### Test cases
We can get a working pixel streaming unreal planbase operational.


keifgwinn:
The demo is complete, but not sure if we need to pull this down.

keifgwinn:
@mosment3DRepo ?

Charence:
Worth keeping this up in case there's a need to do a demo

carmenfan:
Just to add a note so we don't forget - We've asked @3drepo/product-team  and the verdict is that we should keep this running incase we have to do some last minute demos, at least until the end of the year.

This will require a rewrite for productization, but there's not appetite for this right now

keifgwinn:
@3drepo/product-team what's the current situation with this?

carmenfan:
closing as it is old and stale and not a business priority

PR #2859, opened 2021-11-10
ISSUE #2858 only initialise v5 services on the first instance of api service

### LINKED ISSUE #2858: A single model import error triggers over 16 Notification messages
### Description

I have uploaded a file that's failing to process. In normal circumstances, I would receive a single warning message. Currently, I'm receiving up to 16 individual messages about the same error.

- This is happening in: prod
- This requires a hotfix to production: not sure

### Steps to replicate
- Login to 3D Repo
- Upload an unsupported file to any model container
- See the error message come up multiple times

### Current Behaviour
- Too many error messages

### Expected Behaviour
- Only one warning message

image

carmenfan:
This fixes #2858

#### Description
only initialise v5 services on the first instance of API server.

#### Test cases
- we should no longer get crazy amount of notifications if the 3drepo.io server has multiple api services.

### DIFF
--- a/backend/src/v4/3drepo.js
+++ b/backend/src/v4/3drepo.js
@@ -231,9 +231,18 @@ function createChat(serverConfig) {
 	require(service).createApp(server, serverConfig);
 }
 
+let firstAPIServer = true;
+
 function createService(subDomainApp, serverConfig) {
 	const service = `./services/${serverConfig.service}.js`;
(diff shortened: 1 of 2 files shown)


mosment3DRepo (approved):
Tested corrupt and normal files, notifications appear once in all instances compared to >15 on prod.

PR #2876, opened 2021-11-22
Issue 2352 - metadata schema change

### LINKED ISSUE #2352: Look into performance of the metadata rules query
### Description
We need to look into reducing the latency of the API request for metadata rules.
Doing this is taking 4s, and this model is not big: 
image

### db migration
Part of v4.22 migration. it should run with 
`yarn run-migration 4.22` (NOTE: assuming server config is specified in env_var)

carmenfan:
This fixes #2352

#### Description
Implemented changes required on the IO side for https://github.com/3drepo/3drepobouncer/issues/514

- Migration scripts to:
    - remove existing indicies that are no longer viable 
    - add new indicies for metadata queries
    - remove `current` from revision bson, append `rev_id` to individual nodes instead
    - add sharedID to mesh maps so we no longer have to fetch the unoptimised scene in asset bundle creator
- Added functions in db handler to facilitate migration scripts, and the use of `distinct` to get all possible values from a key
- Changed metadata queries to use the new schema, rewrote smart rules to query generator into v5
- Changed all queries that utilises history.current to just query the scene collection instead
- Updated toy project references

#### Test cases
- All groups, metadata related API endpoints should still work as before
- Speed of complicated/heavy queries should increase by a good margin
- Toy project should import correctly

#### Before Merge
<del> merge branch ISSUE_2352_tests into 1.13_test in the test repository. </del>


carmenfan:
@mosment3DRepo  this is ready for QA.

#### Note
Existing models will be incompatible with this branch - as they are still running under the old schema. Only models imported using the bouncer hooked up to this branch works here. We don't want to run the db migration script until this is merged in, as those models will stop working on staging branches.

So to test this, you will need to import a new model (if you're testing with toy project, if you create a new account through this branch, it will import a version of toy project that has the schema updated.). Alternatively, let me know which model (or entire teamspace) you want to convert, and i can run the script on an ad hoc basis.

#### Tests
- Toy project should import and function fine in this branch
- All metadata endpoints should function correctly

mosment3DRepo:
@carmenfan did we say that revision/master/head/meta/all.json would still return the same key:value schema rather than the new one?

carmenfan:
> @carmenfan did we say that revision/master/head/meta/all.json would still return the same key:value schema rather than the new one?

If any of the v4 endpoints give back different results it's a bug 😆 

Fixed.

backend/src/scripts/migrations/4.22/updateMetadataSchema.js
```diff
@@ -0,0 +1,54 @@
...
+const { v5Path } = require('../../../interop');
+const { getTeamspaceList, getCollectionsEndsWith } = require('../utils');
+
+const { find, updateOne } = require(`${v5Path}/handler/db`);
+const { logger } = require(`${v5Path}/utils/logger`);
+
+const processModel = async (teamspace, scene) => {
+	const meta = await find(teamspace, scene, { type: 'meta' }, { metadata: 1 });
+	const proms = meta.map(({ _id, metadata }) => {
+		if (!Array.isArray(metadata)) {
```
Charence, line 27: minor point, but this could use the array check in typeCheck?

Charence (commented):
Code looks ok. Just wondering how to test the migration scripts.

Charence (dismissed):

mosment3DRepo (dismissed):
reviewed as per https://github.com/3drepo/3D-Repo-Product-Team/issues/167

carmenfan:
Holding off until the release after

Charence (approved):

keifgwinn (approved):
Tested and approved

PR #2878, opened 2021-11-24
Issue 2810 - Add / Delete federations

### LINKED ISSUE #2810: Ability to add and delete federations
#### Add a new federation
v4 endpoint : https://3drepo.github.io/3drepo.io/#api-Model-createModel
`POST /teamspaces/:teamspace/projects/:project/federations`

payload (note: similar to v4 but not identical!)
````js
{
   name: "awesomeModel",
   unit: "ft",
   desc: "This is an awesome model!",
   code: "awe12",
   type: "Mechanical"
}
````
response (success) 
````js
{
    _id: <federation id as a string>
}
````

NOTES:
- name, units and type is compulsory, if missing, return INVALID_ARGUMENTS with bespoke messages
- make sure we also do a type check, if failed, return INVALID_ARGUMENTS with bespoke messages
- units should be recognisible (ie. must be m, mm, cm, dm or ft... or whatever is the list)
- input validity should be done as a middleware inside dataConverter, use [yup](https://www.npmjs.com/package/yup)
- ensure we convert uuid back to string before we return the response (this should be done in routes file, not processor)
- you can only create a federation if you are an admin (should be same as v4), NOT_AUTHORIZED otherwise

#### Remove a federation
Corresponding v4 endpoint: https://3drepo.github.io/3drepo.io/#api-Model-deleteModel

`DELETE /teamspaces/:teamspace/projects/:project/federations/:federation`

NOTES:
- only admins can do this.
- just return `templates.ok` upon success

### Goals
- The user should be able to create a new federation if they have necessary permissions
- The user should be able to delete a federation if they have necessary permissions

### Tasks
Add federation
  - [x] Implement endpoint
  - [x] Add unit tests
  - [x] Add e2e tests
  - [x] Add swagger doc

Delete federation
  - [x] Implement endpoint
  - [x] Add unit tests
  - [x] Add e2e tests
  - [x] Add swagger doc

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #2810 

#### Description
Endpoints for adding and deleting federations
Unit tests added
E2E tests added
Swagger documentation added

#### Test cases
The user should be able to add and delete federations provided they send valid data

### CODE REVIEW

File: backend/src/v5/models/projects.js [hunk 1]
```diff
@@ -36,6 +36,13 @@ Projects.removeModelFromProject = (ts, project, model) => updateOneProject(
 	{ $pull: { models: model } },
 );
 
+Projects.checkProjectExists = async (ts, project, projection) => {
+	const res = await findOneProject(ts, { _id: project }, projection);
```
carmenfan, line 40: is this not identical to getProjectById just below it? I think we can just use that instead?

backend/tests/v5/e2e/routes/teamspaces/projects/federations/federations.test.js
```diff
@@ -127,7 +128,12 @@ const modelSettings = [
 			},
 		},
 	},
-
+	// NOTE: this model gets deleted after deleteContainer test
```
carmenfan, line 131:
```suggestion
	// NOTE: this model gets deleted after deleteFederation test
```

carmenfan (changes requested):

carmenfan (approved):

PR #2890, opened 2021-12-03
ISSUE #2889  - writeConcern fix

### LINKED ISSUE #2889: [MONGODB DRIVER] Warning: Top-level use of w, wtimeout, j, and fsync is deprecated. Use writeConcern instead.
### Description

Since the recent DB upgrades, I've seen this warning

`[MONGODB DRIVER] Warning: Top-level use of w, wtimeout, j, and fsync is deprecated. Use writeConcern instead.
`
### Goals
- [x] Warning shouldn't throw

### Tasks

- [x] Identify all use of  use of w, wtimeout, j, and fsync and update to current standards

keifgwinn:
This fixes #2889

#### Description
- writeConcern updated
- searched for other instances but couldn't find any others.

#### Test cases
running an invite didn't throw an error


backend/src/v4/models/invitations.js
```diff
@@ -34,7 +34,7 @@ const MODELS_PERMISSION = ["collaborator", "commenter", "viewer"];
 
 const getCollection = async () => {
 	const coll = await db.getCollection("admin", "invitations");
-	coll.createIndex({ "teamSpaces.teamspace": 1 }, { "background": true, "w":1 });
+	coll.createIndex({ "teamSpaces.teamspace": 1 }, { "background": true, writeConcern: { w: "1" } });
```
carmenfan, line 37: @sanmont3drepo  I did a git blame and this is your code... 😆 

Do you remember why we are trying to create an index everytime we do getCollection here? and why did you have to specify write concern?

carmenfan (approved):

backend/src/v4/models/invitations.js
sanmont3drepo: I Dont remember. If I have to guess it was probably because a warning or an error that I got if that parameter wasnt there.

PR #2896, opened 2021-12-08
ISSUE #2868 - Import Animations bug

### LINKED ISSUE #2868: Import animations is broken
### Description
Import animation option that appears when you try to upload an spm file is not being respected on the backend, no matter whether you check the check box or not, the model is being imported with animations

### Steps to replicate
- upload an spm file
- uncheck import animations check box

### Current Behaviour
- Model is imported with animations

### Expected Behaviour
- Model should not be imported with animations

ChristosTsiotsias:
This fixes #2868 
#### Description
Bug fix added so that importAnimations is not always true when uploading new revisions

#### Test cases
Upload a new revision into a model sending an SMP file
In the request body (or from the UI) set the import animations option to false
Animations should not be uploaded


carmenfan:
@ChristosTsiotsias  
This is not quite right, if importAnimations is unset, it needs to evaluate to true, so it should only evaluate to false if it's specifically set that way.


Thinking about it, instead of fixing this, can you look into removing the v4 code, and get the v4 route to call v5 functions instead? I did something similar for group import/export as you can see [here](https://github.com/3drepo/3drepo.io/blob/staging/backend/src/v4/routes/group.js#L487) It's just calling functions imported from v5.

This will then automatically fix this, and also makes your next task easier (as you won't have to alter 2 endpoints.

ChristosTsiotsias:
> @ChristosTsiotsias This is not quite right, if importAnimations is unset, it needs to evaluate to true, so it should only evaluate to false if it's specifically set that way.
> 
> Thinking about it, instead of fixing this, can you look into removing the v4 code, and get the v4 route to call v5 functions instead? I did something similar for group import/export as you can see [here](https://github.com/3drepo/3drepo.io/blob/staging/backend/src/v4/routes/group.js#L487) It's just calling functions imported from v5.
> 
> This will then automatically fix this, and also makes your next task easier (as you won't have to alter 2 endpoints.

Ah okay I thought the default value was false. Sure I will try to do it the way you say

backend/src/v4/routes/model.js
```diff
@@ -1714,7 +1714,7 @@ router.patch("/:model/upload/ms-chunking/:corID", middlewares.hasUploadAccessToM
  * ------WebKitFormBoundarySos0xligf1T8Sy8I-- *
  *
  */
-router.post("/:model/upload", middlewares.hasUploadAccessToModel, uploadModel);
+router.post("/:model/upload", middlewares.chechPermsAndFormatV5NewRevisionsData, validateNewRevisionData, uploadModel);
```
carmenfan, line 1717: Let's keep them separate 
```suggestion
router.post("/:model/upload", middlewares.hasUploadAccessToModel, middlewares.formatV5NewRevisionsData, validateNewRevisionData, uploadModel);
```

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/commons/revisions.js
```diff
@@ -53,36 +54,37 @@ Revisions.validateUpdateRevisionData = async (req, res, next) => {
...
 		});
 
 	try {
 		req.body = await schema.validate(req.body);
 		if (!req.file) throw createResponseCode(templates.invalidArguments, 'A file must be provided');
+
+		const { teamspace } = req.params;
+		await sufficientQuota(teamspace, req.file.size);
+
+		const isTagValid = await isValidTag(req.params.teamspace, req.params.container, req.body.tag);
```
carmenfan, line 82: Can we do that as part of the test in L70?

I.e. line 70 should look something like this:

````js
  tag: YupHelper.types.strings.code.required().test("tag-not-in-use", "Revision name is already used by an existing revision", async () => { /* the tag test here*/})
````

File: backend/src/v5/models/revisions.js [hunk 3]
```diff
@@ -67,4 +67,13 @@ Revisions.updateRevisionStatus = async (teamspace, model, revision, status) => {
 	}
 };
 
+Revisions.isValidTag = async (teamspace, model, tag) => {
```
carmenfan, line 70: Naming of the function is a bit weird, what we're testing here is if tag is already in use. so maybe it's better to call it `tagInUse` or something

carmenfan (changes requested):

PR #2920, opened 2021-12-20
Issue 2892 - SPM file revisions have Timezones assigned

### LINKED ISSUE #2892: SPM upload timezone selection
### Description

Currently Synchro records all activity in UTC time, this means that when they are uploaded to 3D Repo the epoch times display incorrectly if a user is working outside of UTC or in a timezone which uses daylight savings.

This can be fixed by overriding the browser timezone sensor to UTC, but this is not a sustainable workflow

### Goals

- [x] Allow users to set the Timezone when uploading an SPM file
  - [x] Support 2 dropdowns in the user interface to first allow selection of the country, followed by the timezone (to avoid a long list of timezones) using this list or similar as a basis https://en.wikipedia.org/wiki/List_of_time_zones_by_country
    - We should be able to utilise something like this for the information https://www.npmjs.com/package/countries-and-timezones
  - [x] Set the default timezone when uploading via the 3D Repo UI to the users timezone as presented by the browser

- [x] add extra parameters on the model upload request to pass in timezone information
  - [x]  Use Timezone naming standards from tz database https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

### Related Resources
image

ChristosTsiotsias:
This fixes #2892 

#### Description
Two new dropdowns added in the New Revision window for SPM revisions: Country and Timezone.
The user can now select a country and the specific timezone which will be assigned to the revision.
The country and timezone dropdowns have default selection based on the info we get from the browser.
New param added to the upload endpoint to accept timezone (input validation added).

#### Test cases
The new revision window (SPM upload) should have the new dropdowns.
The dropdowns should be preselected correctly.
A new upload request with valid timezone should store the timezone in the DB.
A new upload request with invalid timezone should throw error.


ChristosTsiotsias:
I have managed to get the timezone info from the browser, that means the timezone DD should always have the correct default value. However I have not managed to get the country info. The reason is that I can only get the lat long coordinates and we then have to make a 3rd party api call to get the country which is paid. 

What happens now is that I get the timezone and set the country as the first country to use this timezone. I think most timezones correspond to one country but there might be cases where a timezone applies to more than one.

mosment3DRepo:
@ChristosTsiotsias after speaking with George and testing out the PR we think it would be better to switch to a single, longer dropdown.

In order to make sense for the end user we should implement the Google Calendar method of listing the timezones with their GMT+/- hours prefixed to the name:

image

carmenfan:
@ChristosTsiotsias  I merged in the animation flag fix - can you take a look at fixing the conflicts?

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/commons/revisions.js
```diff
@@ -77,6 +78,15 @@ const validateRevisionUpload = async (req, res, next) => {
 				}),
 			desc: YupHelper.types.strings.shortDescription,
 			importAnimations: Yup.bool().default(true),
+			timezone: Yup.string().test('valid-timezone',
+				'The timezone provided is not valid',
+				(value) => {
+					if (value !== undefined) {
+						const allTimezones = tz.getAllTimezones();
+						return Object.values(allTimezones).map((t) => t.name).includes(value);
```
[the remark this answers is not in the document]
carmenfan: actually, would it be  possible to use `getTimezone` instead?

Looking at the docs, this should work? (assuming they return undefined if it's an invalid timezone name?)

````js
return !!tz.getTimezone(value);
````

File: backend/tests/v5/unit/middleware/dataConverter/inputs/teamspaces/projects/models/commons/revisions.test.js [hunk 2]
```diff
@@ -113,7 +114,7 @@ Quota.sufficientQuota.mockImplementation((ts) => (ts === 'noQuota' ? Promise.rej
 RevisionsModel.isTagUnique.mockImplementation(() => true);
 
 const testValidateNewRevisionData = () => {
-	const standardBody = { tag: '123', description: 'this is a model', importAnim: false };
+	const standardBody = { tag: '123', description: 'this is a model', importAnimations: false, timezone: 'Europe/Berlin' };
```
carmenfan, line 117: could we add tests where we pass in `timezone` as:
1. null
2. some other type (like a number)

frontend/src/v4/routes/teamspaces/components/uploadModelFileDialog/uploadModelFileDialog.component.tsx
```diff
@@ -93,6 +108,18 @@ export class UploadModelFileDialog extends React.PureComponent<IProps, IState> {
...
+
+		const timezones = allTimezones.map((tz) => {
+			return {
+				name: tz.name,
+				label: `(GMT ${tz.utcOffsetStr}) ${tz.name}`
+			}
+		});
+
+		const defaultTimezone = timezones.find((tz) => tz.name === browserTimezone)
+			??  timezones.find((tz) => tz.name === 'Europe/London');
```
carmenfan, line 121: We should probably default to `Etc/UTC`, as London could be BST in the summer

frontend/src/v4/routes/teamspaces/components/uploadModelFileDialog/uploadModelFileDialog.component.tsx
```diff
@@ -93,6 +108,18 @@ export class UploadModelFileDialog extends React.PureComponent<IProps, IState> {
 		const { modelId, teamspaceName, fetchModelSettings, fetchRevisions } = this.props;
 		fetchRevisions(teamspaceName, modelId, true);
 		fetchModelSettings(teamspaceName, modelId);
+		const browserTimezone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+
+		const timezones = allTimezones.map((tz) => {
```
carmenfan, line 113: with L120 and this we're looping the timezones twice. 

We could just deduce the default timezone whilst we're restructuring the data

frontend/src/v4/routes/teamspaces/components/uploadModelFileDialog/uploadModelFileDialog.component.tsx
```diff
@@ -179,11 +227,13 @@ export class UploadModelFileDialog extends React.PureComponent<IProps, IState> {
...
 			this.setState({
-				allowImportAnimations: true
+				allowImportAnimations: true,
+				allowSetTimezone: true
 			}, () => setFieldValue('importAnimations', true));
 		} else if (this.state.allowImportAnimations) {
 			this.setState({
-				allowImportAnimations: false
+				allowImportAnimations: false,
+				allowSetTimezone: false
```
carmenfan, line 236: I think this should be a separate else if?

or we just do else without the condition

frontend/src/v4/routes/teamspaces/components/uploadModelFileDialog/uploadModelFileDialog.component.tsx
```diff
@@ -213,7 +263,7 @@ export class UploadModelFileDialog extends React.PureComponent<IProps, IState> {
 		return (
 			<Formik
 				onSubmit={this.handleFileUpload}
-				initialValues={{ revisionName: '', revisionDesc: '', file: '', importAnimations: false, }}
+				initialValues={{ revisionName: '', revisionDesc: '', file: '', importAnimations: false }}
```
carmenfan, line 266: should we not be setting `allowSetTimezone` too?

frontend/src/v4/routes/teamspaces/components/uploadModelFileDialog/uploadModelFileDialog.container.ts
```diff
@@ -24,7 +24,7 @@ import {
 	selectIsPending,
 	selectRevisions,
 	selectSettings,
-	ModelActions,
+	ModelActions
```
carmenfan, line 27: I think you are probably making the linter cry by removing these commas 😆

backend/tests/v5/unit/middleware/dataConverter/inputs/teamspaces/projects/models/commons/revisions.test.js
ChristosTsiotsias: if I pass null, it interprets it as undefined for some reason and then in the validation its the same as not including timezone at all :/ can I skip null ?

carmenfan: no, it would be good to add it, it's more for the situations where people alter the code in the future, if it break this test then we know

carmenfan:
Looks fine, waiting on https://github.com/3drepo/3drepobouncer/issues/526 to test the full pipeline

carmenfan:
@mosment3DRepo  @miadibe  this is ready for testing

carmenfan (approved):

mosment3DRepo (approved):
Tested as per:

https://github.com/3drepo/3D-Repo-Product-Team/issues/163

PR #2925, opened 2022-01-05
ISSUE #2862 Smart groups export bug fix

### LINKED ISSUE #2862: Can't import Smart Groups JSON (exported from another federation)
### Description

I'm trying to export a select few smart groups from Residential Federation in 3D Repo. When I upload the JSON export to another federation, it doesn't work

- This is happening in: prod

### Steps to replicate
- Open this model in 3drepo.io [Residential Federation](www.3drepo.io/viewer/Demo_3D_Repo/19ce3220-6e89-11ea-9f08-1387f50c0299)
- Open the Groups card
- Filter for Smart Groups ONLY
- In the search bar type in CS and hit enter
- Export the groups
- Open this model in 3drepo.io [MEP vs STR](www.3drepo.io/viewer/Demo_3D_Repo/2e30c430-4308-11ec-80ea-2de55b1b3ca1)
- Open Smart Groups Card
- Import groups (recently exported JSON)

### Current Behaviour
- Fails to import "Request failed with status code 400"

### Expected Behaviour
-  Import Groups

[Smart Groups JSON Export](https://github.com/3drepo/3drepo.io/files/7521789/Federation_11_11_2021_15_57_50_groups.zip)

**Groups to Export**
Smart Groups Bug

**Import Error**
Smart Groups Bug 2

**3D Repo Error Message**
Import Smart Groups

ChristosTsiotsias:
This fixes #2862 

#### Description
Bug fix added so that smart groups export Json has all dates in integer format

#### Test cases
Find a smart group that has a createdAt property in date format and no updatedAt property in the database
Export the group as json and import it in another model
The export and import should be successful

### DIFF
--- a/backend/src/v5/middleware/dataConverter/schemas/groups.js
+++ b/backend/src/v5/middleware/dataConverter/schemas/groups.js
@@ -143,6 +143,13 @@ const schema = (group, strict = false) => {
 };
 
 Group.validateSchema = (group) => schema(group, true).validate(group);
-Group.castSchema = (group) => schema(group).cast(group);
+Group.castSchema = (group) => {
+	const castedSchema = schema(group).cast(group);
 
+	if (typeof castedSchema.updatedAt !== 'number') {
+		castedSchema.updatedAt = castedSchema.createdAt;
+	}
+
+	return castedSchema;
+};
 module.exports = Group;
--- a/backend/src/v5/middleware/dataConverter/schemas/groups.js
+++ b/backend/src/v5/middleware/dataConverter/schemas/groups.js
@@ -143,13 +143,11 @@ const schema = (group, strict = false) => {
 };
 
 Group.validateSchema = (group) => schema(group, true).validate(group);
+
 Group.castSchema = (group) => {
-	const castedSchema = schema(group).cast(group);
-
-	if (typeof castedSchema.updatedAt !== 'number') {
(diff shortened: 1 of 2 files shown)


backend/src/v5/middleware/dataConverter/schemas/groups.js
```diff
@@ -143,6 +143,13 @@ const schema = (group, strict = false) => {
 };
 
 Group.validateSchema = (group) => schema(group, true).validate(group);
-Group.castSchema = (group) => schema(group).cast(group);
+Group.castSchema = (group) => {
+	const castedSchema = schema(group).cast(group);
```
carmenfan, line 147: It doesn't make a lot of sense to anyone who's reading this to see that if `updatedAt` is not a number then it should be taking the `createdAt` value 😆 

It's probably better to just prefill the value before casting:

````js
const groupToCast = {...group};
groupToCast.updatedAt = groupToCast.updatedAt || groupToCast.createdAt;
return schema(group).cast(group);
````

carmenfan (changes requested):

backend/src/v5/middleware/dataConverter/schemas/groups.js
ChristosTsiotsias: sure that makes sense :P

carmenfan (approved):

PR #2929, opened 2022-01-06
ISSUE #2919 - units renamed to unit in container stats endpoint and tests

### LINKED ISSUE #2919: Revamp - Fix inconsistent naming of "units" in endpoints
### Description
Endpoints are using various names for a container's units.
getContainerStats uses `"units"`
addContainer, updateSettings, getSettings use `"unit"`

### Expected Behaviour
- All endpoints should use the same term (most likely `"unit"` since that's what the majority use)

ChristosTsiotsias:
This fixes #2919 

#### Description
Units property renamed to units in container stats endpoint. Similar changes added to tests


carmenfan (changes requested):
I think this breaks the v5 front end - @The-Daniel @sanmont3drepo  can someone go in and take a look?

carmenfan (approved):

Issue #2927, opened 2022-01-05
Responsiveness in search bar in Federations


sanmont3drepo:
### Description
When the browser gets resized the search bar in the containers lists doesnt look good

### Goals
- [ ]  The search bar should look good at all resolutions

### Tasks
- [ ] Fix the search bar accordingly to design in figma

### Related Resources
[Design](https://www.figma.com/file/djgxvRKOaMMyutswn5yywU/Teamspace-(live)?node-id=3279%3A48088)

PR #2939, opened 2022-01-12
ISSUE #2904 - Responsive federations list

### LINKED ISSUE #2904: Revamping -  Federations responsiveness
### Description
Federations should be able to handle  when the browser is resized

### Goals
- [ ] Columns should scale the same as containers
- [ ] Responsive: Content needs margin left and right at smaller breakpoints.

### Tasks
- [ ] ...

### Related Resources
image

[MockUp](https://www.figma.com/file/djgxvRKOaMMyutswn5yywU/Teamspace-(live)?node-id=3274%3A46976)

sanmont3drepo:
This fixes #2904 

#### Description
- Responsive desing for federations
- It hides the columns of the federations list as the user resizes the window

#### Test cases
- Resize the window and see how it responds in this [link](https://issue-2904.dev.3drepo.io/v5/dashboard)

### DIFF
--- a/frontend/src/v5/ui/components/dashboard/dashboardList/dashboardListItem/components/dashboardListItemButton/dashboardListItemButton.component.tsx
+++ b/frontend/src/v5/ui/components/dashboard/dashboardList/dashboardListItem/components/dashboardListItemButton/dashboardListItemButton.component.tsx
@@ -18,27 +18,21 @@
 import React, { Dispatch, ReactNode, SyntheticEvent } from 'react';
 import { FixedOrGrowContainer } from '@controls/fixedOrGrowContainer';
 import { Tooltip } from '@material-ui/core';
-import { Display } from '@/v5/ui/themes/media';
+import { IFixedOrGrowContainer } from '@controls/fixedOrGrowContainer/fixedOrGrowContainer.component';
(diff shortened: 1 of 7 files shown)


mosment3DRepo:
### Odd drop at 1024px wide

When you resize the width below 1024 the area above the accordion doubles in height
ResizeDrop1

george-hadfield:
Yeah I'd be happy with that. Approved from my end.

mosment3DRepo:
# New Federation Priority over Search box

I think we should fix the New Federation button and reduce the size of the search box on width changes @george-hadfield  thoughts?

new proposal on top, current implementation on the bottom:
search compress

george-hadfield:
The current implementation is not how I've seen this previously. Should
work how it does on the containers section. Has there been a deployment
issue with an old version?

sanmont3drepo:
> # New Federation Priority over Search box
> I think we should fix the New Federation button and reduce the size of the search box on width changes @george-hadfield thoughts?
> 
> new proposal on top, current implementation on the bottom: search compress

The problems with that bit will be adressed in this issue:
https://github.com/3drepo/3drepo.io/issues/2927

Regarding to the list is there anything else?

george-hadfield (approved):

carmenfan (approved):

PR #2941, opened 2022-01-14
Issue 2924 - Profile endpoints

### LINKED ISSUE #2924: Profile endpoints
### Description
Create v5 endpoints to facilitate the[ profile page.](https://www.figma.com/file/djgxvRKOaMMyutswn5yywU/Teamspace-(live)?node-id=3676%3A29600)

#### GET user profile
v4 endpoint: https://3drepo.github.io/3drepo.io/#api-User-getProfile

route: `GET /user` 
payload: none
response: same as v4. (note: the current v4 payload also generates an intercom hash - let's ignore for now, i feel like it should come from a different endpoint)

#### Update user details
v4 endpoint: https://3drepo.github.io/3drepo.io/#api-Account-updateUser
route: `PUT /user`
payload: same as v4 (note:  password update is also done via this endpoint, but it doesn't show on the documentation)
response: none (just http code is fine)

Notes:
- make sure we check character length, email validation, password strength etc on the data.
- this endpoint should **NOT** be accessible via API key authentication

#### Generate API Key
v4 endpoint : https://3drepo.github.io/3drepo.io/#api-User-generateApiKey

route: `POST /user/key`
payload: none
response: same as v4

NOTE: 
- this endpoint should **NOT** be accessible via API key authentication

#### Delete API key
v4 endpoint: https://3drepo.github.io/3drepo.io/#api-User-deleteApiKey_HTTP_1_1_200_OK___

route: `DELETE /user/key`
payload: none
response: none (http code only, unless errored)

NOTE: 
- this endpoint should **NOT** be accessible via API key authentication
- 
#### GET Avatar
v4 endpoint : https://3drepo.github.io/3drepo.io/#api-Account-getAvatar

route: `GET /user/avatar`
payload: none
response: same as v4.

#### Upload a new avatar
v4 endpoint: https://3drepo.github.io/3drepo.io/#api-Account-uploadAvatar 

route: `PUT /user/avatar` (Note this is a **PUT** when it used to be **POST**)
payload: same as v4 - apart from we don't need `user` field, just a file is fine.

- Make sure we are still imposing the file size limit
- check supported format

### Goals
- [x] The user should be able to get their profile information
  - [x] This endpoint should work with API keys authentication
- [x] The user should be able to update their profile information
  - [x] This endpoint should **not** work with API keys authentication
  - [x] Payload data should be validated, and relevant error input if it failed
- [x] The user should be able to generate/delete api keys
  - [x] This endpoint should **not** work with API keys authentication
- [x] The user should be able to view their current avatar
  - [x] returns 404 not found if there is no avatar
- [x] The user should be able to update their avatar
  - [x] validation check should be in place for supported format, file size.

### Tasks
_To be filled in by the developer_

ChristosTsiotsias:
This fixes #2924 

#### Description
Implementation, swagger doc, unit tests and e2e tests added for profile endpoints.

#### Test cases
The user should be able to reach all endpoints as long as they provide valid data.


backend/tests/v5/e2e/routes/users.test.js
```diff
@@ -152,16 +156,225 @@ const testGetUsername = () => {
...
+		});
+
+		test('should fail if the user has a session via an API key', async () => {
+			const data = { firstName: 'newName' };
+			const res = await agent.put(`/v5/user?key=${testUser.apiKey}`)
+				.send(data).expect(templates.notLoggedIn.status);
+			expect(res.body.code).toEqual(templates.notLoggedIn.code);
+		});
+
+		test('should fail if the update data have invalid email', async () => {
```
carmenfan, line 203: might be better to group these ones (the ones where you have logged in) into another describe, so you can [do some setup and tear down ](https://jestjs.io/docs/setup-teardown), something like 
````js
describe("With valid authentication", () => {
    beforeAll(async () => {
        //log in before doing all the tests that assumes the user has logged in
        await testSession.post('/v5/login/').send({ user: testUser.user, password: testUser.password });
   });
    afterAll(async() => {
         // log out after this block of tests to ensure we don't contaminate other test cases
        await testSession.post('/v5/logout/');
    });

    test(/*...*/);
    test(/*...*/);
    test(/*...*/);
})
````

Describe can be nested so you should be able to just have another describe block here

File: backend/tests/v5/e2e/routes/users.test.js [hunk 2]
```diff
@@ -152,16 +156,225 @@ const testGetUsername = () => {
...
+			const data = { oldPassword: testUser.password, newPassword: testUser.password };
+			await testSession.post('/v5/login/').send({ user: testUser.user, password: testUser.password });
+			const res = await testSession.put('/v5/user/').send(data).expect(templates.invalidArguments.status);
+			expect(res.body.code).toEqual(templates.invalidArguments.code);
+		});
+
+		test('should update the profile if the user is logged in', async () => {
+			const data = { firstName: 'newName' };
+			await testSession.post('/v5/login/').send({ user: testUser.user, password: testUser.password });
+			await testSession.put('/v5/user/').send(data).expect(200);
```
carmenfan, line 255: Can we do a get here to check that the name really has been updated?

backend/tests/v5/e2e/routes/users.test.js
```diff
@@ -152,16 +156,225 @@ const testGetUsername = () => {
...
+		});
+
+		test('should update the profile and change password if the user is logged in', async () => {
+			const data = { firstName: 'newName', oldPassword: testUser.password, newPassword: 'Passport123!' };
+			await testSession.post('/v5/login/').send({ user: testUser.user, password: testUser.password });
+			await testSession.put('/v5/user/').send(data).expect(200);
+			await testSession.put('/v5/user/').send({ oldPassword: 'Passport123!', newPassword: testUser.password });
```
carmenfan, line 262: same here, do a test to make sure it has changed

backend/tests/v5/e2e/routes/users.test.js
```diff
@@ -152,16 +156,225 @@ const testGetUsername = () => {
...
+			const res = await agent.post(`/v5/user/key?key=${testUser.apiKey}`)
+				.expect(templates.notLoggedIn.status);
+			expect(res.body.code).toEqual(templates.notLoggedIn.code);
+		});
+
+		test('should create and return a new Api key if the user is logged in', async () => {
+			await testSession.post('/v5/login/').send({ user: testUser.user, password: testUser.password });
+			const res = await testSession.post('/v5/user/key').expect(200);
+			const userProfileRes = await testSession.get('/v5/user');
+			expect(res.body).toEqual({ apiKey: userProfileRes.body.apiKey });
```
carmenfan, line 284: Can we verify the api key works?

carmenfan: Also add a test where there's already an existing api key and generating a new one means the old one should stop working

carmenfan: Be good to call logout after this test to we're not contaminating the next test it runs (applies to other places but i'll stop pointing it out)

backend/tests/v5/e2e/routes/users.test.js
```diff
@@ -152,16 +156,225 @@ const testGetUsername = () => {
...
+			const res = await agent.delete(`/v5/user?key=${testUser.apiKey}`)
+				.expect(templates.notLoggedIn.status);
+			expect(res.body.code).toEqual(templates.notLoggedIn.code);
+		});
+
+		test('should delete the Api key if the user is logged in', async () => {
+			await testSession.post('/v5/login/').send({ user: testUser.user, password: testUser.password });
+			await testSession.delete('/v5/user/key').expect(200);
+			const userProfileRes = await testSession.get('/v5/user');
+			expect(userProfileRes.body.apiKey).toEqual(undefined);
```
carmenfan, line 306: can we also do a check that the api key will not longer be valid? (i.e. do a get request with it and see)

backend/tests/v5/e2e/routes/users.test.js
```diff
@@ -152,16 +156,225 @@ const testGetUsername = () => {
...
+
+const testGetAvatar = () => {
+	describe('Get the avatar of the logged in user', () => {
+		test('should fail if the user is not logged in', async () => {
+			const res = await agent.get('/v5/user/avatar').expect(templates.notLoggedIn.status);
+			expect(res.body.code).toEqual(templates.notLoggedIn.code);
+		});
+
+		test('should fail if the user has a session via an API key', async () => {
+			const res = await agent.get(`/v5/user/avatar?key=${testUser.apiKey}`)
```
carmenfan, line 319: I think we said the user should be able to get the avatar with api key?

carmenfan (changes requested):

backend/tests/v5/e2e/routes/users.test.js
ChristosTsiotsias: should I do that for the other tests where I update the email as well ?

carmenfan: yea any PUT/POST if we can do a GET to verify that'd be good

backend/tests/v5/e2e/routes/users.test.js
ChristosTsiotsias: you mean testuser.ApiKey or userProfileRes.body.apiKey ? cause the second is undefined

carmenfan: i mean once you call delete on the key, use that key to do a http request again and see if you get an error. So it would be the api key that you would get from the endpoint prior to you deleting it

backend/src/v5/models/users.js
```diff
@@ -158,4 +162,53 @@ User.deleteFavourites = async (username, teamspace, favouritesToRemove) => {
...
+	const updateUserCmd = {
+		updateUser: username,
+		pwd: newPassword,
+	};
+
+	await db.runCommand('admin', updateUserCmd);
+	await updateUser(username, { $unset: { 'customData.resetPasswordToken': 1 } });
+};
+
+User.updateProfile = async (username, updatedProfile) => {
```
carmenfan, line 175: can we move this to the processor?

Basically models should be tightly knitted to the schema of the database, this is mostly trying to convert data format from what we get from network to what we need to do in the datbaase, which is more of a proccessor's  job

So the processor function should just call a `changePassword` function, and/or `update` function in here.

backend/src/v5/routes/users.js
```diff
@@ -103,11 +165,176 @@ const establishRoutes = () => {
...
+	* /user/avatar:
+	*   get:
+	*     description: Gets the avatar of the logged in user
+	*     tags: [Auth]
+	*     operationId: getAvatar
+	*     responses:
+	*       401:
+	*         $ref: "#/components/responses/notLoggedIn"
+	*       200:
+	*         description: Updates the details of the user
```
carmenfan, line 314: this doesn't render:
image

carmenfan (changes requested):

carmenfan (changes requested):
@ChristosTsiotsias  sorry, we had a review of this UI yday and I realised we're missing 2 new fields on the endpoint:

image


Can we add them? in the database, they reside in `customData.billing.billingInfo`

backend/src/v5/middleware/dataConverter/inputs/users.js
```diff
@@ -65,6 +66,10 @@ Users.validateUpdateData = async (req, res, next) => {
 				}
 				return true;
 			}),
+		company: Yup.string(),
```
carmenfan, line 69: We should impose lengths. We should never allow user to give us indefinite number of characters

use `strings.title`

carmenfan (changes requested):
@ChristosTsiotsias can we also add a e2e that tests those fields have been updated?

carmenfan (approved):

PR #2956, opened 2022-01-19
ISSUE #2955 - Dropdown shows correct selected project

### LINKED ISSUE #2955: Wrong selection the project dropdown
### Description
If a project different from the first one from the list of projects has been selected, the dropdown shows the wrong project being selected

### Steps to replicate
- navigate to a project which is not the first one from the list
- click the project dropdown from the navigations breadcrumb

### Current Behaviour
- The first project is being shown as selected

### Expected Behaviour
- It should show the actual selected project as the selected in the list

image

Amantini1997:
This fixes #2955

#### Description
The selected project is correctly highlighted from the dropdown menu

#### Test cases
Select a project which is not the first one and see that it gets correctly highlighted


sanmont3drepo:
- I know that this issue specifies only selected project, but , can we see if there's a way to make it work with both the project and the teamspace?

Problem with teamspace:
image


- There's a commit  (as well as the PR name) that has the wrong issue number. To fix this you can use https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

Amantini1997:
> 



> * I know that this issue specifies only selected project, but , can we see if there's a way to make it work with both the project and the teamspace?
> 
> Problem with teamspace: image
> 
> * There's a commit  (as well as the PR name) that has the wrong issue number. To fix this you can use https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History ( the interactive bit)

I fixed the commits issue number.

I also fixed the selection for all the items of the breadcrumbs, but I adopted a non-dynamic solution. This means that, should the breadcrumbs display more (or different) items other than _teamspace_ and _project_, the `navigationMenu.component` should be modified to account for that.

frontend/src/v5/ui/components/shared/navigatonMenu/navigationMenu.component.tsx
```diff
@@ -42,11 +44,19 @@ export const NavigationMenu = ({ anchorEl, handleClose, list }: INavigationMenu)
 			horizontal: 'left',
 		},
 	};
+	const { project, teamspace } = useParams();
+	const selectedItem = project || teamspace;
```
sanmont3drepo, line 48: Can we move this to breadcrumbs.component and pass a selectedItem prop to this component?  
I think we might need to refactor that component (I dont believe is easy to read), but for now I think is better to have this decision there. Which is the component that 'knows' which option has been selected.

sanmont3drepo (changes requested):

frontend/src/v5/ui/components/shared/navigatonMenu/navigationMenu.component.tsx
Amantini1997: Done

sanmont3drepo (approved):

PR #2981, opened 2022-01-27
#2980 travis block merge when the rabbitmq is set

### LINKED ISSUE #2980: Add a check on travis to see if the branch is hooked up to a specific queue
### Description
Some features require a feature branch 3drepobouncer for testing, which requires a change in `.azure/build-and-deploy.yaml`

When the PR is ready to merge, we don't actually want to merge in bespoke queue setting.

As a quick fix, we need something to alert the user to revert the change before merging in the branch - and the quickest way to do this right now is probably add a test on the travis yaml to see if the queue dns has been overridden - fail the test if it is.

### Goals
- travis failure if dns queue is overridden

keifgwinn:
This fixes #2980

#### Description

- [x] Add new test

#### Test cases

- [x] Travis should fail if rabbitmq is mentioned in the customHelmOverride variable
- [x] Travis should succeed if the customHelmOverride  does not contain rabbitmq

### DIFF
--- a/.azure/build-and-deploy.yaml
+++ b/.azure/build-and-deploy.yaml
@@ -31,7 +31,7 @@ variables:
   branchName: master
   group: tests-group
   customHelmOverride: |
-    config.ADDITIONAL_CONFIG=hello
+     config.APP_QUEUE_DNS=issue-xxx-rabbitmq,config.ADDITIONAL_CONFIG=hello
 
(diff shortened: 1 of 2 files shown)


keifgwinn:
Successfully failing with rabbitmq set.

carmenfan (approved):

PR #3011, opened 2022-02-08
ISSUE #3010 create index at init, remove primary write concern

### LINKED ISSUE #3010: we should not be enforcing any mongo write concern in the code
### Description
We're currently enforcing a write concern on an index creation on the invitations collection: https://github.com/3drepo/3drepo.io/blob/master/backend/src/v4/models/invitations.js#L37

firstly, the syntax is incorrect as it should be { w: 1} (1 should not be a string).
secondly, it is throwing error for non replica set deployments:
image

Unless there's a legitimate reason (which should be documented in the code), we should not be deciding read/write concerns on the code base, this should be left to the considerations of DevOps.

We also should not be attempting to create an index everytime we get the collection.

carmenfan:
This fixes #3010

#### Description
- moved index creation to when the file is executed (i.e. init)
- removed write concern specification

#### Test cases
- sign up should no longer error
- create invitations should still work

### DIFF
--- a/backend/src/v4/models/invitations.js
+++ b/backend/src/v4/models/invitations.js
@@ -32,11 +32,7 @@ const { omit } = require("lodash");
 
 const MODELS_PERMISSION = ["collaborator", "commenter", "viewer"];
 
-const getCollection = async () => {
-	const coll = await db.getCollection("admin", "invitations");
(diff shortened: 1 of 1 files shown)


keifgwinn:
LGTM

keifgwinn (approved):
LGTM

PR #3061, opened 2022-02-22
Issue 3027 - Fetch container views endpoint 

### LINKED ISSUE #3027: Endpoints to fetch views from containers
### Description
endpoint to fetch view list from federations was adding in #2789

We need to do the same for containers.

NOTE: this issue will branch from #2789 for simplicity and to avoid conflicts

### Goals
- [ ] Fetching view list from containers should succeed

### Tasks
- [ ] Create endpoint to fetch views list
- [ ] refactor file-type to utils

ChristosTsiotsias:
This fixes #3027

#### Description
Fetch container views endpoint implemented.
Swagger documentation and unit - e2e tests added.

#### Test cases
The user should be able to fetch container views provided they pass correct details
The user should get an empty array if there are no views in a container
The user should get an error if they pass invalid details

### DIFF
--- a/backend/src/v5/middleware/dataConverter/multer.js
+++ b/backend/src/v5/middleware/dataConverter/multer.js
@@ -16,8 +16,8 @@
  */
 
 const { createResponseCode, templates } = require('../../utils/responseCodes');
-const FileType = require('file-type');
 const Multer = require('multer');
+const { fileTypeFromBuffer } = require('../../utils/helper/typeCheck');
 const { respond } = require('../../utils/responder');
 const { fileUploads: uploadConfig } = require('../../utils/config');
 const { validateMany } = require('../common');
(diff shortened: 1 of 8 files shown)


backend/src/v5/middleware/dataConverter/multer.js
```diff
@@ -61,7 +61,7 @@ const imageFilter = (req, file, cb) => {
 const ensureFileIsImage = async (req, res, next) => {
 	const fileBuffer = req.file?.buffer;
 	if (fileBuffer) {
-		const type = await FileType.fromBuffer(fileBuffer);
+		const type = await fileTypeFromBuffer(fileBuffer);
```
carmenfan, line 64: Can we extrapolate this a bit more

maybe split it into 2 functions

one that returns the extension, the other returns the MIME type

carmenfan (changes requested):
Looks fine, just a small change on the file-type refactoring

backend/src/v5/middleware/dataConverter/multer.js
ChristosTsiotsias: something like this ?
image

carmenfan: sure, i don't think you need the top one though

You just need another function that returns you the mime type

PR #3064, opened 2022-02-22
ISSUE #3016 - Page will go blank if the error from a fetch is 404

### LINKED ISSUE #3016: Page will go blank if the error from a fetch is 404
### Description
When there's a request and the result is 404 error the whole page will go blank. I'm not sure if its a major problem, but it could potentially be. I found out while switching branches and a 404 error returned (because that feature was not in the branch and I didn't restart the server) 

### Steps to replicate
- Change a fetch  request to a non existing endpoint

### Current Behaviour
- The page goes blank

### Expected Behaviour
- An alert modal with the 404 error returned from the server should pop up

The-Daniel:
This fixes #3016

#### Description
- Fixed the alert modal so that it can handle responses that do not contain the message in JSON format

#### Test cases
- Set the API destination for a get request (e.g. `fetchContainers()` in _/services/api/containers.ts_) to a non-existent location. 
- Observe that the alert modal appears and can be exited.
- Set off a different error (e.g. attempt to delete one of the sample containers, or create a container with a name that already exists)
- Observe that the modal is unchanged

### DIFF
--- a/frontend/src/v5/ui/components/shared/modals/templates/alertModal/alertModal.component.tsx
+++ b/frontend/src/v5/ui/components/shared/modals/templates/alertModal/alertModal.component.tsx
@@ -20,23 +20,31 @@ import { Button, DialogContent, DialogContentText, DialogTitle } from '@material
 import WarningIcon from '@assets/icons/warning.svg';
 import { FormattedMessage } from 'react-intl';
 import { Container, Actions, Details, Status } from '@/v5/ui/components/shared/modals/modals.styles';
+import { AxiosError } from 'axios';
 
 interface IAlertModal {
 	onClickClose?: () => void,
 	currentActions?: string
 	errorMessage?: string;
-	error?: {
-		request: {
-			response: string;
-		};
-	};
+	error: AxiosError;
 	details?: string
 }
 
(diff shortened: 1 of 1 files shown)


frontend/src/v5/ui/components/shared/modals/templates/alertModal/alertModal.component.tsx
```diff
@@ -20,23 +20,31 @@ import { Button, DialogContent, DialogContentText, DialogTitle } from '@material
...
-	};
+	error: AxiosError;
 	details?: string
 }
 
 export const AlertModal: React.FC<IAlertModal> = ({ onClickClose, currentActions = '', error, details, errorMessage }) => {
-	const responseData = error?.request?.response ? JSON.parse(error?.request?.response) : {};
-	const { message, status, code } = responseData;
-	const errorStatus = `${status} - ${code}`;
+	let message; let code;
```
Amantini1997, line 34: Do we have any rules that force us to declare variables on different lines?

The-Daniel: Do you mean ES Lint rules? If so then no

Amantini1997 (approved):

PR #3067, opened 2022-02-23
ISSUE #3003 - Realtime service for v5 

### LINKED ISSUE #3003: Realtime system
### Description
We need a clean system to hook up to chat events in v5. 

### Goals
- [ ] It's easy to setup 
- [ ] It's easy to add new events, shouldn't feel hacky.
- [ ] Should be transparent for sagas or redux

### Tasks
- [ ] Try out several designs

sanmont3drepo:
This fixes #3003 

#### Description
- Provides a way to subscribe to realtime events
- Added an example of how to use it in federations

### DIFF
--- a/frontend/src/v5/services/realtime/realtime.service.ts
+++ b/frontend/src/v5/services/realtime/realtime.service.ts
@@ -85,7 +85,7 @@ export const unsubscribeToEvent = (event, callback) => {
 	socket.off(event, callback);
 };
 
-export const subscribeToRoomEvent = (roomType: IRoomType, event, callback) => {
+export const subscribeToRoomEvent = (roomType: IRoomType, event: string, callback) => {
 	joinRoom(roomType);
 	subscribeToEvent(event, callback);
 
@@ -95,11 +95,21 @@ export const subscribeToRoomEvent = (roomType: IRoomType, event, callback) => {
 	};
 };
 
-export const subscribeToDm = (callback) => {
-	subscribeToEvent('message', callback);
+interface IDirectMessage {
+	event: string;
+	data: any;
+}
+
+export const subscribeToDM = (event: string, callback) => {
+	const dmCallback = (message: IDirectMessage) => {
+		if (message.event !== event) return;
+		callback(message.data);
+	};
+
+	subscribeToEvent('message', dmCallback);
 
 	return () => {
-		unsubscribeToEvent('message', callback);
+		unsubscribeToEvent('message', dmCallback);
 	};
 };
 
--- a/frontend/src/v5/services/realtime/realtime.service.ts
+++ b/frontend/src/v5/services/realtime/realtime.service.ts
@@ -24,6 +24,11 @@ interface IRoomType {
 	notifications?: boolean
 }
 
+export enum SocketEvents {
+	CONNECT = 'connect',
+	DISCONNECT = 'disconnect',
+}
+
 let socket:Socket = null;
 const roomsJoined:Record<string, number> = {};
 
@@ -41,25 +46,19 @@ interface IChatConfig {
 	reconnectionAttempts: number
 }
 
-export const initializeSocket = ({ host, path, reconnectionAttempts }: IChatConfig) => {
-	socket = io(host, {
-		path,
-		reconnectionAttempts,
-		transports: ['websocket'],
-		reconnection: true,
-		reconnectionDelay: 500,
-	});
-
-	// eslint-disable-next-line @typescript-eslint/no-use-before-define
-	socket.on('connect', reJoinRooms);
-};
-
 export const joinRoom = (roomType : IRoomType) => {
 	const joinedCount = (roomsJoined[roomTypeToId(roomType)] || 0);
 	roomsJoined[roomTypeToId(roomType)] = joinedCount + 1;
 	if (joinedCount > 0) return;
 
 	socket.emit('join', roomType);
+};
+
+const reJoinRooms = () => {
+	const roomTypes = Object.keys(roomsJoined).map(idToRoomType);
+	roomTypes.forEach((roomType) => {
+		socket.emit('join', roomType);
+	});
 };
 
 export const leaveRoom = (roomType : IRoomType) => {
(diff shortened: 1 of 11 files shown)


frontend/src/v5/services/realtime/realtime.service.ts
```diff
@@ -0,0 +1,74 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { io, Socket } from 'socket.io-client';
+
+const CHAT_SERVER = 'http://local.3drepo.io:3000';
```
carmenfan, line 20: This needs to be extracted from the config?

frontend/src/v5/services/realtime/realtime.service.ts
```diff
@@ -0,0 +1,74 @@
...
+
+interface IRoomType {
```
carmenfan, line 22: I don't know whether you want to think about this now,

but as it stands, we have 2 rooms

{ teamspace, project, model } - for model related notifications
{ notification: true } - for user's notifications

So technically everything is optional here

frontend/src/v5/services/realtime/realtime.service.ts
```diff
@@ -0,0 +1,74 @@
...
+	project: string;
+}
+
+let socket:Socket = null;
+const roomsJoined:Record<string, number> = {};
+
+const roomTypeToId = (roomType:IRoomType) => `${roomType.teamspace}.${roomType.project}.${roomType.model}`;
+
+export const initializeSocket = () => {
+	socket = io(CHAT_SERVER, {
```
carmenfan, line 34: we need to handle reconnection by hooking into the connect event, [like v4](https://github.com/3drepo/3drepo.io/blob/staging/frontend/src/v4/modules/chat/chat.sagas.ts#L41)

carmenfan: message event needs to be listened to for logout events https://github.com/3drepo/3drepo.io/blob/staging/frontend/src/v4/modules/chat/chat.sagas.ts#L43

carmenfan: Another thing is that the backend sends back error messages like this: https://github.com/3drepo/3drepo.io/blob/master/backend/src/v4/services/chat.js#L129


But i'm thinking now whether I should change it in v5 to be more like the logout events? (so it's a direct message with the details)

Otherwise you'd have to hook onto every possible error event to get any error handling.

carmenfan (changes requested):

frontend/src/v5/services/realtime/realtime.service.ts
```diff
@@ -0,0 +1,106 @@
...
+	socket = io(host, {
+		path,
+		reconnectionAttempts,
+		transports: ['websocket'],
+		reconnection: true,
+		reconnectionDelay: 500,
+	});
+
+	// eslint-disable-next-line @typescript-eslint/no-use-before-define
+	socket.on('connect', reJoinRooms);
```
carmenfan, line 54: It doesn't look like `reJoinRooms` function calls this function , so can't we just move the `reJoinRooms` declaration above this?

frontend/src/v5/services/realtime/realtime.service.ts
```diff
@@ -0,0 +1,106 @@
...
+
+	if (joinedCount !== 0) return;
+	delete roomsJoined[roomTypeToId(roomType)];
+
+	socket.emit('leave', roomType);
+};
+
+const reJoinRooms = () => {
+	const roomTypes = Object.keys(roomsJoined).map(idToRoomType);
+	roomTypes.forEach(joinRoom);
```
carmenfan, line 77: I don't think you actually want to call `joinRoom` here, that will +1 to your roomsJoined counter unnecessarily? You'll end up with zombie counts due to reconnects

I think you just want to call `socket.emit` here, that will also save you from converting the id back to the room type

frontend/src/v5/services/realtime/realtime.service.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export const subscribeToEvent = (event, callback) => {
+	socket.on(event, callback);
+};
+
+export const unsubscribeToEvent = (event, callback) => {
+	socket.off(event, callback);
+};
+
+export const subscribeToRoomEvent = (roomType: IRoomType, event, callback) => {
```
carmenfan: for non-room events or dm, we're presenting the user with a `subscribeToEvent` and a `unsubscribeToEvent` function

But for room events and dms, we're presenting a subscribe function and a callback function to unsubscribe..

Considering we're exposing all of them to the outside world, the inconsistency is odd.

carmenfan (changes requested):

frontend/src/v5/services/realtime/realtime.service.ts
sanmont3drepo: Yes, I just didn't liked that in that place is not grouped in my own mental map of logical blocks (all the join functions together)

carmenfan: in that case maybe move this function down? or would that be odd too?

frontend/src/v5/services/realtime/realtime.service.ts
carmenfan: Not sure if it's really a good solution, but you could just make subscribeToEvent return a callback to unsubscribe and not expose unsubscribe? 

At least that'd be consistent

frontend/src/v5/services/realtime/realtime.service.ts
sanmont3drepo: 🧟‍♂️

frontend/src/v5/services/realtime/realtime.service.ts
sanmont3drepo: I did that, but also added a enum for those  events as (I think) they should be limited.

carmenfan (approved):

Issue #3072, opened 2022-02-25
test mongo 3.2 with RC


carmenfan:
this is a testing branch to see if we 3.2 is compatible with the latest RC for 4.22

carmenfan:
closing as this is a testing branch to test 3.2 compatibility

PR #3074, opened 2022-02-25
#3072 - testing rc with mongo 3.2

keifgwinn:
This tests #3072

#### Description
We're connecting this branch to an older version of mongo to check backwards compatibility. This pull request exists to  use the automation.

#### Test cases
full test suite passes.


mosment3DRepo:
### Logging into Plugins kicks the user out of the web

When a user logs in to the plugins (Publish Model) while the are logged in to .io in the browser, it kicks them out of the browser session.

This works with 4.20 Plugins as well as 4.22 Plugins

https://user-images.githubusercontent.com/62879582/155766440-ae637f8a-416e-4657-90e3-519aeb03f9ec.mp4

carmenfan:
@keifgwinn  i had to do an update as Matt found a bug in the source code. Can you re-do your magic please 😆

keifgwinn:
done, and fixed automation so if you need to do more changes it should work now.

PR #3097, opened 2022-03-09
Issue 2891 - script to dump out members of all teamspaces

### LINKED ISSUE #2891: A script to fetch teamspace info
### Description
Wrote a script to fetch members in teamspaces for @mosment3DRepo , comitting this so it can be run easily in the future

This relies on #2352 , as it relies on some db functions that were introduced there.

carmenfan:
This fixes #2891

#### Description
- added a utility script that will produce a CSV file of all the members of each teamspaces

#### Test cases
The following script should work and produce a csv (will require to set NODE_ENV to the config folder being used):
> yarn run-script .\src\scripts\utility\teamspaces\findAllMembersInAllTeamspaces.js


backend/src/scripts/common/utils.js
```diff
@@ -0,0 +1,35 @@
...
+const { v5Path } = require('../../interop');
+
+const { listDatabases, listCollections } = require(`${v5Path}/handler/db`);
+const { USERNAME_BLACKLIST } = require(`${v5Path}/models/users.constants`);
+
+const Utils = {};
+
+Utils.getTeamspaceList = async () => {
+	const dbList = await listDatabases();
+	return dbList.flatMap(({ name: db }) => (USERNAME_BLACKLIST.includes(db) ? [] : db));
```
ChristosTsiotsias, line 27: couldn't we just do 

return dbList.flatMap(({ name }) => (USERNAME_BLACKLIST.includes(name) ? [] : name)); ?

ChristosTsiotsias, line 27: couldn't we just do 

return dbList.flatMap(({ name }) => (USERNAME_BLACKLIST.includes(name) ? [] : name)); ?

carmenfan:
> We could also include the role ie admin or team_member, or we dont care ?

don't care.

This is used for a very specific purpose

ChristosTsiotsias (approved):

PR #3105, opened 2022-03-14
Issue 3085 - Forgot password endpoints

### LINKED ISSUE #3085: Forgot password endpoints
### Description
create 2 endpoints for v5 version of forgotten password

#### Trigger request password
v4 reference: https://3drepo.github.io/3drepo.io/#api-Account-forgotPassword
v5 route: `POST /user/password`
payload:
````js
{
    user: "<username or email>"
}
````

- This should return 200 OK regardless of whether the user exists.
- Use v4 mailer/template for now (v5 email design TBC)

#### Reset the password
v4 reference: https://3drepo.github.io/3drepo.io/#api-Account-forgotPassword
v5 route: `PUT /user/password`
payload:
````js
{
    user: "<username>",
    token: "<token from email>",
    newPassword: "xxx"
}
````

- the new password should conform to the schema restriction

### Goals
- The user should be able to request password for a valid username or email
- The user should be able to reset the password with a valid token
- The endpoint should return failure if the payload is invalid
- The endpoint should return ok if the user request password for an invalid username/email
- the endpoint should fail with a relevant error message if the token has expired (or is invalid)

### Tasks

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #3085 

Endpoints for sending forgot password email and resetting password added.
Unit and E2E tests added.
Swagger documentation added

#### Test cases
The user should get an email with a reset password link using the post 'user/password' endpoint and should be able to reset their password using the put 'user/password' endpoint provided they pass correct details

### CODE REVIEW

File: backend/src/v5/middleware/dataConverter/inputs/users.js [hunk 1]
```diff
@@ -120,6 +120,45 @@ const validateAvatarData = async (req, res, next) => {
...
+
+		next();
+	} catch (err) {
+		respond(req, res, createResponseCode(templates.invalidArguments, err?.message));
+	}
+};
+
+Users.validateResetPasswordData = async (req, res, next) => {
+	const schema = Yup.object().shape({
+		token: Yup.string().required(),
```
carmenfan, line 148: check token validity here.

the processor should just change the password.

backend/src/v5/models/users.js
```diff
@@ -20,7 +20,9 @@ const config = require('../utils/config');
 const db = require('../handler/db');
 const { events } = require('../services/eventsManager/eventsManager.constants');
 const { generateHashString } = require('../utils/helper/strings');
+const { logger } = require('../utils/logger');
 const { publish } = require('../services/eventsManager/eventsManager');
+const { sendResetPasswordEmail } = require('../../v4/mailer/mailer');
```
carmenfan, line 25: should be using v5 mailer (even if it's just importing v4 mailer)

There should be no reference to v4 from v5 (apart from the ones that are importing the whole file e..g mailer, db)

backend/src/v5/models/users.js
```diff
@@ -207,4 +209,35 @@ User.getAvatar = async (username) => {
 
 User.uploadAvatar = (username, avatarBuffer) => updateUser(username, { $set: { 'customData.avatar': { data: avatarBuffer } } });
 
+User.resetPasswordToken = async (username) => {
```
carmenfan, line 212: Naming is a bit ambiguous here. I think what we're doing is not resetting a password token, but generating a token

```suggestion
User.generateResetPasswordToken = async (username) => {
```

backend/src/v5/models/users.js
```diff
@@ -207,4 +209,35 @@ User.getAvatar = async (username) => {
...
+		const expiryAt = new Date();
+		expiryAt.setHours(expiryAt.getHours() + config.tokenExpiry.forgotPassword);
+
+		resetPasswordToken = {
+			token: generateHashString(64),
+			expiredAt: expiryAt,
+		};
+
+		await updateUser(username, { $set: { 'customData.resetPasswordToken': resetPasswordToken } });
+	} catch {
```
carmenfan, line 230: you don't need this if you checked the user exists at the middleware

backend/src/v5/models/users.js
```diff
@@ -207,4 +209,35 @@ User.getAvatar = async (username) => {
 
 User.uploadAvatar = (username, avatarBuffer) => updateUser(username, { $set: { 'customData.avatar': { data: avatarBuffer } } });
 
+User.resetPasswordToken = async (username) => {
+	let email; let resetPasswordToken; let
```
carmenfan, line 213: you don't need the first try, so you shouldn't need the declarations here

backend/src/v5/models/users.js
```diff
@@ -207,4 +209,35 @@ User.getAvatar = async (username) => {
...
+		firstName;
+
+	try {
+		const { customData } = await User.getUserByQuery({ user: username }, { user: 1, 'customData.email': 1, 'customData.firstName': 1 });
```
carmenfan, line 217: Don't need the next 2 lines

```suggestion
		const { customData: { email, firstName } } = await User.getUserByQuery({ user: username }, { user: 1, 'customData.email': 1, 'customData.firstName': 1 });
```

backend/src/v5/models/users.js
```diff
@@ -207,4 +209,35 @@ User.getAvatar = async (username) => {
...
+		email = customData.email;
+		firstName = customData.firstName;
+
+		const expiryAt = new Date();
+		expiryAt.setHours(expiryAt.getHours() + config.tokenExpiry.forgotPassword);
+
+		resetPasswordToken = {
+			token: generateHashString(64),
```
carmenfan, line 225: default is fine, there's no reason for it to be 64 characters long.

```suggestion
			token: generateHashString(),
```

backend/src/v5/middleware/dataConverter/inputs/users.js
```diff
@@ -120,6 +120,45 @@ const validateAvatarData = async (req, res, next) => {
...
+		.required();
+
+	try {
+		await schema.validate(req.body);
+
+		try {
+			const usernameOrEmail = req.body.user;
+			const { user } = await getUserByQuery({ $or: [{ user: usernameOrEmail }, { 'customData.email': usernameOrEmail }] });
+			req.body.user = user;
+		} catch {
```
carmenfan, line 136: just respond `templates.ok` here and call next() after L135 instead of L140

```suggestion
                   next();
		} catch {
		   respond(req,res, templates.ok);
```

backend/src/v5/processors/users.js
```diff
@@ -72,4 +73,16 @@ Users.getAvatar = getAvatar;
 
 Users.uploadAvatar = uploadAvatar;
 
+Users.resetPasswordToken = (username) => resetPasswordToken(username);
```
carmenfan, line 76:
```suggestion
Users.resetPasswordToken = resetPasswordToken;
```

backend/src/v5/processors/users.js
```diff
@@ -72,4 +73,16 @@ Users.getAvatar = getAvatar;
...
+
+Users.resetPassword = async (username, token, newPassword) => {
+	const { customData } = await getUserByUsername(username);
```
carmenfan, line 79: Few things here:
1. test the user validity/token validity at middleware So at the processor you're just performing the operation instead of testing preconditions.
2. you want to throw invalid token even if username doesn't exist.
3. Let the database query check the token validity for you by passing in the token/date you expect - so you don't have to do L81 manually.

backend/src/v5/processors/users.js
```diff
@@ -72,4 +73,16 @@ Users.getAvatar = getAvatar;
...
+
+	if (customData?.resetPasswordToken?.token !== token || customData.resetPasswordToken.expiredAt < new Date()) {
+		throw templates.invalidToken;
+	}
+
+	await updatePassword(username, newPassword);
```
carmenfan, line 85: you also need to wipe the token as you update. tokens should be single use. (can just call `$unset` in the query)

backend/src/v5/routes/users.js
```diff
@@ -361,6 +385,64 @@ const establishRoutes = () => {
...
+	*                 type: string
+	*                 description: The username of the user
+	*                 example: username123
+	*               newPassword:
+	*                 type: string
+	*                 description: The new password of the user
+	*                 example: newPassword123!
+	*               token:
+	*                 type: string
+	*                 description: The reset passsword token
```
carmenfan, line 436:
```suggestion
	*                 description: The reset password token
```

backend/src/v5/routes/users.js
```diff
@@ -361,6 +385,64 @@ const establishRoutes = () => {
...
+	*                 example: c0f6b97ae5a9c210ee050a9ada3faabc
+	*     responses:
+	*       401:
+	*         $ref: "#/components/responses/notLoggedIn"
```
carmenfan, line 440: not a good error example since it's not an endpoint that requires logging in

backend/src/v5/routes/users.js
```diff
@@ -361,6 +385,64 @@ const establishRoutes = () => {
...
+	*           schema:
+	*             type: object
+	*             properties:
+	*               user:
+	*                 type: string
+	*                 description: The username or email of the user
+	*                 example: nick.wilson@email.com
+	*     responses:
+	*       401:
+	*         $ref: "#/components/responses/notLoggedIn"
```
carmenfan, line 407: not a good error example since it's not an endpoint that requires logging in

backend/src/v5/utils/responseCodes.js
```diff
@@ -29,10 +29,12 @@ ResponseCodes.templates = {
 	notAuthorized: { message: 'You do not have sufficient access rights for this action', status: 401 },
 	licenceExpired: { message: 'Licence expired.', status: 401 },
 	tooManyLoginAttempts: { message: 'Too many unsuccessful login attempts! Account locked', status: 400 },
-	userNotVerified: { message: 'Account not yet verified. Please check your email.', status: 400 },
+	userNotVerified: { message: 'Account not yet verified. Please check your email', status: 400 },
+	accountLocked: { message: 'Account locked. Please try again later', status: 400 },
```
carmenfan, line 33: this is not used?

backend/tests/v5/e2e/routes/users.test.js
```diff
@@ -433,6 +441,82 @@ const testDeleteApiKey = () => {
...
+		});
+
+		test('should fail if the new password is too weak', async () => {
+			const res = await agent.put('/v5/user/password').send({ newPassword: 'Abcdef1!', token: 'some random token', user: testUserWithToken.user })
+				.expect(templates.invalidArguments.status);
+			expect(res.body.code).toEqual(templates.invalidArguments.code);
+		});
+
+		test('should fail if user is not found', async () => {
+			const res = await agent.put('/v5/user/password').send({ newPassword: 'Abcdefghijk1!', token: 'some random token', user: 'invalid user' })
```
carmenfan, line 490: return invalid token here

we don't want people to use this endpoint to figure out usernames

backend/tests/v5/e2e/routes/users.test.js
```diff
@@ -433,6 +441,82 @@ const testDeleteApiKey = () => {
...
+			await agent.post('/v5/user/password').send({ user: testUser.user })
+				.expect(templates.ok.status);
+		});
+	});
+};
+
+const testResetPassword = () => {
+	describe('Reset user password', () => {
+		test('should fail if a token is not provided', async () => {
+			const res = await agent.put('/v5/user/password').send({ newPassword: 'Abcdefghijk1!', user: testUserWithToken.user })
```
carmenfan, line 466: Don't have to change this (unless you want to)

but should use `generateRandomString()` on any strings you don't care about. (that's what it's for afterall!)
```suggestion
			const res = await agent.put('/v5/user/password').send({ newPassword: generateRandomString(), user: testUserWithToken.user })
```

backend/tests/v5/e2e/routes/users.test.js
```diff
@@ -433,6 +441,82 @@ const testDeleteApiKey = () => {
...
+
+		test('should fail if user token is different than the one provided', async () => {
+			const res = await agent.put('/v5/user/password').send({ newPassword: 'Abcdefghijk1!', token: 'different token', user: testUserWithToken.user })
+				.expect(templates.invalidToken.status);
+			expect(res.body.code).toEqual(templates.invalidToken.code);
+		});
+
+		test('should reset user password', async () => {
+			await agent.put('/v5/user/password').send({ newPassword: 'Abcdefghijk1!', token: validPasswordToken.token, user: testUserWithToken.user })
+				.expect(templates.ok.status);
```
carmenfan, line 515: call again with the same token and expect failure

carmenfan: also should try to log on to see if the password has really changed.

config/testV5/config.js
```diff
@@ -154,6 +154,23 @@ module.exports = {
 		{title: "Cookies", type: "haveRead", page: "cookies", fileName: "cookies.html"}
 	],
 	userNotice: false,
-	termsUpdatedAt: 12345678
+	termsUpdatedAt: 12345678,
+
+	mail: {
```
carmenfan, line 159: i dont' think you need to put in actual mailer info?

your unit test is mocking the mailer

e2e may fail, but it returns ok to the user anyway?

carmenfan (changes requested):

config/testV5/config.js
carmenfan: if you have to, put in some super dummy data with https://ethereal.email/

don't spam google smtp :P

backend/src/v5/processors/users.js
ChristosTsiotsias: this is happening in updatePassword model method

backend/src/v5/routes/users.js
ChristosTsiotsias: true 🤣

config/testV5/config.js
ChristosTsiotsias: I think we said that if an email fails to be sent we should let the user know

carmenfan: This is the config for v5 tests, has nothing to do with actual processing 

Yes you need mailer configured when you run it, Just wondering if we need to set something for the test (i.e. does it fail in the wrong way if it does? if the answer is yes, use etheral to generate dummy but valid credentials)

carmenfan: Just generate the ethereal mail on the fly. nodemailer has an api for that - so it changes everytime and  no one can log in and snoop on the emails.

They have an example on the link provided. 
image

carmenfan (changes requested):

backend/src/v5/middleware/dataConverter/inputs/users.js
```diff
@@ -120,6 +120,55 @@ const validateAvatarData = async (req, res, next) => {
...
+		user: Yup.string().required(),
+	}).strict(true).noUnknown()
+		.required();
+
+	try {
+		await schema.validate(req.body);
+
+		try {
+			const usernameOrEmail = req.body.user;
+			const { user } = await getUserByQuery({ $or: [{ user: usernameOrEmail }, { 'customData.email': usernameOrEmail }] });
```
carmenfan, line 134: Found a bug here.

Email search needs to be case insensitive. if i add random caps in my email it doesn't find me. 

Can we move this query into model/users.js and make sure it's case insensitive. We're going to run into the same problem when we do registration so let's make sure it's consistent.

Also make sure we add a e2e test case for it

carmenfan: PS: we should add projection here, we only watn the username, shouldn't drag the whole user document out

carmenfan (changes requested):
@ChristosTsiotsias 

I made some tiny changes for clean up. But I found a bug that needs addressing 😆

backend/src/v5/middleware/dataConverter/inputs/users.js
ChristosTsiotsias: we already have getUserByUsername should I just convert that to getUserByUsernameOrEmail ?
image

carmenfan: nah we want to make sure we don't check for email if the intention is just the username

so probably a new `getUserByUsernameOrEmail`

carmenfan (approved):

PR #3109, opened 2022-03-16
Issue 3056 - Refactor types in frontend

### LINKED ISSUE #3056: Split out types files
### Description
Right now the types files (for example containers or federations) have all the types in the same file regarding that objects domain. I think splitting them out and moving the types to files which those types have more context (like the IProps in the components) can make the types files more managable and improve readbility.

### Goals
- [ ] Everything should work as before
- [ ] Types should be in the proper context

### Tasks
- [ ] move the payload types to the api files
- [ ] move the action types to the redux files
- [ ] leave the shared types in the same file

sanmont3drepo:
This fixes #3056 

#### Description
- Renamed some types
- Grouped some types for more reusability
- Moved out some types to files that provides a bit more context


frontend/src/v5/store/store.types.ts
```diff
@@ -0,0 +1,31 @@
...
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export type TeamspaceId = { teamspace: string };
+export type ProjectId = { projectId: string };
+export type ContainerId = { containerId: string };
+export type FederationId = { federationId: string };
+
+export type TeamspaceAndProjectId = TeamspaceId & ProjectId;
```
The-Daniel, line 23: I really like this method :)

frontend/src/v5/services/api/containers.ts
```diff
@@ -17,24 +17,49 @@
...
+	ContainerStats,
+	NewContainer,
 } from '@/v5/store/containers/containers.types';
+import { TeamspaceAndProjectId, TeamspaceProjectAndContainerId } from '@/v5/store/store.types';
 import api from './default';
 
+/**
+ * Types
+*/
+type FetchContainersParams = TeamspaceAndProjectId;
```
The-Daniel, line 29: What's the main advantage of renaming these types instead of just using them (e.g. TeamspaceAndProjectId) directly in the redux functions below?

frontend/src/v5/store/federations/federations.types.ts
```diff
@@ -14,8 +14,6 @@
  *  You should have received a copy of the GNU Affero General Public License
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-import { Action } from 'redux';
-
 export interface IFederation {
```
The-Daniel, line 17: Is there a logic behind where we use the "I" names? (e.g. *IFederationsState* instead of just *FederationsState*). I understand that these are interfaces and not types, but in this same file there is the interface *SurveyPoint*. Should this be *ISurveyPoint*?

Could it also be worth using the "I" names for types as well? Just so it's clear that FederationStats is a type and not an object containing federation stats? I suppose in the context it's used it should always be quite clear, but it's something to consider

frontend/src/v5/services/api/federations.ts
```diff
@@ -16,32 +16,37 @@
...
+	FederationBackendSettings,
+	MinimumFederation,
+	FederationView,
+	FederationStats,
 } from '@/v5/store/federations/federations.types';
+import { TeamspaceAndProjectId, TeamspaceProjectAndFederationId } from '@/v5/store/store.types';
 import { AxiosResponse } from 'axios';
 import api from './default';
 
+type FavouriteParams = TeamspaceProjectAndFederationId;
```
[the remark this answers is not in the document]
The-Daniel: Maybe also for all the redux files?

frontend/src/v5/store/federations/federations.sagas.ts
```diff
@@ -16,21 +16,10 @@
  */
 
 import { all, put, takeEvery, takeLatest } from 'redux-saga/effects';
-import { FederationsActions, FederationsTypes } from '@/v5/store/federations/federations.redux';
+import { AddFavouriteAction, DeleteFederationAction, FederationsActions, FederationsTypes, FetchFederationsAction, FetchFederationSettingsAction, FetchFederationStatsAction, FetchFederationViewsAction, RemoveFavouriteAction, UpdateFederationContainersAction, UpdateFederationSettingsAction } from '@/v5/store/federations/federations.redux';
```
The-Daniel, line 19: Obviously not that important. But this is a very long line lol

frontend/test/containers/containers.redux.spec.ts
```diff
@@ -15,8 +15,7 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { INITIAL_STATE, reducer as containersReducer, ContainersActions } from '@/v5/store/containers/containers.redux';
-import { IContainersState } from '@/v5/store/containers/containers.types';
+import { INITIAL_STATE, reducer as containersReducer, ContainersActions, IContainersState } from '@/v5/store/containers/containers.redux';
```
The-Daniel, line 18: Is there a reason why we name each individual reducer as "reducer" when whenever we seem to import it end up renaming it anyway?

frontend/src/v5/services/api/containers.ts
sanmont3drepo: No reason to be honest!

frontend/src/v5/store/federations/federations.types.ts
sanmont3drepo: I changes SurveyPoint to ISurveyPoint to match the style with interfaces.  I think I prefer without the I in general, but if I remember correctly there was a rule to enforce the I in interfaces.

frontend/test/containers/containers.redux.spec.ts
sanmont3drepo: The reason is that I continued to use the style of v4, but I changed it.

frontend/src/v5/services/api/federations.ts
```diff
@@ -115,3 +103,12 @@ export const updateFederationContainers = async ({
 		containers,
 	})
 );
+
+/**
+ * Types
+*/
+type UpdateFederationContainersParams = TeamspaceProjectAndFederationId & { containers: string[] };
+type UpdateFederationSettingsParams = TeamspaceProjectAndFederationId & {settings: FederationBackendSettings};
```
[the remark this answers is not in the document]
sanmont3drepo: I moved in MinimumContainer to the containers.type file and FetchContainersResponse  to the bottom of the file following the style as before

The-Daniel (approved):

PR #3110, opened 2022-03-16
Issue 3050 - Fetch only views and other federations related data only when its needed

### LINKED ISSUE #3050: Fetch only views and other federations related data only when its needed
### Description
When navigating to the federations page there are fetches for the federation settings and views being done before opening a federation setting modal. These should be done only when necessary (when the form is opened

### Steps to replicate
- Navigate to federations (in a teamspace that has federations)
- Open network view in the developer console of the browser

### Current Behaviour
- It shows that there's been several fetches (views and settings) for each federation

### Expected Behaviour
- It should only fetch views and settings when the federation settings form modal is open

image

Amantini1997:
This fixes #3050

### Description
When navigating to the federations page there are fetches for the federation settings and views being done before opening a federation setting modal. These should be done only when necessary.

**Instead of doing it after opening the settings modal, it does it before so that the `desc`(description) appears under the federation modal**


sanmont3drepo:
Now it doesnt seem to show the saved settings values when opening the settings form

sanmont3drepo (approved):

PR #3111, opened 2022-03-16
Issue 3058 - Changed 'implicit-arrow-linebreak' rule in eslint

### LINKED ISSUE #3058: Change 'implicit-arrow-linebreak' lint rule setting
### Description
Right now lint has the rule 'implicit-arrow-linebreak' set to beside. This means that in a arrow function the body of the function should be on the side. This rule often conflicts with our other rule of maximum character length. Lets change it to force the body to be below instead of on the side. I also think that is more consistent to have the body of all the functions in the same format, regardless if it is an arrow function or not.

### Goals
- [ ] Avoid more character length ignores in the code
- [ ] Improve the readability

### Tasks
- [ ] Change the rule from 'implicit-arrow-linebreak': ['error', 'beside'] to 'implicit-arrow-linebreak': ['error', 'below'].
- [ ] Update code to fix linting problems.

sanmont3drepo:
This fixes #3058 
#### Description
- Changed 'implicit-arrow-linebreak' rule in eslint in frontend
- Changed 'implicit-arrow-linebreak' rule in eslint in backend


sanmont3drepo:
I dont think It makes anything better,

PR #3115, opened 2022-03-23
ISSUE #3113 - Managing queued status on empty containers

### LINKED ISSUE #3113: Show queue status in containers
### Description
Right now there is no feedback showing the status of containers, just the status of the latest revision. That means that if you have created a container and uploaded a file and is still enqueued.
It should show up where it says "empty container".

### Goals
- [ ] As a user I want to see the status of the container

### Tasks
- [ ] 
### Related Resources
- in v4:
image
- in v5:
image

sanmont3drepo:
This fixes #3113 

#### Description
- When the container is empty and there's a revision that is queued now it shows the status in the containers list
- When the container is empty and there's a revision that is queued now it doesn't show the upload button in the revisions details and instead shows a message explaining the situation.
- When the revision is queued or being processed it has the upload new revision button disabled for that container in the ellipsis menu.

### DIFF
--- a/frontend/src/v5/store/containers/containers.helpers.ts
+++ b/frontend/src/v5/store/containers/containers.helpers.ts
@@ -29,15 +29,13 @@ export const filterContainers = (federations: IContainer[], filterQuery: string)
 	) => [name, code, type].join('').toLowerCase().includes(filterQuery.trim().toLowerCase()))
 );
 
-export const isBusyInBackend = (status: UploadStatuses) => {
-	const busyInBackend = [
-		UploadStatuses.QUEUED,
-		UploadStatuses.PROCESSING,
-		UploadStatuses.QUEUED_FOR_UNITY,
-		UploadStatuses.UPLOADING,
-		UploadStatuses.GENERATING_BUNDLES,
+export const canUploadToBackend = (status: UploadStatuses) => {
+	const statusesForUpload = [
+		UploadStatuses.OK,
+		UploadStatuses.FAILED,
 	];
-	return busyInBackend.includes(status);
+
+	return !statusesForUpload.includes(status);
 };
 
 export const prepareSingleContainerData = (
--- a/frontend/src/v5/store/containers/containers.helpers.ts
+++ b/frontend/src/v5/store/containers/containers.helpers.ts
@@ -29,6 +29,17 @@ export const filterContainers = (federations: IContainer[], filterQuery: string)
 	) => [name, code, type].join('').toLowerCase().includes(filterQuery.trim().toLowerCase()))
 );
 
+export const isBusyInBackend = (status: UploadStatuses) => {
+	const busyInBackend = [
+		UploadStatuses.QUEUED,
+		UploadStatuses.PROCESSING,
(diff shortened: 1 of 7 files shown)


frontend/src/v5/ui/components/shared/revisionDetails/revisionDetails.component.tsx
```diff
@@ -58,7 +61,19 @@ export const RevisionDetails = ({ containerId, revisionsCount = 1 }: IRevisionDe
 		}
 	}, []);
 
-	if (!isLoading && revisions && revisions.length === 0) {
+	if (revisionsCount === 0) {
+		if (isBusyInBackend(status)) {
```
Amantini1997, line 65: In other components we used a lot the notation
```jsx
{isBusyInBackend(status) && ( 
   <Component>
       ...
   </Component>
}
```

Maybe we can use that here too to be more consistent?

frontend/src/v5/store/containers/containers.helpers.ts
```diff
@@ -29,6 +29,17 @@ export const filterContainers = (federations: IContainer[], filterQuery: string)
 	) => [name, code, type].join('').toLowerCase().includes(filterQuery.trim().toLowerCase()))
 );
 
+export const isBusyInBackend = (status: UploadStatuses) => {
+	const busyInBackend = [
```
carmenfan, line 33: It would be better to do the reverse, i.e. !(undefined || ok || failed)

This is because we may add more processing stages in the future (or rename them), there will be less things to change.

sanmont3drepo: Make sense!

sanmont3drepo:
> looks ok, just 2 questions (maybe for @mosment3DRepo and @george-hadfield )
> 
> image
> 
> 1. should this be `Upload New Revision`? unlike `in`, new isn't a preposition so to me it makes more sense to be in caps too
> 2. should we add a tooltip to show the reason it's disabled? otherwise it may be unclear to the user why this is happening

Regarding 2, just my opinion; I feel that it would be too "dropdowny", I mean to me it feels like a dropdown within a dropdown.

keifgwinn:
it looks like it's listing 'Load Container in 3D Viewer' twice in that screen shot as well?

carmenfan:
> it looks like it's listing 'Load Container in 3D Viewer' twice in that screen shot as well?

that's true, I completely blanked that 😆  @sanmont3drepo

george-hadfield:
> looks ok, just 2 questions (maybe for @mosment3DRepo and @george-hadfield )
> 
> image
> 
> 1. should this be `Upload New Revision`? unlike `in`, new isn't a preposition so to me it makes more sense to be in caps too
Let's keep it as it is. The 'N' doesn't need to be capitalised.

> 2. should we add a tooltip to show the reason it's disabled? otherwise it may be unclear to the user why this is happening
For now, let's keep it as it is without a tooltip.

carmenfan (approved):

Amantini1997 (approved):

PR #3142, opened 2022-04-06
ISSUE #3141 Updated unity build with more robust integer ops

### LINKED ISSUE #3141: Possible precision problem with idmaps inside super meshes
### Description
Counterpart of https://github.com/3drepo/3drepounity/issues/419 created to test fix

sebjf:
This fixes #3141 

#### Description
Contains an updated build of Unity with the supermesh map calculations modified to be use more robust pseudo-integer operations.


mosment3DRepo (dismissed):
Tested on customers problem models, large models and very small models. All tree selection and object selection works as expected

Issue #3118, opened 2022-03-24
Mailer V5


carmenfan:
### Description
At the moment, we're just bringing in the v4 mailer into v5. 

We want to rewrite it and only support functions that are currently needed.

#### Make a reusable template
In v4, we have full templates of every single email we send out as a [separate template](https://github.com/3drepo/3drepo.io/tree/master/backend/src/v4/mailer/templates). In reality, they share a lot of common components. Going forward, we want to ensure that we standardise the look and feel, and each context specific template should only contain the context specific components.

For e.g. we should have a general 3D Repo Email template:
![image](https://user-images.githubusercontent.com/11945337/159956113-b60d08f7-a642-4a7f-906b-ebfc828fc6c7.png)

and a Reset password email template:
![image](https://user-images.githubusercontent.com/11945337/159956179-534f87fb-f97f-4232-9c55-d5ce5c74c187.png)

The general template should be a full html, containing an area to insert the core email component, and the reset password email should only contain the text and reset password specific components.

#### Generalise function calls
In v4, we have [specific function calls](https://github.com/3drepo/3drepo.io/blob/master/backend/src/v4/mailer/mailer.js#L98) for each email to send 

In v5, we should generalise this so there is one function call to send email, and it takes in a constant to denote which email template to use. for e.g. to send an reset password email,  the caller should do something like this:

````js
mailer.send("recipient@mail.com", templates.RESET_PASSWORD, { username, token });
`````

In the mailer, we should only have a `send` function, and there should be a constants file that declare the list of templates available. So subsequent emails we create we just have to create new templates and utilise the same `send` function.

### Goals
- [ ] Create a v5 mailer service
- [ ] Support the following emails:
  - [ ] Email verification for sign up (for #3106) (mockup [here](https://www.figma.com/file/djgxvRKOaMMyutswn5yywU/Teamspace-(live)?node-id=5990%3A42695))
  - [ ] Forgot password email (for #3085) (mockup [here](https://www.figma.com/file/djgxvRKOaMMyutswn5yywU/Teamspace-(live)?node-id=6304%3A42192))

### Tasks
_to be filled in by the developer_

PR #3147, opened 2022-04-07
ISSUE #3112 - Forgot Password/Change Password

### LINKED ISSUE #3112: Forgot password
### Description
Forgot password for v5

### Goals
- [ ] As a user I want a way to reset the password I've forgotten
- [ ] As a user I want to enter my email and receive a link to reset the password
- [ ] As a user I want to enter the enter my new password and save it

### Tasks
- [x] Add a route for v5 forgot password
- [x] Create the form for sending the reset password link reusing the components from login 
- [ ] Create state for the link sent
- [x] Add a new API function for sending the email (see #3085 )
- [x] Add a route for v5 resetting the password (it should take token and username as query params)
- [x] Create the form for saving the new password
- [x] add a new API function for resetting the password (see #3085 )

### Notes
I dont think is necessary to create a saga or redux bit for this because the state doesnt need to be share with a different  part of the  app

### Related Resources
[MockUp](https://www.figma.com/file/djgxvRKOaMMyutswn5yywU/Teamspace-(live)?node-id=3825%3A31341)

The-Daniel:
This fixes #3112

#### Description
- adds screen for when a user forgets password (navigable from login)
- Entering a username/email will trigger an email containing a link to change your password
- Clicking this link should take you to the password change page (with username and credentials stored in the URL)
- Entering both your username and your email will trigger an email to your linked email address.
- After entering your username a message shows detailing your next steps
- If you enter a non-existant email the exact same happens (no indication that this user doesn't exist)
- Accessing the change password page with query parameters will show inputs for creating a new password
- Submit is disabled for when the passwords do not match and an error message is shown
- When a suitable password is given and confirmed the user is redirected to the login whereupon they can login with their new password

#### Test cases
- Accessing the change password page without the query parameters shows an error message (i.e. accessing just `/v5/password-change`)
- If password is too weak there is an error
- If password exceeds 65 characters there is an error
- If the query parameters are wrong (or the token has expired) there is an error

#### NOTES!!!
- The email is currently the same one as v4 including the v4 hyperlink. Simply click the link and manually add */v5/* before *password-change*
(The URL should look something like this `http://local.3drepo.io/v5/password-change?username=JasonVoorhees&token=35a206903ac383dc4cec994838180bfb`) This will change with #3118


frontend/src/v5/ui/controls/submitButton/submitButton.component.tsx
```diff
@@ -29,7 +29,7 @@ export const SubmitButton = ({
 		fullWidth
 		variant={variant}
 		color={color}
-		style={{ marginLeft: 0, marginRight: 0 }}
+		style={{ margin: '15px 0' }}
```
Amantini1997, line 32: Could you please override the component using `styled(SubmitButton)` in the respective  *.style.ts files so we are more consistent?

frontend/src/v5/validation/auth.ts
```diff
@@ -16,11 +16,41 @@
...
+			formatMessage({
+				id: 'passwordChange.error.passwordRequired',
+				defaultMessage: 'Password is required',
+			}),
+		)
+		.max(65,
+			formatMessage({
+				id: 'passwordChange.error.tooLong',
+				defaultMessage: 'Password is limited to 65 characters',
+			})),
```
Amantini1997, line 45: I believe the password validation schema should also check the password strength. This is what I wrote for the registration form: https://github.com/3drepo/3drepo.io/pull/3149/files#diff-96fc3af36608ace028221cd1949c5fb813cf03dc95a25b58a72c84e3b9743880

frontend/src/v5/validation/auth.ts
```diff
@@ -16,11 +16,41 @@
...
+	newPasswordConfirm: Yup.string()
+		.required(formatMessage({
+			id: 'passwordChange.error.passwordRequired',
+			defaultMessage: 'Password is required',
+		}))
+		.oneOf([Yup.ref('newPassword')],
+			formatMessage({
+				id: 'passwordChange.error.notMatching',
+				defaultMessage: 'The two entered passwords do not match',
```
Amantini1997, line 54: In the signup form, I wrote this error message for when the passwords do not match: "Passwords must match". Maybe we could ask George what to put there, and we both use what he says so that we keep it consinstent.

frontend/src/v5/ui/routes/login/passwordForgot/passwordForgot.component.tsx
```diff
@@ -0,0 +1,72 @@
...
+				{
+					isSubmitted ? (
+						<AuthParagraph>
+							<FormattedMessage
+								id="auth.forgotPassword.passwordSent.p1"
+								defaultMessage="A password change request has been sent. You will receive an email
+								shortly with a link to change your password."
+							/>
+							<br />
+							<br />
```
Amantini1997, line 52: In a similar situation, I used a component (which I called `<Gap />`) instead of `<br />` so that I can create exactly the amount of the space needed. I thought that was a better practice. Let's see what Santiago says about it, in case I'll change mine

Amantini1997 (changes requested):

frontend/src/v5/validation/auth.ts
The-Daniel: I've copied the password requirements from the V5 backend (https://github.com/3drepo/3drepo.io/blob/ISSUE_3112/backend/src/v5/utils/helper/yup.js)

As for the password strength. I let the backend handle the password and if it's too weak (or any other errors occur) I use the error messages from theere. I'm not sure if my method is best practise so I'll ask Santiago about it

keifgwinn: but to be clear, do it on both api and frontend as we did have a previous case of the api not honouring the password strength https://github.com/3drepo/3drepo.io/issues/2453

keifgwinn: also pet peeve I see happen a lot with the client side password validation is that it kicks in as soon as people start typing putting ugly red boxes up and possibly moving/redrawing elements and it's super annoying. The validation should kick in on mouse out/stopped typing for a few seconds to make sure the users ceased input.

george-hadfield:
### Change "Sign out" to "Log out" from the profile dropdown

george-hadfield:
### Change 2px border to 1px when active on an error

image

george-hadfield:
### Prevent input field from shifting up to accommodate error text
There is already sufficient space to allow for the error text without a shift

image

george-hadfield:
### Align issue text to the left, inline with 'Password' form label

image

george-hadfield:
### Change form label font-size to 14px from 12px
This should be a global change to all form labels and buttons

george-hadfield:
### Increase the body text from 12px to 14px
Make this a global change throughout v5

carmenfan:
@george-hadfield  ready for another look after azp has finished deploying (~10mins), this should now have the v5 email templates (and links to v5 pages)

george-hadfield:
###Back to login text should use link colour like other links

image

george-hadfield:
### Reduce spacing between error text and the next field below. 
Reduce by around 12px

image

The-Daniel:
@Amantini1997 I have addressed your feedback

george-hadfield (approved):

Amantini1997 (changes requested):

carmenfan:
removing the deployment as george's done his feedback and just waiting on code review.

rerun azp if needed

Amantini1997 (approved):

sanmont3drepo (approved):

sanmont3drepo (approved):

PR #3150, opened 2022-04-08
Issue 3118 - V5 Mailer 

### LINKED ISSUE #3118: Mailer V5

ChristosTsiotsias:
This fixes #3118 

#### Description
Mailer implementation for V5.
Currently supports Forgot Password, Verify User and Missing file (to be deleted) emails.
A base standard email template is used for all emails and specific html is added on top of it based on the email to be sent.

#### Test cases
When V5 signup or reset password endpoints are called the respective emails should be sent and they should look identical to the V4 ones.


backend/tests/v5/unit/services/mailer/mailer.test.js
```diff
@@ -0,0 +1,63 @@
...
+
+		test('should fail if config.mail.smtpConfig is not set and config.mail.generateCredentials is false', async () => {
+			config.mail.generateCredentials = false;
+			await expect(Mailer.sendEmail(FORGOT_PASSWORD, recipient, data, attachments))
+				.rejects.toEqual({ message: 'config.mail.smtpConfig is not set' });
+			config.mail.generateCredentials = true;
+		});
+
+		test('should send email if attachments are provided', async () => {
+			await Mailer.sendEmail(FORGOT_PASSWORD, recipient, data, attachments);
```
carmenfan, line 48: you should mock nodemail to check what we're sending to the nodemail api call.

backend/tests/v5/unit/services/mailer/mailer.test.js
```diff
@@ -0,0 +1,63 @@
...
+		});
+
+		test('should send email if attachments are not provided', async () => {
+			await Mailer.sendEmail(FORGOT_PASSWORD, recipient, data);
```
carmenfan, line 52: ditto

backend/tests/v5/unit/services/mailer/mailer.test.js
```diff
@@ -0,0 +1,63 @@
...
+		});
+
+		test('should send email if the subject of template is a function', async () => {
+			await Mailer.sendEmail(FILE_MISSING, recipient, data, attachments);
```
carmenfan, line 56: ditto

backend/tests/v5/unit/services/mailer/mailer.test.js
```diff
@@ -0,0 +1,63 @@
...
+		});
```
carmenfan, line 57: add test to show error handling, what if sending email failed? (which you should be able to do once you've mocked the nodemailer

backend/tests/v5/unit/services/mailer/mailer.test.js
```diff
@@ -0,0 +1,63 @@
...
+	});
+};
+
+describe('services/mailer/mailer', () => {
```
carmenfan, line 61: rename to mailer/index (includes the filename)

backend/tests/v5/unit/services/mailer/templates/baseTemplate.test.js
```diff
@@ -0,0 +1,960 @@
...
+	describe('get base template html', () => {
+		test('should get base template html', async () => {
+			const data = {
+				firstName: generateRandomString(),
+				emailContent: generateRandomString(),
+			};
+
+			const res = await BaseTemplate.html(data);
+
+			expect(res).toEqual(
```
carmenfan, line 33: This is a bit excessive

could we just check if the string is an html? https://www.npmjs.com/package/is-html

Also we should check data was provided has all the data we need, so we don't send out dodgy emails

backend/tests/v5/unit/services/mailer/templates/fileMissing.test.js
```diff
@@ -0,0 +1,67 @@
...
+				model: generateRandomString(),
+				collection: generateRandomString(),
+				refId: generateRandomString(),
+				link: generateRandomString(),
+				domain: generateRandomString(),
+			};
+
+			const res = await FileMissing.html(data);
+
+			expect(res).toEqual(
```
carmenfan, line 37: similar comments here

backend/tests/v5/unit/services/mailer/templates/forgotPassword.test.js
```diff
@@ -0,0 +1,167 @@
...
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Affero General Public License as
+ *  published by the Free Software Foundation, either version 3 of the
+ *  License, or (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
```
carmenfan, line 13: ditto

backend/tests/v5/unit/services/mailer/templates/verifyUser.test.js
```diff
@@ -0,0 +1,177 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const { src } = require('../../../../helper/path');
+const { generateRandomString } = require('../../../../helper/services');
+
```
carmenfan, line 20: ditto

backend/tests/v5/unit/services/mailer/mailer.test.js
carmenfan: also another one for unrecognised template name

backend/src/v5/models/fileRefs.js
```diff
@@ -62,7 +64,13 @@ FileRefs.fetchFileStream = async (teamspace, model, extension, fileName) => {
 		return { readStream: stream, size: entry.size };
 	} catch {
 		logger.logError(`Failed to fetch file from ${entry.type}. Trying GridFS....`);
-		sendFileMissingError({ teamspace, model, collection, refId: entry._id, link: entry.link }).catch((err) => {
+		await sendEmail(FILE_MISSING, config.contact?.email, {
```
carmenfan, line 67: Actually can you just remove this line (and any corresponding files)

This will be easier to clean up once the other issue is merged, otherwise I'd have to remember to go clean up the mailer.

carmenfan (changes requested):

backend/src/v5/services/mailer/templateNames.js
```diff
@@ -15,8 +15,10 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-const { v4Path } = require('../../interop');
-// eslint-disable-next-line import/no-dynamic-require, security/detect-non-literal-require, require-sort/require-sort
-const Mailer = require(`${v4Path}/mailer/mailer`);
+const TemplateNames = {};
```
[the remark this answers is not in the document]
carmenfan: just for the sake of pointing out obvious, i don't mean implementating  the html/subject functions here, just import them into this file instead (and can do it up top)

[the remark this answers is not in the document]
ChristosTsiotsias: I now understand what you mean. But in that case why can I not get rid of the name and pass the whole template in sendEmail method ? 

image

then I can use template.html and template.subject directly in sendEmail without having to even get it from constants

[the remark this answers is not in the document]
carmenfan: Because you don't want to just take the word of the callee for it - if we do that we're just calling whatever function that is passed down to the function - which could be naything

We want to check it's a valid template.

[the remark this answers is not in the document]
ChristosTsiotsias: I see...

backend/src/v5/services/mailer/mailer.js
```diff
@@ -0,0 +1,75 @@
...
+const { createTestAccount } = require('nodemailer');
+const { logger } = require('../../utils/logger');
+const nodemailer = require('nodemailer');
+
+let transporter;
+
+const Mailer = {};
+
+Mailer.sendEmail = async (templateName, to, data, attachments) => {
+	if (!config?.mail?.sender) {
```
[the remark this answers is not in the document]
ChristosTsiotsias: my only concern is await createTestAccount()... I can make an async function and call it on initialization but without awaiting it... is that ok ?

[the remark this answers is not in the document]
carmenfan: no, the point is if the initialisation failed, we want the service to shutdown..

If you need a promise on that, print the error message and  do a process.exit on failure

backend/tests/v5/unit/services/mailer/mailer.test.js
ChristosTsiotsias: for unrecognised template should I throw an exception manually or let it return 

 [TypeError: Cannot read property 'html' of undefined]

carmenfan: throw an exception manually

carmenfan:
Couple of things we need to do before merging:

1. update the email templates - it's currently still using v4 templates - I asked @george-hadfield if he can provide us html exports of the mock ups
2. need to change the urls to v5 on the email

carmenfan (approved):

PR #3192, opened 2022-05-03
Issue 566 - Default role assigned to existing and new users

### LINKED ISSUE #566: Fix "Creating a user without roles is deprecated in MongoDB >= 2.6" in backend integration tests
Mongo does not like us creating users without specifying a role

We should create a generic user role in `system.roles` that has no privileges, and automatically assign any newly created users with this role.

We should also have a migration script to ensure every users are assigned to this role

ChristosTsiotsias:
This fixes #566

#### Description
New users created in both v4 and v5 are assigned the default user role upon signup.
Existing users that don't have the role already are assigned it via a migration script.
Users created in tests are also assigned a default role.

#### Test cases
A newly created user should have the default role assigned to them.
Existing users should be assigned the default role when migration script is run.

### DIFF
--- a/backend/src/scripts/migrations/4.26/addAndAssignDefaultRole.js
+++ b/backend/src/scripts/migrations/4.26/addAndAssignDefaultRole.js
@@ -32,7 +32,7 @@ const createDefaultRole = async () => {
 
 const addAndAssignDefaultRole = async () => {
 	const users = await db.find('admin', 'system.users',
-		{ 'roles.role': { $nin: [DEFAULT] } }, { user: 1 });
+		{ 'roles.role': { $ne: DEFAULT } }, { user: 1 });
 
 	const role = { role: DEFAULT, db: 'admin' };
 	await Promise.all(users.map(async ({ user }) => {
--- a/backend/src/scripts/migrations/4.26/addAndAssignDefaultRole.js
+++ b/backend/src/scripts/migrations/4.26/addAndAssignDefaultRole.js
@@ -31,16 +31,14 @@ const createDefaultRole = async () => {
 };
 
 const addAndAssignDefaultRole = async () => {
-	const users = await db.find('admin', 'system.users', {}, { roles: 1, user: 1 });
+	const users = await db.find('admin', 'system.users',
+		{ 'roles.role': { $nin: [DEFAULT] } }, { user: 1 });
 
-	await Promise.all(users.map(async (user) => {
-		logger.logInfo(`\t\t-${user.user}`);
-
-		const role = { role: DEFAULT, db: 'admin' };
-		if (!user.roles.includes(role)) {
-			const grantRoleCmd = { grantRolesToUser: user.user, roles: [role] };
-			await db.runCommand('admin', grantRoleCmd);
-		}
+	const role = { role: DEFAULT, db: 'admin' };
+	await Promise.all(users.map(async ({ user }) => {
+		logger.logInfo(`\t\t-${user}`);
+		const grantRoleCmd = { grantRolesToUser: user, roles: [role] };
+		await db.runCommand('admin', grantRoleCmd);
 	}));
 };
 
--- a/backend/src/scripts/migrations/4.26/addAndAssignDefaultRole.js
+++ b/backend/src/scripts/migrations/4.26/addAndAssignDefaultRole.js
@@ -15,40 +15,38 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-const { DEFAULT, TEAM_MEMBER } = require('./roles.constants');
-const db = require('../handler/db');
+const { v5Path } = require('../../../interop');
+
+const db = require(`${v5Path}/handler/db`);
+const { logger } = require(`${v5Path}/utils/logger`);
+const { DEFAULT } = require(`${v5Path}/models/roles.constants`);
 
 const createDefaultRole = async () => {
 	const roleFound = await db.findOne('admin', 'system.roles', { _id: `admin.${DEFAULT}` });
 
-	// istanbul ignore next
 	if (!roleFound) {
 		const createRoleCmd = { createRole: DEFAULT, privileges: [], roles: [] };
 		await db.runCommand('admin', createRoleCmd);
 	}
 };
 
-createDefaultRole();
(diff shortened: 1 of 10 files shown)


backend/src/scripts/migrations/4.26/addAndAssignDefaultRole.js
```diff
@@ -0,0 +1,52 @@
...
+	const roleFound = await db.findOne('admin', 'system.roles', { _id: `admin.${DEFAULT}` });
+
+	if (!roleFound) {
+		const createRoleCmd = { createRole: DEFAULT, privileges: [], roles: [] };
+		await db.runCommand('admin', createRoleCmd);
+	}
+};
+
+const addAndAssignDefaultRole = async () => {
+	const users = await db.find('admin', 'system.users', {}, { roles: 1, user: 1 });
```
carmenfan, line 34: You should add a query to filter out any users who already have the role, then you won't have to do the check at L40

backend/src/scripts/migrations/4.26/addAndAssignDefaultRole.js
```diff
@@ -0,0 +1,52 @@
...
+
+	await Promise.all(users.map(async (user) => {
```
carmenfan, line 36: I don't think you need roles after previous change.

```suggestion
	await Promise.all(users.map(async ({user}) => {
```

backend/src/v4/models/user.js
```diff
@@ -496,9 +496,10 @@ User.createUser = async function (username, password, customData, tokenExpiryTim
 	};
 
 	cleanedCustomData.billing = await UserBilling.changeBillingAddress(cleanedCustomData.billing || {}, billingInfo);
+	const defaultRole = { db: "admin", role: C.DEFAULT_USER_ROLE };
```
carmenfan, line 499: might be better to store the whole object as the default role instead of we always seems to use it in this context

call it `defaultRoleObj` or something if you need to make it obvious it's an object

backend/src/v5/models/users.js
```diff
@@ -128,7 +128,7 @@ User.getFavourites = async (user, teamspace) => {
 
 User.getAccessibleTeamspaces = async (username) => {
 	const userDoc = await User.getUserByUsername(username, { roles: 1 });
-	return userDoc.roles.map((role) => role.db);
+	return userDoc.roles.flatMap((role) => (role.db !== 'admin' ? [role.db] : []));
```
carmenfan, line 131: This is probably my doing, but since we're changing this line 😆 
```suggestion
	return userDoc.roles.flatMap(({db}) => (db !== 'admin' ? [db] : []));
```

backend/tests/v5/helper/services.js
```diff
@@ -56,19 +56,31 @@ queue.purgeQueues = async () => {
 	}
 };
 
+const defaultTestRoleName = 'userTest';
+
 // userCredentials should be the same format as the return value of generateUserCredentials
 db.createUser = async (userCredentials, tsList = [], customData = {}) => {
 	const { user, password, apiKey, basicData = {} } = userCredentials;
 	const roles = tsList.map((ts) => ({ db: ts, role: 'team_member' }));
+	roles.push({ db: 'admin', role: defaultTestRoleName });
```
carmenfan, line 65: What's this for?

backend/tests/v5/helper/services.js
```diff
@@ -56,19 +56,31 @@ queue.purgeQueues = async () => {
...
 	const adminDB = await DbHandler.getAuthDB();
 	return adminDB.addUser(user, password, { customData: { ...basicData, ...customData, apiKey }, roles });
 };
 
 db.createTeamspaceRole = (ts) => createTeamSpaceRole(ts);
 
+db.createDefaultRole = async () => {
```
carmenfan, line 72: also unsure why this is added

backend/tests/v5/helper/services.js
```diff
@@ -56,19 +56,31 @@ queue.purgeQueues = async () => {
...
+		const createRoleCmd = { createRole: defaultTestRoleName, privileges: [], roles: [] };
+		await DbHandler.runCommand('admin', createRoleCmd);
+	}
+};
+
 // breaking = create a broken schema for teamspace to trigger errors for testing
-db.createTeamspace = (teamspace, admins = [], breaking = false) => {
+db.createTeamspace = async (teamspace, admins = [], breaking = false) => {
 	const permissions = admins.map((adminUser) => ({ user: adminUser, permissions: TEAMSPACE_ADMIN }));
+	await db.createDefaultRole();
```
carmenfan, line 83: and this

backend/tests/v5/unit/models/users.test.js
```diff
@@ -49,6 +50,7 @@ const testGetAccessibleTeamspaces = () => {
 				roles: [
 					{ db: 'ts1', role: 'a' },
 					{ db: 'ts2', role: 'b' },
+					{ db: 'admin', role: 'user' },
```
carmenfan, line 53: use generateRandomString if you don't care about the role name, if we do then we should probably be using the constant from `src`

backend/tests/v5/unit/models/users.test.js
```diff
@@ -542,6 +544,7 @@ const testAddUser = () => {
 				permissions: [],
 			};
 
+			db.createUser.mockImplementationOnce(() => {});
```
carmenfan, line 547: I think you can merge it with the next line? `spyOn` returns a [mock function](https://jestjs.io/docs/mock-function-api)
```suggestion
			const fn = jest.spyOn(db, 'createUser').mockImplementationOnce(() => {});
```

carmenfan (changes requested):

backend/src/v4/models/user.js
ChristosTsiotsias: you mean like this 

image

backend/tests/v5/helper/services.js
ChristosTsiotsias: I added these to get rid of these warnings

image

carmenfan: you shouldn't need to do this with the change where we create the defaultRole before creating the user

carmenfan (approved):

PR #3199, opened 2022-05-04
ISSUE #3198 - fix case in elastic loginRecord

### LINKED ISSUE #3198: Elastic loginRecord is case sensitive
### Description

The mapping and the documents must match on case or elastic recognises them as different things.

- This is happening in: both 
- This requires a hotfix to production: no  (but it would be good to try and get this into the next major release)

### Steps to replicate
- log into a system with the elastic records configured.
- observe the document mapping and contents have different cases.

### Current Behaviour
- Cases not matched

### Expected Behaviour
- Cases matched

image
image

keifgwinn:
This fixes #3198

#### Description

- Fixed the case

#### Test cases
- elastic records match mapping


carmenfan (approved):

keifgwinn:
ok, having tested it would be better to move the case the other way, with mapping going cased and message matching.

PR #3202, opened 2022-05-05
Issue 3195 - User agent parsing change

### LINKED ISSUE #3195: Unhandled except in isUserAgentFromPlugin
### Description

There's an UnhandledPromiseRejectionWarning in ./backend/src/v5/utils/helper/userAgent.js:67:60 if the useragent isn't set

```
[www-588c4b6c87-bprs6] (node:43) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'split' of undefined
[www-588c4b6c87-bprs6]     at UserAgent.isUserAgentFromPlugin (/home/node/3drepo.io/backend/src/v5/utils/helper/userAgent.js:67:60)


```
- This is happening in: both

### Steps to replicate
- Log in without useragent set

### Current Behaviour
- UnhandledPromiseRejectionWarning on splitting the empty

### Expected Behaviour
- Gracefully handle it

ChristosTsiotsias:
This fixes #3195

Change in the way we parse user agent string so that an exception is not thrown when it is empty string

#### Test cases
Making a login request with an empty user agent should not throw an unhandled exception

### DIFF
--- a/backend/src/v5/models/loginRecord.js
+++ b/backend/src/v5/models/loginRecord.js
@@ -16,8 +16,7 @@
  */
 
 // detects edge as browser but not device
-const { getUserAgentInfoFromBrowser, getUserAgentInfoFromPlugin,
-	isUserAgentFromPlugin } = require('../utils/helper/userAgent');
+const { getUserAgentInfo} = require('../utils/helper/userAgent');
 const db = require('../handler/db');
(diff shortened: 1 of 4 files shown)


carmenfan (dismissed):
Looks fine but it's failing linter

carmenfan (approved):

Issue #3208, opened 2022-05-09
Update viewer background


sebjf:
### Description

Created for a deployment with an updated viewer background for V5 development.

### Related Resources

Unity Issue: `https://github.com/3drepo/3drepounity/issues/429`

PR #3209, opened 2022-05-09
Issue 3208 - Viewer background update

sebjf:
This addresses #3208. This PR is created only to have a build made. It should not be merged.


sebjf:
\azp run

PR #3249, opened 2022-05-13
ISSUE #3248 don't generate credentials

### LINKED ISSUE #3248: Tests stalls on validating mailer config occasionally
### Description
Very occasionally (1 in 100s of times), some mailer associated e2e tests (e.g. forget password) timesout. This is due to a the promise of automatically generating a mailer test account doesn't return in time.

### Steps to replicate
- write a script to run tests continuously until the last process returns faillure, e.g. (powershell script)
````
do{
	yarn test:v5-e2e routes/users.test --collectCoverage false -t "should send forgot password email with valid username"
}while($?)
````
- let it run

### Current Behaviour
- Eventually it will fail on a timeout

### Expected Behaviour
shoudl never fail

### PR DESCRIPTION
**Pull request by carmenfan**
This fixes #3248

#### Description
Hard code a dummy mailer account instead of generating credentials as that endpoint can sometime stall.

#### Test cases
should (hopefully) pass the test continuously.

### CODE REVIEW

File: backend/tests/v5/unit/services/mailer/mailer.test.js [hunk 1]
```diff
@@ -61,6 +61,7 @@ const testSendEmail = () => {
 		test('should fail if config.mail.smtpConfig is not set and config.mail.generateCredentials is false', async () => {
 			Mailer.reset();
 			config.mail.generateCredentials = false;
+			delete config.mail.smtpConfig;
 			await expect(Mailer.sendEmail(emailTemplates.FORGOT_PASSWORD.name, recipient, data, attachments))
 				.rejects.toThrow('config.mail.smtpConfig is not set');
 			config.mail.generateCredentials = true;
```
ChristosTsiotsias, line 67: I would keep the config in the same state after each test. Could we not set generateCredentials to true by default ?

ChristosTsiotsias (changes requested):

backend/tests/v5/unit/services/mailer/mailer.test.js
carmenfan: This line is actually redundant because we reset the config before each test: https://github.com/3drepo/3drepo.io/blob/9b91d06dd362fc1f72de785a7f65b0356750991c/backend/tests/v5/unit/services/mailer/mailer.test.js#L47

I'll remove it

ChristosTsiotsias: ah true

ChristosTsiotsias (changes requested):

backend/tests/v5/unit/services/mailer/mailer.test.js
```diff
@@ -44,26 +44,43 @@ const testSendEmail = () => {
 		const mailerConfig = { ...config.mail };
 
 		beforeEach(() => {
-			config.mail = mailerConfig;
+			config.mail = { ...mailerConfig };
 		});
 
 		test('should fail if config.mail.sender is not set', async () => {
 			Mailer.reset();
```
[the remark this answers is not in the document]
carmenfan: short answer is yes.

this call resets the promise. I call it before to ensure that a previous test has not already done any operation that would've initialised the mailer, and it is called after to ensure it resets the state of the test to make sure it doesn't cause another test to fail because of this test.

Initially I wanted to put it in beforeEach, but I needed a test that would pick up the init without triggering it for code coverage, hence the state of this.

backend/tests/v5/unit/services/mailer/mailer.test.js
```diff
@@ -44,26 +44,43 @@ const testSendEmail = () => {
...
 		});
 
 		test('should fail if config.mail.smtpConfig is not set and config.mail.generateCredentials is false', async () => {
 			Mailer.reset();
 			config.mail.generateCredentials = false;
+			delete config.mail.smtpConfig;
 			await expect(Mailer.sendEmail(emailTemplates.FORGOT_PASSWORD.name, recipient, data, attachments))
 				.rejects.toThrow('config.mail.smtpConfig is not set');
+			Mailer.reset();
```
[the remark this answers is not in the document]
carmenfan: it's there for the reason above

PR #3260, opened 2022-05-17
ISSUE #2805 - Having non existing users in the db does not block you from updating permissions

### LINKED ISSUE #2805: Teamspace admin can't assign users permissions to a select few models
### Description
API request to update a user permission fails if the existing permission array consist of a user that doesn't exist/no longer in the teamspace

### Steps to replicate
- go to the model settings of the model in the database
- Add a new permission entry for a user that doesn't exist
- try to update the permissions of that model (e.g. add an existing user into the model as a collaborator)

### Current Behaviour
- API request fails with USER_NOT_FOUND

### Expected Behaviour
- API request should pass.

<details>
  <summary>Bogdan's Original bug report</summary>

### Description

I open the User management panel and head over to model & federations permissions of a particular project. Most of the models are working fine. However, there are 2 models that won't allow you to change user permissions. It throws an error message "User Not Found"

- This is happening in: prod

### Steps to replicate
- Open the [User Management window](https://www.3drepo.io/dashboard/user-management/John_Sisk_TeamSpace/projects?project=Project%20Kells&view=1)
- Select models 21.9.27 or Main 4D Model
- Change the permission level for any of the users
- Get an error message

### Current Behaviour
- Only works with some models

### Expected Behaviour
- Expect the model permissions to work across all of the models

[User Management Issue](https://www.dropbox.com/preview/3D%20Repo%20Team%20Folder/Support%20Team/GitHub%20Issues/Images/Sisk%20User%20Permissions.png)

**Warning Message**
image

</details>

ChristosTsiotsias:
This fixes #2805

#### Description
When updating model permissions, only the permissions changed are validated. Thus, when there are permissions that include users that no longer exist / have access to the TS no errors are being thrown.

#### Test cases
In a model that has permissions with non existing user, update a permission.


ChristosTsiotsias:
I have copied some functionality from changePermissions to updatePermissions so there are some duplicate lines. The other alternative I thought and have saved is to change the changePermissions method to only accept permissions to be updated but all places that call it need to be adjusted.

backend/src/v4/models/modelSetting.js
```diff
@@ -413,58 +420,48 @@ ModelSetting.updateHeliSpeed = async function(account, model, newSpeed) {
...
+				} else {
+					action.$pull = { permissions: setting.permissions[index] };
+				}
+			} else if (permission.permission) {
+				action.$push = { permissions: permission };
 			}
-		} else if ("" !== permissionUpdate.permission) {
-			setting.permissions.push(permissionUpdate);
-		}
+			db.updateOne(account, MODELS_COLL, { _id: model }, action);
```
carmenfan, line 459: if we have n permissions, we're doing n updates to the single document.  don't think this is a good idea


Also the fact that you could be overwriting the whole array with $set at L452 means you might end up in a weird state where some $push/pull gets overwritten.

You are probably better off doing a single $set at the end of the loop at this rate.

backend/src/v4/routes/model.js
```diff
@@ -722,7 +722,7 @@ router.put("/:model", middlewares.hasEditAccessToFedModel, middlewares.formatV5N
  *    }
  * ]
  */
-router.post("/models/permissions", middlewares.hasEditPermissionsAccessToMulitpleModels, updateMultiplePermissions);
```
carmenfan, line 725: does this make sense? If i'm reading you correctly, you're making this POST request do the same thing as the PATCH request.

Which shouldn't be the case? POST should overwrite the array where as PATCH should make changes to it?

carmenfan (changes requested):

backend/src/v4/models/modelSetting.js
```diff
@@ -438,33 +445,42 @@ ModelSetting.updateMultiplePermissions = async function(account, modelIds, updat
...
+	permissions = _.uniq(permissions, "user");
 
-		if (-1 !== userIndex) {
-			if ("" !== permissionUpdate.permission) {
-				setting.permissions[userIndex].permission = permissionUpdate.permission;
-			} else {
-				setting.permissions.splice(userIndex, 1);
+	const promises = [];
+	const dbUser = await findByUserName(account);
+	permissions.forEach(permission => {
```
carmenfan, line 463: instead of using a `forEach` where every iteration we doing a `push` to an array of promise, we should use a map.

backend/src/v4/models/modelSetting.js
```diff
@@ -438,33 +445,42 @@ ModelSetting.updateMultiplePermissions = async function(account, modelIds, updat
...
+		checkPermissionIsValid(permission);
+		checkUserHasPermissionTemplate(dbUser, permission);
+
+		promises.push(teamspaceMemberCheck(permission.user, dbUser.user).then(() => {
```
carmenfan, line 467: use async/await syntax instead of .then 

```suggestion
		promises.push(teamspaceMemberCheck(permission.user, dbUser.user).then(() => {
```

carmenfan (changes requested):

carmenfan (approved):

PR #3266, opened 2022-05-17
Issue 3232 - Make navbar less opaque

### LINKED ISSUE #3232: Set main nav bar to 0.95 transparency in v5 viewer
We would like the model to be able to move underneath the main blue nav bar and have it slightly visible. 0.95 transparency would work perfectly, but the model needs to be able to move underneath

Amantini1997:
This fixes #3232 

#### Description
Main navbar opacity has been reduced to 95%


carmenfan:
@mosment3DRepo @george-hadfield  I think the correct way to get what you want is to reduce the unity's camera [field of view](https://docs.unity3d.com/ScriptReference/Camera-fieldOfView.html), so we make the user see less vertically than before.

Hiding the top strip of the canvas behind the nav bar doesn't sound like it's the best idea for the following reasons:
   - To only let transparency happen on viewer, we're adding extra random if statements around to ensure it's opaque anywhere outside of the viewer.
  - The screenshot aspect ratio won't match the user's POV, so when the user creates screenshot it will never match exactly what they're after: 
image
  - Similarly when they click on a view it'll never exactly match the image, it'll always feel like you're seeing something "lower" than the screenshot as the user's centre will always be ~72px lower than the actual camera centre.
  - centre to object will never "centre" to object, you'll always be centred ~72px below where you want to centre

PR #3275, opened 2022-05-19
ISSUE #3183 - if container has no revisions you cannot open it in viewer

### LINKED ISSUE #3183: Link working on empty containers
### Description
Link working on empty containers and containers that are currently being processed even though there's nothing to load.

- 
### Steps to replicate
- Create an empty container
- Click to enter the container

### Current Behaviour
- The user can enter the container and get an error

### Expected Behaviour
- The user shouldn't be able enter the container and get an error

The-Daniel:
This fixes #3183

#### Description
- Prevents the user from accessing the viewer for containers that do not have any revisions


george-hadfield:
Remove underline hover state from Container title when there are no revisions. Make the container title 'base.light' if there are no revisions. Change text 'Latest revision' to 'No revisions'

Build
image

Design
image

george-hadfield (changes requested):
Small changes

The-Daniel:
@george-hadfield how do you want it to work for when a container is in the process of uploading its first container (i.e. queued, processing, maybe even failed?). I assume the link should still not be clickable, but do you want the colours to change?

george-hadfield:
If you can't load a model, please use the disabled grey as above. 
If there is a status such as those you mentioned, then use the colour as built now

image

george-hadfield (approved):

sanmont3drepo (approved):

PR #3280, opened 2022-05-20
Issue 3256 - files manager v5

### LINKED ISSUE #3256: files manager v5
### Description
Migration of the files manager service into v5.

We should refactor this a bit so we have a data storage service that manages this, instead of relying on model/fileRef.

### Goal
- Migrate file services into v5
- add tests to ensure it is robust and stable

### Tasks
- [x] Add a files manager into services
  - [x] Decouple relevant code from file ref

carmenfan:
This fixes #3256

#### Description
Migrate files manager to v5.

Slight rework from v4:
This is effectively `ExternalServices` of v4, but it also takes the responsibility of reading the `fileRefs` on the callee's behalf. Essentially any processors that needs to retrieve a file should now talk to this instead of `fileRefs`.

FileRefs should be a thin mapping to the schema in database, it should not be retrieving files on behalf of users.

#### Test cases
download model files should work as before
deleting models should work as before


backend/src/v5/services/filesManager.js
```diff
@@ -0,0 +1,81 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const { getAllRemovableEntriesByType, getRefEntry } = require('../models/fileRefs');
+const FSHandler = require('../handler/fs');
```
ChristosTsiotsias, line 19: We just retrieve the methods we want to use

backend/tests/v5/unit/models/fileRefs.test.js
```diff
@@ -70,116 +71,65 @@ const testGetTotalSize = () => {
...
-				expect(fnAggregate.mock.calls[i][2]).toEqual(query);
-			});
-		});
+const testGetAllRemovableEntriesByType = () => {
+	describe('Get all removable entries by type', () => {
+		const refCol = generateRandomString();
 
-		test('should not fail if the the ref collection has no links', async () => {
-			const teamspace = 'someTS';
+		test('should remove files from collection', async () => {
```
ChristosTsiotsias, line 78: Is this description valid ?

ChristosTsiotsias (changes requested):

backend/src/v5/services/filesManager.js
carmenfan: Reason why It's not done it's because FSHandler and GridFSHandler have the same named functions (see ~L30)

so if I do that then i'd have to rename all the functions to avoid clash 

And as this file grow it'll very most likely be requiring all the functions from those files, so it's less of a concern

backend/tests/v5/unit/services/filesManager.test.js
```diff
@@ -0,0 +1,195 @@
...
+				.mockImplementation((ts, col) => Promise.resolve(col === refCol1 ? refCol1Data : [{ _id: 'fs', links: [] }]));
+
+			const teamspace = generateRandomString();
+
+			await FilesManager.removeAllFilesFromModel(teamspace, model);
+
+			expect(db.listCollections).toHaveBeenCalledTimes(1);
+			expect(db.listCollections).toHaveBeenCalledWith(teamspace);
+
+			expect(FileRefs.getAllRemovableEntriesByType).toHaveBeenCalledTimes(2);
```
ChristosTsiotsias, line 92: Do we wanna test the method calls that are happening outside of the method ? I thought we would only wanna test the calls of removeAllFilesInCol, thats what I do in all unit tests

backend/tests/v5/unit/services/filesManager.test.js
```diff
@@ -0,0 +1,195 @@
...
+			expect(FileRefs.getAllRemovableEntriesByType).toHaveBeenNthCalledWith(2, teamspace, refCol2);
+
+			expect(FSHandler.removeFiles).toHaveBeenCalledTimes(1);
+			expect(FSHandler.removeFiles).toHaveBeenCalledWith(refCol1Data[0].links);
+
+			expect(GridFSHandler.removeFiles).toHaveBeenCalledTimes(1);
+			expect(GridFSHandler.removeFiles).toHaveBeenCalledWith(teamspace, refCol1, refCol1Data[1].links);
+		});
+
+		test('Should error if the type of storage is unknown', async () => {
```
ChristosTsiotsias, line 103: Should throw* error

backend/tests/v5/unit/services/filesManager.test.js
```diff
@@ -0,0 +1,195 @@
...
+		test('should throw error if the revision has no entry', async () => {
+			FileRefs.getRefEntry.mockRejectedValueOnce(templates.fileNotFound);
+			await expect(FilesManager.getFileAsStream(
+				generateRandomString(),
+				generateRandomString(),
+				generateRandomString(),
+			)).rejects.toEqual(templates.fileNotFound);
+		});
+
+		test('return throw an error if the storage type is unrecognised', async () => {
```
ChristosTsiotsias, line 140: should throw*

backend/tests/v5/unit/services/filesManager.test.js
```diff
@@ -0,0 +1,195 @@
...
+			const fileEntry = { type: generateRandomString() };
+			FileRefs.getRefEntry.mockResolvedValueOnce(fileEntry);
+			await expect(FilesManager.getFileAsStream(
+				generateRandomString(),
+				generateRandomString(),
+				generateRandomString(),
+			)).rejects.toEqual(templates.fileNotFound);
+		});
+
+		test('should return a stream of the reference is found', async () => {
```
ChristosTsiotsias, line 150: if*

ChristosTsiotsias (changes requested):

backend/src/v5/services/filesManager.js
```diff
@@ -0,0 +1,81 @@
...
+			return Promise.resolve();
+		},
+	);
+
+	return Promise.all(deletePromises);
+};
+
+FilesManager.removeAllFilesFromModel = async (teamspace, model) => {
+	const collList = await listCollections(teamspace);
+	const refCols = collList.filter(({ name }) => {
```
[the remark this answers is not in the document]
carmenfan: yes - i'll merged them into one but done it slightly differently (e.g. moved the regex out as it's always the same)

backend/tests/v5/unit/services/filesManager.test.js
carmenfan: We're not testing that function, that function is mocked. we're just testing that that function is called

carmenfan: this is akin to checking db.find() etc are called when we unit test files in the `models` folder

PR #3284, opened 2022-05-23
ISSUE #2897 - Removed silent for jest tests

### LINKED ISSUE #2897: Revamping - Errors in sagas test
### Description
When you run the frontend tests without --silent (sneaky!) the sagas tests throws an error. I was able to track where does this comes from and is related to `src\v5\services\api\default.ts` and its dependency on v4. Probably the tests might run faster if this problem is solved.

### Steps to replicate
- remove --silent option in the 'scripts'->'test' in  package.json
- run yarn test

### Current Behaviour
- Throws erros in console regardless of the tests passing

### Expected Behaviour
- Just green tests without errors 

image

sanmont3drepo:
This fixes #2897 

#### Description
- Removed silent param in jest tests
- Silencing console.log only in federations.saga.tests because faker is console logging an unnnecessary warning.

### DIFF
--- a/frontend/test/federations/federations.sagas.spec.ts
+++ b/frontend/test/federations/federations.sagas.spec.ts
@@ -19,7 +19,7 @@ import * as FederationsSaga from '@/v5/store/federations/federations.sagas';
 import { expectSaga } from 'redux-saga-test-plan';
 import { FederationsActions } from '@/v5/store/federations/federations.redux';
 import { mockServer } from '../../internals/testing/mockServer';
-import { omit, pick, times } from 'lodash';
+import { before, omit, pick, times } from 'lodash';
 import {
 	federationMockFactory,
 	prepareMockRawSettingsReply,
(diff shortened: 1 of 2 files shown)


frontend/test/federations/federations.sagas.spec.ts
```diff
@@ -19,7 +19,7 @@ import * as FederationsSaga from '@/v5/store/federations/federations.sagas';
 import { expectSaga } from 'redux-saga-test-plan';
 import { FederationsActions } from '@/v5/store/federations/federations.redux';
 import { mockServer } from '../../internals/testing/mockServer';
-import { omit, pick, times } from 'lodash';
+import { before, omit, pick, times } from 'lodash';
```
Amantini1997, line 22: I don't think `before` is used anywhere

Amantini1997 (changes requested):

Amantini1997 (approved):

PR #3304, opened 2022-05-26
Issue 3274 - Revisions cannot be uploaded while model is on queued or processing stage

### LINKED ISSUE #3274: Trying to upload a new revision on a model that is currently processing one should fail
### Description
Currently we allow users to upload multiple revisions simultaneously via the API.

This cases race conditions in the model processing status and we end up losing track some of the processing. @3drepo/product-team  has decided that we no longer want to allow the user to do this and thus we should enforce it on the backend.

### Goals
- The user should not be able to upload a new revision if there is one processing within the model

### Tasks
- Add a middleware test in v5 to check if the model status is processing/queuing a model
- Ensure v4 is also enforcing this

ChristosTsiotsias:
This fixes #3274

#### Description
While a container status is 'queued' or 'processing' no new revisions should be uploaded
This should apply to both v4 and v5

#### Test cases
Upload a revision to a container
Try to upload a new revision while the status of the container is 'queued' or 'processing'
An invalid arguments error should be returned


backend/src/v4/routes/model.js
```diff
@@ -1716,7 +1717,8 @@ router.patch("/:model/upload/ms-chunking/:corID", middlewares.hasUploadAccessToM
  * ------WebKitFormBoundarySos0xligf1T8Sy8I-- *
  *
  */
-router.post("/:model/upload",  middlewares.hasUploadAccessToModel, middlewares.formatV5NewModelRevisionsData, validateNewModelRevisionData, uploadModel);
+router.post("/:model/upload",  middlewares.hasUploadAccessToModel, middlewares.formatV5NewModelRevisionsData,
```
carmenfan, line 1720: This is a bit overkill i think, just do it as part of `formatV5NewModelRevisionsData` ?

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/containers.js
```diff
@@ -40,6 +40,21 @@ Containers.canDeleteContainer = async (req, res, next) => {
 	}
 };
 
+Containers.checkModelStatus = async (req, res, next) => {
+	try {
+		const { teamspace, container } = req.params;
+		const { status } = await getModelById(teamspace, container, { _id: 0, status: 1 });
+
+		if (status === 'queued' || status === 'processing') {
```
carmenfan, line 48: do it the other way round, we have more status messages than queued and processing.

So it's safer (and more expandable to say) if status is not `ok`, `undefined` or `failed`

carmenfan: Can we add model status as a constant in `modelSettings.constants` and use it from there?

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/containers.js
```diff
@@ -40,6 +40,21 @@ Containers.canDeleteContainer = async (req, res, next) => {
...
+			throw templates.revisionProcessed;
```
carmenfan, line 49: this status code seems redundant since it'll always get caught to return with invalidArguments?

carmenfan (changes requested):
We should also impose this on editing federation as they go through the same kind of pipeline

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/containers.js
ChristosTsiotsias: true but we want the message right ? shall I just do throw  'A revision is already being processed.' ?

carmenfan: Sure, but I think if you do it through the other middleware and we just fit it into the Yup schema it becomes a non issue

PR #3314, opened 2022-05-27
Issue #3288 - Added the subscription to logged out 

### LINKED ISSUE #3288: Realtime event when the user gets kicked out by logging in in a different browser
### Description
As policy we don't allow users to log in into the a different browser. For this we signoff the user in the other browser the user has already signed in as well as sending a realtime event to reflect that .  This is already implemented in v4, but we need to implement it in v5.

### Goals
- [ ] As a user I want to know if was automatically signed off the platform in realtime

### Tasks
- [ ] Add the logoff event in v5
- [ ] Suscribe to the logoff event in v5 and logoff the user
- [ ] Show a modal explaining why the user has been kicked out of the platform

sanmont3drepo:
This fixes #3288 

#### Description
- Added subscription to logged out
- A modal is shown stating that you were kicked out because you logged in in a different browser

##### Login in a different browser - steps:
- Login in a browser
- Without closing that browser login in a different browser
- In the first browser you should have been logged out and an alert modal should be displayed


frontend/src/v5/ui/components/shared/modals/templates/warningModal/warningModal.component.tsx
```diff
@@ -0,0 +1,50 @@
...
+				defaultMessage="Warning!"
+			/>
+		</DialogTitle>
+		<DialogContent>
+			<DialogContentText>
+				{ message }
+			</DialogContentText>
+		</DialogContent>
+		<Actions>
+			<Button autoFocus variant="contained" color="primary" onClick={() => onClickClose()}>
```
The-Daniel, line 42: Not really important, but the onClick can be simplified to `onClick={onClickClose}`, also the FormattedMessage id is for the alert modal

frontend/src/v5/store/auth/auth.sagas.ts
```diff
@@ -98,8 +98,19 @@ function* logout() {
 	yield put(AuthActions.setPendingStatus(false));
 }
 
+function* kickedOut() {
+	yield put({ type: 'RESET_APP' });
+	yield put(DialogsActions.open('warning', {
+		message: formatMessage({
+			id: 'auth.logout.kickedOut',
+			defaultMessage: 'You have been logged out because your account has signed in elsewhere',
```
The-Daniel, line 106: Looking at the README in `src\v5\services\chat\README.md` it seems to suggest that the logged out event returns an object containing a _reason_. If so, would it be useful to pass the _reason_ into this message in the case that there are multiple reasons a user is kicked out?

frontend/src/v5/ui/components/shared/modals/templates/warningModal/warningModal.component.tsx
sanmont3drepo: Yup about the onclick. Not sure about the formattedMessage. Because we kind of want to keep the same wording through out the modals...? Let me see what George says about the wording/modal style

george-hadfield:
I didn't get logged out, I was able to be logged into both Safari and Chrome at the same time.

george-hadfield:
image

Worked but not instantly

frontend/src/v5/store/dialogs/dialogs.redux.ts
```diff
@@ -53,7 +53,7 @@ export const dialogsReducer = createReducer(INITIAL_STATE, {
 /**
  * Types
  */
-interface IDialogActions {
+export interface IDialogsActionsCreators {
```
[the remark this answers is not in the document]
sanmont3drepo: There was a version in which I was using this. But yeah, Ill get rid of the export and fix the typing for consistency.

george-hadfield:
Can we also change wording to:

**You've been logged out**
Your account is being used in a different browser

sanmont3drepo:
> image
> 
> Worked but not instantly

That's actually a bug and not the proper dialog.

sanmont3drepo:
> Can we also change wording to:
> 
> **You've been logged out** Your account is being used in a different browser

image


So, not this one?

The-Daniel (approved):

The-Daniel (approved):

PR #3339, opened 2022-06-08
ISSUE #3320 - Safeti base - Shapes

### LINKED ISSUE #3320: Safeti base - Shapes
### Description
Safeti base's UI is broken  in the viewer   in v5 - we need to update it to make it usable for users. This is for the shapes tab.

### Goals
- [ ]  As a user I want to use the shapes functionality from safetibase in v5

### Tasks
- [ ] Fix alignment of save and clear icons
image

- [ ] Fix in colour picker not showing the complete colour in the textfield
image

### Related Resources

[MockUp](https://www.figma.com/file/J5I3hYQTrLCjEVP1iqcyZK/3D-Viewer?node-id=136550%3A176546) 

### Note
**This is broken also in issues, so this should fix both**

The-Daniel:
This fixes #3320

#### Description
- Styles the shapes tab of viewer to fit V5 style
- Fix position of action buttons when editing shape name
- Fix padding in the hex value input inside the colour picker modal


george-hadfield (dismissed):

frontend/src/v5/ui/routes/viewer/theme.ts
```diff
@@ -134,6 +134,18 @@ export const theme = createTheme(
 							'&.react-autosuggest__suggestions-container': {
 								boxShadow: 'none',
 							},
+							'&.color-picker__panel': { // Colour picker modal
```
Amantini1997, line 137: If you are writing the color picker style in the theme, could you please get rid of this file then ? `frontend\src\v5\ui\v4Adapter\dialogs\colorPickerDialog.overrides.ts`

It was used to tackle the color-picker style in the screenshot dialog, but if you put the style in a theme, it is no longer necessary

Amantini1997 (changes requested):

frontend/src/v5/ui/routes/viewer/theme.ts
The-Daniel: Ah, I hadn't seen that file. Only the colorPicker.overrides file

Amantini1997 (approved):

PR #3415, opened 2022-07-15
Issue 3395 - SSO Sign Up Endpoints

### LINKED ISSUE #3395: Single sign on : Sign up
### Description
To support user sign up through sign on(SSO) using microsoft accounts

### Goals
- [ ] Users should be able to sign up to 3D Repo through single sign on
- [ ] Users should be given an error if they try to sign up with single sign on with an email that is already used

#### MSAL
This example works and can be adapted to what we need
https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/msal-v1.4.16/samples/msal-node-samples/auth-code

The general gist of how this works:
- On an endpoint, redirect the user to microsoft to sign in
- Upon success, microsoft will redirect the user back to an endpoint specified

#### Identifying user accounts that are using SSO
Since an SSO user are not allowed to modify some of the fields and cannot log in through password, we need to identify them on the record, add a field in user's customData:
````js
{
   //...
  customData: {
    sso: {
        type: "aad",
        id: "" //unique id from microsoft
     }
  }
}
````

#### Sign up
```mermaid
graph TD;
  A["User click on sign up with SSO"]
  B["Redirect to AAD to authenticate"]
  D["redirect user to sign up page with some fields hidden (email, firstname/lastname etc)"]
  F["User fills in rest of the data (username, company name etc), signs up"]
  E["Fetch user data from microsoft (email, id, firstname, lastname)"]
  G["Create User, bypass email confirmation"]
  A-->B;
  B-->D;
  D-->F;
  F-->E;
  E-->G
```

##### End points
- General AAD Authentication endpoint:
  route:  `GET /sso/aad/authenticate`
  - should redirect to MS for authentication
  - should then redirect to the redirection endpoint
  - this is actually our login endpoint, but don't worry about it for now, just ensure microsoft authenticates the user properly
- Redirect endpoint
  route: `GET /sso/aad/authenticate-post`
   - this is the endpoint we expect MSAL to call upon authentication
   - we should've [stored in the state](https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-js-pass-custom-state-authentication-request) a front end url to redirect user back to
   - This is again, will be our login endpoint where we check whether the user exists in 3drepo, but don't worry for now
- Sign up with AAD
   route: `POST /sso/aad/signup`
   - we expect the same payload as before, without:
     - email
     - password
     - firstname
     - lastname
   - In this route, we should authenticate with MS again, which this time should embed another backend endpoint for redirection
   - The state should store all the parameters the user sent so we can retrieve it on the direct
- Sign up redirect
   route: `GET /sso/aad/signup-post`
   - the redirect url to finish off sign up once authenticated, this should be the point where we create the user 

##### NOTE:
- Code this in mind that we may add other sso providers in the future, so code placement/the way the logic work should minimise duplication when we add a new provider
- [This example](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/msal-v1.4.16/samples/msal-node-samples/auth-code) will probably be useful
- We only expect users to be able to sign up via v5 interface, but if we added any new restrictions, these should be applied to v4 as well.
- We should ensure sso.type/sso.id must be unique across users
- We will probably want to make the microsoft endpoints middleware as we will probably use them multiple times in different endpoints
- These endpoints should error appropriate if AAD is not configured in the app

### Tasks

ChristosTsiotsias:
This fixes #3395

#### Description
Enpoints for SSO Sign Up added as well as swagger documentation. (Tests are missing until code feedback is added)

Endoints added:

- sso/aad/authenticate 
- sso/aad/authenticate-post 
- sso/aad/signup
- sso/aad/signup-post

#### Test cases

- sso/aad/authenticate 
Should redirect the user to MS authentication page with the state param set to 3D Repo SSO signup page

- sso/aad/authenticate-post 
To be called upon successful MS authentication. Should redirect to the URL stored in req.query.state

- sso/aad/signup
Should receive signup fields apart from email, firstname, lastname, password. Should redirect to microsoft authentication again with the state set to signup fields and set to redirect to  sso/aad/signup-post endpoint.

- sso/aad/signup-post
Should retrieve signup fields from state, and obtain email, firstname, lastname fields from MS using an auth token. Should then sign the user up


backend/src/v5/routes/sso/aad/aad.js
```diff
@@ -0,0 +1,165 @@
...
+
+const authenticateRedirectUri = 'http://localhost/api/v5/sso/aad/authenticate-post';
+const signupRedirectUri = 'http://localhost/api/v5/sso/aad/signup-post';
+// Frontend uri for signup form
+const sigupUri = '';
+
+const authenticate = (req, res) => {
+	try {
+		const params = { redirectUri: authenticateRedirectUri, state: sigupUri };
+		const authenticationCodeUrl = getAuthenticationCodeUrl(params);
```
carmenfan, line 33: I'm pretty sure this is a function that returns a promise, so you're redirecting the response to a promise...?

image

carmenfan: also I would put `getAuthenticationCodeUrl` in services  once again and expose it as a function

backend/src/v5/routes/sso/aad/aad.js
```diff
@@ -0,0 +1,165 @@
...
+const { validateSignUpData } = require('../../../middleware/dataConverter/inputs/users');
+
+const authenticateRedirectUri = 'http://localhost/api/v5/sso/aad/authenticate-post';
+const signupRedirectUri = 'http://localhost/api/v5/sso/aad/signup-post';
+// Frontend uri for signup form
+const sigupUri = '';
+
+const authenticate = (req, res) => {
+	try {
+		const params = { redirectUri: authenticateRedirectUri, state: sigupUri };
```
carmenfan, line 32: can we put the signUpUri (typo btw) inside a stringified json so we are consistent with how we're treating state?

backend/src/v5/routes/sso/aad/aad.js
```diff
@@ -0,0 +1,165 @@
...
+		respond(req, res, err);
+	}
+};
+
+const signup = (req, res) => {
+	try {
+		const { body } = req;
+		const params = {
+			redirectUri: signupRedirectUri,
+			state: JSON.stringify({
```
carmenfan, line 55: I think you also need to encodeURI this, as the state becomes a query string

backend/src/v5/routes/sso/aad/aad.js
```diff
@@ -0,0 +1,165 @@
...
+
+	/**
+	* @openapi
+	* /sso/aad/authenticate:
+	*   get:
+	*     description: Redirects the user to Microsoft's authentication page and calls authenticate-post endpoint upon successful authentication
+	*     tags: [Aad]
+	*     operationId: authenticate
+	*/
+	router.get('/authenticate', checkAadConfig, authenticate);
```
carmenfan, line 92: not convince `checkAadConfig` need to exist

it's probably an error to throw on all the functions that will be in services/sso/aad

backend/src/v5/routes/sso/aad/aad.js
```diff
@@ -0,0 +1,165 @@
...
+
+	/**
+	 * @openapi
+	 * /sso/aad/authenticate-post:
+	 *   get:
+	 *     description: Redirects the user to 3D Repo's SSO signup page
+	 *     tags: [Aad]
+	 *     operationId: authenticatePost
+	 */
+	router.get('/authenticate-post', checkAadConfig, authenticatePost);
```
carmenfan, line 102: i think `authenticate-post` needs be a constant so you don't risk renaming this and not renaming the redirect URI

backend/src/v5/routes/sso/aad/aad.js
```diff
@@ -0,0 +1,165 @@
...
+};
+
+const establishRoutes = () => {
+	const router = Router({ mergeParams: true });
+
+	/**
+	* @openapi
+	* /sso/aad/authenticate:
+	*   get:
+	*     description: Redirects the user to Microsoft's authentication page and calls authenticate-post endpoint upon successful authentication
```
carmenfan, line 88:
```suggestion
	*     description: Redirects the user to Microsoft's authentication page and redirects the user to a provided URI upon success
```

carmenfan: we need to tell the reader how to give us the redirect URL

backend/src/v5/routes/sso/aad/aad.js
```diff
@@ -0,0 +1,165 @@
...
+	*     tags: [Aad]
+	*     operationId: authenticate
+	*/
+	router.get('/authenticate', checkAadConfig, authenticate);
+
+	/**
```
carmenfan, line 94: don't document this, this is an internal api that we don't ever expect anyone to call but microsoft

backend/src/v5/routes/sso/aad/aad.js
```diff
@@ -0,0 +1,165 @@
...
+	 *                 example: true
+	 *               captcha:
+	 *                 type: string
+	 *                 description: The reCAPTCHA token generated from the sign up form
+	 *                 example: 5LcN0ysfAAAAAHpnld1tAweI7DKU7dswmwnHWYcB
+	 *     responses:
+	 *       401:
+	 *         $ref: "#/components/responses/notLoggedIn"
+	 */
+	router.post('/signup', checkAadConfig, signup);
```
carmenfan, line 145: we should be validating the payload here

backend/src/v5/routes/sso/aad/aad.js
```diff
@@ -0,0 +1,165 @@
...
+
+	/**
```
carmenfan, line 147: again, no need to document

backend/src/v5/routes/sso/aad/aad.js
```diff
@@ -0,0 +1,165 @@
...
+	 *     description: Retrieves user email, firstname and lastname from Microsoft Graph API, signs a user up and sends a verification email to the email address provided
+	 *     tags: [Aad]
+	 *     operationId: signup-post
+	 *     responses:
+	 *       400:
+	 *         $ref: "#/components/responses/invalidArguments"
+	 *       200:
+	 *         description: Retrieves user email, firstname and lastname from Microsoft Graph API, signs a user up and sends a verification email to the email address provided
+	 */
+	router.get('/signup-post', checkAadConfig, getUserDetailsAndValidateEmail, validateSignUpData, signupPost);
```
carmenfan, line 160: put name of route in constant

carmenfan: At this state, i don't think we need to do another validation of the payload. we just make sure user email is not taken already, piece together the info and call signup.

We also need to differentiate between a user that already signed up, and a user that already signed with SSO and give different error messages as the UI needs to respond differently

backend/src/v5/utils/sso/aad.js
```diff
@@ -0,0 +1,40 @@
...
+
+const config = require('../config');
+const msal = require('@azure/msal-node');
+
+const Aad = {};
+
+let clientApplication;
+
+Aad.getClientApplication = () => {
+	const clientAppConfig = { auth: config.sso.aad };
```
carmenfan, line 26: could we also configure the logging to log into our logger (with the right labelling)

backend/src/v5/middleware/dataConverter/inputs/sso/aad/index.js
```diff
@@ -0,0 +1,68 @@
...
+
+Aad.checkAadConfig = async (req, res, next) => {
+	if (!config.sso?.aad) {
+		respond(req, res, createResponseCode(templates.unknown, 'Microsoft Authentication is not configured'));
+	}
+
+	await next();
+};
+
+Aad.getUserDetailsAndValidateEmail = async (req, res, next) => {
```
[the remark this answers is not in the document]
ChristosTsiotsias: The next middleware is validating all the data Im getting from MS, so If I get them in the processor how would I be able to validate them using yup ?

backend/src/v5/middleware/dataConverter/inputs/sso/aad/index.js
```diff
@@ -0,0 +1,68 @@
...
+				Authorization: `Bearer ${response.accessToken}`,
+			},
+		});
+
+		req.body = {
+			...JSON.parse(req.query.state),
+			email: user.data.mail,
+			firstName: user.data.givenName,
+			lastName: user.data.surname,
+			password: generateHashString(),
```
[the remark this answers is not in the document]
carmenfan: https://github.com/3drepo/3drepo.io/pull/3415#discussion_r922186199

data validation should be on user data, not data we impose on the object to make things work

we should impose a different version of the schema for sso

backend/src/v5/routes/sso/aad/aad.js
```diff
@@ -0,0 +1,165 @@
...
+const { respond } = require('../../../utils/responder');
+const { templates } = require('../../../utils/responseCodes');
+const { validateSignUpData } = require('../../../middleware/dataConverter/inputs/users');
+
+const authenticateRedirectUri = 'http://localhost/api/v5/sso/aad/authenticate-post';
+const signupRedirectUri = 'http://localhost/api/v5/sso/aad/signup-post';
+// Frontend uri for signup form
+const sigupUri = '';
+
+const authenticate = (req, res) => {
```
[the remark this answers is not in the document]
ChristosTsiotsias: but how will I set the params in the middleware ?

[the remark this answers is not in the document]
ChristosTsiotsias: 2 different middleware methods for authenticate and signup right ?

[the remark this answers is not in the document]
carmenfan: single middleware, with the 2 functions from route utilising them

backend/src/v5/routes/sso/aad/aad.js
ChristosTsiotsias: I think thats happening authomatically ... is it needed ?
image

carmenfan: if msal do it for you then i guess it's not needed

backend/src/v5/routes/sso/aad/aad.js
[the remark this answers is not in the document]
ChristosTsiotsias: but they have different params.. what am I missing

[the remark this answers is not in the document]
carmenfan: It's just code, you can do whatever you like. you just need a function before the middleware to setup the parameters.

Something like this?

NB: don't copy the function names... it's just there to show you what it does 😆 

If you're stuck, leave it - i'll fix it

````js
router.post("signUp", functionToSetupTheParams, authenticateMiddleware, ...);
route.get("authenticate", functionToSetupTheParams2, authenticateMiddleware, ...);

const functionToSetupTheParams = (req, res, next) => {
   const params = {/*...*/};
   req.authParams = params;
   next()
}

const authenticateMiddleware = (req, res, next) => {
   const params = req.authParams;
  //...
}
````

File: backend/src/v5/middleware/dataConverter/inputs/sso/aad/index.js [hunk 15]
```diff
@@ -0,0 +1,68 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const { createResponseCode, templates } = require('../../../../../utils/responseCodes');
+const axios = require('axios');
```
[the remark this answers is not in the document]
ChristosTsiotsias: should I replace httpsReq in v4 and make v5 use it ?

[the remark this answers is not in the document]
carmenfan: no instead of importing v4 [here](https://github.com/3drepo/3drepo.io/blob/staging/backend/src/v5/utils/httpsReq.js) 

write the functions you need using axios (and make sure whatever else is relying on this file is changed appropriately)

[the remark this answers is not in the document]
carmenfan: Maybe rename it to webRequests, we don't need to have httpsReq specifically, if we start calling http protocol it should live there too

backend/tests/v5/e2e/routes/sso/aad/aad.test.js
```diff
@@ -0,0 +1,180 @@
...
+const newUserEmail = 'newUserEmail@email.com';
+
+const setupData = async () => {
+	await ServiceHelper.db.createUser(testUser, [], { email: userEmail });
+	await ServiceHelper.db.createUser(testUserSso, [], { email: userEmailSso,
+		sso: { type: aad, id: generateRandomString() } });
+};
+
+const testAuthenticate = () => {
+	describe('Sign Up Authenticate', () => {
```
carmenfan, line 49:
```suggestion
	describe('Authenticate', () => {
```

backend/tests/v5/e2e/routes/sso/aad/aad.test.js
```diff
@@ -0,0 +1,180 @@
...
+		test('should redirect the user to Microsoft authentication page', async () => {
```
carmenfan, line 50: add a test for when the redirect URI is not provided

backend/src/v5/utils/responseCodes.js
```diff
@@ -22,6 +22,7 @@ const ResponseCodes = {};
 
 ResponseCodes.templates = {
 	ok: { message: 'OK', status: 200 },
+	found: { message: 'Found', status: 302 },
```
carmenfan, line 25: this isn't an actual respond we send out and only used within tests to get the status

so just use 302 instead of this

backend/tests/v5/e2e/routes/sso/aad/aad.test.js
```diff
@@ -0,0 +1,180 @@
...
+			const signupUri = generateRandomString();
+			const res = await agent.get(`/v5/sso/aad/authenticate?signupUri=${signupUri}`)
+				.expect(templates.found.status);
+
+			const redirectUri = res.headers.location;
+			expect(redirectUri).toEqual(
```
carmenfan, line 56: parse the URL and test for the parts you want to test https://nodejs.org/docs/latest-v14.x/api/url.html#url_the_whatwg_url_api

backend/tests/v5/e2e/routes/sso/aad/aad.test.js
```diff
@@ -0,0 +1,180 @@
...
+	});
+};
+
+const testAuthenticatePost = () => {
+	describe('Sign Up Authenticate Post', () => {
+		test('should redirect the user to ', async () => {
+			const state = { redirectUri: generateRandomString() };
+			const res = await agent.get(`/v5/sso/aad/authenticate-post?state=${encodeURIComponent(JSON.stringify(state))}`)
+				.expect(templates.found.status);
+			expect(res.headers.location).toEqual(expect.stringMatching(state.redirectUri));
```
carmenfan, line 75: why do you need stringMatching here?

backend/tests/v5/e2e/routes/sso/aad/aad.test.js
```diff
@@ -0,0 +1,180 @@
...
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const SuperTest = require('supertest');
+const ServiceHelper = require('../../../../helper/services');
+const { src } = require('../../../../helper/path');
+const { generateRandomString } = require('../../../../helper/services');
+const { authenticateRedirectUri, signupRedirectUri } = require('../../../../../../src/v5/services/sso/aad/aad.constants');
+
+jest.mock('../../../../../../src/v5/services/sso/aad', () => ({
```
carmenfan, line 24: are you providing it valid config values to do the test? if so that sounds like a terrible idea?

backend/tests/v5/e2e/routes/sso/aad/aad.test.js
```diff
@@ -0,0 +1,180 @@
...
+
+			const redirectUri = res.headers.location;
+			expect(redirectUri).toEqual(
+				expect.stringContaining('https://login.microsoftonline.com/common/oauth2/v2.0/authorize'),
+			);
+			expect(redirectUri).toEqual(
+				expect.stringContaining(`redirect_uri=${encodeURIComponent(signupRedirectUri)}`),
+			);
+			expect(redirectUri).toEqual(
+				expect.stringContaining(`state=${encodeURIComponent(JSON.stringify({
```
carmenfan, line 113: i don't think you need to spell out the json?

```suggestion
				expect.stringContaining(`state=${encodeURIComponent(JSON.stringify(newUserData))})
```

backend/tests/v5/e2e/routes/sso/aad/aad.test.js
```diff
@@ -0,0 +1,180 @@
...
+		agent = await SuperTest(server);
+		await setupData();
+	});
+
+	afterAll(() => ServiceHelper.closeApp(server));
+
+	testAuthenticate();
+	testAuthenticatePost();
+	signup();
+	signupPost();
```
carmenfan, line 179: Def good to have separate tests for each routes, but e2e tests should really replicate the user's experience, we should have a test where it goes through the 2 routes and ultimately redirects to the URI user provided, that's why it's called end to end 😄

backend/tests/v5/e2e/routes/sso/aad/aad.test.js
```diff
@@ -0,0 +1,180 @@
...
+
+		const newUserDataFromAad = {
+			mail: newUserEmail,
+			givenName: generateRandomString(),
+			surname: generateRandomString(),
+			id: generateRandomString(),
+		};
+
+		test('should fail if the email already exists', async () => {
+			const state = { ...newUserData };
```
carmenfan, line 141: i think you should have a function to generate data instead of reusing the same data everywhere

backend/tests/v5/e2e/routes/sso/aad/aad.test.js
```diff
@@ -0,0 +1,180 @@
...
+		};
+
+		const newUserDataFromAad = {
+			mail: newUserEmail,
+			givenName: generateRandomString(),
+			surname: generateRandomString(),
+			id: generateRandomString(),
+		};
+
+		test('should fail if the email already exists', async () => {
```
carmenfan, line 140: I think all these errors needs to be a query string on the redirected URI

otherwise it probably will be difficult for front end to manage the error handling

backend/tests/v5/e2e/routes/sso/aad/aad.test.js
```diff
@@ -0,0 +1,180 @@
...
+			const res = await agent.get(`/v5/sso/aad/signup-post?state=${encodeURIComponent(JSON.stringify(state))}`)
+				.expect(templates.invalidArguments.status);
+			expect(res.body.code).toEqual(templates.invalidArguments.code);
+		});
+
+		test('should sign a new user up', async () => {
+			const state = { ...newUserData };
+			Aad.getUserDetails.mockResolvedValueOnce({ data: newUserDataFromAad });
+			await agent.get(`/v5/sso/aad/signup-post?state=${encodeURIComponent(JSON.stringify(state))}`)
+				.expect(templates.ok.status);
```
carmenfan, line 160: can we verify a user has been created? (like call getUserByUsername or something)

backend/src/v5/middleware/dataConverter/inputs/users.js
```diff
@@ -183,33 +183,17 @@ const generateSignUpSchema = () => {
...
-		lastName: types.strings.name.required().transform(formatPronouns),
 		countryCode: types.strings.countryCode.required(),
 		company: types.strings.title.optional(),
 		mailListAgreed: Yup.bool().required(),
 		...(captchaEnabled ? { captcha: Yup.string().required() } : {}),
-	})
-		.noUnknown().required();
+	}).noUnknown().required();
 
 	return captchaEnabled
```
carmenfan, line 198: Can we strip the captcha field after testing it, then you can just pass the whole body into the state for sso endpoints without worry

backend/src/v5/middleware/dataConverter/inputs/users.js
```diff
@@ -234,6 +239,16 @@ Users.validateSignUpData = async (req, res, next) => {
 	}
 };
 
+Users.validateSsoSignUpData = async (req, res, next) => {
```
carmenfan, line 242: a lot of duplication with validate `validateSignUpData`

do something like this

````js

const validateSignUpData = (isSSO) => {
try {
		const schema = generateSignUpSchema(isSSO); //Merge this function too.
		req.body = await schema.validate(req.body);
		await next();
	} catch (err) {
		respond(req, res, createResponseCode(templates.invalidArguments, err?.message));
	}
}

Users.validateSignUpData = validateSignUpData(false);
Users.validateSsoSignUpData = validateSignUpData(true);
````

File: backend/src/v5/processors/users.js [hunk 29]
```diff
@@ -30,9 +30,16 @@ const { publish } = require('../services/eventsManager/eventsManager');
 const { sendEmail } = require('../services/mailer');
 const { templates } = require('../services/mailer/mailer.constants');
 
-Users.signUp = async (newUserData) => {
+Users.signUp = async (newUserData, generatePassword) => {
```
carmenfan, line 33: I don't think we should have a flag here like this

we could just check the data, if it's sso we know we need to generate password?

backend/src/v5/middleware/dataConverter/inputs/sso/aad/index.js
```diff
@@ -0,0 +1,90 @@
...
+const { respond } = require('../../../../../utils/responder');
+
+const Aad = {};
+
+Aad.getUserDetailsAndCheckEmailAvailability = async (req, res, next) => {
+	const { data: { mail, givenName, surname, id } } = await getUserDetails(req.query.code,
+		signupRedirectUri, req.session.pkceCodes?.verifier);
+
+	try {
+		const user = await getUserByQuery({ 'customData.email': mail }, { 'customData.sso': 1 });
```
carmenfan, line 31: since we use this more than once, we should just add a function `getUserByEmail` in `models/users`

backend/src/v5/middleware/dataConverter/inputs/sso/aad/index.js
```diff
@@ -0,0 +1,90 @@
...
+const { authenticateRedirectUri, signupRedirectUri } = require('../../../../../services/sso/aad/aad.constants');
+const { createResponseCode, templates } = require('../../../../../utils/responseCodes');
+const { getAuthenticationCodeUrl, getUserDetails } = require('../../../../../services/sso/aad');
+const { aad } = require('../../../../../services/sso/sso.constants');
+const { getUserByQuery } = require('../../../../../models/users');
+const { respond } = require('../../../../../utils/responder');
+
+const Aad = {};
+
+Aad.getUserDetailsAndCheckEmailAvailability = async (req, res, next) => {
```
carmenfan, line 26: can we just name this `validateUserDetails`

backend/src/v5/middleware/dataConverter/inputs/sso/aad/index.js
```diff
@@ -0,0 +1,90 @@
...
+		const user = await getUserByQuery({ 'customData.email': mail }, { 'customData.sso': 1 });
+		const message = user.customData.sso ? 'Email already exists from SSO user' : 'Email already exists';
+		respond(req, res, createResponseCode(templates.invalidArguments, message));
+		return;
+	} catch {
+		// do nothing
+	}
+
+	req.body = {
+		...JSON.parse(req.query.state),
```
carmenfan, line 40: considering the front end redirect goes into this, we need to strip it out

carmenfan: it is quite psosible for this JSON parser to fail, which as it stands will break the server, we should really put this in a try

backend/tests/v5/unit/middleware/dataConverter/inputs/sso/aad/aad.test.js
```diff
@@ -0,0 +1,171 @@
...
+			},
+			session: { pkceCodes: { verifier: generateRandomString() } },
+		};
+
+		test(`should respond with ${templates.invalidArguments.code} if the email already exists`, async () => {
+			AadServices.getUserDetails.mockResolvedValueOnce(aadUserDetails);
+			UsersModel.getUserByQuery.mockResolvedValueOnce({ customData: {} });
+			const mockCB = jest.fn();
+			await Aad.getUserDetailsAndCheckEmailAvailability(req, {}, mockCB);
+			expect(mockCB.mock.calls.length).toBe(0);
```
carmenfan, line 62:
```suggestion
			expect(mockCB).not.toHaveBeenCalled()
```

backend/tests/v5/unit/middleware/dataConverter/inputs/sso/aad/aad.test.js
```diff
@@ -0,0 +1,171 @@
...
+			expect(Responder.respond).toHaveBeenCalledTimes(1);
+			expect(Responder.respond.mock.results[0].value.code).toEqual(templates.invalidArguments.code);
```
carmenfan, line 64:
```suggestion
			expect(Responder.respond).toHaveBeenCalledWith(req, {}, createResponseCode(templates.invalidArguments, 'Email already exists'));
```

backend/tests/v5/unit/middleware/dataConverter/inputs/sso/aad/aad.test.js
```diff
@@ -0,0 +1,171 @@
...
+			expect(Responder.respond.mock.results[0].value.message).toEqual('Email already exists');
+		});
+
+		test(`should respond with ${templates.invalidArguments.code} if the email already exists`, async () => {
```
carmenfan, line 68: Same as above for all the tests

backend/tests/v5/unit/middleware/dataConverter/inputs/sso/aad/aad.test.js
carmenfan: but these are probably all N/A as we need to redirect with the error messages

backend/tests/v5/unit/middleware/dataConverter/inputs/sso/aad/aad.test.js
```diff
@@ -0,0 +1,171 @@
...
+			);
+		});
+	});
+};
+
+const testSetAuthenticateAuthParams = () => {
+	describe('Set auth params for authenticate endpoint', () => {
+		const req = {
+			query: { signupUri: generateRandomString() },
+			session: { pkceCodes: { challenge: generateRandomString(), challengeMethod: generateRandomString() } },
```
carmenfan, line 103: the amount of times you have to do this you may as well write a helper for it

config/testV5/config.js
```diff
@@ -167,4 +167,11 @@ module.exports = {
 		},
 		sender: '"3D Repo" <dummyEmail@3drepo.org>'
 	},
+
+	sso: {
+		aad: {
+			clientId: "4caba479-bfa6-46bc-9214-a0dc1dcbf8c7",
```
carmenfan, line 173: unless these are keys that AAD let everyone use, i don't think we should be putting genuine ids here (even if it's just a sandbox application.

carmenfan (changes requested):
I get an unhandled promise rejection and my request never returns if my config isn't configured with sso and i'm trying to run a sso endpoint:
image

If I have it configured, it's complaining about the redirect_uri:
image

base on my config, my config.api_server.url is `{3drepo_api}`

backend/src/v5/middleware/dataConverter/inputs/sso/aad/index.js
```diff
@@ -0,0 +1,90 @@
...
+		codeChallenge: req.session.pkceCodes.challenge,
+		codeChallengeMethod: req.session.pkceCodes.challengeMethod,
+	};
+
+	req.authParams = params;
+
+	await next();
+};
+
+Aad.setSignupAuthParams = async (req, res, next) => {
```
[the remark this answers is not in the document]
ChristosTsiotsias: but I can try to do it slightly differently

backend/src/v5/processors/users.js
ChristosTsiotsias: you mean check if newUserData.password is empty then generate it ? so no flag at all ?

carmenfan (approved):
@ChristosTsiotsias  I finished off this PR with some changes

Main thing to be aware of is that I moved the sso middleware from `middleware/dataConverter/input/sso` to `middleware/sso`, this is likely to cause conflicts with your other branch

PR #3473, opened 2022-07-28
#3471 - send in the branchName to the action and protect major branches

### LINKED ISSUE #3471: update close PR system to send in the issue number as a variable and protect against accidental staging/master deletion
### Description

Currently the pipeline for destroy is wiping the staging and would wipe master also. We can protect it with some conditions on the pipeline. It's also not getting the correct branchName to destroy.

### Steps to replicate
- Merge a pull request

### Current Behaviour
- branch that's being merged to is removed, not the source, but we'll protect those with conditions

### Expected Behaviour
- staging/master unaffected by merges

keifgwinn:
This fixes #3471 

#### Description
- the action sends in the issue number as branchName now
- additional conditional branch protection for major branches

#### Test cases
- hopefully now won't remove things it shouldn't and get the correct branch name to remove

### DIFF
--- a/.github/workflows/onPRClose.yml
+++ b/.github/workflows/onPRClose.yml
@@ -34,4 +34,4 @@ jobs:
         azure-devops-project-url: https://dev.azure.com/3drepo/3drepo.io
         azure-pipeline-name: 'destroy'
         azure-devops-token: ${{ secrets.AZURE_DEVOPS_TOKEN }}
-        azure-pipeline-variables:  '{"branchName": "issue-${{needs.tagOnPRMerge.outputs.issue-number}}"}'
+        azure-pipeline-variables:  '{"branchName": "issue-${{ needs.tagOnPRMerge.outputs.issue-number }}"}'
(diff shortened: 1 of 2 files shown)


.github/workflows/onPRClose.yml
```diff
@@ -24,10 +26,12 @@ jobs:
   destroy:
     name: Call Azure Pipeline
     runs-on: ubuntu-latest
+    needs: tagOnPRMerge
```
carmenfan, line 29: super tiny in the grand scheme of things...

If you need the issue number extracted from `tagOnPRMerge` it migth be better to split it into 3 jobs (extract issue number, apply staging label, destory pipeline), where the last 2 needs the first one.

as it stands i think it'll need to be performed serially and you may as well add this as a step in the same job, and if for any reason if the staging label step failed it probably won't run the destroy pipeline

carmenfan (changes requested):

.github/workflows/onPRClose.yml
keifgwinn: made those changes

carmenfan (approved):

PR #3474, opened 2022-07-28
ISSUE #3406 - Unexpected behaviour in upload revisions modal

### LINKED ISSUE #3406: Unexpected behaviours in upload revisions modal
### Description
- [x] Selecting a destination, clearing the selection, and then trying to choose the initial destination again doesn't work. It says it's already in use.
- [x] Selecting an existing container. Then changing to creating a new container doesn't disable the sidebar fields.
- [x] The "Latest Revision" in the destination autocomplete actually shows the container name, not the revision name
- [x] When an error occurs in the revision description the field turns red, but an error message does not appear.
- [x] The modal maintains all data when closed and reopened. It should instead wipe the data.
- [x] The modal doesn't respond well to small screen sizes
- [x] Long container names are not truncated in destination list
- [x] Using the destination autocomplete creates errors in the console

### Steps to replicate
- Open the revision modal

image
Here there is an error (revision description too long), but there is no error text
image
Here you can see that the container Empty Container says it is in use. However there are no other files in the modal, and the container shows that there are no active uploads

The-Daniel:
This fixes #3406 

#### Description
- Cleaned up a lot of code
- Form is cleared when modal is closed
- The latest revision in the Destination dropdown now shows the latest revision instead of the container name
- Revision description errors are now visible
- Modal responds better to smaller screen sizes
- Long container names are now truncated in destination list
- Removed console errors when using autocomplete
- **FIXED**: Selecting a destination, clearing the selection, and then trying to choose the initial destination again doesn't work. It says it's already in use.
- **FIXED**: Selecting an existing container. Then changing to creating a new container doesn't disable the sidebar fields.

#### Test cases
- Open the uploads modal
- Try uploading files
- See how smaller screens look in both stages of the upload screen


frontend/src/v5/ui/routes/dashboard/projects/containers/uploadFileForm/uploadFileForm.component.tsx
```diff
@@ -179,61 +180,63 @@ export const UploadFileForm = ({ openState, onClickClose }: IUploadFileForm): JS
...
 							/>
 						</Padding>
-					</ScrollArea>
+					</UploadsListScroll>
 					<Sidebar
-						open={Number.isInteger(selectedIndex) && !isUploading}
+						open={sidebarOpen}
 						onClick={() => setSelectedIndex(null)}
-						noButton={!(Number.isInteger(selectedIndex))}
+						noButton={!sidebarOpen}
```
Amantini1997, line 232: After the changes, `noButton` is true when the `sidebarOpen` is false.
I think you can get rid of it then

frontend/src/v5/ui/controls/sideBar/sidebar.component.tsx
```diff
@@ -35,14 +35,14 @@ export const Sidebar = ({
 	hidden = false,
 	children,
 }: ISidebar): JSX.Element => (
-	<Container className={className} open={open} hidden={hidden}>
+	<SidebarContainer className={className} open={open} hidden={hidden}>
 		{!noButton && (
```
Amantini1997, line 39: I think you can get rid of the `noButton` prop (as well as the conditional render) as it is true when the prop `open` is false, and have just 
```ts
	<SidebarContainer className={className} open={open} hidden={hidden}>
		<ExpandButton onClick={onClick} variant="main">
			<ExpandIcon />
		</ExpandButton>
		<SidebarContent>
			{children}
		</SidebarContent>
	</SidebarContainer>

```

frontend/src/v5/ui/routes/dashboard/projects/containers/uploadFileForm/uploadFileForm.component.tsx
```diff
@@ -71,12 +68,14 @@ export const UploadFileForm = ({ openState, onClickClose }: IUploadFileForm): JS
...
-	}, [isUploading]);
+		remove();
+		setSelectedIndex(null);
+		setIsUploading(false);
+	}, [open]);
 
 	useEffect(() => {
-		if (fields.some((field) => filesizeTooLarge(field.file))) {
-			setError('uploads', { type: 'custom', message: '' });
+		if (fields.some(({ file }) => filesizeTooLarge(file))) {
```
Amantini1997, line 77: I am not sure if it would make the code cleaner in the overall, but for this goal you could use a `useState` variable, like `filesTooLarge` and here you just call 
```ts
setFilesTooLarge(filesTooLarge(fields.some(({ file }) => filesizeTooLarge(file)))
```

and then use that variable to set the disableButton disabled state.

This way you don't need to alter the form errors. But again, I am not sure in the overall it would make the code much cleaner or just cleaner, so feel free to ignore this comment in case

Amantini1997 (changes requested):

frontend/src/v5/ui/controls/sideBar/sidebar.component.tsx
The-Daniel: I initially did this in case the sidebar is used somewhere else where you want the button to be visible when the sidebar is closed (and act as an open sidebar button). Since it's currently not being used elsewhere I'll get rid of it. If in the future we decide it's needed then it shouldn't be too hard to re-add

frontend/src/v5/ui/routes/dashboard/projects/containers/uploadFileForm/uploadFileForm.component.tsx
The-Daniel: This let me clean up the code quite a bit, thanks!

sanmont3drepo (approved):

Amantini1997 (approved):

PR #3484, opened 2022-07-29
#3480 - destroy pipeline error handling

### LINKED ISSUE #3480: destroy pipeline - add check on if the instance exists to reduce failed runs
### Description

We could check to see if the instance exists before trying to delete it to reduce failed runs, but actually in testing and setting it up it seemed like over kill so now I've set it to just not error if the instance is missing.

### Goals

- [x] Do not error out if the instance doesn't exist

### Tasks

- [x] Update .azure/destroy.yaml

keifgwinn:
This fixes #3480

#### Description

- Actually decided that it's simplest to just supress the error if the branch doesn't exist.

#### Test cases
close PR doesn't generate a failure email when run if the branch is already missing or was never created.


keifgwinn:
to test we should "azp run" on monday and then pull/close this PR.

carmenfan (approved):

PR #3505, opened 2022-08-08
Issue 3499 - more utility scripts

### LINKED ISSUE #3499: add more utility scripts for teamspace management
### Description
- a script to provide a list of teamspaces with active subscriptions
- extend `removeTeamspaceAndOwner` to allow configuration option to only remove the teamspace 
  - drop user should also remove the avatar
  - drop database is not removing fileshare files... we need to remove all refs prior to dropping

### Goals
- should be able to only remove a teamspace without removing the owner from the db
- should be able to get a list of active subscriptions

carmenfan:
This fixes #3499 

#### Description
- changed removeTeamspaceAndOwner script to be able to just delete the teamspace without deleting the owner
- add a script to dump out a list of teamspaces with active licenses
- some refactoring to push the functions back in the main code base and tested

#### Test cases
- should be able to remove a teamspace without removing the owner
- should be able to remove a teamspace and the owner
- should not fail if the teamspace doesn't exist
- should not fail if the owner doesn't exist
- should be able to print out a list of teamspaces and their active licenses


backend/tests/v5/unit/models/loginRecords.test.js
```diff
@@ -16,22 +16,23 @@
...
-const username = 'someUsername';
-const ipAddress = '290.241.146.180';
-const browserUserAgent = 'browser user agent';
-const pluginUserAgent = 'PLUGIN: plugin user agent';
-const referrer = 'www.google.com';
-const LoginRecord = require(`${src}/models/loginRecord`);
+const LoginRecord = require(`${src}/models/loginRecords`);
 
 const testSaveLoginRecord = () => {
+	const sessionId = '123456';
```
ChristosTsiotsias, line 30: we could use random strings here

backend/tests/v5/unit/models/loginRecords.test.js
```diff
@@ -93,6 +94,21 @@ const testSaveLoginRecord = () => {
...
+const testRemoveAllUserRecords = () => {
+	describe('Remove all user login records', () => {
+		test('Should just drop the user collection within loginRecords', async () => {
+			const fn = jest.spyOn(db, 'dropCollection').mockResolvedValue(undefined);
+
+			const username = generateRandomString();
+			await expect(LoginRecord.removeAllUserRecords(username)).resolves.toBeUndefined();
+
+			expect(fn).toHaveBeenCalledTimes(1);
+			expect(fn).toHaveBeenCalledWith('loginRecords', username);
```
ChristosTsiotsias, line 106: we could store 'loginRecords' in a constant somewhere

backend/tests/v5/unit/models/notifications.test.js
```diff
@@ -0,0 +1,41 @@
...
+const testRemoveAllUserNotifications = () => {
+	describe('Remove all user notifications', () => {
+		test('Should just drop the user collection within notifications', async () => {
+			const fn = jest.spyOn(db, 'dropCollection').mockResolvedValue(undefined);
+
+			const username = generateRandomString();
+			await expect(Notifications.removeAllUserNotifications(username)).resolves.toBeUndefined();
+
+			expect(fn).toHaveBeenCalledTimes(1);
+			expect(fn).toHaveBeenCalledWith('notifications', username);
```
ChristosTsiotsias, line 34: we could store 'notifications' in a constant

backend/tests/v5/unit/models/notifications.test.js
```diff
@@ -0,0 +1,41 @@
...
+		});
+	});
+};
+
+describe('models/loginRecords', () => {
```
ChristosTsiotsias, line 39: 'models/notifications'

backend/tests/v5/unit/models/teamspaces.test.js
```diff
@@ -427,6 +427,17 @@ const testRemoveUserFromAdminPrivileges = () => {
 	});
 };
 
+const testGetAllTeamspacesWithActiveLicenses = () => {
+	describe('Get all teamspaces with active licenses', () => {
+		test('Should perform a query to find all teamspaces with active subscriptions', async () => {
+			const expectedRes = [generateRandomString()];
+			const fn = jest.spyOn(db, 'find').mockResolvedValueOnce(expectedRes);
+			await expect(Teamspace.getAllTeamspacesWithActiveLicenses()).resolves.toEqual(expectedRes);
+			expect(fn).toHaveBeenCalledTimes(1);
```
ChristosTsiotsias, line 436: we could check the find params ?

backend/tests/v5/unit/models/users.test.js
```diff
@@ -241,6 +241,13 @@ const testDeleteFromFavourites = () => {
 			expect(fn.mock.calls.length).toBe(0);
 		});
 
+		test('Should not return error when trying to remove all favourites from a specific teamspace but the user doesn\'t have any', async () => {
```
ChristosTsiotsias, line 244: It will still throw an error if the user has no favourites right ? It will not throw error if favouritesToRemove param is empty. Maybe the description needs an adjustment ?

backend/tests/v5/unit/models/users.test.js
```diff
@@ -241,6 +241,13 @@ const testDeleteFromFavourites = () => {
...
+			jest.spyOn(db, 'findOne').mockResolvedValue({ customData: favouritesData });
```
ChristosTsiotsias, line 245: we could use mockResolvedValueOnce on both findOne and updateOne

ChristosTsiotsias (changes requested):

backend/src/scripts/utility/teamspaces/removeTeamspaceAndOwner.js
```diff
@@ -56,37 +33,47 @@ const removeUserFromAllTeamspaces = async (user) => {
...
+const run = async (teamspaces, removeOwners) => {
+	const teamspaceArr = teamspaces.split(',');
+	const tsList = await getTeamspaceList();
+	for (const teamspace of teamspaceArr) {
+		logger.logInfo(`-${teamspace}`);
+		if (tsList.includes(teamspace)) {
+			// eslint-disable-next-line no-await-in-loop
+			await removeTeamspace(teamspace);
+		}
+		if (removeOwners) {
```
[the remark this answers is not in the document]
carmenfan: Hmmm I don't agree here 😆 

If we can't find the teamspace, the owner might still exist. This can happen if we removed the teamspace by running the script once with removeOwner set to false, then run it again to remove the owner account.

backend/src/scripts/utility/teamspaces/allActiveLicenses.js
```diff
@@ -0,0 +1,69 @@
...
+const writeResultsToFile = (results, outFile) => new Promise((resolve) => {
+	const currDate = new Date();
+	logger.logInfo(`Writing results to ${outFile}`);
+	const writeStream = FS.createWriteStream(outFile);
+	writeStream.write('Teamspace,Type, Data(MB),Seats,ExpiryDate\n');
+	results.forEach(({ user, customData }) => {
+		const subs = customData.billing.subscriptions;
+		Object.keys(subs).forEach((subType) => {
+			const { collaborators, expiryDate, data } = subs[subType];
+			if (!expiryDate || expiryDate > currDate) {
```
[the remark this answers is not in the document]
carmenfan: no i guess not

backend/tests/v5/unit/models/loginRecords.test.js
carmenfan: bit here or there. We already have it as constant in the code, it's just not exposed, and i don't think we want to expose it for the test

We can do that if we use it in more than one place in the tests in the future.

backend/tests/v5/unit/models/notifications.test.js
carmenfan: same as above

backend/tests/v5/unit/models/teamspaces.test.js
carmenfan: the query is too complex, the copy and paste isn't worth it imo. it just makes the test harder to read

can add it if you really think it's worth while

backend/tests/v5/unit/models/users.test.js
carmenfan: it should throw an error if i'm trying to remove a specific favourite (i.e. provided a fed/model id) - this is a separate existing test.

it shouldn't if i am trying to remove any favourites within a certain teamspace (that's the change i made within the codebase)

I think this describe accurately described the situation?

ChristosTsiotsias: Right I see.. However the findOne mock is still returning some favourites. Shouldnt it be mocked to return null ?

backend/tests/v5/unit/models/teamspaces.test.js
ChristosTsiotsias: sure its ok I guess

backend/tests/v5/unit/models/users.test.js
carmenfan: doesn't really matter , the teamspace in question (`teamspace3`) has no entry on the data we mocked (teamspace1 and teamspace2).

updated as you wish

ChristosTsiotsias (approved):

PR #3510, opened 2022-08-08
Issue #3498 - Fixed loading bug that was caused  because viewer was re-rendered multiple times 

### LINKED ISSUE #3498: Sometimes I get 404 when loading a model/ needs research
### Description
Sometimes I get a 404 error while loading a model. 

- This is happening in: locally 

### Steps to replicate 
*Needs research*
-
 
### Current Behaviour
- 

### Expected Behaviour
- 

image

sanmont3drepo:
This fixes #3498

#### Description
- Avoided updating the store if there is no new data


frontend/src/v5/store/containers/containers.sagas.ts
```diff
@@ -66,8 +69,12 @@ export function* fetchContainers({ teamspace, projectId }: FetchContainersAction
 	try {
 		const { containers }: FetchContainersResponse = yield API.Containers.fetchContainers({ teamspace, projectId });
 		const containersWithoutStats = prepareContainersData(containers);
+		const storedContainers = yield select(selectContainers);
```
The-Daniel, line 72: If you have no containers in a project then the list remains pending forever. The same is true for no federations in fetchFederations

The-Daniel (changes requested):
It's so much faster to navigate the app with this change. Since fetchFederations and fetchContainers are called in the federations list you can navigate from here to the containers list and the data is already loaded! Speedy! (hint hint)

The-Daniel (approved):

The-Daniel (approved):

PR #3516, opened 2022-08-12
ISSUE #3322 - Upload a revision to a particular container

### LINKED ISSUE #3322: Upload revision to a particular container from the containers list
### Description
In the containers list there are several places from which a a user can upload a revision, but we are only  using one (from the collapsible menu). Also this way of triggering the modal is not specific for a particular container. 

### Proposal
For the particular revision triggers (ellipsis menu and empty revision list) I think it should directly trigger a file selection (if possible) and auto select the container from which the upload dialog was triggered

### Goals
- [ ] As a user I want the system file selection dialog to open when I select "Upload New Revision" so that I can find the local file I want to upload
- [ ] As a user I want to be limited to a single file selection because I have identified a single container to upload to
- [ ] As a user once the file has been selected I want it to populate the "Upload Files" modal
- [ ] As a user I want the "Destination" container to be pre-populated with the target container I selected in the first step
- [ ] As a user I want the remainder of the workflow to be consistent with bulk file upload
- [ ] As a user I want to see the option to add additional files during the upload process so that users that are unfamiliar with bulk upload (because it is a new feature) will see a hint that it is possible

### Tasks
- [ ] Add the props to the upload form so it can receive a container id and just select that from the start
- [ ] When opening the upload form trigger the browse files if there is a default container pass as a prop

### Related Resources
fig 1
image

fig 2
image

fig 3
image

The-Daniel:
This fixes #3322 

#### Description
- Changed method of calling upload revision modal to use redux actions
- The uploads modal can be opening in both the ellipses menu and the revisions description on an empty container
- A file selection window opens up first. Exiting here will cancel the process
- If a file is selected then the upload revisions modal opens with the file and desination container already selected
- You can only pick one file in the initial file window. Afterwards you can add more files and containers in the usual way

#### Test cases
- Upload a file to a specific container
- Upload files to the specific container and extra containers
- You should not be able to add a file to a processing container
- The old method of bulk uploading files should be unaffected


mosment3DRepo (approved):

frontend/src/v5/ui/routes/dashboard/projects/containers/uploadFileForm/uploadFileForm.helpers.ts
```diff
@@ -0,0 +1,33 @@
...
+	f.type = 'file';
+	f.accept = ClientConfig.acceptedFormat.map((format) => `.${format}`).toString();
+	f.onchange = (e: Event) => {
+		const presetFile = (<HTMLInputElement>e.target).files[0];
+		DialogsActionsDispatchers.open(UploadFileForm, {
+			presetFile,
+			presetContainerId,
+		});
+	};
+	f.click();
```
sanmont3drepo, line 32: nice trick!

sanmont3drepo (approved):

PR #3518, opened 2022-08-12
ISSUE #3492 - Rewrite Redux and Selector tests: Auth 

### LINKED ISSUE #3492: Rewrite Redux and Selector tests: Auth
### Description

### Goals

### Tasks
- [ ] Create a new test file for auth

### Related Resources

The-Daniel:
This fixes #3492

#### Description
- Created test file for auth which combines redux action testing and selector testing


frontend/test/auth/auth.store.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+import { authMockFactory } from './auth.fixtures';
+import { selectAuthenticationFetched, selectIsAuthenticated, selectIsPending, selectLoginError, selectReturnUrl } from '@/v5/store/auth/auth.selectors';
+import reducers from '@/v5/store/reducers';
+
+describe('Auth: store', () => {
+	let dispatch, getState = null;
+
+	const mockAuth = authMockFactory();
+
+	beforeEach(() => {
```
The-Daniel, line 29: I'm wondering if we can move this to the test.helpers.ts file

sanmont3drepo: Yes, that would be a good idea,

sanmont3drepo: If you do so, can you please modify the xisting test to use this?

sanmont3drepo (changes requested):

sanmont3drepo (approved):

sanmont3drepo (approved):

PR #3530, opened 2022-08-16
Issue 3494 - Rewrite Redux and Selector tests: Projects

### LINKED ISSUE #3494: Rewrite Redux and Selector tests: Projects
### Description

### Goals

### Tasks
- [ ] Create a new test file for projects

### Related Resources

Amantini1997:
This fixes #3494

#### Description
Rewritten project tests to remove redux unit tests in favour of integration tests


frontend/test/projects/projects.store.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+import { ProjectsActions } from '@/v5/store/projects/projects.redux';
+import { selectCurrentProject, selectCurrentProjectDetails, selectCurrentProjects } from '@/v5/store/projects/projects.selectors';
+import { TeamspacesActions } from '@/v5/store/teamspaces/teamspaces.redux';
+import { times } from 'lodash';
+import { projectMockFactory } from './projects.fixtures';
+import { createTestStore } from '../test.helpers';
+
+const projectListIncludesProject = (projectList, project) => (
+	projectList.map(({ _id }) => _id).includes(project._id)
+);
```
sanmont3drepo, line 27: Maybe make this a helper function for all arrays that have objects with `{_id}` , like `containsId(arr, id)` and re use it in containers.

sanmont3drepo (changes requested):

sanmont3drepo (approved):

sanmont3drepo (approved):

sanmont3drepo (approved):

sanmont3drepo (approved):

The-Daniel (approved):

PR #3531, opened 2022-08-16
Issue 3491- Rewrite Redux and Selector tests: Users

### LINKED ISSUE #3491: Rewrite Redux and Selector tests: Users
Description

### Goals

### Tasks
- [ ] Create a new test file for users

### Related Resources

Amantini1997:
This fixes #3491

#### Description
Rewritten users tests to remove redux unit tests in favour of integration tests


frontend/test/users/users.store.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+		({ dispatch, getState } = createTestStore());
+	});
+
+	it('should set users', () => {
+		const username = 'username';
+		const mockUsers = [
+			userWithoutAvatarMockFactory({ user: username }),
+			userWithAvatarMockFactory({ user: username.toUpperCase() })
+		];
+		dispatch(UsersActions.fetchUsersSuccess(teamspace, mockUsers));
```
sanmont3drepo, line 38: can we add a case without setting the users? how the selectors would react?

sanmont3drepo (changes requested):

sanmont3drepo (approved):

PR #3532, opened 2022-08-16
Issue 3495 - Rewrite Redux and Selector tests: CurrentUser

### LINKED ISSUE #3495: Rewrite Redux and Selector tests: CurrentUser
### Description

### Goals

### Tasks
- [ ] Create a new test file for currentUser

### Related Resources

Amantini1997:
This fixes #3495 

#### Description
Rewritten currentUser user test to replace redux unit tests with integration test


frontend/test/currentUser/currentUser.store.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+import { createTestStore } from '../test.helpers';
+
+describe('CurrentUser: store', () => {
+	let dispatch, getState;
+
+	beforeEach(() => {
+		({ dispatch, getState} = createTestStore());
+	})
+
+	it('should set currentUser', () => {
```
sanmont3drepo, line 35: can we add a case of selecting the current user without setting it ? just to guarantee the selector doesnt break

frontend/test/currentUser/currentUser.store.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+		dispatch(CurrentUserActions.fetchUserSuccess(mockCurrentUser));
+		const currentUser = selectCurrentUser(getState());
+		expect(currentUser).toEqual(mockCurrentUser);
+	})
+
+	describe('Updating currentUser attributes:', () => {
+		it('should update personal data', () => {
+			const mockCurrentUser = currentUserMockFactory();
+			dispatch(CurrentUserActions.fetchUserSuccess(mockCurrentUser));
+			const { firstName: mockFirstName } = generatePersonlData();
```
sanmont3drepo, line 46: `generatePersonlData();` has a typo, should be `generatePersonalData();`

frontend/test/currentUser/currentUser.store.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+			dispatch(CurrentUserActions.updateUserSuccess({ firstName: mockFirstName }));
+			const firstName = selectFirstName(getState());
```
sanmont3drepo, line 48: instead of using `selectFirstName`  can we use `selectCurrentUser(getState())` and test it against `{...mockCurrentUser, firstName:mockFirstName}` so you can test that `updateUserSuccess` doesnt breake anything else

sanmont3drepo (changes requested):

frontend/test/currentUser/currentUser.store.spec.ts
Amantini1997: Sure, but just so you know, I think the selector uses a default value `|| {}` which is never used given the `INITIAL_STATE` set for currentUser
image

currentUser starts as a non-empty object which properties are empty strings

sanmont3drepo: Yea, I know. I just wanted to be sure that the selector wont break

sanmont3drepo (approved):

sanmont3drepo (approved):

sanmont3drepo (approved):

PR #3539, opened 2022-08-17
ISSUE #3489 - Rewrite Redux and Selector tests: Federations

### LINKED ISSUE #3489: Rewrite Redux and Selector tests: Federations
### Description

### Goals

### Tasks
- [ ] Create a new test file for federations

### Related Resources

The-Daniel:
This fixes #3489 

#### Description
- Add store test file which test both redux reducers and selectors
- Fix broken updateFederationSuccess reducer


frontend/test/federations/federations.store.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+		expect(federationFromState.status).toEqual(stats.status);
+		expect(federationFromState.category).toEqual(stats.category);
+	});
+
+	it('should fetch a federation\'s views', () => {
+		const { _id: federationId } = createAndAddFederationToStore();
+		const mockViews = prepareMockViews();
+		dispatch(FederationsActions.fetchFederationViewsSuccess(projectId, federationId, mockViews));
+		const viewsFromState = selectFederationById(getState(), federationId).views;
+		expect(viewsFromState).toBe(mockViews);
```
Amantini1997, line 84: According to the jest docs: `Use .toBe to compare primitive values or to check referential identity`, The tests still pass, but I think it's better to use `toEqual` which does a deep comparison when it comes to objects

Amantini1997, line 84: According to the jest docs: `Use .toBe to compare primitive values or to check referential identity`, The tests still pass, but I think it's better to use `toEqual` which does a deep comparison when it comes to objects

Amantini1997 (changes requested):

frontend/test/federations/federations.store.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+		const mockFederation = federationMockFactory();
+		dispatch(FederationsActions.createFederationSuccess(projectId, prepareMockNewFederation(mockFederation), mockFederation._id));
+		const federationIsIncluded = selectFederationById(getState(), mockFederation._id);
+		expect(federationIsIncluded).toBeTruthy();
+	});
+
+	it('should fetch federations successfully', () => {
+		const mockFederations = times(5, () => federationMockFactory());
+		dispatch(FederationsActions.fetchFederationsSuccess(projectId, mockFederations));
+		const federations = selectFederations(getState());
```
Amantini1997, line 54: Just to let you know, https://github.com/3drepo/3drepo.io/pull/3532#discussion_r948155194
in this PR Santiago mentioned to test this kind of selector before populating the store. I may be worth it to add him to review this PR to see if he wants to request this change here too

sanmont3drepo (approved):

PR #3541, opened 2022-08-18
ISSUE 3527 - Migration script to add vertex counts

### LINKED ISSUE #3527: Create a migration script to add vertex counts to mesh BSONs
### Description
As part of https://github.com/3drepo/3drepobouncer/issues/570, @sebjf has added vertex counts into the mesh bson objects. We need a migration script to add this property to every existing mesh object.

We can calculate this field base on the m_map in the super mesh objects: 
image

each entry in m_map has a v_from and v_to to express the number of vertices + their position in the buffer. so `v_to`-`v_from` will give you the vertex count of the original mesh (with the mesh id being stored in `map_id`)

### NOTE:
- Some mesh may be too big and span across multiple super meshes so the script needs to cater for this (these meshes will NOT group with any other mesh so you can assume it may only possibly happen if m_map.length === 1)
- Branch off #3065 and create a PR to merge into #3065
- Implement the migration script for `v4.30`
- The script must run/conform to the latest migration script suite which is located [here](https://github.com/3drepo/3drepo.io/tree/master/backend/src/scripts/migrations)
- Do not right bespoke functions to do this where possible, use whatever is in the [main v5 codebase](https://github.com/3drepo/3drepo.io/tree/master/backend/src/v5) where necessary (as they are CI tested)

### Goals
A script that will go through all teamspaces and add vertex count length where it is lacking.

Charence:
This fixes #3527 

#### Description
- Add migration script for inserting vertex counts


backend/src/scripts/migrations/4.30/addVertexCountsToMeshBSONs.js
```diff
@@ -0,0 +1,88 @@
...
+		} else if (m_map.length === 1) {
+			const verticesCount = m_map[0].v_to - m_map[0].v_from;
+			spanningMeshVerticesCounts[UUIDToString(m_map[0].map_id)] = (
+				spanningMeshVerticesCounts[UUIDToString(m_map[0].map_id)] || 0
+			) + verticesCount;
+		}
+	});
+
+	// eslint-disable-next-line camelcase
+	meshUpdates = meshUpdates.concat(Object.keys(spanningMeshVerticesCounts).map((map_id) => updateOne(
```
carmenfan, line 58: probably more ES6 style to do this

```suggestion
	const singularMeshUpdates = Object.keys(spanningMeshVerticesCounts.map((map_id) => updateOne(
	// ...
	
	await Promise.all([...meshUpdates, ...singularMeshUpdates]);
```

backend/src/scripts/migrations/4.30/addVertexCountsToMeshBSONs.js
```diff
@@ -0,0 +1,88 @@
...
+const { v5Path } = require('../../../interop');
+
+const { UUIDToString, stringToUUID } = require(`${v5Path}/utils/helper/uuids`);
+const { getTeamspaceList, getCollectionsEndsWith } = require('../../utils');
+
+const { find, updateOne } = require(`${v5Path}/handler/db`);
+const { logger } = require(`${v5Path}/utils/logger`);
+
+const processModel = async (teamspace, model) => {
+	const stashDB = `${model}.stash.3drepo`;
```
carmenfan, line 27: can we add a test here to see if there's any meshes in the `.scene` collection that doesn't have vertex_counts? 

it's quite likely we will have to rerun the script, we'd want to early return and don't do all the ops if we've done it already.

backend/src/scripts/migrations/4.30/addVertexCountsToMeshBSONs.js
```diff
@@ -0,0 +1,88 @@
...
+};
+
+const processTeamspace = async (teamspace) => {
+	const scenes = await getCollectionsEndsWith(teamspace, '.scene');
+	for (const { name: scene } of scenes) {
+		const model = scene.slice(0, -('.scene'.length));
+		logger.logInfo(`\t\t\t${model}`);
+		// eslint-disable-next-line no-await-in-loop
+		await processModel(teamspace, model);
+		// eslint-disable-next-line no-await-in-loop
```
carmenfan, line 75: don't think this line is neccessary

carmenfan (changes requested):

carmenfan (approved):

carmenfan:
Travis is failing but this seems to be to do with the parent branch changes - ok to merge back into parent branch

PR #3558, opened 2022-08-22
ISSUE #3552 - Rewrite redux file to use Immer.js: Revisions

### LINKED ISSUE #3552: Rewrite redux file to use Immer.js: Revisions
### Description
Updating a redux state the standard way is a nuisance. Using Immer.js makes the code cleaner and more intuitive.

### Goals
- [ ] Make update redux states easier and cleaner;
- [ ] Avoid increases in code complexity/length;

### Tasks
- [ ] Use Immer produce-syntax to update redux states
- [ ] Finding a suitable single point modification to avoid using `produce` for every single reducer
- [ ] EXTRA, the first parameter of a reducer is the current state which has a default value `state = INITIAL_STATE`. The default value is not necessary as the state always falls back to that value because of redux-sauce's `createReducers`. Remove that assignment.

Use #3524 as a reference if needed

Amantini1997:
This fixes #3552 

#### Description
Re-write redux file

### DIFF
--- a/frontend/src/v5/store/revisions/revisions.redux.ts
+++ b/frontend/src/v5/store/revisions/revisions.redux.ts
@@ -63,10 +63,9 @@ export const updateRevisionSuccess = (state, {
 	containerId,
 	data,
 }: UpdateRevisionSuccessAction) => {
-	const revision = state.revisionsByContainer[containerId].find(
-		({ _id }) => _id === data._id,
-	);
-	Object.assign(revision, data);
+	const revisions = state.revisionsByContainer[containerId];
+	const index = revisions.findIndex(({ _id }) => _id === data._id);
+	revisions[index] = { ...revisions[index], ...data };
 };
 
 export const setIsPending = (state, { isPending, containerId }: SetIsPendingAction) => {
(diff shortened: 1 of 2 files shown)


frontend/src/v5/store/revisions/revisions.redux.ts
```diff
@@ -40,106 +41,68 @@ export const INITIAL_STATE: IRevisionsState = {
...
-			isComplete,
-			errorMessage,
-		},
-	},
-});
-
-export const revisionProcessingSuccess = (state = INITIAL_STATE, {
+}: SetUploadCompleteAction) => {
+	const uploads = state.revisionsUploadStatus;
+	uploads[uploadId] = { ...(uploads[uploadId] || {}), isComplete, errorMessage };
```
sanmont3drepo, line 81: I tried doing this in the console
````
var uploads = {}
var pp = {...uploads['2323'], isComplete:true}
````
And it didnt cause an error... 🤷‍♂️


```suggestion
	uploads[uploadId] = { ...uploads[uploadId], isComplete, errorMessage };
```

sanmont3drepo (changes requested):

frontend/src/v5/store/revisions/revisions.redux.ts
Amantini1997: uploads may be `undefined` and I thought spreading it caused issues, but it doesn't, so I changed the code as you suggested

sanmont3drepo (approved):

sanmont3drepo (approved):

PR #3586, opened 2022-08-26
ISSUE #3496 - Include icons in storybook

### LINKED ISSUE #3496: Get icons into storybook
### Description
Sometimes for finding icons is a bit of a hassle. Having a searchable sample list in storybook would help to dint them easily

### Goals
- [ ] As a developer/designer I want to be able to find icons easily to know which ones are available and which ones I need to add

### Tasks
- [ ] Research if there is a plugin for storybook
- [ ] In case there is not, or is not applyable to our case design a solution

Amantini1997:
This fixes #3496

#### Description
Storybook now detects and render all the icons under `@assets\icons\`.

Icons are displayed all together (see `allICons`) or according to the folder structure curently used

icons/
├─ controls/
├─ .../
└─ twoToned/ 
image

Based on this, I would recommend to gather all the icons that follow a similar pattern (outlined, filled, etc.) together, so that they will also be displayed together in the `allIcons` story

#### Limitations
Due to the nature of rendering, it is not possible to detect the specific props for each single icon (see the rendering of `controls/checkbox` for reference). Also, the approach used is dynamic in terms of the icons contained in each folder, but it needs to know the folder structure at compile time as that has to be hardcoded.

image
> Checkbox not receiving the border-color prop

#### Extra
In order to simplify the content visibility, the background uses a non-white colour which can be changed from the controls (the same goes for the icons' sizes)
image


sanmont3drepo (dismissed):

frontend/stories/icons/icons.helper.tsx
```diff
@@ -0,0 +1,55 @@
...
+		// eslint-disable-next-line
+		Icon: require(`@assets/icons/${formattedDir}${name.replace(/\.tsx/, '')}`).default,
+		name: getDisplayName(name),
+	}));
+	const svgIcons = svgs.map((name) => ({
+		// eslint-disable-next-line
+		Icon: () => <img src={require(`@assets/icons/${formattedDir}${name}`)} alt={name} />,
+		name: getDisplayName(name),
+	}));
+	return [...tsxIcons, ...svgIcons];
```
The-Daniel, line 54: My only complaint is that the svg icons always come after the tsx icons, so if you're looking for 'helicopter.svg' in storybook you might not think it's not there because it's at the end instead of in alphabetical order. Perhaps a sort could fix this?

The-Daniel (approved):

sanmont3drepo (approved):

PR #3588, opened 2022-08-30
ISSUE #3566 - Improve list performance (Federations, Update Federation, Containers)

### LINKED ISSUE #3566: Memoize list items (improve performance)
### Description
As a user I want a streamline experience and performance always imspact this. Because we have so many lists if we can improve on the performance of these we can deliver a bit of value to the ux
This is also detailed here: #3450 

### Goals
- [ ]  Navigate lists without feeling lag or anything of the sort

### Tasks
- [x] Memoize containers list
- [x] Memoize federations list
- [x] Memoize edit federations (containers list)
- [ ] Memoize upload containers list

### Notes
- Be aware that filtering (including highlighting of the results) should keep on working
- Take in consideration also that ordering by columns should work

The-Daniel:
This fixes #3566 

#### Description
- This improves the render performance of the federations list, update federation's containers list, and containers list.

#### Test cases
- Use react dev tools to view component updates
- Change list sort
- Filter list
- Update a list item (name, code, etc.)


frontend/src/v5/ui/routes/dashboard/projects/federations/federationsList/federationsList.component.tsx
```diff
@@ -126,14 +114,14 @@ export const FederationsList = ({
 				</DashboardListHeader>
 				<DashboardList>
 					{!isEmpty(sortedList) ? (
-						sortedList.map((federation, index) => (
+						sortedList.map((federation, index) => (federation.hasStatsPending ? (
+							<SkeletonListItem delay={index / 10} key={federation._id} />
+						) : (
```
sanmont3drepo, line 119: I'm not against this, but why the change?

sanmont3drepo: Also why moving **onFavouriteChange** ?

sanmont3drepo (approved):

frontend/src/v5/ui/routes/dashboard/projects/federations/federationsList/federationsList.component.tsx
The-Daniel: 1. This way you don't have to pass in 'index' as a prop to FederationListItem. If the index were to change (i.e. by adding or deleting a previous item) then the component would rerender. Also I'm not a fan of conditionally rendering outside of the main return (if that makes sense) because you have to make sure all your hooks occur before this otherwise you get vague errors that always take me a minute to decipher.

2. I could just wrap it in a useCallback() and pass it in as a prop but it seems overcomplex when you can just define the function lower in the DOM where it is actually used

PR #3614, opened 2022-09-08
ISSUE #3592 - Captcha Bug fix 

### LINKED ISSUE #3592: Staging V5 Sign-Up Captcha fail
### Description

Highlighting objects on a federation node that would highlight multiple sub models is not functioning.

### Steps to replicate
- Go to https://staging.dev.3drepo.io/v5/signup
- Fill in all 3 sections of the form
- Clicking "Create Account" generates the following API error:
```
{
	"message": "Invalid captcha",
	"status": 400,
	"code": "INVALID_ARGUMENTS",
	"value": "INVALID_ARGUMENTS",
	"place": "POST /api/v5/user"
}
```

https://user-images.githubusercontent.com/62879582/187468879-89e6529a-5f79-47d2-bf18-8931f8ab74d6.mp4

ChristosTsiotsias:
This fixes #3592

#### Description
Captcha post request now is made with url-encoded params and works properly 

#### Test cases
In v5 (capctha enabled) sign up using the UI
The account should be successfully created

### DIFF
--- a/backend/src/v5/middleware/dataConverter/inputs/users.js
+++ b/backend/src/v5/middleware/dataConverter/inputs/users.js
@@ -216,14 +216,11 @@ const generateSignUpSchema = (isSSO) => {
 
 	return captchaEnabled
 		? schema.test('check-captcha', 'Invalid captcha', async (body) => {
-			const checkCaptcha = post(config.captcha.validateUrl, {
-				secret: config.captcha.secretKey,
-				response: body.captcha,
-			});
-
+			const url = `${config.captcha.validateUrl}?secret=${config.captcha.secretKey}&response=${body.captcha}`;
+			const checkCaptcha = post(url);
 			const result = await checkCaptcha;
 
-			return result.success;
+			return result.data.success;
 		})
 		: schema;
 };
--- a/backend/src/v5/middleware/dataConverter/inputs/users.js
+++ b/backend/src/v5/middleware/dataConverter/inputs/users.js
@@ -216,11 +216,17 @@ const generateSignUpSchema = (isSSO) => {
 
 	return captchaEnabled
 		? schema.test('check-captcha', 'Invalid captcha', async (body) => {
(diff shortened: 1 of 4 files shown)


backend/src/v5/middleware/dataConverter/inputs/users.js
```diff
@@ -216,14 +216,11 @@ const generateSignUpSchema = (isSSO) => {
 
 	return captchaEnabled
 		? schema.test('check-captcha', 'Invalid captcha', async (body) => {
-			const checkCaptcha = post(config.captcha.validateUrl, {
-				secret: config.captcha.secretKey,
-				response: body.captcha,
-			});
-
+			const url = `${config.captcha.validateUrl}?secret=${config.captcha.secretKey}&response=${body.captcha}`;
```
carmenfan, line 219: It would be clearer to do it like before, instead of appending all the parameters in the query string. You can still do this with axios https://axios-http.com/docs/req_config

backend/src/v5/middleware/dataConverter/inputs/users.js
```diff
@@ -216,14 +216,11 @@ const generateSignUpSchema = (isSSO) => {
...
+			const checkCaptcha = post(url);
 			const result = await checkCaptcha;
```
carmenfan, line 221:
```suggestion
			const { data }= await checkCaptcha;
			return data.success
```

carmenfan (changes requested):

backend/src/v5/middleware/dataConverter/inputs/users.js
ChristosTsiotsias: cause then we wont be using the webReuqests class

carmenfan: > cause then we wont be using the webReuqests class

Shoudl still be using WebRequest class,it's just a abstraction to axios?

carmenfan (approved):

PR #3615, opened 2022-09-08
ISSUE #3546 - resolved crashes when static 404's

### LINKED ISSUE #3546: ERR_STREAM_WRITE_AFTER_END - Can't set headers after they are sent.
### Description

If you visit the billing or favicon route you get an error about the headers being modified after they're sent

[https://3drepo.local/favicon.ico](https://3drepo.local/favicon.ico)
[https://3drepo.local/dashboard/billing/subscription](https://3drepo.local/dashboard/billing/subscription)

- This is happening in: both 

### Steps to replicate
- start up a local copy of IO
- repeatedly visit the billing endpoint

### Current Behaviour
- sometimes it will just error with a soft warning
- after a certain amount of visits it will crash node entirely.

### Expected Behaviour

- we shouldn't crash or attempt to modifier headers after send

image

[https://bobbyhadz.com/blog/javascript-error-cannot-set-headers-after-they-are-sent-to-client](https://bobbyhadz.com/blog/javascript-error-cannot-set-headers-after-they-are-sent-to-client)

keifgwinn:
This fixes #3546 

#### Description
- Slightly changed route for static's to ensure that the headers aren't sent early
- Removed a broken image on the /billing/ route and slightly changed the wording
- tweaked favicon route settings

#### Test cases
visit 

1. https://3drepo.local/favicon.ico
2. https://3drepo.local/dashboard/billing/subscription

and review the server log to see no further errors

*** further investigation showed it wasn't the /favicon.ico route itself, but the browser requesting sw.js ( I couldn't trace the reason for this) but the crash was caused by the 404 handling on /sw.js (or any other 404)

### DIFF
--- a/backend/src/v4/services/frontend.js
+++ b/backend/src/v4/services/frontend.js
@@ -59,9 +59,9 @@ module.exports.createApp = function () {
 
 		if (req.path.indexOf(".") !== -1) {
 			next();
+		} else {
+			res.sendFile(path.resolve(publicDir + "/index.html"));
 		}
-		res.sendFile(path.resolve(publicDir + "/index.html"));
-
 	});
 
 	return app;
--- a/backend/src/v4/services/frontend.js
+++ b/backend/src/v4/services/frontend.js
@@ -38,6 +38,8 @@ module.exports.createApp = function () {
 	const configRoute = require("../routes/config");
 	const cors = require("cors");
 
+	app.use(favicon(`${__dirname}/../../../resources/images/favicon.ico`));
+
 	app.use(compression());
 	app.use("/config", configRoute);
 
@@ -47,7 +49,6 @@ module.exports.createApp = function () {
 		extended: true
 	}));
 	app.use(bodyParser.json());
-	app.use(favicon(`${__dirname}/../../../resources/images/favicon.ico`));
 
 	docsService(app);
 
--- a/backend/src/v4/services/frontend.js
+++ b/backend/src/v4/services/frontend.js
@@ -31,19 +31,20 @@ module.exports.createApp = function () {
 	const express = require("express");
 	const compression = require("compression");
 	const bodyParser = require("body-parser");
-	const compress = require("compression");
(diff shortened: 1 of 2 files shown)


backend/src/v4/services/frontend.js
```diff
@@ -31,39 +31,37 @@ module.exports.createApp = function () {
...
 	const favicon = require("serve-favicon");
 	const app = express();
 	const path = require("path");
 	const configRoute = require("../routes/config");
 	const cors = require("cors");
 
-	app.use(compression());
+	app.use(compression({ level: 9 }));
+	const publicDir =  __dirname + "/../../../../public";
+	app.use(favicon(`${publicDir}/assets/images/favicon.ico`));
```
carmenfan, line 42: https://app.travis-ci.com/github/3drepo/3drepo.io/jobs/582251044#L694

This is breaking v5 test because we don't pull favicon from the public directory (and v5 test doesn't build frontend so that favicon doesn't exists)

frontend/src/v4/routes/subscription/components/subscriptionForm.component.tsx
```diff
@@ -355,7 +355,7 @@ export class SubscriptionForm extends PureComponent<IProps, IState> {
 								<FormInfo>** Subject to VAT where applicable</FormInfo>
 							</FormInfoContainer>
 							<PayPalInfoContainer>
-								<ConfirmContainer>
+								{/* <ConfirmContainer>
```
carmenfan, line 358: if you were to comment it out you might as well remove it.

I don't think it ever enables anyway though?

carmenfan (changes requested):

frontend/src/v4/routes/subscription/components/subscriptionForm.component.tsx
keifgwinn: ok, removed.

PR #3616, opened 2022-09-09
Issue 3564 - Update Ticket Endpoint

### LINKED ISSUE #3564: Custom ticket: update a ticket
### Description
We need an endpoint to update a ticket. 

route: `HTTP PATCH /teamspaces/{teamspace}/projects/{project}/federations/{federation}/tickets/{ticket}`
payload:
````js
{
    title: "a new title for this existing ticket",
    properties: {
        propToUpdate: "updated value"
    },
    modules: {
        risk: {
            propToUpdate: 123
        }
    }
}
````
NOTES:
- template type cannot be updated
- read only values cannot be updated
- deprecated fields cannot be updated
- type/data constraints/casts that we impose on creating a ticket should apply here
- only properties provided should be updated, except:
  - `updated at` should be updated everytime someone updates the ticket
  - If `risk consequence` and or `risk likelihood` has been updated, we need to recalculate `level of risk`
  - if `treated risk consequence` and or `treated risk likelihood` has been updated, we need to recalculate `treated level of risk`
- A lot of the existing code in addTicket can be reused here

### Goals
- [ ] Create an endpoint that allows the user to update a ticket

### Tasks

### Related Resources
This issue depends on https://github.com/3drepo/3drepo.io/issues/3429

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #3564

#### Description
Endpoint to update a ticket added
Swagger doc, unit and e2e tests added

#### Test cases
Calling the endpoint with payload that complies to the template should update the ticket

### CODE REVIEW

File: backend/tests/v5/unit/middleware/dataConverter/inputs/teamspaces/settings.test.js [hunk 1]
```diff
@@ -419,8 +423,8 @@ const testCheckTicketTemplateExists = () => {
 			expect(Responder.respond).not.toHaveBeenCalled();
 		});
 
-		test(`Should respond with ${templates.templatesNotFound} if template does not exist`, async () => {
-			TemplateModelSchema.getTemplateById.mockRejectedValueOnce(templates.templatesNotfound);
+		test(`Should respond with ${templates.templateNotFound} if template does not exist`, async () => {
```
carmenfan, line 426: I think it needs to be this?

```suggestion
		test(`Should respond with ${templates.templateNotFound.code} if template does not exist`, async () => {
```

backend/tests/v5/unit/middleware/dataConverter/inputs/teamspaces/settings.test.js
```diff
@@ -433,9 +437,115 @@ const testCheckTicketTemplateExists = () => {
...
+			expect(req.templateData).toBeUndefined();
+			expect(Responder.respond).toHaveBeenCalledWith(
+				req, res, templates.templateNotFound,
+			);
+		});
+	});
+};
+
+const testCheckTicketExists = () => {
+	describe('Check if ticket exists', () => {
```
carmenfan, line 449: as mentioned in the source code this needs to move to the tickets middleware

backend/tests/v5/unit/middleware/dataConverter/inputs/teamspaces/settings.test.js
```diff
@@ -433,9 +437,115 @@ const testCheckTicketTemplateExists = () => {
...
+				{ properties: 1, modules: 1, config: 1 });
+			expect(TicketModelSchema.getTicketById).toHaveBeenCalledWith(req.params.teamspace, req.params.project,
+				req.params.container, req.params.ticket, { type: 1, modules: 1, properties: 1 });
+			expect(next).toHaveBeenCalledTimes(1);
+			expect(req.templateData).toEqual(template);
+			expect(req.ticketData).toEqual(ticket);
+			expect(Responder.respond).not.toHaveBeenCalled();
+		});
+
+		test(`Should respond with ${templates.ticketNotFound} if ticket does not exist`, async () => {
```
carmenfan, line 487: pretty sure we need .code? ticketNotfound would be an object and it'd probably show up as `{object object}`

```suggestion
		test(`Should respond with ${templates.ticketNotFound.code} if ticket does not exist`, async () => {
```

backend/tests/v5/unit/schema/tickets/index.test.js
```diff
@@ -243,43 +244,64 @@ const testValidateTicket = () => {
 	describe('Validate ticket', () => {
 		const propertyTypeTestData = [
 			['Text', { type: propTypes.TEXT }, generateRandomString(), generateRandomString(121)],
+			['Text (unset)', { type: propTypes.TEXT }, null, generateRandomString(121), false],
```
carmenfan, line 247: I think we should be splitting this in a different array and doing them separately

we want these to fail if it's new ticket, but pass when it's updating.

the good test bad test that is imposed this array also doesn't have much meaning.

carmenfan (changes requested):

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/commons/tickets.js
```diff
@@ -54,7 +53,20 @@ const templateIDToParams = async (req, res, next) => {
 	}
 };
 
-TicketsMiddleware.validateNewTicket = validateMany([templateIDToParams, checkTicketTemplateExists, validateNewTicket]);
+const setSafetibaseModValues = (req, res, next) => {
+	const oldTicketSafetibaseMod = req.ticketData.modules?.safetibase;
+	const updatedTicketSafetibaseMod = req.body.modules?.safetibase;
+
+	if (oldTicketSafetibaseMod && updatedTicketSafetibaseMod) {
```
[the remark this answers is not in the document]
ChristosTsiotsias: because if someone want to update just one safetibase prop I want to retrieve all the rest of existing values to calculate levelOfRisk and treatedLevel... otherwise the param received in processReadOnlyValues will have only the prop to be updated

backend/src/v5/schemas/tickets/index.js
```diff
@@ -62,20 +62,23 @@ const generatePropertiesValidator = async (teamspace, properties) => {
 				}
 
 				if (prop.type === propTypes.ONE_OF) {
-					validator = validator.oneOf(values);
+					validator = validator.oneOf(isNewTicket ? values : values.concat(null));
```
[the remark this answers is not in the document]
ChristosTsiotsias: it fails with this message

image

but I could try to change it to the proper one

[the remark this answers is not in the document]
carmenfan: I'm not talking about the error message, the error message is because you don't append `nullable()` further down if it's isn't required. You're appending `null` as an option here even if the field is an required field, which not be accepting null

I think this line should be 
```suggestion
					validator = validator.oneOf(!isNewTicket && prop.required ? values.concat(null): values);
```


Then your error message problem will also naturally resolve itself.

carmenfan (approved):

PR #3655, opened 2022-09-23
Issue 3578 - Update ticket log records 

### LINKED ISSUE #3578: Custom tickets: record change log
### Description
Everytime there is a change on the ticket, we want to add a record to log it.

For e.g. if the following update is applied on a ticket:
````js
{
  properties: {
    xyz: "new val"
  }
}
````

Then we want to store a record in `tickets.logs` that shows:
````
{
  _id: "generate an id", //uuid
  teamspace: "teamspace",
  project: "project", //note this should be UUID not string
  model: "model",
  ticket: "ticket id" , //note this should be UUID not string
  author: "username", // the user who made the change
  from: {
    properties: { xyz: "old val" }
  },
  to: { 
    properties: {xyz: "new val"}
  },
  date: new Date(), //date at which this was updated, this should be identifical to `properties.Updated at` at that point in time.
}
````

#### NOTES:
- notice no model ID on the collection - we're moving towards storing all tickets from all con/fed in the same collection.
- when we remove a model, we need to remove the ticket data from the collection
- it's possible that there are multiple properties/module properties being updated
- No need to add a log for new ticket - only updates

### Goals
- [ ] Add a system log every time a ticket has been updated

### Tasks

### Related Resources
Will depend on #3564

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #3578

#### Description
A log record is added everytime a ticket gets updated. Unit and e2e tests added to test this. 

#### Test cases
Everytime a ticket gets updated a new record should be inserted into the database

### CODE REVIEW

File: backend/tests/v5/e2e/routes/teamspaces/projects/containers/tickets.test.js [hunk 1]
```diff
@@ -510,14 +510,15 @@ const testUpdateTicket = () => {
...
 				delete ticket.properties['Updated at'];
 				delete updatedTicket.properties['Created at'];
 				delete updatedTicket.properties.Owner;
-				expect({ ...updatedTicket }).toEqual({
-					...ticket,
-					...payloadChanges,
-				});
+				expect({ ...updatedTicket }).toEqual({ ...ticket, ...payloadChanges });
+
+				const ticketLog = await findOne(teamspace, 'tickets.logs', { date: new Date(updatedDate) });
```
carmenfan, line 520: maybe we shoudl refactor this out to be a function as you have a copy and pasted code down at L551

backend/tests/v5/unit/models/tickets.logs.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+const testAddTicketLog = () => {
+	describe('Add ticket log', () => {
+		test('Should add a new ticket log', async () => {
+			const teamspace = generateRandomString();
+			const ticketLog = { from: generateRandomString(), to: generateRandomString() };
+			const fn = jest.spyOn(db, 'insertOne').mockResolvedValueOnce(undefined);
+
+			await TicketLogs.addTicketLog(teamspace, ticketLog);
+			ticketLog._id = fn.mock.calls[0][2]._id;
```
carmenfan, line 32: instead of doing this i think it's clearer if you just do `{_id: fn.mock.calls[0][2]._id, ...ticketLog}` down at L34

backend/tests/v5/unit/models/tickets.test.js
```diff
@@ -163,99 +168,113 @@ const testGetAllTickets = () => {
...
-					$unset: {
-						'modules.module.propToUnset': 1,
-						propToUnset: 1,
-						'properties.propToUnset': 1,
+					$set: {
+						'modules.module.propToUpdate': newPropValue,
+						'properties.Updated at': date,
 					},
 				});
+			expect(EventsManager.publish).toHaveBeenCalledTimes(1);
```
carmenfan, line 227: can we add a test to check the publish isn't called if something errored?

just mock the `updateOne` function to throw

carmenfan (changes requested):

backend/src/v5/models/tickets.js
```diff
@@ -35,11 +39,32 @@ Tickets.addTicket = async (teamspace, project, model, ticket) => {
...
 
-Tickets.updateTicket = async (teamspace, ticketId, data) => {
+const formatRecordLogData = (oldTicket, newTicket, updateObj) => {
+	const from = {};
+	const to = cloneDeep(newTicket);
+
+	const date = to.properties['Updated at'];
+
+	// eslint-disable-next-line no-param-reassign
+	delete updateObj['properties.Updated at'];
```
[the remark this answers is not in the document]
ChristosTsiotsias: I thought we wanted to exclude it from the record log, no ?

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.js
```diff
@@ -17,18 +17,17 @@
...
-const { removeFile } = require('../../../../../services/filesManager');
 
 const Tickets = {};
 
 const processBinaryProperties = (template, oldTicket, updatedTicket) => {
 	const toRemove = [];
 	const toAdd = [];
 
-	const updateReferences = (templateProperties, oldProperties = {}, updatedProperties) => {
+	const updateReferences = (templateProperties, oldProperties = {}, updatedProperties = {}) => {
```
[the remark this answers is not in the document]
ChristosTsiotsias: Because of module properties.. the update data may not include a module

image

backend/src/v5/schemas/tickets/index.js
```diff
@@ -114,7 +115,7 @@ Tickets.validateTicket = async (teamspace, template, data, isNewTicket) => {
 	const validator = Yup.object().shape({
 		title: isNewTicket ? types.strings.title.required() : types.strings.title,
 		properties: await generatePropertiesValidator(teamspace, fullTem.properties, isNewTicket),
-		modules: Yup.object(moduleSchema).default({}),
+		modules: Yup.object(moduleSchema).default(isNewTicket ? {} : undefined),
```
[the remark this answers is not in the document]
ChristosTsiotsias: In that case If I just update the title, the updateData in model will end up being 

{ title: '...', properties: { Updated at: '...' }, modules : { } }

If I keep if like this, modules will only be initialised if its a new ticket

backend/src/v5/models/tickets.js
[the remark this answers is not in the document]
carmenfan: Sure, but you're changing a function argument object without making it an output, it's kind of dangerous (hence the linter rule).

There are other ways we could've achieved the same thing.


But this is all N/A if we rewrite the way we create this object, so see how it goes

backend/tests/v5/e2e/routes/teamspaces/projects/containers/tickets.test.js
ChristosTsiotsias: true, do you think I should take out the whole suceed clause ?

carmenfan: if it's the same everywhere sure

carmenfan (changes requested):
- Empty ticket update (i.e. passing in {} as the body) is going through as success with an empty change log (should've through invalid arguments - can we add an e2e as this is a regression):
image

- Can we check that if the field hasn't changed, we remove the property from the update? 
image

carmenfan (approved):

PR #3659, opened 2022-09-23
ISSUE 3205 - Remove incomplete revision script

### LINKED ISSUE #3205: Utility scripts - more clutter clean up
Branching this out from #3139 as it is getting a bit big, and these are less urgent

- [ ] Add new scripts to:
  - [ ] Remove any incomplete (failed, not processing) revisions

### Remove any incomplete (failed, not processing) revisions
Depending on at what point did the model import process failed, we may end up with data in the database/fileshare that is corrupted, not useful (e.g. we imported the scene but it didn't reach unity, or the process timed out when we were uploading bson objects)

We want a script that will go through the scenes and remove any revision data that is associated with revisions that are incomplete and is older than 14 days (i.e. very unlikely to be sitting in the queue still

Charence:
This fixes #3205 

#### Description
Removes all DB entries and files associated with an incomplete model revision

#### Test cases
- Should remove incomplete revisions > a fortnight old
- Should remove activities/sequences from SPM upload


backend/src/scripts/utility/teamspaces/removeIncompleteRevisions.js
```diff
@@ -0,0 +1,186 @@
...
+const processModelSequences = async (teamspace, model, revId) => {
+	const sequences = await find(teamspace, `${model}.sequences`, { rev_id: revId }, { frames: 1 });
+	const sequenceFilenames = [];
+
+	for (const { _id, frames } of sequences) {
+		if (frames) {
+			for (const { state } of Object.values(frames)) {
+				sequenceFilenames.push(state);
+				// eslint-disable-next-line no-await-in-loop
+				await processCollection(teamspace, `${model}.activities`, { sequenceId: _id }, '_extRef');
```
carmenfan, line 119: I don't think this should be in the loop? should just be within L114?

backend/src/scripts/utility/teamspaces/removeIncompleteRevisions.js
```diff
@@ -0,0 +1,186 @@
...
+				// eslint-disable-next-line no-await-in-loop
+				await processFilesAndRefs(teamspace, `${model}.activities.ref`, undefined, { _id: UUIDToString(_id) });
```
carmenfan, line 121: Not that it makes any difference, but isn't this just filename?

I think ti's a bit confusing to have both filename and filters, just have filters?

backend/src/scripts/utility/teamspaces/removeIncompleteRevisions.js
```diff
@@ -0,0 +1,186 @@
...
+		await processFilesAndRefs(teamspace, `${model}.stash.json_mpc.ref`, assets.map((filename) => filename.replace(`/${teamspace}/${model}/`, '').replace('unity3d', 'json.mpc')));
+		// eslint-disable-next-line no-await-in-loop
+		await processFilesAndRefs(teamspace, `${model}.stash.src.ref`, assets.map((filename) => filename.replace(`/${teamspace}/${model}/`, '').replace('unity3d', 'src.mpc')));
+	}
+
+	await processCollection(teamspace, `${model}.stash.unity3d`, { _id: revId }, '_extRef');
+	await processCollection(teamspace, `${model}.stash.3drepo`, { rev_id: revId }, '_extRef');
+};
+
+const processModelSequences = async (teamspace, model, revId) => {
```
carmenfan, line 110: Again, have a think here what needs to be done serially and what can be parallelised. too many unnecessary awaits.

backend/src/scripts/utility/teamspaces/removeIncompleteRevisions.js
```diff
@@ -0,0 +1,186 @@
...
+			timestamp: { $lt: new Date(Date.now() - 1000 * 60 * 60 * 24 * 14) },
+		};
+		// eslint-disable-next-line no-await-in-loop
+		const badRevisions = await find(teamspace, name, incompleteRevisionFilter, { rFile: 1 });
+
+		for (const { _id, rFile } of badRevisions) {
+			const model = name.slice(0, -('.history'.length));
+			logger.logInfo(`\t\t-${model}::${UUIDToString(_id)}`);
+
+			// eslint-disable-next-line no-await-in-loop
```
carmenfan, line 150: Have a think what can be done in parallel, I don't think you need to await for everything like this

carmenfan (changes requested):

backend/src/scripts/utility/teamspaces/removeIncompleteRevisions.js
```diff
@@ -0,0 +1,186 @@
...
+		await processFilesAndRefs(teamspace, `${model}.stash.unity3d.ref`, assets.map((filename) => filename.replace(`/${teamspace}/${model}/`, '')));
+		// eslint-disable-next-line no-await-in-loop
+		await processFilesAndRefs(teamspace, `${model}.stash.json_mpc.ref`, jsonFiles.map((filename) => filename.replace(`/${teamspace}/${model}/`, '')));
+		// eslint-disable-next-line no-await-in-loop
+		await processFilesAndRefs(teamspace, `${model}.stash.json_mpc.ref`, assets.map((filename) => filename.replace(`/${teamspace}/${model}/`, '').replace('unity3d', 'json.mpc')));
+		// eslint-disable-next-line no-await-in-loop
+		await processFilesAndRefs(teamspace, `${model}.stash.src.ref`, assets.map((filename) => filename.replace(`/${teamspace}/${model}/`, '').replace('unity3d', 'src.mpc')));
+	}
+
+	await processCollection(teamspace, `${model}.stash.unity3d`, { _id: revId }, '_extRef');
```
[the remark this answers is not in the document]
Charence: I thought I came across `_extRefs` but I can't seem to find it anymore, so removing it.

backend/src/scripts/utility/teamspaces/removeIncompleteRevisions.js
Charence: Changed to just filters. The `filenames` was there as `GridFS.removeFiles()` takes a list of filenames but I guess `FilesManager.removeFilesWithMeta()` should also handle GridFS.

backend/src/scripts/utility/teamspaces/removeIncompleteRevisions.js
```diff
@@ -0,0 +1,186 @@
...
+const { removeFilesWithMeta } = require(`${v5Path}/services/filesManager`);
+const { UUIDToString } = require(`${v5Path}/utils/helper/uuids`);
+
+const processFilesAndRefs = async (teamspace, collection, filenames = [], filter) => {
+	try {
+		if (filenames.length > 0) {
+			await removeFiles(teamspace, collection, filenames);
+		}
+
+		await removeFilesWithMeta(teamspace, collection, filter || { _id: { $in: filenames } });
```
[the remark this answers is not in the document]
Charence: Changed to just take `filter` and use `removeFilesWithMeta` only.

backend/src/scripts/utility/teamspaces/removeIncompleteRevisions.js
```diff
@@ -0,0 +1,214 @@
...
+	));
+
+	const supermeshIds = (await find(teamspace, `${model}.stash.3drepo`, { rev_id: revId, type: 'mesh' }, { _id: 1 })).map((r) => UUIDToString(r._id));
+	supermeshIds.map((supermeshId) => {
+		modelStashPromises.push(processFilesAndRefs(teamspace, `${model}.stash.3drepo.ref`, { _id: { $in: [
+			`${supermeshId}_faces`,
+			`${supermeshId}_normals`,
+			`${supermeshId}_vertices`,
+		] } }));
+		modelStashPromises.push(processFilesAndRefs(teamspace, `${model}.stash.json_mpc.ref`, { _id: { $in: [
```
carmenfan, line 107: instead of hard coding all these, just regex for `supermeshId*` for all of them

backend/src/scripts/utility/teamspaces/removeIncompleteRevisions.js
```diff
@@ -0,0 +1,214 @@
...
+
+	modelStashPromises.push(removeRecords(teamspace, `${model}.stash.unity3d`, { _id: revId }));
+	modelStashPromises.push(removeRecords(teamspace, `${model}.stash.3drepo`, { rev_id: revId }, '_extRef'));
+
+	await Promise.all(modelStashPromises);
+};
+
+const processModelSequences = async (teamspace, model, revId) => {
+	const sequences = await find(teamspace, `${model}.sequences`, { rev_id: revId }, { frames: 1 });
+	const sequenceFilenames = [];
```
carmenfan, line 129: Think this is better named as stateFileNames

backend/src/scripts/utility/teamspaces/removeIncompleteRevisions.js
```diff
@@ -0,0 +1,214 @@
...
+
+	for (const { _id, frames } of sequences) {
+		if (frames) {
+			for (const { state } of frames) {
+				sequenceFilenames.push(state);
+			}
+		}
+
+		// eslint-disable-next-line no-await-in-loop
+		await Promise.all([
```
carmenfan, line 139: I don't think we need to delete this one sequence at a time 

Just collect all the info and run it in a single query

backend/src/scripts/utility/teamspaces/removeIncompleteRevisions.js
```diff
@@ -0,0 +1,214 @@
...
+
+		for (const { _id, rFile } of badRevisions) {
+			const model = name.slice(0, -('.history'.length));
+			// eslint-disable-next-line no-await-in-loop
+			await removeRevision(teamspace, model, _id);
+
+			// eslint-disable-next-line no-await-in-loop
+			await processFilesAndRefs(teamspace, name, { _id: { $in: rFile } });
+			// eslint-disable-next-line no-await-in-loop
+			await removeRecords(teamspace, name, incompleteRevisionFilter);
```
carmenfan, line 183: I think this is in the wrong place?

As far as I can tell we're removing all the revisions we've found everytime we loop through one revision.

backend/src/scripts/utility/teamspaces/removeIncompleteRevisions.js
```diff
@@ -0,0 +1,214 @@
...
+	]);
+};
+
+const processTeamspace = async (teamspace, revisionAge) => {
+	const cols = await getCollectionsEndsWith(teamspace, '.history');
+
+	for (const { name } of cols) {
+		const incompleteRevisionFilter = {
+			incomplete: { $exists: true },
+			timestamp: { $lt: new Date(new Date().setDate(new Date().getDate() - revisionAge)) },
```
carmenfan, line 170: I don't think you need 3 dates here.... looks a bit mad.

Just instantiate one date and manipulate that

backend/src/v4/handler/gridfs.js
```diff
@@ -20,36 +20,36 @@
 const DB = require("./db");
 const utils = require("../utils");
 
+const cleanColName = (col) => {
```
carmenfan, line 23:
```suggestion
const cleanColName = (col) => col.endsWith(".ref") ? col.slice(0, -4) : col;
```

carmenfan (changes requested):

carmenfan (approved):

carmenfan:
This is ready for merge, pending testing on live databases https://github.com/3drepo/DevOps/issues/493

carmenfan (dismissed):

backend/src/scripts/utility/scene/removeIncompleteRevisions.js
```diff
@@ -191,6 +192,19 @@ const processTeamspace = async (teamspace, revisionAge) => {
 };
 
 const run = async (revisionAge) => {
+	if (revisionAge < 2) {
+		logger.logWarning(`Revision Age is ${revisionAge}; currently processing jobs will be removed when revision age is set to 0.`);
+
+		const rl = readline.createInterface({ input: process.stdin, output: process.stdout });
```
carmenfan, line 198: @keifgwinn does reading stdIn works for you? or you rather have a confirmation via extra command argument?

From what I recall you don't log into the machine to run this

backend/src/scripts/utility/scene/removeIncompleteRevisions.js
```diff
@@ -191,6 +192,19 @@ const processTeamspace = async (teamspace, revisionAge) => {
...
+		const response = await new Promise((resolve) => rl.question('Continue? (y/N): ', (ans) => {
+			rl.close();
+			resolve(ans);
+		}));
+		if (!response.match(/^y/i)) {
+			process.abort();
```
carmenfan, line 204: can't we just return here?

it's a bit drastic to use `process`

carmenfan (changes requested):

backend/src/scripts/utility/scene/removeIncompleteRevisions.js
carmenfan: as discussed, add a flag so skip this if it's in place

backend/src/scripts/utility/scene/removeIncompleteRevisions.js
carmenfan: wouldn't have thought linter would complain about return unless you're trying to return with a value, which rule did it break?

backend/src/scripts/utility/scene/removeIncompleteRevisions.js
Charence: @carmenfan Added. I've put it inside the `revisionAge < 2` condition so that the warning will still be logged.

backend/src/scripts/utility/scene/removeIncompleteRevisions.js
Charence: @carmenfan It doesn't break anymore :scream:

carmenfan (approved):

keifgwinn (approved):
worked well

PR #3663, opened 2022-09-26
ISSUE #3662 - Create Github action to auto add issues to the Sprint Project

### LINKED ISSUE #3662: Create Github action to auto add issues to the Sprint Project
### Description

In order to make it easier to manage issues we should auto assign all issues to the Sprint board to allow them to be triaged

### Related Resources
https://github.com/marketplace/actions/assign-to-one-project

mosment3DRepo:
This fixes #3662


.github/workflows/AutoAssignToSprint.yml
```diff
@@ -0,0 +1,18 @@
...
+    types: [opened]
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  assign_one_project:
+    runs-on: ubuntu-latest
+    name: Assign to One Project
+    steps:
+    - name: Assign NEW issues to sprint proect
```
carmenfan, line 14: typo

.github/workflows/AutoAssignToSprint.yml
```diff
@@ -0,0 +1,18 @@
+name: Auto Assign to Sprint Project
+
+on:
+  issues:
```
carmenfan, line 4: will be good to trigger it on PRs

.github/workflows/AutoAssignToSprint.yml
```diff
@@ -0,0 +1,18 @@
...
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  assign_one_project:
+    runs-on: ubuntu-latest
+    name: Assign to One Project
+    steps:
+    - name: Assign NEW issues to sprint proect
+      uses: srggrs/assign-one-project-github-action@1.2.1
+      if: github.event.action == 'opened'
```
carmenfan, line 16: do you need this? 

I think this always evaluates to true. on L5 you already specified the trigger is that an issue is opened to be the action. 

I think the example you followed only did it because it was also being triggered when someone added a label

carmenfan (changes requested):

.github/workflows/AutoAssignToSprint.yml
carmenfan: Also it'll be good to rename this to  `OnNewIssueOrPR` as that's the trigger, we can add more jobs that's irrelevant to assigning projects in the future in the same script

.github/workflows/AutoAssignToSprint.yml
mosment3DRepo: just the if statement on L16? I assume we still need the 'uses' on L15?

carmenfan: yep

carmenfan:
@mosment3DRepo Have you tried this script?
it doesn't work for me: 
image


Looking at the github issues on the repo, the API doesn't support v2 github projects: https://github.com/srggrs/assign-one-project-github-action/issues/89 

The issue creator did add a link to suggest how to make it work (or maybe there's another github action script?)

carmenfan (changes requested):
see comment above

mosment3DRepo:
> @mosment3DRepo Have you tried this script? it doesn't work for me: image
> 
> Looking at the github issues on the repo, the API doesn't support v2 github projects: [srggrs/assign-one-project-github-action#89](https://github.com/srggrs/assign-one-project-github-action/issues/89)
> 
> The issue creator did add a link to suggest how to make it work (or maybe there's another github action script?)

just realised I was testing it on a v1 project on my personal account

it looks like this one might work https://github.com/marketplace/actions/add-to-github-projects

carmenfan (approved):
This is working. We need to setup an organisation secret to run this.

Assigning the PR to @keifgwinn 

We need a secret with `repo` and `project` privileges with the name `PROJ_MANAGEMENT_TOKEN`, accessible by all repos and it should just work

Can merge once the secret is in place

PR #3684, opened 2022-10-11
Issue 3683 - Streaming Milestone 2

sebjf:
#### Description
This PR is created so we have a deployment to test streaming milestone 2 features.


carmenfan:
> Hi @carmenfan, I am still investigating the loading bottleneck but my latest tests suggest strongly its inside the viewer, so it should be safe to send this branch to the frontend team.
> 
> Things that are hacky:
> 
> 1. `webpack.common.config.js` has been given a new entry which causes it to output a new file, `indexeddbworker.js` to `dist`. The hacky bit is that the '[chunkname]' has been removed from the output filename/the filename is not 'properly' controlled. `unity-indexedbcache.ts` just relies on it being in `dist`.
> 2. `unity-indexeddb-worker.ts` is meant to be fed into a `WebWorker`, which has a different [scope](https://developer.mozilla.org/en-US/docs/Web/API/DedicatedWorkerGlobalScope) to regular scripts. Right now TypeScript is not aware of this and shows a bunch of errors in the IDE (though it does run).
> 3. I have not touched anything that builds the standalone `unity-util.js` for customers embedding the viewer. These customers/consumers of `unity-util.js` will need the worker script too. Not sure if they'll be able to resolve it as it is...
> 
> (I don't think the actual implementation of the indexeddb cache class and the worker it runs is hacky, but this isn't my area so someone should make sure it isn't too!)

@sanmont3drepo

frontend/src/globals/unity-indexedbcache.ts
```diff
@@ -0,0 +1,177 @@
...
+	 * class gets its own worker. */
+	worker: any;
+
+	// IndexedDb actions are offloaded to a WebWorker. This is for performance
+	// reasons. WebWorkers run individual scripts in their own thread, and
+	// interact with the main thread (of their creator) by posting messages.
+	// When messages are passed, a deep copy is created.
+
+	createWorker() {
+		this.worker = new Worker(new URL('dist/indexeddbworker.js', window.location.origin));
```
carmenfan, line 57: could we put this in the unity folder instead of dist?

frontend/src/globals/unity-indexeddb-worker.ts
```diff
@@ -0,0 +1,359 @@
...
+		request.onsuccess = (event: any) => {
+			this.db = event.target.result;
+
+			// Populate our index
+			const transaction = this.db.transaction(this.objectStoreName, 'readonly');
+			const objectStore = transaction.objectStore(this.objectStoreName);
+			const req = objectStore.getAllKeys();
+			req.onsuccess = () => {
+				this.index = {};
+				req.result.forEach((x) => this.index[x] = 1); // Give any value here - we will only check the existence of the key (property)
```
carmenfan, line 96: Linter complained: 
```suggestion
				req.result.forEach((x) => { this.index[x] = 1; }); // Give any value here - we will only check the existence of the key (property)
```

frontend/src/globals/unity-indexeddb-worker.ts
```diff
@@ -0,0 +1,359 @@
...
+	 * store(s). This is initialised when the database is opened and updated
+	 * when set requests are made. It is used to perform early rejects on get
+	 * requests we know will fail (because there is no key in the store),
+	 * avoiding expensive database transactions. */
+	index: any;
+
+	createIndexedDbCache() {
+		const request = indexedDB.open('3DRepoCacheDb', 1);
+
+		request.onerror = () => {
```
carmenfan, line 74: could we make these camel case or is the naming convention imposed on us by something else?

frontend/src/globals/unity-indexeddb-worker.ts
```diff
@@ -0,0 +1,359 @@
...
+		} else {
+			// We believe the key is in the database, so go ahead and request it
+			try {
+				const readTransaction = this.db.transaction(this.objectStoreName, 'readonly');
+				const objectStore = readTransaction.objectStore(this.objectStoreName);
+				const request = objectStore.get(key);
+
+				request.onsuccess = () => {
+					if (request.result === undefined) {
+						this.sendGetTransactionComplete({ // Where a key is not found, we want to handle this outside the error chain
```
carmenfan, line 148: so it's possible for us to think the key is in the db but it was not there after all?

frontend/src/globals/unity-indexeddb-worker.ts
```diff
@@ -0,0 +1,359 @@
...
+				readTransaction.oncomplete = () => {
+					this.numReadTransactions--;
+					this.commitMemoryCache();
+				};
+
+				readTransaction.onerror = () => {
+					this.numReadTransactions--;
+				};
+
+				readTransaction.onabort = () => {
```
carmenfan, line 193:
```suggestion
				readTransaction.onabort =readTransaction.onerror;
```

frontend/src/globals/unity-indexeddb-worker.ts
```diff
@@ -0,0 +1,359 @@
...
+
+			// eslint-disable-next-line guard-for-in
+			for (const key of keys) {
+				const record = this.memoryCache[key];
+				const request = objectStore.put(record, key);
+				request.onsuccess = () => {
+					delete this.memoryCache[key]; // Delete the local copy
+					this.index[key] = 1; // Update the index as we go
+				};
+				if (numRequests++ > 5) { // Don't try and commit too much in one transaction otherwise we may block users' get requests
```
carmenfan, line 281: it looks like every iteration would create a request? wouldn't it be better to just loop through `keys.slice(0, 5)` instead of doing a break?

then you wouldn't need to declare numRequests either

frontend/src/globals/unity-indexeddb-worker.ts
```diff
@@ -0,0 +1,359 @@
...
+			transaction.oncomplete = () => {
+				this.committing = false;
+				this.commitMemoryCache(); // Try to chain additional commits. If there a pending reads the chain will be broken at the start of commitMemoryCache().
+			};
+
+			transaction.onerror = () => {
+				this.committing = false;
+			};
+
+			transaction.onabort = () => {
```
carmenfan, line 295:
```suggestion
			transaction.onabort = transaction.onerror
```

frontend/src/globals/unity-indexeddb-worker.ts
```diff
@@ -0,0 +1,359 @@
...
+	}
+}
+
+onmessage = (e) => {
+	if (e.data.message === 'createIndexedDb') {
+		self.repoCache = new IndexedDbCacheWorker();
+	}
+
+	if (e.data.message === 'Set') {
+		const { key } = e.data;
```
carmenfan, line 349:
```suggestion
		const { key, record } = e.data;
```

frontend/src/globals/unity-indexeddb-worker.ts
```diff
@@ -0,0 +1,359 @@
...
+		const { record } = e.data;
+		self.repoCache.memoryCache[key] = record; // Will be written later by commitMemoryCache
+	}
+
+	if (e.data.message === 'Get') {
+		const { key } = e.data;
```
carmenfan, line 355:
```suggestion
		const { key, id } = e.data;
```

frontend/src/globals/unity-util.ts
```diff
@@ -16,10 +16,14 @@
  */
 
 /* eslint-disable no-underscore-dangle */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
carmenfan, line 19: 🤔  bit confused with this one, your next line is a comment?

frontend/src/globals/unity-util.ts
```diff
@@ -16,10 +16,14 @@
...
+/* eslint-enable no-var */
```
carmenfan, line 20: I would have thought no-var is already enabled as standard? @sanmont3drepo

carmenfan (changes requested):

frontend/src/globals/unity-indexeddb-worker.ts
sebjf: Yes! It should be incredibly rare, but it is possible... 
One example is if the user clears the database out from under the tab, either manually or because an onVersionChanged event rebuilds it. When the database is reopened the dictionary should be rebuilt, but anything that happens between then could use out of date data.
Another possibility is that the commit transaction fails after one or more individual set operation callbacks complete, which will roll back all the changes in that transaction.

frontend/src/globals/unity-indexeddb-worker.ts
sebjf: Updated!

frontend/src/globals/unity-indexeddb-worker.ts
sebjf: Updated!

frontend/src/globals/unity-indexeddb-worker.ts
sebjf: Updated!

frontend/src/globals/unity-util.ts
sebjf: Me as well I think that may have been a merge artefact!

frontend/src/globals/unity-indexeddb-worker.ts
sebjf: Done!

frontend/src/globals/unity-indexeddb-worker.ts
sebjf: Done!

frontend/src/globals/unity-indexeddb-worker.ts
sebjf: Yes, that is much neater.

frontend/src/globals/unity-indexeddb-worker.ts
sebjf: It is the `IDBOpenDBRequest` [API](https://developer.mozilla.org/en-US/docs/Web/API/IDBOpenDBRequest) unfortunately!

frontend/src/globals/unity-indexedbcache.ts
sebjf: Updated! I added a new member to UnityUtil (I also finally understand WebPack enough now to move the folder!)

carmenfan (approved):

PR #3686, opened 2022-10-12
ISSUE #3579 - Tickets resources taken into consideration in quota info

### LINKED ISSUE #3579: Custom ticket: quota calculation update
### Description
We've added a new `tickets.resources.ref` that stores blob data for tickets, this needs to be added into consideration when we calculate the quota

### Goals
- [ ] Ensure v4 quota calculation considers data in custom tickets
- [ ] Ensure v5 quota calculation considers data in custom tickets

### Tasks
- change v4/v5 quota calculation to consider resources in custom tickets
- add extra tests where necessary

### Related Resources
Depends on #3429

ChristosTsiotsias:
This fixes #3579

#### Description

Ticket resources are now taked into consideration in quota endpoints in v4 and v5
v4 endpoint now uses v5 method to return the quota

#### Test cases
Calculate spaced used for a user without any ticket resources
Add some ticket resource
Calculate space used again
It should return an increased space used

### DIFF
--- a/backend/src/v5/utils/quota.js
+++ b/backend/src/v5/utils/quota.js
@@ -59,7 +59,7 @@ Quota.getQuotaInfo = async (teamspace) => {
 };
 
 Quota.getSpaceUsed = async (teamspace) => {
-	const colsToCount = ['.history.ref', '.issues.ref', '.risks.ref', '.resources.ref'];
+	const colsToCount = ['.history.ref', '.issues.ref', '.risks.ref', '.resources.ref', 'tickets.resources.ref'];
 	const collections = await DBHandler.listCollections(teamspace);
 	const promises = [];
 	collections.forEach(({ name }) => {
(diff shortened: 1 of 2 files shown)


backend/src/v5/utils/quota.js
```diff
@@ -59,7 +59,7 @@ Quota.getQuotaInfo = async (teamspace) => {
 };
 
 Quota.getSpaceUsed = async (teamspace) => {
-	const colsToCount = ['.history.ref', '.issues.ref', '.risks.ref', '.resources.ref'];
+	const colsToCount = ['.history.ref', '.issues.ref', '.risks.ref', '.resources.ref', 'tickets.resources.ref'];
```
carmenfan, line 62: judging by this logic, it was already using taking into account tickets usage? Didn't seem like the change was necessary afterall

carmenfan (changes requested):

backend/src/v5/utils/quota.js
ChristosTsiotsias: ah true

carmenfan (dismissed):

carmenfan:
@ChristosTsiotsias  it failed linter 
image

carmenfan (approved):

PR #3689, opened 2022-10-13
ISSUE #3665 - Clipping planes added in view prop of Custom Tickets

### LINKED ISSUE #3665: Custom tickets - views are missing clipping planes
### Description
Validator in the backend is missing clipping planes https://github.com/3drepo/3drepo.io/blob/staging/backend/src/v5/schemas/tickets/validators.js#L49

### Steps to replicate
N/A

### Current Behaviour
one cannot save a view with clipping planes inside a custom ticket

### Expected Behaviour
clipping planes should be stored (same format as v4, inside `camera`

ChristosTsiotsias:
This fixes #3665

#### Description
A view property of a custom ticket can now have clipping plane

#### Test cases
Create a ticket with a view prop and add clippingPlanes property.
The ticket should be created and the clipping planes should be stored in the db

### DIFF
--- a/backend/src/v5/schemas/tickets/validators.js
+++ b/backend/src/v5/schemas/tickets/validators.js
@@ -53,14 +53,14 @@ Validators.propTypesToValidator = {
 			forward: types.position.required(),
 			up: types.position.required(),
 			size: Yup.number().when('type', (type, schema) => (type === CameraType.ORTHOGRAPHIC ? schema.required() : schema.strip())),
-			clippingPlanes: Yup.array().of(
-				Yup.object().shape({
-					normal: types.position.required(),
-					distance: Yup.number().required(),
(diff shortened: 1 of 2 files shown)


backend/src/v5/schemas/tickets/validators.js
```diff
@@ -53,6 +53,13 @@ Validators.propTypesToValidator = {
 			forward: types.position.required(),
 			up: types.position.required(),
 			size: Yup.number().when('type', (type, schema) => (type === CameraType.ORTHOGRAPHIC ? schema.required() : schema.strip())),
+			clippingPlanes: Yup.array().of(
```
carmenfan, line 56: Can we move this out of the camera

so it's at the same level as `screenshot`, `state` etc

carmenfan (changes requested):

carmenfan (approved):

Issue #3710, opened 2022-10-26
Custom tickets - Tickets details - Pin modules functional


sanmont3drepo:
### Description
As a user of tickets I want to be able to associate a ticket with a pin and see that reflected in the viewer.

### Goals
- [ ] As a Collaborator+ I can add a pin when there is no pin currently uploaded
- [ ] As a Collaborator+ I can change the position of a pin which has already been placed
- [ ] As a Collaborator+ once I click "Add Pin" or "Change Pin Location" I will enter a _pin placement mode_. This will turn the stroke of the Pin module teal and enable the position selection tool in Unity. Once I select a position in the model it will update the location of the pin and exit _pin placement mode_.
- [ ]  As a Collaborator+ if I return to the list of tickets when I am in _pin placement mode_ it will cancel out of the mode and return back to the default model behaviour
- [ ]  As a Collaborator+ if I use the navigation arrows inside the ticket to navigate to a different ticket when I am in _pin placement mode_ it will cancel out of the mode and return back to the default model behaviour
- [ ]  As a Collaborator+ if I click the Custom Tickets bubble on the left of the page to close the Tickets card when I am in _pin placement mode_ it will cancel out of the mode and return back to the default model behaviour
- [ ]  As a Collaborator+ if I click the "X" button inside the module next to "Select new location on model" when I am in _pin placement mode_ it will cancel out of the mode and return back to the default model behaviour
- [ ] As a Collaborator+ if I select "Delete Pin" I see a confirmation pop-up to confirm my choice

### Tasks
- [ ] Render the filtered pins 
- [ ] When editing a ticket, only show that pin
- [ ] Add the add a pin feature
- [ ] Add delete a pin
- [ ] Add the change pin feature

### Related Resources
[MockUp](https://www.figma.com/file/3FjNEOdQbcHwzquVZxUqMl/3D-Repo?node-id=646%3A6011)

sanmont3drepo:
Merged in main branch

PR #3712, opened 2022-10-26
ISSUE #3630 - Custom tickets - Tickets details Pin modules component

### LINKED ISSUE #3630: Custom tickets - Tickets details - Pin modules component
### Description
We need a component to have the pin property represented. This issue is for the visual component.

### Goals
- [ ] As a developer I want to have a component to use to represent pin property and edition 

### Tasks
- [ ] Create the empty state (if the pin prop is empty)
- [ ] Create the add pin state (if the pin prop is empty and the internal state is in edit mode)
- [ ] Create the change pin state (if the pin prop is not empty)
- [ ] Create the changing pin state (if the pin prop is not empty and the internal state is in edit mode)

### Related Resources
[Mockup](https://www.figma.com/file/3FjNEOdQbcHwzquVZxUqMl/3D-Repo?node-id=646%3A6011)

The-Daniel:
This fixes #3630 

#### Description
- adds component for adding pins to a ticket
NOTE: None of the logic for adding an actual pin is done here (that is done in #3710). As such there are some temporary hacks here that will be done in the functionality issue.

#### Test cases
- Open a ticket's details and fiddle with the pin component


frontend/src/v5/ui/components/viewer/cards/tickets/pinDetails/pinDetails.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+import DeleteIcon from '@assets/icons/delete.svg';
+import { FormattedMessage } from 'react-intl';
+import { PinAction, PinActions, PinContainer, PinName, SettingLocationText } from './pinDetails.styles';
+
+export const PinDetails = () => {
+	const [isEditMode, setIsEditMode] = useState(false);
+	const [hasPin, setHasPin] = useState(false); // This can be removed when pin logic added
+
+	const onClickCancelSetPin = () => setIsEditMode(false);
+	const onClickDelete = () => setHasPin(false);
```
Amantini1997, line 31: I think this should also set the edit mode to false, in fact, if you add and then delete a pin, it will still be given as selected

Amantini1997 (changes requested):

frontend/src/v5/ui/components/viewer/cards/tickets/pinDetails/pinDetails.component.tsx
The-Daniel: This stuff is temporary until the functionality for pins is added. I just did it in a way where all states could be achieved. You're right that deleting a pin should disable edit mode, but it'll probably get overwritten any way when the pin logic is added so I think it's best to leave it for now.

sanmont3drepo:
@The-Daniel can you add a storybook story for George to see? I believe you need the viewer theme, ask @Amantini1997 how to add that to a story.

sanmont3drepo:
I think is missing this case, when there's no pin and you click "add pin" for the first time:

image

The-Daniel:
> I think is missing this case, when there's no pin and you click "add pin" for the first time:
> 
> image

Yes, so the states are currently a bit of a hack using react useState so that the different states can be accessed whilst the pin logic is undeveloped.

If I'm making the storybook component of this I can set the states there and remove this hasPin hack.

george-hadfield (approved):

PR #3713, opened 2022-10-27
Issue 3682 - New Ticket

### LINKED ISSUE #3682: Custom tickets - New ticket - Functional
### Description
We need to a way to create new tickets . This first approach (without everything in the design) should provide a way to select a template, create a ticket form with that and be able to save it 

### Goals
- [ ] Create a custom ticket

### Tasks
- [x] Add a select box with the list of templates in the ticket list
- [x] Add a new ticket button in the ticket list
- [x] Create a new ticket card that will render a template form based on the template previously selected in the ticket's list
- [ ] Create the schema based on the properties definition and the template modules
- [ ] Add a save button in the create new ticket card, when the form is valid, save the ticket

### Notes
- Do not implement the ui design as it is in the link this is a guide of the flow only 
- For an example of the validator of a property you can take a look at this 
https://github.com/3drepo/3drepo.io/blob/ISSUE_3657/frontend/src/v5/store/tickets/tickets.helpers.ts#L27

### Related Resources
[MockUp](https://www.figma.com/file/3FjNEOdQbcHwzquVZxUqMl/3D-Repo?node-id=789%3A5282)

Amantini1997:
This fixes #3682 

#### Description
It is now possible to create a new ticket using the existing templates.
The template will be rendered in a form and the default values should be populated.
The fields are validated onChange and trigger standard errors. 
As the whole form becomes valid, it is possible to save (create) the ticket.
Once the ticket is successfully created, the user should be redirected to the edit ticket view.

#### Test cases
To use default values, follow the example below in the creation of a template (I have provided the date and the coords, which are the 2 most interesting cases perhaps)
```json
{
	"name": "Default date and Coords",
	"code": "HLW",
	"config": {
	  "comments": true,
	  "issueProperties": true,
	  "defaultView": true,
	  "defaultImage": true,
	  "pin": true
	},
	"deprecated": false,
	"properties": [
	  {
		"name": "Coords",
		"type": "coords",
		"deprecated": false,
		"required": true,
		"default": [1,0,-1]
	  },
	  {
		"name": "Halloween date",
		"type": "date",
		"deprecated": false,
		"required": true,
		"default": "Mon Oct 31 2022 22:00:00 GMT+0000 (Greenwich Mean Time)"
	  }
	]
  }
```


frontend/src/v5/ui/routes/viewer/tickets/newTicket/newTicket.component.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+	const goToTicketDetails = (ticketId) => {
+		contextValue.setCardView(TicketsCardViews.Details, { ticketId });
+	};
+
+	const changeTemplate = () => {
+		contextValue.setCardView(TicketsCardViews.Templates);
+	};
+
+	const getBaseTicket = (): EditableTicket => {
```
sanmont3drepo, line 93: This one's wrong it equating ticket with template.
I see two ways of doing this:

1 . I would take out the "defaultvalue" prop of the properties . And create a base ticket using the default value information from the template.

2. If Im not mistaken (I guess the defaultValue is set if the value is undefined) just pass an empty ticket and default value would take care of the rest? 

Also, move this one to the tickets helper

frontend/src/v5/ui/routes/viewer/tickets/newTicket/newTicket.component.tsx
```diff
@@ -0,0 +1,165 @@
...
+	const templateId = contextValue.props.template._id;
+	const { teamspace, project, containerOrFederation } = useParams<ViewerParams>();
+	const isFederation = modelIsFederation(containerOrFederation);
+	const template = TicketsHooksSelectors.selectTemplateById(containerOrFederation, templateId);
+	const isLoading = !('config' in template);
+
+	const getEditableProperties = () => {
+		const NON_EDITABLE_PROPERTIES = ['Owner', 'Created at', 'Updated at'];
+		return template.properties?.filter(({ name }) => !NON_EDITABLE_PROPERTIES.includes(name));
+	};
```
sanmont3drepo, line 50: Filter out the properties by "readOnly" instead of enumerating tthese.

sanmont3drepo: Also do the same thing with the modules.

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/manyOfProperty.component.tsx
```diff
@@ -17,18 +17,17 @@
...
 export const ManyOfProperty = ({
 	property: { name, readOnly, required, values },
-	value: defaultValues,
-}: { property: PropertyDefinition, value: any }) => {
-	const selectedValues = (defaultValues || []) as string[];
-
+	defaultValue = [],
+	...props
+}: PropertyProps) => {
 	if (values === 'jobsAndUsers') {
```
sanmont3drepo, line 31: regarding this. I would create a basic Many off (which would use values as values. And a JobsAndUsers select. This component would return one or the other

sanmont3drepo (changes requested):

frontend/src/v5/store/tickets/tickets.helpers.ts
```diff
@@ -15,8 +15,135 @@
...
+
+const parseModule = (module) => {
+	const parsedModule = {};
+	Object.entries(module)
+		// skip nullish values that are not 0s or false
+		.filter((entry) => ![null, undefined, ''].includes(entry[1] as any))
+		.forEach(([key, value]) => {
+			if (Array.isArray(value)) {
+				if (value.length === 0) return;
+				// manyOf should not allow empty or nullish values, but coords may.
```
The-Daniel, line 116: I could be wrong here, but I think manyOf should allow empty values? With the assignees the value can be an empty array to show that it is part of the template but no assignees have been assigned yet. In this case the assignees list component will display 'Unassigned' , whereas if the assignees doesn't exist on the properties it will not be rendered.

A similar thing also happens for the dueDate, where it can be null and the ticket list item will show that the ticket can have a due date but it's jut unset.

@sanmont3drepo do you have any thoughts on this? Alternatively the ticket list item could check if the ticket's template includes assignees/dueDate but they're missing in the ticket, and then render the empty states

carmenfan: >  but I think manyOf should allow empty values?

Yes it should allow empty values unless required is set to true

carmenfan: in fact `coords` does not allow empty or nullish values so that comment is very wrong 😆 

For coords you either don't have it (again, if required is set to true), or it has to be [x, y, z]

carmenfan: See [here](https://github.com/3drepo/3drepo.io/blob/staging/backend/src/v5/schemas/tickets/validators.js#L89), Coords is of `type.position` which is of[ type array of numbers with length 3](https://github.com/3drepo/3drepo.io/blob/staging/backend/src/v5/utils/helper/yup.js#L70)

Amantini1997: I now understand that my comment is not as clear as I thought. I'll rephrase it here and then fix the comment in the code.

What I mean by that is that manyOf takes a value, which may indeed be `undefined` (and if so, it is filter right before the forEach)
```ts
Object.entries(module)
	// skip nullish values that are not 0s or false
	.filter((entry) => ![null, undefined, ''].includes(entry[1] as any)) // <- This line here
	.forEach(([key, value]) => {
```
or an empty array (which would also not be a problem)
```ts
if (Array.isArray(value)) {
	if (value.length === 0) return;  // <- This line here
```
However, when its value is a non empty array, the elements of that array should each represent the value of a selected option from the dropdown menu. Those very values should not be undefined

frontend/src/v5/store/tickets/tickets.helpers.ts
```diff
@@ -15,8 +15,135 @@
...
+import { trimmedString } from '@/v5/validation/shared/validators';
+import { isEmpty, isUndefined } from 'lodash';
+import * as Yup from 'yup';
+import { PropertyDefinition } from './tickets.types';
 
 export const modelIsFederation = (modelId: string) => (
 	!!FederationsHooksSelectors.selectContainersByFederationId(modelId).length
 );
+
+const MAX_TEXT_LENGTH = 50;
```
carmenfan, line 29: Not sure where you got these numbers from?

In the backend,[ TEXT and LONG_TEXT](https://github.com/3drepo/3drepo.io/blob/staging/backend/src/v5/schemas/tickets/validators.js#L33) are of type `title` and `longDescription` [which are 120 and 1200 respectively.](https://github.com/3drepo/3drepo.io/blob/staging/backend/src/v5/utils/helper/yup.js#L70)

Amantini1997: I don't remember exactly why I put that to TEXT be have a maxLenght of 50, but I got the title length from the figma design 
image

Amantini1997: I got the title length from the figma design 
image, but since it can span over multiple lines, I assumed that the title was a longText type, hence, the 120

As for the 50 chars length, I used this [validator](https://github.com/3drepo/3drepo.io/blob/ISSUE_3657/frontend/src/v5/store/tickets/tickets.helpers.ts#L27) without realising that was not the actual length and I just kept it

carmenfan: title is of type strings.title https://github.com/3drepo/3drepo.io/blob/staging/backend/src/v5/schemas/tickets/index.js#L138

So if anything it falls in the `TEXT` type, nothing to do with  `LONG_TEXT`

carmenfan: backend validators are always made under consultation with the product team which is more probably reliable the source of truth than anything else

if there's any questions/doubts, just look it up on the code or ask me or christos!

frontend/src/v5/store/tickets/tickets.helpers.ts
```diff
@@ -15,8 +15,206 @@
...
 );
+
+export const getTicketDefaultValues = ({ title, properties, modules }: ITicket) => ({
+	title,
+	properties,
+	...modules,
+});
+
+export const filterNonEditablePropertiesFromTemplate = (template) => {
+	const propertyIsEditable = ({ readOnly }) => !readOnly;
```
sanmont3drepo, line 37: aren't you missing the properties a template level?

sanmont3drepo: Lets simplify the name a bit   getEditableProperties

frontend/src/v5/store/tickets/tickets.helpers.ts
```diff
@@ -15,8 +15,206 @@
...
+	properties.reduce(
+		(ticketProperties, prop) => ({
+			...ticketProperties,
+			[prop.name]: prop.default,
+		}),
+		{},
+	)
+);
+
+export const getEditableTicketFromTemplate = (template: ITemplate): EditableTicket => {
```
sanmont3drepo, line 59: No need to specify "fromTemplate", template  is alrady in the the argument. Would change the naming still getDefaultTicket(template). To me, that tells me that its a ticket (based on the template as it is an argument) that has the default properties.

frontend/src/v5/ui/routes/viewer/tickets/ticketDetails/ticketsDetailsCard.component.tsx
```diff
@@ -116,8 +123,10 @@ export const TicketDetailsCard = () => {
 					<CircleButton size="medium" variant="viewer" onClick={goNext}><ChevronRight /></CircleButton>
 				</HeaderButtons>
 			</CardHeader>
-			<FormProvider {...useForm()}>
-				<TicketForm template={template} ticket={ticket} />
+			<FormProvider {...formData}>
+				<Form>
```
sanmont3drepo, line 127: Why is this ``<Form/>`` needed here?

sanmont3drepo (changes requested):

frontend/src/v5/ui/routes/viewer/tickets/ticketDetails/ticketsDetailsCard.component.tsx
Amantini1997: Functionally, it is not required, but its a semantic thing since the `<FormProvider>` only creates a context, but it does not add a `<form>` in the actual DOM

The-Daniel (approved):

The-Daniel:
image
If you keep pressing enter in the ticket title input it can take over the entire card. Maybe there should be a limit to the number of lines?

sanmont3drepo:
> image I don't think this tooltip is our design. Also there should perhaps be error text when you enter the value, not just a tooltip on form submit

To be honest, that one seem on that is your browser doing stuff.

carmenfan:
> image If you keep pressing enter in the ticket title input it can take over the entire card. Maybe there should be a limit to the number of lines?

@Amantini1997 title should only span across one line for editing - the user shouldn't be able to record newlines within the title, that's for LONG_TEXT only.

Essentially it's a special case of TEXT, where we show the text on multiple lines if necessary, but it's not designed to be longer than one sentence

PR #3719, opened 2022-10-31
Issue 3573 - Merge implementation/tests of group end points

### LINKED ISSUE #3573: Merge container/federation implementations - Groups
They are so similar we end up with a lot of duplicated code.

The problem is that we also would need to auto generate swagger documentation to fit.... perhaps we can write them with enum variable parameter?

Trying this out with groups as a first test to see if it is possible

carmenfan:
This fixes #3573
#### Description
- merge routes `federations/groups` and `containers/groups` into a single implementation with a flag to denotes whether it's a federation
- Similar thing done in the e2e tests

#### Test cases
- the routes should work as before
- swagger should still make sense


backend/tests/v5/e2e/routes/teamspaces/projects/models/common/groups.test.js
```diff
@@ -0,0 +1,215 @@
...
+	const project = {
+		id: ServiceHelper.generateUUIDString(),
+		name: ServiceHelper.generateRandomString(),
+	};
+
+	const container = ServiceHelper.generateRandomModel();
+	const fed = ServiceHelper.generateRandomModel({ isFederation: true });
+
+	const groups = [
+		ServiceHelper.generateGroup(teamspace, container._id),
```
ChristosTsiotsias, line 134: shouldn't we include some federation groups as well ?

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/groups.test.js
```diff
@@ -0,0 +1,215 @@
...
+		id: ServiceHelper.generateUUIDString(),
+		name: ServiceHelper.generateRandomString(),
+	};
+
+	const container = ServiceHelper.generateRandomModel();
+	const containerNoGroups = ServiceHelper.generateRandomModel();
+	const fed = ServiceHelper.generateRandomModel({ isFederation: true });
+	const fedNoGroups = ServiceHelper.generateRandomModel({ isFederation: true });
+
+	const groups = [
```
ChristosTsiotsias, line 46: shouldn't we include some federation groups as well ?

ChristosTsiotsias (changes requested):

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/groups.test.js
carmenfan: that model id is for the objects identifier (i.e. given a geometry UUID, which model did it come from) - this function just generates some group data, it doesn't save it to a database

federation is just a federated list of models, it'll never own any geometry (i.e. it'll always come from a container)

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/groups.test.js
carmenfan: same as above! 😆

backend/src/v5/routes/teamspaces/projects/models/common/groups.js
```diff
@@ -94,24 +125,22 @@ const establishRoutes = () => {
...
+	router.post('/export', hasReadAccessToModel(isFed), validateGroupsExportData, exportGroups(isFed), serialiseGroupArray);
 	/**
 	 * @openapi
-	 * /teamspaces/{teamspace}/projects/{project}/federations/{federation}/groups/import:
+	 * /teamspaces/{teamspace}/projects/{project}/{type}/{model}/groups/import:
 	 *   post:
-	 *     description: Import a list of groups into the federation
-	 *     tags: [Federations]
-	 *     operationId: ImportFederationGroups
+	 *     description: Import a list of groups into the model
```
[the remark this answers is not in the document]
carmenfan: I'm just gonna remove `into the model` -  the intention is pretty self explanatory

ChristosTsiotsias (dismissed):

PR #3734, opened 2022-11-03
ISSUE #3721 - Custom Tickets - Show unset due dates and assignees in tickets list

### LINKED ISSUE #3721: Custom tickets - show unset due dates and assignees in tickets list
### Description
Right now we are showing the due dates and assignees in the ticket list only when they are set. We need to fetch the templates details and show the unset perroperties for templates that has these properties

### Goals
- [ ] As a user I want to know a first glance if a ticket should have an asignee or a due date.

### Tasks
- [ ] Fetch the templates details in the tickets list (maybe just add the fetch in the fetch templates saga)
- [ ] For tickets that are based on templates that hase due dates and asignees add the blank field.

The-Daniel:
This fixes #3721 

#### Description
- The ticket template details are loaded when opening the tickets list
- Issue properties are only rendered if the template has `issueProperties: true`
- If a ticket has issue properties but does not include either Due Date or Assignees the empty states for these are rendered
- Updated saga test for fetchTickets

#### Test cases
View a variety of tickets of different template types (issueProperties: true/false)
View issue tickets with unset due dates and assignees


Amantini1997 (approved):

sanmont3drepo:
because now we are fetching the templates details in fetchtemplates I think is no longer neede to fetch the template details in the newticket and details card.

sanmont3drepo (approved):

PR #3742, opened 2022-11-09
ISSUE #3740 use the full template for binary extraction

### LINKED ISSUE #3740: Custom ticket: Default image isn't being stored correctly
### Description
Default image isn't being taken out and stored in files manager, it is being stored in side the db as a binary instead.

The endpoint still gives the user a resource reference to fetch, but that reference isn't found and results in a 404 error

We should also add a e2e test for this test case to avoid regression

### Steps to replicate
- Create a template with `defaultImage: true` set in the config
- Create a ticket with that template, saving a default image
- GET the resource Id for the image
- try to get the resource

### Current Behaviour
- `RESOURCE_NOT_FOUND` is returned
- default image is stored as a binary field
image

### Expected Behaviour
- the original image should be returned
- the image should be stored in file storage like a custom image field

carmenfan:
This fixes #3740

#### Description
- pull the full template for addTicket and updateTicket in order to validate and extract the images/binaries from default properties
- added e2e test for the failure case

#### Test cases
- you should be able to upload a default image
  - it should no longer be stored in the database
  - you should be able to be retrieved by the given resourceId

### DIFF
--- a/backend/tests/v5/unit/middleware/dataConverter/inputs/teamspaces/projects/models/commons/tickets.test.js
+++ b/backend/tests/v5/unit/middleware/dataConverter/inputs/teamspaces/projects/models/commons/tickets.test.js
@@ -27,6 +27,9 @@ const SettingsMW = require(`${src}/middleware/dataConverter/inputs/teamspaces/se
 jest.mock('../../../../../../../../../../src/v5/schemas/tickets');
 const TicketSchema = require(`${src}/schemas/tickets`);
 
+jest.mock('../../../../../../../../../../src/v5/schemas/tickets/templates');
+const TemplatesSchema = require(`${src}/schemas/tickets/templates`);
+
 jest.mock('../../../../../../../../../../src/v5/models/tickets.templates');
 const TemplateModelSchema = require(`${src}/models/tickets.templates`);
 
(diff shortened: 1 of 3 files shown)


carmenfan:
@ChristosTsiotsias  image update bug fixed and ready for review

backend/tests/v5/unit/middleware/dataConverter/inputs/teamspaces/projects/models/commons/tickets.test.js
```diff
@@ -97,6 +104,8 @@ const testValidateNewTicket = () => {
 				await next();
 			});
 
+			TemplatesSchema.generateFullSchema.mockImplementationOnce((template) => template);
```
ChristosTsiotsias, line 107: expect(generateFullSchema).toHaveBeenCalledTimes is missing in some tests in this file

ChristosTsiotsias (changes requested):
generateFullSchema call from shcemas/tickets  validateTicket method can go

backend/tests/v5/unit/middleware/dataConverter/inputs/teamspaces/projects/models/commons/tickets.test.js
carmenfan: I changed my mind, moved the `generateFullSchema` call inside processor

Worried about regression if we have future functions calling them in different places they may not generate the full schema before calling.

PR #3752, opened 2022-11-10
Issue 3735 - fix error 500 when the user does not have permissionTemplates

### LINKED ISSUE #3735: Editing model permissions in v4 returns 404
### Description
Trying to change the permission for a model in v4returns a 500 error code.

### Steps to replicate
In v4 
- go to User Management;
- select projects;
- go to models permissions;
- select a project;
- tick one model from the list, and then
- try to change any permission

### Current Behaviour
- The server returns a 404

### Expected Behaviour
- Permission should update with no error

8b091a00-c1ae-4a51-be08-20a2f91249ac

carmenfan:
This fixes #3735

#### Description
Problem arises because anyone who signed up on v5 would not have permissionTemplates in their bson, but v4 logic still assumes its presence.

- hardcoded permissionTemplates in the code instead of trying to fetch it from the database
- v4 no longer add permissionTemplates in the user bson at sign up
- removed functions to add/remove permission templates as they are now constants (and was never used)
- migration script added to remove all permissionTemplates

#### Test cases
- situation described in the issue should work for both
  - users signed up via v4 endpoint
  - users signed up via v5 endpoint

### DIFF
--- a/backend/src/scripts/migrations/4.28/index.js
+++ b/backend/src/scripts/migrations/4.28/index.js
@@ -16,6 +16,7 @@
  */
 
 const moveLoginRecords = require('./moveLoginRecords');
+const removePermissionTemplates = require('./removePermissionTemplates');
 const indexLinksInRef = require('./indexLinksInRef');
 const removeUnityAssetsJSON = require('./removeUnityAssetsJSON');
 const removeGridFSBackUps = require('./removeGridFSBackUps');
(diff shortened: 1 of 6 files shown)


ChristosTsiotsias (changes requested):
There are still the createTemplate and deleteTemplate routes in v4. These should be deleted too right ?

ChristosTsiotsias (changes requested):
should we also remove the intergrated tests for adding / removing templates in v4 ?

ChristosTsiotsias (dismissed):

PR #3763, opened 2022-11-11
ISSUE_3672 - update Status and fix issue title

### LINKED ISSUE #3672: GitHub Actions - auto change status of issue if there's an update on the PR
### Description
Add github action to change the status of an issue base on the status of the PR

### Goals
- [x]  if someone has been assigned a reviewer -> change issue status to "awaiting approval"
- [x]  if someone has came back with change requests -> change issue status to "address feedback"
- [x]  if it has been merged in, change issue status to "done"

### Tasks
- [x] review if we can do this in onPRReviews.yml
- [x] identify issue updating component in github actions
- [x] add the three tasks

keifgwinn:
This fixes #3672

#### Description
- get PR numbers and node-ids
- ported all the actions to our own org to ensure consistent behaviour
- patched those to account for changes to set-output

#### Test cases
- issues that are going for review update in the project
- issues that are merged are marked as done
- issues that are returned set in progress


.github/workflows/onPRReviews.yml
```diff
@@ -1,23 +1,120 @@
...
-    types: [review_requested, review_request_removed]
+    types: [ready_for_review, review_requested, review_request_removed]
   pull_request_review:
     types: [edited, dismissed, submitted]
 jobs:
+  getPRNumber:
+    name: Get Current PR Number
+    runs-on: ubuntu-latest
+    outputs: 
+      issue-number: ${{ steps.condval.outputs.value }}
```
carmenfan, line 12:
```suggestion
      pr-number: ${{ steps.condval.outputs.value }}
```

.github/workflows/onPRReviews.yml
```diff
@@ -1,23 +1,120 @@
...
           if_true: ${{ github.event.number }}
           if_false: ${{ github.event.pull_request.number }}
       - name: Check reviews
-        uses: carmenfan/all-reviewers-approved@v1.0
+        id: checkReviews
+        uses: 3drepo/all-reviewers-approved@v1
         with:
           pr: ${{ steps.condval.outputs.value }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Update status
```
carmenfan, line 56: wrong place for this. just because all the reviewers approved doesn't necessarily mean the issue is done

The issue should only be considered done when it's merged - the best place to do this is probably in https://github.com/3drepo/3drepo.io/blob/master/.github/workflows/onStagingCommits.yml

.github/workflows/onPRReviews.yml
```diff
@@ -1,23 +1,120 @@
...
+        with:
+          github_token: ${{ secrets.PROJ_MANAGEMENT_TOKEN }}
+          organization: 3drepo
+          project_number: 22
+          content_id: ${{ needs.getIssueNumber.outputs.content-id }}
+          field: Status
+          value: 'Done'
 
+#  if someone has been assigned a reviewer -> change issue status to "awaiting approval"
+  updateProjectTicketStatus:
```
carmenfan, line 69: Issue is probably less confusing than Ticket

.github/workflows/onPRReviews.yml
```diff
@@ -1,23 +1,120 @@
...
+        run: |
+          set -ex 
+          case $ACTION in
+            review_requested ) 
+              echo "status=Awaiting Approval" >> $GITHUB_OUTPUT
+              ;;
+            ready_for_review ) 
+              echo "status=Awaiting Approval" >> $GITHUB_OUTPUT
+              ;;
+            opened )
```
carmenfan, line 89: don't think we need this, at the point of the user setting the PR to open state the issue should naturally be `In progress`?

.github/workflows/onPRReviews.yml
```diff
@@ -1,23 +1,120 @@
...
+              echo "status=In Progress" >> $GITHUB_OUTPUT
+              ;;
+            review_request_removed )
```
carmenfan, line 92: don't need this either - it's not obvious what the intention is if they removed someone from a review request so I don't think we should change the state

.github/workflows/onPRReviews.yml
```diff
@@ -1,23 +1,120 @@
...
+        id: setStatusOutput   
+        run: |
+          set -ex 
+          case $ACTION in
+            review_requested ) 
+              echo "status=Awaiting Approval" >> $GITHUB_OUTPUT
+              ;;
+            ready_for_review ) 
+              echo "status=Awaiting Approval" >> $GITHUB_OUTPUT
+              ;;
```
carmenfan, line 88: We're missing `Address Feedback` if someone posted a review requesting for further changes

.github/workflows/onPRReviews.yml
```diff
@@ -1,23 +1,120 @@
...
+      content-id: ${{ needs.getIssueNumber.outputs.content-id }}
+      status: ${{ steps.setStatusOutput.outputs.status }}
+    needs: [ getIssueNumber, checkMergeReady ]
+    steps:
+      - name: Set status message from input 
+        id: setStatusOutput   
+        run: |
+          set -ex 
+          case $ACTION in
+            review_requested )
```
carmenfan, line 83: could we also add something to change the PR status to `To Review` if someone requested a review?

This would be for situations where the dev has addressed the feedback and waiting for someone to have another look

.github/workflows/onPRReviews.yml
```diff
@@ -1,23 +1,120 @@
 name: On review update
 on:
   pull_request:
-    types: [review_requested, review_request_removed]
+    types: [ready_for_review, review_requested, review_request_removed]
   pull_request_review:
     types: [edited, dismissed, submitted]
 jobs:
+  getPRNumber:
```
carmenfan, line 8: camel case here but kebab case on the other PR 

I don't mind which, but we should make them all consistent

carmenfan (changes requested):

.github/workflows/onPRClose.yml
```diff
@@ -3,41 +3,75 @@ on:
...
+
+  tag-on-pr-merge:
     name: Tag merged issues as In staging if applicable
     runs-on: ubuntu-latest
-    needs: extractIssueNumber
+    needs: get-issue-number
     steps:
       - name: Apply in staging label
-        if: needs.extractIssueNumber.outputs.issue-number
+        if: needs.get-issue-number.outputs.issue-number
```
carmenfan, line 42: If I'm reading [this](https://github.com/3drepo/extract-issue-number-from-PR/blob/v1/index.js) correctly, you're extracting the `issue-number`[ regardless of whether the issue is being merged into staging](https://github.com/3drepo/extract-issue-number-from-PR/blob/v1/index.js#L72). So provided the branch name matches the regex it'll set the `issue-number` output. Meaning we will be tagging pretty much every issue as `In Staging`

So I think you need to check for the output `merged` here? [which is only provided if it's merged](https://github.com/3drepo/extract-issue-number-from-PR/blob/v1/index.js#L78) and the base matches the target base.

I didn't test it, so I could be missing something

.github/workflows/onPRClose.yml
```diff
@@ -3,41 +3,75 @@ on:
...
+          issue-number: ${{ needs.get-issue-number.outputs.issue-number }}
           label: 'in staging'
           github-token: ${{ secrets.GITHUB_TOKEN }}
-  destroy:
-    name: Call Azure Pipeline
+
+  update-project-status:
+    name: Update status of the issue in the project
+    runs-on: ubuntu-latest
+    needs: get-issue-number
```
carmenfan, line 52: Similar to above, Unless i've mistaken, I could've closed the issue without merging and this would erroneously tagged this as done.

carmenfan: But this case is slightly different, we want to set it to done if it's been merged, we don't care about whether it's been merged to staging.

.github/workflows/onPRReviews.yml
```diff
@@ -1,23 +1,119 @@
...
+  get-issue-number:
+    name: Get Current Issue Number
+    needs: [ get-pr-number ]
+    runs-on: ubuntu-latest
+    outputs: 
+      issue-number: ${{ steps.commitMsgParser.outputs.issue-number }}
+      content-id: ${{ steps.commitMsgParser.outputs.content-id }}
+    steps:
+      - name: Extract Issue Number from PR
+        uses: 3drepo/extract-issue-number-from-PR@v1
```
carmenfan, line 35: Since you use this in so many different context I think you should keep this simple and only do what it says on the tin - extract the issue number base on the branch name.

Should probably move the is merged or not part out, it's confusing why we need to pass in a `base` if all we're doing is extract an issue number from a branch name

.github/workflows/onPRReviews.yml
```diff
@@ -1,23 +1,119 @@
...
+
+  check-merge-ready:
     name: Ensure all reviewers have approved the PR
     runs-on: ubuntu-latest
+    needs: [ get-issue-number ]
+    outputs: 
+      approved: ${{ steps.checkReviews.outputs.approved }}
     steps:
-      - uses: haya14busa/action-cond@v1
+      - uses: 3drepo/action-cond@v1
```
carmenfan, line 49: this is now a duplicate as you already pull PR number out at `get-pr-number`

.github/workflows/onPRReviews.yml
```diff
@@ -1,23 +1,119 @@
...
       - name: Check reviews
-        uses: carmenfan/all-reviewers-approved@v1.0
+        id: checkReviews
+        uses: 3drepo/all-reviewers-approved@v1
         with:
           pr: ${{ steps.condval.outputs.value }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check reviews 
+        if: failure()
+        id: setStatusOutputToTrue
```
carmenfan, line 63: This is not very easy to read. I'm actually unsure what this is suppose to do doing. Does it stop the script if it's not fully reviewed?

.github/workflows/onPRReviews.yml
```diff
@@ -1,23 +1,119 @@
...
+      - name: Set status message from input 
+        id: setStatusOutput   
+        run: |
+          set -ex 
+          case $ACTION in
+            submitted ) 
+              echo "status=Address Feedback" >> $GITHUB_OUTPUT
+              ;;
+            review_requested ) 
+              echo "status=To Review" >> $GITHUB_OUTPUT
```
carmenfan, line 85: This block of code is not very easy to read, but i'm guessing what you're saying here is set issue status to  `To Review` if the trigger was `review_requested`?

`To Review` should never be set on a issue (this is a PR status), so it should be `Awaiting Approval`.

carmenfan (changes requested):

.github/workflows/onPRReviews.yml
keifgwinn: I'm currently not sure which action status this is, I put the debug on the end to try and catch it in operation.

keifgwinn: https://stackoverflow.com/questions/44159555/how-do-we-know-a-pull-request-is-approved-or-rejected-using-api-in-github suggests we need to query graphql

.github/workflows/onPRClose.yml
keifgwinn: yes, this is because I didn't actually touch this as it was already working previously, but then I did a search and replace which changed the action to the 3D Repo one with the new behaviour to always return the issue number and merged as optional, as I needed it all the time and the previous action only returned it if it was merged. Anyway, yes, it should check merged and only fire on that.

.github/workflows/onPRReviews.yml
keifgwinn: always() is because the previous step may fail or succeed based on reviewer approval, I added the addition id: setStatusOutputToTrue to try and stop it erroring when the reviewers aren't ready yet and producing the notification noise all the time.

.github/workflows/onPRReviews.yml
carmenfan: not sure i follow, if we can figure out here https://github.com/3drepo/all-reviewers-approved/blob/main/index.js#L57 surely we can figure it out.

.github/workflows/onPRReviews.yml
carmenfan: does that mean it fake passes when it fails?

.github/workflows/onPRReviews.yml
keifgwinn: I misunderstood your previous note for setting to Review on the PR and did it on the issue, so if we want to modify both status then I need a second block of this for PR action and one for Issue action. Is that what you're after?

carmenfan: it's suppose to be done in the issue, what you done is correct?

the PR status is done via some actions wired in the project already all you have to worry about is the issue ones.

to be clear:

when a review is requested/the PR is first created, the issue should be in `awaiting approval`
someone done a review and it was not approved, the issue should be in `addressing feedback`

carmenfan: ah I guess if someone re-requested a review the PR needs to be set to `to review`, is that what you are trying to do here?

.github/workflows/onPRReviews.yml
keifgwinn: renamed

.github/workflows/onPRReviews.yml
keifgwinn: Yes, that was the intention

.github/workflows/onPRClose.yml
```diff
@@ -3,41 +3,77 @@ on:
   pull_request:
     types: [closed]
 jobs:
-  extractIssueNumber:
-    name: Extract Issue Number
+
+  get-pr-number:
```
carmenfan, line 7: @keifgwinn  why do you need this? 

Unless i'm mistaken as this yml is only ran on PR close, `github.event.number` should always get you the PR number?

carmenfan (approved):

PR #3781, opened 2022-11-23
Issue 3776 - move dispatchers and selectors in the same files

### LINKED ISSUE #3776: Move all hookselectors and actiondispatchers to the same file
### Description
Every hookselector has its own file but the actual content of the file is just one line. So I think it could be good if we just one file with all the hook selectors instead of having so many files . Same with the action dispatchers.

### Goals
- [ ] As a dev I want it to be easier to navigate files and to add hooksselectors and action dispatchers

### Tasks
- [ ] Move all hook selectors to one file 
- [ ] Move all action dispatchers to one file

Amantini1997:
This fixes #3776 

#### Description
Moved selections and dispatchers in the same files


The-Daniel (approved):
Easily one of the most exciting PRs I have ever reviewed

PR #3782, opened 2022-11-23
Issue 3730 - V5 notifications

### LINKED ISSUE #3730: V5 Notifications MVP
### Description

We need to have a mechanism for accessing notifications from inside V5. At the moment we are not looking to redefine the notifications, but just allow them to be accessed

### Goals

- [ ] Allow notifications to be accessed from V5
- [ ] Where possible maintain styling with V5
- [ ] If the page opens in a new tab, avoid users from accessing other elements of V4
- [ ] Add the notification count to the V5 toolbar

Amantini1997:
This fixes #3730

#### Description
Notifications are now available in V5


george-hadfield:
### Reduce size of "Mark as read" circle
Reduce to same size as the "X"

image

The-Daniel (commented):
Have you looked into using the custom scrollbar for the notifications list?

Amantini1997:
> Have you looked into using the custom scrollbar for the notifications list?

To add the scrollbar, I must add it to the v4 as I am in fact importing that.
@sanmont3drepo should I do that?

george-hadfield (approved):

sanmont3drepo (approved):

sanmont3drepo:
There are some icons that doesnt show. I think is because the extension of these are `.svg` instead of `.svg.tsx`, nevertheless I think they should show up in the list:

image

sanmont3drepo (changes requested):
Wrong comment

The-Daniel (approved):

sanmont3drepo (approved):

PR #2994, opened 2022-02-02
ISSUE #2914 - Upload containers: Upload item list items

The-Daniel:
This fixes #2914

#### Description
- Uploads are displayed in a list
- Clicking a list item shows it as selected
- You can click the "filename" header to sort by filename (Note: You'll be able to sort by other headings when I make the changes to #2698 to make the field objects flat)
- Errors in the revision tag show a red error icon with the error message in the tooltip (container name too, but this is just temporary and might change with #2943)
- Filename has a tooltip to help when long filenames are cut off with small screens

#### Test cases
- Upload items to show the list. (either using the browse button or drag and drop)
- Clicking on 'filename' heading changes list order
- Clicking an item changes its appearance
- Clicking the edit button opens the settings sidebar
- Clicking delete removes the item
- Deleting all the items reverts the modal to its original state (no list headings)
- Create errors in revision box (fewer than 2 letters, more than 120), and hover over error icon to see tooltip

PR #3783, opened 2022-11-24
Issue 3778 - Refactor Circle Button

### LINKED ISSUE #3778: Refactor CircleButton
### Description
CircleButton button has become a little messy. We need to take a look at it and see if it can make it a bit 

### Goals
- [ ] As a dev I want to be able to reuse CircleButton  easily 

### Tasks
- [ ] Refactor CircleButton

Amantini1997:
This fixes #3778

### Description
CircleButton uses less css code and avoid the `&&` in the styling to allow css styling override easier in other components


The-Daniel:
There are a few places in the app which have circular buttons but which don't use the CircleButton component. For example there are the EllipsisButton and the ChevronButton which use a styled Mui IconButton. Althouh perhaps ChevronButton shouldn't as it seems a bit distinct... @sanmont3drepo?

You might also be able to address [this comment](https://github.com/3drepo/3drepo.io/pull/2994#discussion_r809128199) from Santiago

The-Daniel (changes requested):

The-Daniel (approved):
Nice. I think this is a lot better now

frontend/src/v5/ui/controls/circleButton/circleButton.styles.ts
```diff
@@ -15,83 +15,39 @@
...
-	if (buttonSize) {
-		return css`
-			height: ${SIZE_MAP[size]}px;
-			width: ${SIZE_MAP[size]}px;
-			flex-shrink: 0;
-		`;
+import styled from 'styled-components';
+import { Button } from '@mui/material';
+
+export const BaseCircleButton = styled(Button)<{ disabled?: boolean }>`
```
sanmont3drepo, line 21: what happened with the size? It wasn't used anywhere?

sanmont3drepo (approved):

frontend/src/v5/ui/controls/circleButton/circleButton.styles.ts
Amantini1997: That's right. The only size used when I was tackling this issue was large, so there was no point in keeping it.

However, in custom tickets, I saw that a new variant of circleButton was created ("viewer") which used another size ("medium"). As such, I created a new component called ViewerButton to address the different size. In general, however, it feels nicer to override the css sizes rather than having a prop to control that, what do you think?

sanmont3drepo (approved):

The-Daniel (approved):

Amantini1997:
> I think this one fixes this, right? [#3817 (comment)](https://github.com/3drepo/3drepo.io/pull/3817#issuecomment-1346616241)

@sanmont3drepo Ish.
In the v5 notifications PR, the badge was positioned as follows
image
And it was approved.

Since V5 notifications PR was approved, this PR is only reverting the styling to that state, fixing something introduced in custom ticket. But in #3808 , the badge looks different, in particular, it looks smaller.
image

As such, I am not sure anymore whether it can be considered fixed

PR #3785, opened 2022-11-24
ISSUE #3779 - Fix Storybook components

### LINKED ISSUE #3779: Review stories in storybook
### Description
Some stories are broken. We need to review them and fix where 

### Tasks
- [ ] Check all the stories in storybook
- [ ] FIx the problems that arised

Amantini1997:
This fixes #3779

#### Description
cleaned up stories


frontend/stories/buttons/Button.stories.tsx
```diff
@@ -57,21 +55,9 @@ export default {
 } as ComponentMeta<typeof Button>;
 
 // More on component templates: https://storybook.js.org/docs/react/writing-stories/introduction#using-args
-const Template: ComponentStory<typeof Button> = (args) => <Button {...args}>Material ui button</Button>;
+const Template: ComponentStory<typeof Button> = (args) => <Button {...args}>Controls Button</Button>;
 
-export const Primary = Template.bind({});
 // More on args: https://storybook.js.org/docs/react/writing-stories/args
-Primary.args = {
```
The-Daniel, line 64: Any reason the 'outlined' examples have been removed?
Also you've removed the 'label' and 'label-outlined' variants from the options, so once you change the variant of one of these you can't change back

frontend/stories/outputs/SuccessMessage.stories.tsx
```diff
@@ -24,19 +24,28 @@ export default {
 		children: {
 			description: 'The text or component to display',
```
The-Daniel, line 25: Should this include 'component'?

The-Daniel (changes requested):

The-Daniel (approved):

sanmont3drepo (changes requested):
There are some icons that doesnt show. I think is because the extension of these are `.svg` instead of `.svg.tsx`, nevertheless I think they should show up in the list:

image

sanmont3drepo:
> There are some icons that doesnt show. I think is because the extension of these are `.svg` instead of `.svg.tsx`, nevertheless I think they should show up in the list:
> 
> image

Ok, re tested it in my computer and it worked 🤷

The-Daniel (approved):

PR #3788, opened 2022-11-25
Issue 3705 - View property 

### LINKED ISSUE #3705: Custom tickets - View property
### Description
I want to be able to reflect the view and be able to set the view

### Goals
- [ ] As a Collaborator+ I can create a viewpoint when no viewpoint is currently available
- [ ] As a Collaborator+ I can change the viewpoint of an viewpoint which is already populated
- [ ] As a Collaborator+ if I select "Delete Viewpoint" I see a confirmation pop-up to confirm my choice
- [ ] As a Viewer+ if I select "Go to viewpoint" the Unity camera position will move to the stored position

### Tasks
- [ ] Create a button to create a viewpoint
- [ ] Use unity utils to get the current viewpoint to be able to save it
- [ ] Add button to go to a viewpoint
- [ ] Use unity utils to go to the viewpoint when clicking the go to a viewpoint button

### Notes
This is only to get view working, the design will be applied once #3703 is working

### Related Resources
[Mockup](https://www.figma.com/file/3FjNEOdQbcHwzquVZxUqMl/3D-Repo?node-id=520%3A9837)

sanmont3drepo:
This fixes #3705 

#### Description
- Edit view type properties 
- Supports create viewpoint
- Supports update viewpoint
- Delete viewpoint
- Add/edit/delete  image to the view


The-Daniel:
image
Error pops up when creating a viewpoint in orthographic mode. It still works though

The-Daniel:
Should the viewpoint change to the ticket's default view when opening a ticket?

carmenfan:
> Should the viewpoint change to the ticket's default view when opening a ticket?

yes

sanmont3drepo:
@The-Daniel  fixed both issues

frontend/src/v4/modules/viewpoints/viewpoints.sagas.ts
```diff
@@ -68,10 +68,7 @@ export function* generateViewpoint(teamspace, modelId, name, withScreenshot = fa
 	try {
 		const hiddenGeometryVisible = yield select(selectHiddenGeometryVisible);
 
-		const viewpoint = yield Viewer.getCurrentViewpoint({
-			teamspace,
-			model: modelId
-		});
+		const viewpoint = yield Viewer.getCurrentViewpoint();
```
Amantini1997, line 71: Since this was the only point in the function body using `teamspace` and `modelId`, I think you can remove their definition from the parameters

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/basicTicketImage/ticketImageAction/ticketImageAction.component.tsx
```diff
@@ -15,21 +15,21 @@
...
+	onClick?: (...args) => void;
 };
 export const TicketImageAction = ({
 	onImageChange,
 	disabled: disabledInput,
 	...props
 }: TicketImageActionProps) => {
 	const { isAdmin } = ProjectsHooksSelectors.selectCurrentProjectDetails();
-	const disabled = isUndefined(disabledInput) ? disabledInput : !isAdmin;
+	const disabled = disabledInput || !isAdmin;
```
Amantini1997, line 32: The reason why this was using `isUndefined(disabledInput)` is so that this component knows whether or not `disabledInput` has been set from the parent. If you only use `disabledInput || !isAdmin`, the parent component will not be able to set the action as "enabled" because the `isAdmin` would override that value.

**Example:**
The user (Alice) is not an admin, but the action is `go to viewpoint`, which should be enabled for any users.

If the parent component sets `disabledInput` to `false`, Alice will still not be able to click the action  
> `disabledInput || !isAdmin` = `false || true` = `true`. 

Using the `isUndefined(disabledInput)`, instead, you would still allow Alice to click on that action
> `!isUndefined(disabledInput) ? disabledInput : !isAdmin` = `true ? false : true` = `false`
(**Note**: I forgot to add the `not` in front of `isUndefined`)

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/ticketView/ticketView.component.tsx
```diff
@@ -0,0 +1,117 @@
...
+	error,
+	helperText,
+	required,
+	onBlur,
+	onChange,
+	...props
+}: ITicketView) => {
+	const updateViewpoint = async () => {
+		const currentViewpoint = await ViewerService.getViewpoint();
+		if (isEqual(currentViewpoint, value)) return;
```
The-Daniel, line 57: value can contain a screenshot property whereas currentViewpoint does not. This means if you open a ticket that already has a viewpoint set and instantly try to update the viewpoint (without moving the camera position) you get this error
image
because this condition passes true

Amantini1997:
Required viewpoint has not red asterisk 
image

Amantini1997:
Trying to delete a required viewpoint triggers error 
image

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/ticketView/ticketView.component.tsx
```diff
@@ -0,0 +1,117 @@
...
+		onChange?.(currentViewpoint);
+	};
+
+	const goToViewpoint = async () => {
+		if (!value) return;
```
Amantini1997, line 62: This method is only used by the respective ticketAction, which is disabled when `!(value?.camera)`. Could we perhaps change the disabled condition in the action to be `disabled={!value}` and get rid of this `if`?

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/ticketView/ticketView.component.tsx
```diff
@@ -0,0 +1,117 @@
...
+		await ViewerService.setViewpoint(value);
+	};
+	const deleteViewpoint = async () => {
```
Amantini1997, line 65: What is the `async` for?

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/viewProperty.component.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import { Controller } from 'react-hook-form';
+import { PropertyProps } from './properties.types';
+import { TicketView } from './ticketView/ticketView.component';
+
+export const ViewProperty = ({
+	property: { name, required },
+	formError,
+	defaultValue,
+	...props
```
Amantini1997, line 26: You could do the same as you did for `ImageProperty`, i.e. to convert `property.name` to `label` and then pass the `name` directly without calling `name={props.name}`
```ts 
export const ViewProperty = ({
        property: { name: label, readOnly, required }, // <- name: label
	defaultValue,
	formError,
	name, // <- also this one
```

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/ticketView/ticketView.component.tsx
The-Daniel: I believe this also removes the 'screenshot' from the value which is what causes the image to flicker when updating the viewpoint. If you log `value` in this component you can see that screenshot mysteriously disappears and reappears after the component rerenders a few times

The-Daniel (changes requested):

Amantini1997 (changes requested):

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/basicTicketImage/ticketImageAction/ticketImageAction.component.tsx
carmenfan: why are we checking `isAdmin`?   @sanmont3drepo

The current scope of the work we shouldn't be dealing with permissions, let the request go through and let the api endpoint errors out.

Only allow admins to update a ticket is not a desired behaviour... also the permissions mentioned in the scope is not correct, it should be commenter plus 
image


I didn't sanity check this when you opened the issue because we agreed we were not dealing with the permissions 😆

sanmont3drepo: You are right @carmenfan  we shouldnt worry about this at this point.

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/ticketView/ticketView.component.tsx
sanmont3drepo: There is already that condition to disable the goto button . So this line is not necessary.

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/ticketView/ticketView.component.tsx
sanmont3drepo: Copy and paste error.

frontend/src/v4/modules/viewpoints/viewpoints.sagas.ts
sanmont3drepo: I prefer to leave v4 code untouched as much as possible.

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/ticketView/ticketView.component.tsx
The-Daniel: @sanmont3drepo When I do this now (set a new viewpoint that has the exact same camera) i get the error `properties.Default View.camera.position is a required field`

The-Daniel:
There are still problems when updating the viewpoint when one or more camera properties don't change.

In orthographic mode if you go to a viewpoint, zoom in or out, and create a new viewpoint then you get an error saying "*forward* is a required field"
In orthographic mode if you go to a viewpoint, move about without changing zoom level, create a new viewpoint then you get an error saying "*size* is a required field"
In either mode and creating a viewpoint directly after creating another wihtout moving

Amantini1997:
- I have a required view property which is populated (I have both the view and the image);
- I delete the image first;
- I delete the viewpoint second;
I get this 
image

Amantini1997:
If I try to create a new ticket that contains viewpoints, I am allowed to save the ticket without inserting the title 
image

As a reference, this is my template
```
{
    "_id": "6c1bf430-6f0e-11ed-bb18-15885d43fd33",
    "modules": [],
    "properties": [
        {
            "name": "Description",
            "type": "longText"
        },
        {
            "name": "Owner",
            "type": "text",
            "readOnly": true
        },
        {
            "name": "Created at",
            "type": "date",
            "readOnly": true
        },
        {
            "name": "Updated at",
            "type": "date",
            "readOnly": true
        },
        {
            "name": "Default View",
            "type": "view"
        },
        {
            "name": "Priority",
            "type": "oneOf",
            "values": [
                "None",
                "Low",
                "Medium",
                "High"
            ],
            "default": "None"
        },
        {
            "name": "Status",
            "type": "oneOf",
            "values": [
                "Open",
                "In Progress",
                "For Approval",
                "Closed",
                "Void"
            ],
            "default": "Open"
        },
        {
            "name": "Assignees",
            "type": "manyOf",
            "values": "jobsAndUsers"
        },
        {
            "name": "Due Date",
            "type": "date"
        },
        {
            "name": "Pin",
            "type": "coords"
        },
        {
            "type": "view",
            "name": "View"
        },
        {
            "type": "view",
            "required": true,
            "name": "View - required"
        }
    ],
    "config": {
        "pin": true,
        "defaultView": true,
        "issueProperties": true,
        "comments": true
    },
    "code": "VUE",
    "name": "View"
}
```

Amantini1997:
Same template as before:
- I have a non required view property that has both the viewpoint and the image;
- I delete the viewpoint;
- I create a new viewpoint;
I get this 
image

Amantini1997:
There are some problem with the perspective/ortographic mode:
- In perspective mode, create a new viewpoint (**VP 1**);
- Change mode to ortographic;
- Click "Go to viewpoint" for VP 1;
- **First bug**: the mode in the bottom bar did not update
> The same is also true the other way around

- Change the mode (perspective/ortographic) a couple of times;
- **Second bug**: the model will disappear and you need to interact with it (zoom in/out or move the view with the mouse) to have it back on the screen

https://user-images.githubusercontent.com/43547703/204765411-4a8d569b-a6c7-42c1-b8b9-cda723fddc9a.mp4

carmenfan:
> * I have a required view property which is populated (I have both the view and the image);
> * I delete the image first;
> * I delete the viewpoint second;
>   I get this
>   image

@sanmont3drepo if it helps, [this is the yup schema for views on the backend](https://github.com/3drepo/3drepo.io/blob/staging/backend/src/v5/schemas/tickets/validators.js#L49)

`isNullable` is set to true on update ticket but false on new ticket. `Null` is only an accepted field on non required fields during update as an indication that you want to remove the existing data. If it's a non required field when you're creating a ticket and you do not wish to give any data, it needs to be `undefined`.

Basically you can only set things to null on the outermost properties (screenshot, state, camera, clippingPlanes) If you want to update the camera position we need you to send the whole camera data

sanmont3drepo:
I think there might be a problem when updating a viewpoint 
- This are the things I send when updating a viewpoint:
image

- This is the next fetch for that viewpoint
image

sanmont3drepo:
> I think there might be a problem when updating a viewpoint
> 
> * This are the things I send when updating a viewpoint:
>   image
> * This is the next fetch for that viewpoint
>   image
> 
> You can see that in the first case there are some clipping planes in there while there are not being fetched afterwards.

I think the same thing is happening with a new ticket:

- POST:
image

-GET:
image

frontend/src/v5/ui/routes/viewer/tickets/ticketDetails/ticketsDetailsCard.component.tsx
```diff
@@ -77,14 +78,21 @@ export const TicketDetailsCard = () => {
 
 	const onBlurHandler = () => {
 		const values = dirtyValues(formData.getValues(), formData.formState.dirtyFields);
-		const validVals = removeEmptyObjects(nullifyEmptyStrings(filterErrors(values, formData.formState.errors)));
+		let validVals = removeEmptyObjects(nullifyEmptyStrings(filterErrors(values, formData.formState.errors)));
```
The-Daniel, line 81: I don't think the `let` is necessary here

carmenfan:
> I think there might be a problem when updating a viewpoint
> 
> * This are the things I send when updating a viewpoint:
>   image
> * This is the next fetch for that viewpoint
>   image
> 
> You can see that in the first case there are some clipping planes in there while there are not being fetched afterwards.


This is(should) now be fixed on this branch

Amantini1997:
> There are some problem with the perspective/ortographic mode:
> 
> * In perspective mode, create a new viewpoint (**VP 1**);
> * Change mode to ortographic;
> * Click "Go to viewpoint" for VP 1;
> * **First bug**: the mode in the bottom bar did not update
> 
> > The same is also true the other way around
> 
> * Change the mode (perspective/ortographic) a couple of times;
> * **Second bug**: the model will disappear and you need to interact with it (zoom in/out or move the view with the mouse) to have it back on the screen
> 
>  2022-11-30.09-55-07.mp4

I still get these

sanmont3drepo:
> > There are some problem with the perspective/ortographic mode:
> > 
> > * In perspective mode, create a new viewpoint (**VP 1**);
> > * Change mode to ortographic;
> > * Click "Go to viewpoint" for VP 1;
> > * **First bug**: the mode in the bottom bar did not update
> > 
> > > The same is also true the other way around

Changed the way  v4 reflects the camera projection now so it should have been solved.

> > 
> > 
> > 
> > * Change the mode (perspective/ortographic) a couple of times;
> > * **Second bug**: the model will disappear and you need to interact with it (zoom in/out or move the view with the mouse) to have it back on the screen
> > 
> > 2022-11-30.09-55-07.mp4
> 
> Im still seeing these.

This is not a bug introducecd in this PR and TBH looks like a unity problem. You can create an ISSUE if you want.

Amantini1997:
> > > There are some problem with the perspective/ortographic mode:
> > > 
> > > * In perspective mode, create a new viewpoint (**VP 1**);
> > > * Change mode to ortographic;
> > > * Click "Go to viewpoint" for VP 1;
> > > * **First bug**: the mode in the bottom bar did not update
> > > 
> > > > The same is also true the other way around
> 
> Changed the way v4 reflects the camera projection now so it should have been solved.
> 
> > > * Change the mode (perspective/ortographic) a couple of times;
> > > * **Second bug**: the model will disappear and you need to interact with it (zoom in/out or move the view with the mouse) to have it back on the screen
> > > 
> > > 2022-11-30.09-55-07.mp4
> > 
> > 
> > Im still seeing these.
> 
> This is not a bug introduced in this PR and TBH looks like a unity problem. You can create an issue if you want.


Issue created https://github.com/3drepo/3drepo.io/issues/3812

PR #3796, opened 2022-11-29
Issue 3777 - clean up reacthook-form integration

### LINKED ISSUE #3777: clean up react hook form integration
### Description
At the moment we are repeating code when doing react-hook-form integration. Instead of doing that, we can create just one adapter and make all the integrations to the input components in the same file.

### Goals
- [ ] As a dev I want a consistent clean way to integrate with react hook forms

### Tasks
- [ ] Create an adapter for hooking react-hook forms with the inputs components
- [ ] Move all hooked inputs to one file

Amantini1997:
This fixes #3777

#### Description
All the form inputs now share a common interface and are not automatically wrapped with a <Controller />, meaning they can be used independently.

### DIFF
--- a/frontend/src/v5/ui/v4Adapter/overrides/preview/previewDetails.overrides.ts
+++ b/frontend/src/v5/ui/v4Adapter/overrides/preview/previewDetails.overrides.ts
@@ -17,7 +17,7 @@
 
 import { css } from 'styled-components';
 import { Details, Container as IssueDetails } from '@/v4/routes/viewerGui/components/previewItemInfo/previewItemInfo.styles';
-import { Container, Header, TitleNumber, Grid, Details as Accordion } from '@/v4/routes/viewerGui/components/previewDetails/previewDetails.styles';
+import { Container, Header, TitleNumber, Grid, Details as Accordion, StyledForm } from '@/v4/routes/viewerGui/components/previewDetails/previewDetails.styles';
 import { FieldWrapper } from '@/v4/routes/components/textField/textField.styles';
 import { StyledButton } from '@/v4/routes/viewerGui/components/containedButton/containedButton.styles';
 import { UserIndicator } from '@/v4/routes/components/messagesList/components/message/components/userMarker/userMarker.styles';
@@ -58,6 +58,7 @@ export default css`
 			.MuiAccordionSummary {
 				&-content {
 					margin: 12px 0 5px;
+					max-width: 100%;
 				}
 				
 				&-expandIconWrapper {
@@ -66,6 +67,10 @@ export default css`
 			}
 
 			${Grid} {
+
+				${StyledForm} {
+					width: 94%;
+				}
 
 				${TitleNumber} {
 					color: ${({ theme }) => theme.palette.secondary.main};
@@ -110,6 +115,12 @@ export default css`
 					margin: 0;
 					box-sizing: border-box;
 					height: 100%;
+					width: 100%;
+					text-overflow: ellipsis;
+					white-space: nowrap;
+					overflow: hidden;
+					display: inline-block;
+					line-height: 20px;
 				}
 
 				// weird floating legend
@@ -133,7 +144,7 @@ export default css`
 			// user + status + clone button
 			${Details} {
 				${StyledButton} {
-					margin: 9px 5px 0 0;
+					margin: 9px 4px 0 0;
 				}
 			}
 


frontend/src/v5/ui/routes/dashboard/projects/containers/uploadFileForm/uploadList/uploadListItem/components/uploadListItemDestination/uploadListItemDestination.component.tsx
```diff
@@ -169,9 +169,9 @@ export const UploadListItemDestination = ({
 
 	return (
 		<Autocomplete
```
The-Daniel, line 171: I'm not entirely sure why, and it may be because the uploads form definitely needs a rework, but the destination input is now very laggy when typing. Also the box doesn't change colour on error
This issue: image
Issue-3583: image

frontend/src/v5/ui/routes/dashboard/projects/containers/uploadFileForm/uploadList/uploadListItem/components/uploadListItemRevisionTag/uploadListItemRevisionTag.component.tsx
```diff
@@ -36,21 +37,26 @@ export const UploadListItemRevisionTag = ({
 	defaultValue,
 	...props
 }: IUploadListItemRevision): JSX.Element => (
-	<TextField
+	<ControlledInput
```
The-Daniel, line 40: For some reason this only lets you type one letter before it loses focus

frontend/src/v5/ui/routes/userSignup/userSignupForm/userSignupFormStep/userSignupFormStepAccount/userSignupFormStepAccount.component.tsx
```diff
@@ -108,7 +109,19 @@ export const UserSignupFormStepAccount = ({
 
 	return (
 		<>
-			<FormTextField
+			<ControlledInput
```
The-Daniel, line 112: This loses focus when an error is triggered.
So you can't type a username without it losing focus twice as the username needs to be 2 characters long

The-Daniel: The same happens with other fields in this step. It doesn't seem to be an issue with the fields in step 2

The-Daniel (changes requested):

frontend/src/v5/ui/routes/dashboard/projects/containers/uploadFileForm/uploadList/uploadListItem/components/uploadListItemDestination/uploadListItemDestination.component.tsx
Amantini1997: Are you sure you pulled the latest version? I remember having this problem as well, but I fixed it in one of my latest commits and I cannot reproduce this on my branch

frontend/src/v5/ui/components/shared/userMenu/editProfileModal/editProfilePersonalTab/editProfilePersonalTab.component.tsx
```diff
@@ -127,7 +128,8 @@ export const EditProfilePersonalTab = ({
 	return (
 		<ScrollArea>
 			<EditProfileAvatar user={user} />
-			<FormTextField
+			<ControlledInput
```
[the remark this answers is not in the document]
Amantini1997: This is indeed a problem, but the problem seems to be in the tickets branch. In https://issue-3583.dev.3drepo.io/v5/dashboard/Alessandro/1690b560-7936-11ec-95fe-6f78eefd726c/t/
image

frontend/src/v5/ui/controls/inputs/textArea/textAreaFixedSize.component.tsx
```diff
@@ -0,0 +1,59 @@
...
+	value = '', // this is to be certain that is a controlled field
+	height = 80,
+	...props
+}: TextAreaFixedSizeProps) => (
+	<FormControl required={required} disabled={disabled} error={error}>
+		{label && (
+			<InputLabel id={`${name}-label`}>
+				{label}
+			</InputLabel>
+		)}
```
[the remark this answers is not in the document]
Amantini1997: This problem only arises in Firefox as React-custom-scrollbars hides away and Firefox uses its own scrollbar. This branch is not really concerned with styling, so I guess it's better to create a new issue in case to address this problem

frontend/src/v5/ui/controls/inputs/controlledInput.component.tsx
```diff
@@ -16,32 +16,40 @@
...
+	Input: (props: T) => JSX.Element,
 	name: string,
 	control?: any,
 	formError?: any,
 	defaultValue?: any,
 };
-export const ControlledInput = ({ Input, name, control, formError, ...props }: ControlledInputProps) => (
+
+// eslint-disable-next-line
+export const ControlledInput = <T,>({
```
The-Daniel, line 39: This is interesting. Not saying it's wrong. I'm just curious what the trailing comma is about?

frontend/src/v5/ui/controls/inputs/passwordField/passwordField.component.tsx
```diff
@@ -15,9 +15,8 @@
...
  */
 
-import { TextField } from '@controls/inputs/textField/textField.component';
-import styled from 'styled-components';
+import { TextField, TextFieldProps } from '@controls/inputs/textField/textField.component';
 
-export const PasswordField = styled(TextField).attrs({
-	type: 'password',
-})``;
+export const PasswordField = (props: TextFieldProps) => (
```
The-Daniel, line 20: should login.component.tsx use this, or is the styling too different to not make it worth it?

The-Daniel (changes requested):

frontend/src/v5/ui/controls/inputs/controlledInput.component.tsx
Amantini1997: The trailing comma used this way is a special syntax that allows to use generics, but does not force them to be passed. In this case, `Input` takes some props which are dynamic, so defined by a generic. The generic type is then inferred automatically by typescript

frontend/src/v5/ui/controls/inputs/textArea/textAreaFixedSize.component.tsx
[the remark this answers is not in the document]
The-Daniel: Added the bug to this list: https://github.com/3drepo/3drepo.io/issues/3808

Amantini1997:
> 



> There are some text fields which act very laggy when a key is held down (They feel smooth when typing normally). Ones I've found:
> 
> * The password field in the user sign up (but not the confirm password)
> * All the fields in edit profile
> * Destination and revision name in the upload file form

- The password in the signup is unfortunately expected. This is due to the validation process which is apparently heavy for the password. The reason why you don't see it on `confirm password` as well is that the latter is only compared to the `password` to verify they are the same, so much less heavy computation

Concerning the others, I just wonder to what extent this fix is required. Normally, the user will not keep a key pressed down to fill in this fields, rather, they would type in some strings with different characters, But I will investigate the problem and try to fix

sanmont3drepo (changes requested):

frontend/src/v5/ui/controls/inputs/formInputs.component.tsx
```diff
@@ -0,0 +1,50 @@
...
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { Checkbox } from '@controls/favouriteCheckbox/favouriteCheckbox.styles';
```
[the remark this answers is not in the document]
Amantini1997: Fixed

frontend/src/v5/ui/controls/inputs/checkbox/checkbox.component.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { FormInputProps } from '@controls/inputs/inputController.component';
+import { FormControlLabel, FormControlLabelProps, Checkbox as MuiCheckbox } from '@mui/material';
+
+export type CheckboxProps = FormInputProps & FormControlLabelProps & {
+	onClick?: (event) => any,
+};
+
+export const Checkbox = ({ value, onClick, ...props }: CheckboxProps) => (
+	<FormControlLabel
+		control={<MuiCheckbox checked={value} onClick={onClick} />}
```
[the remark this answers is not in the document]
Amantini1997: Yes, it works. (By the time you reviewed this PR) Most of the event handlers could be passed and they were correctly triggered, with the exception of `onChange` which would have still been triggered, but would have overridden the default onchange behaviour of a controlled element preventing its value update (the same would have been true also for FormTextField). I fixed this in the `InputController` component.

Also, the fact that the `onClick` is extracted from the other props, is simply because if it passed to `FormControlLabel`, it is triggered twice on component click, so it has to be passed directly to the `MuiCheckbox`

The-Daniel:
@Amantini1997

> "The reason why you don't see it in confirm password as well is that the latter is only compared to the password to verify they are the same, so much less heavy computation." 

Then why does the password confirmation seem to lag in the edit profile modal? That only seems to be comparing itself to the new password as well.

> In general, I wonder to what extent this fix is required. Normally, the user will not keep a key pressed down to fill in this fields,

True, the only time it's done is when holding down delete to clear a field

Amantini1997:
> Then why does the password confirmation seem to lag in the edit profile modal? That only seems to be comparing itself to the new password as well.

If that happens while the "new password" is filled, that's because the validation checks are being re-executed on such a field everytime the "confirm password" changes

sanmont3drepo (approved):

The-Daniel (approved):

PR #3819, opened 2022-12-12
ISSUE #3420 - Add ellipsis when viewer card title is too long

### LINKED ISSUE #3420: Viewer Cards too long title not handled very well
### Description
When A Card (like Safetibase or Issues) has a too long title, the titlebar overflows the container

### Steps to replicate
- Open Safetibase
- Change the name to `xcvsxvcsdvvsxvcsdvvsxvsdvvsxvcsdvvsxvcsdvvsxvcsdvxcvsxvcsdvvsxvcsdvvsxvcsdvvsxvcsdvvsxvcsdvvsxvcsdvxcvsxvcsdvvsxvcsdvvs`

### Current Behaviour
- The title bar is too long.

### Expected Behaviour
- The title bar should fit in the container

image

Amantini1997:
This fixes #3420


frontend/src/v5/ui/v4Adapter/overrides/preview/previewDetails.overrides.ts
```diff
@@ -110,6 +115,12 @@ export default css`
 					margin: 0;
 					box-sizing: border-box;
 					height: 100%;
+					width: 100%;
```
The-Daniel, line 118: Not saying we do this in this PR as it should probably be a new PR, but might it be a good idea to add the CSS for truncating elements to the theme, a bit like how we do typography? Since we seem to duplicate these lines of CSS a lot. Of course there is the V4 Truncate component, but I don't think this can be used everywhere.

The-Daniel (approved):

The-Daniel:
Again, not sure if this is for this PR. Group names are truncated but can overflow when nested
image

frontend/src/v5/ui/v4Adapter/overrides/preview/previewDetails.overrides.ts
Amantini1997: Go ahead and list all the strings that overflow their parents, and then we may create a new issue to address them all or, as they all target just the styling (so the eventual PR will not be too hard to review), I might as well fix them all here. Would that be ok @sanmont3drepo?

sanmont3drepo (approved):

frontend/src/v5/ui/v4Adapter/overrides/preview/previewDetails.overrides.ts
sanmont3drepo: Created an issue for this #3827

PR #3820, opened 2022-12-12
ISSUE #3784 - Improve fetch strategy in custom tickets

### LINKED ISSUE #3784: Custom tickets - Improve fetch strategy in custom tickets
### Description
Right now in custom tickets we are fetching everything all the time, so not very efficient. We need to fetch only what's necessary to avoid putting unnecessary strain to our backend.

### Tasks
- [ ] take a look at all the fetchs being made.
- [ ] Manage a way to just make the necessary fetchs

The-Daniel:
This fixes #3784

#### Description
Removes unnecessary api calls for custom tickets. Stop updates to custom tickets redux state from refreshing the list

When the ticket list is opened fetchTickets and fetchTemplates are called
fetchTicket is now only called when opening ticket details
fetchTemplate is now only called when creating a ticket
fetchRiskCategories is now only called once when viewing the tickets list

#### Test cases
Use custom tickets with the network tab of Dev Tools open. Look out for any calls that seem unnecessary.
Close and reopen the tickets list and notice the list is instantly there and doesn't have to rerender as before


frontend/src/v5/ui/routes/viewer/tickets/ticketsList/ticketsListCard.component.tsx
```diff
@@ -47,6 +47,7 @@ export const TicketsListCard = () => {
 			containerOrFederation,
 			isFederation,
 		);
+		TicketsActionsDispatchers.fetchRiskCategories(teamspace);
```
Amantini1997, line 50: If this is moved to `tickets.component.tsx` it can be invoked inside a `useEffect` without dependencies and it would send many less requests to the backend. You can do that because the riskCategories depend on the teamspace rather than a ticket or a template.

The same may be also true for fetchTemplates and fetchTickets, but of course, in a different useEffect as you need the `containerOrFederation` as a dependency

```ts
// in tickets.component.tsx

	useEffect(() => { TicketsActionsDispatchers.fetchRiskCategories(teamspace); }, []);
	useEffect(() => {
		TicketsActionsDispatchers.fetchTickets(
			teamspace,
			project,
			containerOrFederation,
			isFederation,
		);
		TicketsActionsDispatchers.fetchTemplates(
			teamspace,
			project,
			containerOrFederation,
			isFederation,
		);
		TicketsActionsDispatchers.fetchRiskCategories(teamspace);
	}, [containerOrFederation]);
```

Amantini1997 (changes requested):

frontend/src/v5/store/tickets/tickets.redux.ts
```diff
@@ -71,7 +71,11 @@ export const replaceTemplateSuccess = (state: ITicketsState, { modelId, template
 	const modelTemplate = state.templatesByModelId[modelId]
 		.find((loadedTemplate) => loadedTemplate._id === template._id);
 
-	mergeWithArray(modelTemplate, template);
+	if (modelTemplate) {
+		mergeWithArray(modelTemplate, template);
+	} else {
+		state.templatesByModelId[modelId].push(template);
```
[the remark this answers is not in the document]
The-Daniel: I'm not sure I fully understand. Do you mean if a template is made that doesn't have properties, modules, and config? If so:

1. I'm not sure how common it would be (it's not a very useful ticket)
2. I'm not sure how you'd determine that there is nothing more to fetch without first doing a fetch
3. Since this is only one call when opening up ticket details I don't think it's that demanding

Oh, unless you mean if the state already includes B properties then there is no need to make the call again when opening/creating a ticket with that template. That's actually a good point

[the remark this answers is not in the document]
Amantini1997: yep, the last bit is exactly what I am referring to

frontend/src/v5/ui/routes/viewer/tickets/ticketsList/ticketItem/ticketItem.component.tsx
```diff
@@ -37,7 +37,7 @@ export const TicketItem = ({ ticket, onClick, selected }: TicketItemProps) => {
 	const { teamspace, project, containerOrFederation } = useParams<ViewerParams>();
 	const isFederation = modelIsFederation(containerOrFederation);
 	const template = TicketsHooksSelectors.selectTemplateById(containerOrFederation, ticket.type);
-	const hasIssueProperties = template?.config?.issueProperties;
+	const hasIssueProperties = ticket?.properties?.Priority;
```
sanmont3drepo, line 40: why the change in criteria?

sanmont3drepo (changes requested):

frontend/src/v5/ui/routes/viewer/tickets/ticketsList/ticketItem/ticketItem.component.tsx
The-Daniel: The config is only fetched when doing the singular fetchTemplate call which I have now moved into TicketDetails/NewTicket. Before I was running fetchTemplates and then looping over and running fetchTemplate, which was unnecessary.

sanmont3drepo (approved):

Amantini1997 (approved):

PR #3830, opened 2022-12-15
Issue 1699 - Model and Project permissions are removed when a user becomes TS admin

### LINKED ISSUE #1699: Remove lesser permissions if the user is assigned a over encompassing permission
### Description
When a user is assigned a higher level permission (e.g. project admin, teamspace admin), we don't wipe out the existing permissions. This means if the user is revoked of their admin privileges, they still have what they used to have. Which is probably not the desired behaviour.

### Steps to replicate
- give some model level access to a user
- then give them teamspace admin privilege
- revoke teamspace admin privilege
- The user still has the original model level access privileges.

ChristosTsiotsias:
This fixes #1699 

#### Description
When a user becomes admin of another teamspace all the permissions they have in projects/ models in this teamspace are removed.

When a user becomes admin of a project all the permissions they have in the models of this project are removed.

#### Test cases
1) 
Give project or model permissions to another user from your teamspace.
Make them TS admin.
Remove the admin permission.
All the permissions in projects/models should be gone.

2) 
Give model permissions to another user from your teamspace.
Make them Project admin.
Remove the admin permission.
All the permissions in models should be gone.


backend/src/v4/models/project.js
```diff
@@ -469,7 +469,7 @@
 		}
 
 		if (data.permissions) {
-			data.permissions.forEach((permissionUpdate) => {
+			data.permissions.forEach(async (permissionUpdate) => {
```
carmenfan, line 472: you'd need to turn the `forEach` into a `map` with a `Promise.all` if you change this function to async, otherwise we wouldn't be waiting on it

carmenfan (changes requested):

carmenfan (approved):

PR #3832, opened 2022-12-19
Issue 3821 - Filter out removed users from custom ticket assignee inputs

### LINKED ISSUE #3821: Removing a user from a teamspace where they have been assigned to a ticket can cause an error when reassigning users
### Description

Removing a user from a teamspace only removes their assignee status to custom tickets visually. They are still stored on the custom ticket object, so when updating the assignees the PATCH request still includes them which causes a `400 - INVALID_ARGUMENTS` error. The value does not update, so you are unable to change the assignees of any ticket which includes a removed user.

### Steps to replicate
- In the custom tickets card of the viewer assign USER A to a custom ticket.
- Remove the user from the teamspace (either through V4 or the V5 facelift if/when it is merged in)
- Return to the custom tickets card and you will see that the user no longer appears as an assignee
- Add USER B as an assignee to the ticket that contained USER A.
- An error modal will appear

### Current Behaviour
- An error modal appears warning of invalid arguments
- The assignees are not updated

### Expected Behaviour
- No error modal appears
- The assignees are updated

image

The-Daniel:
This fixes #3821

#### Description
Adds filters to two parts of the app which prevent bugs caused by users who have been assigned to a ticket but have then been removed from the teamspace.
The filter in MultiSelect prevents removed users from being included in the API calls. (Before the removed users were not displayed in the select list, however they would still be included in the api call. They could not be removed)
The filter in Assignees prevents removed users from affecting how many assignee circles were displayed before overflow. (If the displayed assignees included a removed user then the max assignees before overflow was reduced)

#### Test cases
Open up some custom tickets and add some assignees (a mix of users and jobs).
Remove some of these assignees from the teamspace (You might need to go to V4 for this). Delete some of the jobs (This is done via V4)
Return to the same custom tickets as before
- The removed users/jobs should not be displayed in the assignees
- Changing the assignees should not cause an error
- Check this in both the tickets list and ticket details


frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/manyOfProperty.component.tsx
```diff
@@ -30,9 +31,10 @@ export const ManyOfProperty = ({ values, ...props }: ManyOfPropertyProps) => {
 	} else {
 		items = (values as string[]).map((value) => ({ value, label: value }));
 	}
+	const filterImpossibleValues = intersection(props.value, items.map((i) => i.value));
```
sanmont3drepo, line 34: Can we change this to selectValues or validValues? 'filterImpossibleValues' has a name that seem more suitable for a function.

Also add a small comment that old values could be different because of changes in members of the teamspace or the template.

frontend/src/v5/ui/controls/assignees/assignees.component.tsx
```diff
@@ -40,11 +40,11 @@ export const Assignees = memo(({
 	disabled,
 	...props
 }: AssigneesProps) => {
-	const [values, setValues] = useState(initialValues);
-	const [open, setOpen] = useState(false);
-	const listedAssignees = initialValues.slice(0, max - 1);
-	const overflowAssignees = initialValues.slice(max - 1);
 	const allUsersAndJobs = UsersHooksSelectors.selectAssigneesListItems();
+	const [values, setValues] = useState(intersection(initialValues, allUsersAndJobs.map((i) => i.value)));
```
sanmont3drepo, line 44: can add a comment stating that initialValues could be invalid because of changes in the members of the teamspace?

sanmont3drepo (changes requested):

sanmont3drepo (approved):

sanmont3drepo (approved):

PR #3856, opened 2023-01-09
Issue 3847 - clone issue/risks from board includes viewpoint

### LINKED ISSUE #3847: Clone issue in v4 kanban doesnt clone the viewpoint
### Description
If you clone an issue in the kanban board,  when you enter the model the view is not changed to the original issue viewpoint.

- This is happening in: both 

### Steps to replicate
- Navigate to Issues & Risks
- Click on an Issue that has a viewpoint
- Click on 'Clone'
- Click on view on model

### Current Behaviour
- The view is not changed to the original view

### Expected Behaviour
- View should be changed to the original view

image

Amantini1997:
This fixes #3847 

#### Description
Cloning a risk or an issue from the board clones the viewpoint as well


frontend/src/v4/modules/risks/risks.sagas.ts
```diff
@@ -412,6 +411,16 @@ function* cloneRisk({ dialogId }) {
 		clonedProperties.descriptionThumbnail = `data:image/png;base64,${base64Image}`;
 	}
 
+	if (clonedProperties.viewpoint?.screenshot) {
+		const base64Image = yield imageUrlToBase64(activeRisk.descriptionThumbnail);
+		clonedProperties.viewpoint.screenshot = `data:image/png;base64,${base64Image}`;
+	}
+
+	if (clonedProperties.viewpoint?.screenshotSmall) {
```
The-Daniel, line 419: I think using the same method of getting the `screenshot` for `screenshotSmall` is causing the thumbnails to be stretched
image

The-Daniel:
A few issues when creating a new Risk (Might all be related). 

1. Click the create new risk button. Before pressing save add an image. Now when you click save this image is lost.
2. The old behaviour was that a viewpoint got made when creating a Risk (with an accompanying screenshot). Now no viewpoint or image is made on creation.
3. If you update the image and then instantly try to clone nothing happens and you cannot even return to the list of risks

Amantini1997:
> 

The 3rd point seems to already be present in staging, it may be worth it to be open a new issue for it

The-Daniel (approved):

sanmont3drepo (approved):

The-Daniel (approved):

sanmont3drepo (approved):

PR #3862, opened 2023-01-10
Issue 3803 - GIS fields are not required and refuse text inputs

### LINKED ISSUE #3803: GIS Survey Point fields should not be required and should not accept text
### Description

The GIS Survey point values in Container/Federation settings should not be required. If their value is empty it should default to 0.

Only number values should be allowed in these fields.

- This is happening in: both 

### Steps to replicate
- Open either the container or federation settings
- Try typing text in one of the number fields
- Try deleting the value of one of the fields (it will block the form from being able to submit)

### Current Behaviour
- Text can be inputted to the number fields
- If a value is deleted the field shows an error and the user cannot submit the form

### Expected Behaviour
- Only numbers can be inputted (Perhaps use the number input that was made for custom tickets)
- A field that is empty defaults to 0 on submit

image
In V4 the fields are not required and default to 0:
image

Amantini1997:
This fixes #3803 

#### Description
- GIS values do not accept text inputs
- GIS values are not required and can be empty (if so they are mapped to 0)

### DIFF
--- a/frontend/src/v5/ui/routes/dashboard/projects/settingsForm/settingsForm.component.tsx
+++ b/frontend/src/v5/ui/routes/dashboard/projects/settingsForm/settingsForm.component.tsx
@@ -16,6 +16,7 @@
  */
 
 import { useEffect, useState } from 'react';
+import { defaults, pick, omitBy, difference, isMatch, mapValues } from 'lodash';
 import { formatMessage } from '@/v5/services/intl';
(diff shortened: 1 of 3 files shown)


The-Daniel:
This is kind of inconsequential but I've noticed if you delete a coordinate value and type in the original value again the "Save changes" button becomes enabled even though there are no changes. I'm not really sure this causes any issues though, just thought I'd raise it

The-Daniel (approved):

PR #3865, opened 2023-01-11
ISSUE #3864 Perform functions in a mongo aggregate on application level instead to support large models

### LINKED ISSUE #3864: Error "Failed to remove data associated with model"
**Description**

Customer gets a "Failed to remove data associated with model (Status Code: 500)" message when trying to delete a model container.

This is happening in production , hotfix is up to Product director to decide.

Steps are difficult to replicate. Customer tried to upload a model which failed and then attempted to delete the model container.

**Current Behaviour**

"Failed to remove data associated with model (Status Code: 500)" message when trying to delete the container

**Expected Behaviour**

For the model container to be removed when the delete button is hit. 

**Pictures/ Gifs of the error**

image

carmenfan:
This fixes #3864

#### Description
The bug is caused by an aggregate pipeline we perform to find all external references we need to remove when we remove a model. The aggregate produce data in the following format:

````js
[{
  _id: "link type", // e.g. "fs" or "gridfs"
  links: ["reference links"] //e.g. ["250/12/b9d175c7-d14b-4436-be83-d505aa2f17b2"] for fs
}]
````
In cases where there are a lot of meshes, where each mesh will typically produce 3 references, we may end up with a lot of links, which may be large enough for BSON sizes to exceed 16MB, hence the aggregate returns an error. The situation @Ramadyfm found had 366108 references in `scene.ref`

This PR fixes the problem by doing a standard `find` instead and let 3drepo.io build the reference objects.

This problem can be reproduced by creating dummy bson documents into a scene via this [script](https://www.dropbox.com/s/68ppyam98t6gm0u/insertManyRefs.js?dl=0) (insert into the utility script folder to run)
image

#### Test cases
the model should now delete successfully (may take a while but it should)

### DIFF
--- a/backend/src/v5/models/fileRefs.js
+++ b/backend/src/v5/models/fileRefs.js
@@ -47,13 +47,19 @@ FileRefs.getTotalSize = async (teamspace, collection) => {
 	return res.length > 0 ? res[0].total : 0;
 };
 
-FileRefs.getAllRemovableEntriesByType = (teamspace, collection) => {
-	const pipeline = [
-		{ $match: { noDelete: { $exists: false }, type: { $ne: 'http' } } },
-		{ $group: { _id: '$type', links: { $addToSet: '$link' } } },
-	];
+FileRefs.getAllRemovableEntriesByType = async (teamspace, collection) => {
+	const query = { noDelete: { $exists: false }, type: { $ne: 'http' } };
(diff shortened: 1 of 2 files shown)


ChristosTsiotsias (approved):

mosment3DRepo (approved):
tested with customer model that was previously failing

PR #3867, opened 2023-01-12
ISSUE #3824 - V5 Repeated calls to containers endpoint when looking at Federations

### LINKED ISSUE #3824: V5 Repeated calls to containers endpoint when looking at Federations
### Description

When accessing the Federations list, the API seems to call the same /containers endpoint for each federation. So 20 federations = 20 duplicate /container queries

- This is happening in: both 

### Steps to replicate
- Make sure there is >1 federation in your teamspace
- Open the federation list in V5 with the Network tool open (https://3drepo.io/v5/dashboard/{teamspace}/{project}/t/federations)
- Notice the repeater /containers calls

image

image

The-Daniel:
This fixes #3824

#### Description

This was caused by the `useContainersData()` in the edit federation modal. By changing the render method of this modal to use the new `DialogsActionsDispatchers.open()` method the component only renders (thus `useContainersData()` is called) when the modal is opened.
Some changes were made to how the teamspace and project are fetched since `useParams()` doesn't work with the action dispatching method of opening modals. Instead I have used redux selectors to get these values. I think this shouldn't cause issues, but if you think otherwise do say.
I also fixed a small problem where the save changes button would be disabled if there were no containers included

#### Test cases

- open up devtools with the Network tab open.
- Go to the federations list and look to see if there are any unnecessary fetchContainers calls
- Open the editFederation modal and check it all works fine


frontend/src/v5/ui/routes/dashboard/projects/federations/editFederationModal/editFederationModal.component.tsx
```diff
@@ -58,10 +51,13 @@ export const EditFederationModal = ({
 		event.preventDefault();
 		onClickClose();
 	};
-
+	const formIsDirty = () => !isEqual(
+		compact(includedContainers).map((c) => c._id),
```
Amantini1997, line 55: The `compact` method is used to filter out falsey values, but do we ever get them here?

The-Daniel: You're right. You used to briefly get undefined values when the form was loading but I fixed it closer to the source of the problem by adding the `if (containers.length)` in editFederation.component. I forgot to remove this "compact" solution, but have now fixed it

Amantini1997 (dismissed):

sanmont3drepo:
There is one false request:
image

It seems that in containers hooks it takes a bit to get the the actual teamspace and project values

sanmont3drepo (changes requested):
There is one request with invalid values:
image

It seems that in containers hooks it takes a bit to get the the actual teamspace and project values.

It happens when you enter the containers url or when you refresh the page

Amantini1997 (approved):

PR #3872, opened 2023-01-13
Issue 3770 - Searchbar in multiselect loses focus

### LINKED ISSUE #3770: Searchbar in Search Select control loses focus when editing the search query
### Description

Trying to filter the items in a search select can be difficult as the searchbar occasionally loses focus. This might be happening when only one selected item remains in the list as the component seems to autoFocus it, however there does seem to be some occasions where this isn't the case (as seen in the steps to replicate below)
Applying `autoFocus={false}` to the menu item doesn't seem to work

### Steps to replicate
- Have a ticket with the assignees "Asset Manager" and "Main Contractor"
- Type in "MA". This will filter to this.
image
The searchbar will still have focus
- Continue to enter "N"
image
The searchbar now loses focus and the "Asset Manager" item gains focus.
- if you were instead to type "I" 
image
There is only one selected item left, yet the searchbar still retains focus.
- And if you type "MAN" as before to lose focus, and then delete the "N" the searchbar again loses focus when there are multiple selected values.
image

(Results may vary if you have users/custom roles which fit these queries)

### Current Behaviour
- The searchbar can lose focus when writing/editing a search query

### Expected Behaviour
- The searchbar does not lose focus when typing/deleting a query

Amantini1997:
This fixes #3770 

#### Description
The searchbar now keeps the focus while typing

### DIFF
--- a/frontend/src/v5/ui/controls/searchSelect/searchSelect.component.tsx
+++ b/frontend/src/v5/ui/controls/searchSelect/searchSelect.component.tsx
@@ -39,7 +39,7 @@ export const SearchSelect = ({ children, ...props }: SelectProps) => {
 		<SearchContextComponent filteringFunction={filterItems} items={children as ReactNode[]}>
 			<SearchContext.Consumer>
 				{ ({ filteredItems }: SearchContextType<typeof MenuItem>) => (
-					<Select {...props}>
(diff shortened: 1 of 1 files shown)


The-Daniel (approved):
This might be for another issue, but it would probably be good if the search bar gained focus when opening the multiselect

PR #3873, opened 2023-01-16
Issue 3063 - Too many modal when trying to access inaccessible model

### LINKED ISSUE #3063: Too many alert modals appear when loading an inaccessible or non-existent model
### Description
- When multiple errors are dispatched multiple alert modals appear.
- This is most notable for containers stats. If the request 404s for all containers then you will get a modal for each container.
- This means you have to cross off many modals to return to the screen
- Once you close all the error's you are then left on a blank 3D viewer screen with a broken bredcrumb
- Clicking the teamspace of the breadcrumb will then take you to a teamspace which you still do not have access to but the UI still displays the teamspace name

### Steps to replicate
- Visit a model you do not have access to (eg. https://staging.dev.3drepo.io/v5/viewer/testmosment3583/1c9a27c0-7494-11ed-b0c6-511e6965aa9f/1cc72b31-7494-11ed-b0c6-511e6965aa9f)
- Click OK on the error messages
- See that you are now in a blank viewer page

### Current Behaviour
- A modal appears for each error

### Expected Behaviour
- Only one error should appear (Or one error per error type?)
- Once you click OK you should be redirected to 3drepo.io/v5/dashboard

Amantini1997:
This fixes #3063

#### Description
When the URL contains wrong segments (like a wrong teamspace/project/resource id) or the user is trying to access a teamspace they do not have access to, such a user is going to be redirected:
- teamspace id is wrong (or not accessible) -> the user will be redirected to the dashboard;
- project id is wrong -> the user will be redirected to the teamspace page;
- resource id is wrong -> the user will be redirected to the project page.

#### Test cases
Access a model in the viewer and try to change the URL segment once for each of the following:
- teamspace;
- project;
- federation/container;
You should get redirected as described above

Then try the same in the dashboard (excluding the federation/container as that bit will not be in the URL)


frontend/src/v5/store/dialogs/dialogs.redux.ts
```diff
@@ -32,6 +33,11 @@ export const { Types: DialogsTypes, Creators: DialogsActions } = createActions({
 }, { prefix: 'MODALS/' }) as { Types: Constants<IDialogsActionCreators>; Creators: IDialogsActionCreators };
 
 export const openHandler = (state, { modalType, props }: OpenAction) => {
+	// avoid opening 2+ redirect modals
```
The-Daniel, line 36: This causes a crash when a modal doesn't contain an error. For instance the confirm delete, upload files, and revision still processing info modals.

The-Daniel (changes requested):

mosment3DRepo (approved):

The-Daniel (approved):

sanmont3drepo (approved):

PR #3874, opened 2023-01-16
ISSUE #3845 - v4 subModels fix and migration

### LINKED ISSUE #3845: v4 submodel records need a migration
### Description
2 things:

#### 1. v4 is writing the submodels record at the point of the API request
when you request to edit a federation, in v4 we immediately record the submodels into the federation settings. This is inaccurate as the federation will not be updated until the bouncer has processed it. 

in v5, we leave it as it is and 3drepo.io only updates the submodels field when the bouncer has finished processing. Since v5 is now responsible for listening to queue messages, v4 no longer need to update the submodel field at all as it will be done by v5. We should remove it for accuracy.

#### 2. We need to write a migration script to unify the way we store submodels in federation settings. 
In legacy v4 schema we store submodels like this:
image

But in v5, we just store an array of model ids. At the moment we are creating some workaround code to make both work. We should instead have a migration script to migration all existing data and remove/change anything that expects [Object].

### Goals
- Migrate old schema in database
- clean up any irrelevant code

Charence:
This fixes #3845 

#### Description
- v4 is writing the submodels record at the point of the API request
- We need to write a migration script to unify the way we store submodels in federation settings.

#### Test cases
- Migrate old schema in database
- clean up any irrelevant code


backend/src/v4/models/helper/model.js
```diff
@@ -560,11 +560,7 @@ function listSubModels(account, model, branch = "master") {
 			findModelSettingById(ref.owner, ref.project, { name: 1 }).then(subModel => {
 				// TODO: Why would this return null?
 				if (subModel) {
-					subModels.push({
-						database: ref.owner,
-						model: ref.project,
-						name: subModel.name
-					});
+					subModels.push(ref.project);
```
carmenfan, line 563: is this right? `getModelSettings` also uses this function and as it stand we wouldn't be giving the submodel info that we used to give by changing this?

backend/src/v4/models/helper/model.js
```diff
@@ -658,7 +654,7 @@ function isSubModel(account, model) {
 
 		feds.forEach(modelSetting => {
 			promises.push(listSubModels(account, modelSetting._id).then(subModels => {
-				return subModels.find(subModel => subModel.model === model);
+				return subModels.find(subModel => subModel === model);
```
carmenfan, line 657: This is a very dumb function to check if a model is a submodel....

All we need is a single query like in v5 https://github.com/3drepo/3drepo.io/blob/ISSUE_3845/backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/containers.js#L32

carmenfan (changes requested):
Did we break BCF? I didn't look too deeply but a quick grep shows it's trying to reference `.model` here:
image

carmenfan (approved):

PR #3890, opened 2023-01-18
Issue 3860 - Create a v5 version of mongo handler

### LINKED ISSUE #3860: Migrate Mongo driver to v5
### Description
At the moment we're just using v4's `handler/db.js` on v5. We should migrate all functions that are currently needed for v5 and write adequate driver tests to ensure services will not be disrupted if we upgrade the database/drivers.

Whilst we're at it, we should also fix the problem where if the app fails to connect to mongo at init it no longer errors out.

### Goals
- [ ] migrate all functions in `handler/db.js` that is used by v5 to v5's file
- [ ] add driver test for all of the functions migrated

carmenfan:
This fixes #3860

#### Description
- v5 db handler no longer relies on v4 db  handler
- added driver tests for all functionality
- no longer expose `runCommand`

#### Test cases
all v5 functionality should function as before


backend/tests/v5/drivers/handler/db.test.js
```diff
@@ -0,0 +1,934 @@
...
+			await expect(DB.deleteOne(dbName, col, { _id: data[5] })).resolves.toBeUndefined();
+			await expect(DB.find(dbName, col, { _id: data[5] })).resolves.toEqual([]);
+		});
+
+		test('Should return successfully if the document was not found', async () => {
+			await expect(DB.deleteOne(dbName, col, { [generateRandomString()]: generateRandomString() }))
+				.resolves.toBeUndefined();
+		});
+
+		test('Should return empty array if collection doesn\'t exist', async () => {
```
ChristosTsiotsias, line 566: return undefined*

backend/tests/v5/drivers/handler/db.test.js
```diff
@@ -0,0 +1,934 @@
...
+			await expect(DB.deleteOne(dbName, generateRandomString(), { })).resolves.toBeUndefined();
+		});
+
+		test('Should return empty array if database doesn\'t exist', async () => {
```
ChristosTsiotsias, line 570: return undefined*

ChristosTsiotsias (changes requested):

backend/tests/v5/unit/models/roles.test.js
```diff
@@ -44,13 +39,11 @@ const testRemoveTeamspaceRole = () => {
 	describe('Remove teamspace role', () => {
 		test('should remove the teamspace role', async () => {
 			const teamspace = generateRandomString();
-			const expectedCommand = {
-				dropRole: TEAM_MEMBER,
-			};
-			const fn = jest.spyOn(db, 'runCommand').mockImplementation(() => { });
+
+			const fn = jest.spyOn(db, 'dropRole').mockImplementation(() => { });
```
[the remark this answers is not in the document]
carmenfan: I don't disagree, I didn't write the tests (these are old tests before we knew better) I just changed the function name 😆

ChristosTsiotsias (approved):

PR #3918, opened 2023-01-27
ISSUE #3878 - Replace embedded dialog forms and use dialog dispatchers

### LINKED ISSUE #3878: Replace embeded dialog forms and use dialog dispatchers to have a more uniform way of using dialogs
### Description
We are using two different ways for formdialogs. One is using an action and the other one is embedding the form dialog in the component that dispatches it. Over time we noticed that using just the dispatcher should cater for every case, providing a standard way of using dialogs.

### Suggested solution
Use dispatch dialogs. Ideally it shouldn't  decrease the readability of the code, each form dialog should be tackled on a case-by-case basis to guarantee that.

### Goal
- [ ] There's only one way of dispatching dialog forms.

### Tasks
- [ ] Take a look at each particular form dialog
- [ ] Replace the open/close logic on the component that dispatches the dialog to a DispatchDialog call

The-Daniel:
This fixes #3878

#### Description
Replaces embedded modals with the `DialogsActionsDispatchers.open()` method
`useParams()` does not work inside the new modal method, so this has been replaced by hooks selectors where necessary

#### Test cases
All modals should work as before. (If there are any `useParams()` still inside any of the modals this will cause errors)
Check there have been no modals missed


frontend/src/v5/ui/routes/dashboard/projects/federations/federationsList/federationListItem/federationListItem.component.tsx
```diff
@@ -78,11 +71,9 @@ export const FederationListItem = memo(({
...
 	};
 
-	const onClickEdit = () => {
-		DialogsActionsDispatchers.open(EditFederationModal, {
-			federation,
-		});
-	};
+	// eslint-disable-next-line max-len
+	const onClickSettings = () => DialogsActionsDispatchers.open(FederationSettingsForm, { federationId: federation._id });
+	const onClickEdit = () => DialogsActionsDispatchers.open(EditFederationModal, { federation });
```
Amantini1997, line 76: We are using _FederationSettings**Form**_ and _EditFederation**Modal**_. I know they were already named this way, but could you please rename them so that they use a common suffix?

frontend/src/v5/ui/routes/dashboard/projects/federations/federationsList/federationListItem/federationListItem.component.tsx
```diff
@@ -78,11 +71,9 @@ export const FederationListItem = memo(({
 		}
 	};
 
-	const onClickEdit = () => {
-		DialogsActionsDispatchers.open(EditFederationModal, {
-			federation,
```
Amantini1997, line 83: why the replacement of prop `federation` with `federationId`?

frontend/src/v5/ui/routes/dashboard/projects/federations/federationSettingsForm/federationSettingsForm.component.tsx
```diff
@@ -17,25 +17,27 @@
...
-	open: boolean;
-	federation: IFederation;
-	onClose: () => void;
+	federationId: string;
+	onClickClose: () => void;
 };
 
 export const FederationSettingsForm = ({
-	federation,
+	federationId,
```
Amantini1997, line 29: why the change from `federation` to `federationId`?
(same for container in the respective component)

frontend/src/v5/ui/routes/dashboard/projects/settingsForm/settingsForm.component.tsx
```diff
@@ -212,29 +210,28 @@ export const SettingsForm = ({
...
-		if (open) {
-			fetchSettings(teamspace, project, containerOrFederation._id);
-			fetchViews(teamspace, project, containerOrFederation._id);
-			reset(getDefaultValues(containerOrFederation, isContainer));
-			setAlreadyExistingNames([]);
-		}
-	}, [open]);
+		fetchSettings(teamspace, project, containerOrFederation._id);
+		fetchViews(teamspace, project, containerOrFederation._id);
+		setAlreadyExistingNames([]);
```
Amantini1997, line 221: I think you can get rid of this line now

Amantini1997 (changes requested):

Amantini1997:
### Closing the modal feels different
- there is a lag between the click to close (on the cross or on the blurred background) and the actual modal close
- the modal used have a fade away effect on close, that effect is no longer there

Amantini1997:
### Visual settings modal doesn't close when clicking on the blurred background
This problem already existed in staging, but I think it may still be related to this more less

frontend/src/v5/ui/routes/dashboard/projects/federations/federationSettingsForm/federationSettingsForm.component.tsx
The-Daniel: A problem with this method is that the props supplied to it can't be mutated. So when the FederationsSettingsForm calls fetchViews and fetchSettings even though the federation is updated in redux it wasn't being updated in the modal prop.

I'm starting to feel like this method might be good for some applications but not others, as it does carry with it a lot of potential to create bugs

sanmont3drepo:
> ### Visual settings modal doesn't close when clicking on the blurred background
> This problem already existed in staging, but I think it may still be related to this more less

Visual settings is a v4 restyling so  is not using the v5 popup system.

The-Daniel:
> > ### Visual settings modal doesn't close when clicking on the blurred background
> > This problem already existed in staging, but I think it may still be related to this more less
> 
> Visual settings is a v4 restyling so is not using the v5 popup system.

It actually does use the v5 system. I've fixed it a long with the other issues that Alessandro raised

Amantini1997:
To fix the issue with the fade away animation not being triggered, the "open" prop was added back to the different modals/dialogs.

I have also noticed that all the dialogs and modals have that blurring bluish background.

I was thinking, what if we have a wrapper component (around the dispatched dialogs) that adds that background and that also takes the open prop? This way we do not have to worry about defining that one prop every time and the different form components are simply forms and not modals anymore, but wrapped up by modals

sanmont3drepo (approved):

Amantini1997 (approved):

PR #3930, opened 2023-01-31
ISSUE #3907 - Search Select shows about 6 items at most

### LINKED ISSUE #3907: Custom tickets: selects can get too large
### Description
Highlighting objects on a federation node that would highlight multiple sub models is not functioning.

- This is happening in: both 

### Steps to replicate
- Create a custom ticket template with lots of options (30+)
- Create a custom ticket using that template 

### Current Behaviour
- The list takes the whole page height

### Expected Behaviour
- The list should show a scrollbar and 5 or 6 options max

image

Amantini1997:
This fixes #3907 

#### Description
The search input is sticky and the whole box comes with a max height

#### Test cases
Edit the assignees list

### DIFF
--- a/frontend/src/v5/ui/controls/searchSelect/searchSelect.component.tsx
+++ b/frontend/src/v5/ui/controls/searchSelect/searchSelect.component.tsx
@@ -39,7 +39,15 @@ export const SearchSelect = ({ children, ...props }: SelectProps) => {
 		<SearchContextComponent filteringFunction={filterItems} items={children as ReactNode[]}>
 			<SearchContext.Consumer>
 				{ ({ filteredItems }: SearchContextType<typeof MenuItem>) => (
-					<Select {...props} MenuProps={{ disableAutoFocusItem: true }}>
+					<Select
+						{...props}
+						MenuProps={{
+							disableAutoFocusItem: true,
+							PaperProps: {
+								style: { maxHeight: 231 },
+							},
+						}}
+					>
 						<SearchInputContainer>
 							<SearchInput
 								placeholder={formatMessage({ id: 'searchSelect.searchInput.placeholder', defaultMessage: 'Search...' })}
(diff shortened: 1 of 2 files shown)


frontend/src/v5/ui/controls/searchSelect/searchSelect.component.tsx
```diff
@@ -39,7 +39,15 @@ export const SearchSelect = ({ children, ...props }: SelectProps) => {
 		<SearchContextComponent filteringFunction={filterItems} items={children as ReactNode[]}>
 			<SearchContext.Consumer>
 				{ ({ filteredItems }: SearchContextType<typeof MenuItem>) => (
-					<Select {...props} MenuProps={{ disableAutoFocusItem: true }}>
+					<Select
+						{...props}
+						MenuProps={{
+							disableAutoFocusItem: true,
+							PaperProps: {
+								style: { maxHeight: 231 },
```
sanmont3drepo, line 47: cant we style this in the searchSelectstyles.ts?
```
export const SearchInput = styled(SearchInputBase)`
	margin: 0;
	padding: 2px 12px 12px;
	${Paper} {
	   max-height: 231px;
	}
	
```

sanmont3drepo (changes requested):

The-Daniel (commented):
Just the thing Santiago suggested

frontend/src/v5/ui/controls/searchSelect/searchSelect.component.tsx
Amantini1997: Unfortunately, this will not work because the `Paper` is not rendered as a descendant of the `Select` component. 
image

It would be possible to style it together with the other `v4Overrides.styles` components, but I think that's not the best idea

sanmont3drepo: Nope that's for v4 restyling

sanmont3drepo (dismissed):

sanmont3drepo:
> > image Could this be an okay PR to suggest that the select list items get the whole max-width/truncation job?
> 
> I guess so. The PR is very small. If @sanmont3drepo is ok, I will do it

Nah, lets close it this way.

sanmont3drepo (approved):

The-Daniel (approved):

PR #3935, opened 2023-02-01
ISSUE #3934 - The custom ticket gets wiped out after closing the panel

### LINKED ISSUE #3934: The custom ticket gets wiped out after closing the panel
### Description
The information on the custom ticket gets wiped out after closing the panel

### Steps to replicate
- Open a fedeeration/container
- Create a new custom ticket or navigate an old one
- Close the panel
- Open it again

### Current Behaviour
- The information on the ticket gets wiped out

### Expected Behaviour
- The information should be still there 

blank ticket

Amantini1997:
This fixes #3934 

#### Description
The problem was this:
- when opening the ticket card, the base templates where always fetched;
- the base template, however, does not have the information needed to render an actual ticket, which is fetched on opening the actual ticket;
- closing the ticket card (while the view was set on the ticket details) and then reopening it, forced the base templates to be refetched (wiping out the exisisting data), but without refetching the template details

#### Test cases
Open up a ticket details, close the ticket card and re-open it

### DIFF
--- a/frontend/src/v5/ui/routes/viewer/tickets/tickets.component.tsx
+++ b/frontend/src/v5/ui/routes/viewer/tickets/tickets.component.tsx
@@ -55,13 +55,13 @@ export const Tickets = () => {
 				enableRealtimeContainerUpdateTicket(teamspace, project, containerOrFederation),
 			);
 		}
-		TicketsActionsDispatchers.fetchTickets(
-			teamspace,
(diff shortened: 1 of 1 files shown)


The-Daniel:
Property values are lost upon reopen
image
image

sanmont3drepo (approved):

PR #3955, opened 2023-02-10
ISSUE #3954 - fix intercom

### LINKED ISSUE #3954: Intercom does not load
### Description
Intercom does not load correctly

- This is happening in: both

### Steps to replicate
- Open the app

### Current Behaviour
- Intercom wither does not appear or it does not work

### Expected Behaviour
- Intercom should load and work correctly

Amantini1997:
This fixes #3954

#### Description
Intercom should work as usual

#### Test cases
1) Log in, refresh the page, the intercom should still be there
2) Click on the intercom, the content should load correctly

### DIFF
--- a/frontend/src/v5/ui/components/intercom/intercom.component.tsx
+++ b/frontend/src/v5/ui/components/intercom/intercom.component.tsx
@@ -54,16 +54,9 @@ const IntercomButton = () => {
 export const Intercom = () => {
 	const { intercomLicense } = clientConfigService;
 
-	let appId = intercomLicense;
-	const commentStartIndex = intercomLicense.indexOf('#');
-
-	if (commentStartIndex) {
-		appId = intercomLicense.slice(0, commentStartIndex).trim();
-	}
-
 	return (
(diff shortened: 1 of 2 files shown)


frontend/src/v5/ui/components/intercom/intercom.component.tsx
```diff
@@ -41,21 +41,29 @@ const IntercomButton = () => {
...
+	}, [isAuthenticated, currentUser.email]);
 
 	return <div />;
 };
 
 export const Intercom = () => {
 	const { intercomLicense } = clientConfigService;
+
+	let appId = intercomLicense;
+	const commentStartIndex = intercomLicense.indexOf('#');
```
carmenfan, line 58: @Amantini1997 get rid of this hack, you should assume what's coming in is correct. It's only a problem in local (create an issue in local so @keifgwinn can deal with it)

carmenfan (changes requested):

carmenfan (approved):

PR #3972, opened 2023-02-15
ISSUE #3971 - Delete /idbfs before viewer loads

### LINKED ISSUE #3971: Delete the idbfs IndexedDb database before the viewer loads
### Description

Before introducing the asynchronous IndexedDb cache, user assets were cached using the System.IO namespace which used the idbfs database.
Emscripten loads this on start-up in order to emulate a synchronous API, meaning even with the new cache, the old files could still have an impact on start-up time and memory.

- This is happening in: both 

### Steps to replicate
- Open any model with a browser that has made extensive use of the viewer before v5.

### Current Behaviour
- The contents of /idbfs are loaded into memory

### Expected Behaviour
- The contents of /idbfs are not loaded into memory

sebjf:
This fixes #3971

#### Description
This PR prepends the viewer initialisation promise chain with a function that deletes the /idbfs database. This is done before loading as the consequences of removing /idbfs on a running emscripten module are not known, and more importantly, emscripten loads the contents of /idbfs into memory at start-up, so start-up times are minimised by deleting the content before the viewer gets to this point.

This snippet will run every time the viewer runs from now on. The viewer does not use the System.IO namespace anymore, so there is nothing to be lost by removing this store each time.

#### Test cases
With this in place, each time the viewer loads the /idbfs should be empty save for one or two small entries created at or after the last time the viewer was started.
Any records dating before the latest run indicate /idbfs is not being cleared as expected.

### DIFF
--- a/frontend/src/globals/unity-util.ts
+++ b/frontend/src/globals/unity-util.ts
@@ -142,6 +142,24 @@ export class UnityUtil {
 	}
 
 	/**
+	 * Removes the IndexedDb database /idbfs, which emulates a synchronous
+	 * filesystem.
+	 * The viewer should not store anything use the File API between runs.
+	 */
+	private static clearIdbfs(): Promise<void> {
+		const deleteRequest = indexedDB.deleteDatabase('/idbfs');
+		return new Promise((resolve) => {
+			deleteRequest.onsuccess = () => {
+				resolve();
+			};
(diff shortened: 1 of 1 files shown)


carmenfan (approved):
Merging in and testing the cache in staging as cache is not available in a brand new branch

Issue #3887, opened 2023-01-17
Release Candidate 5.0.5 


mosment3DRepo:
Previously this was an issue for V5: Remove menus/buttons for non-admin users, but now it will contain all hotfixes needed for 5.0.5

### Description

Currently the APIs block users without the relative permissions from performing certain actions, but the UI should reflect this too

### Goals

- [ ] Remove options for non-admins to access areas of the app/commands they are blocked from performing
- [ ] Teamspace & Project Admins should have access, anyone with Viewer, Commenter, Collaborator access to a container/fed within the teamspace should have the options removed

### Tasks
See https://docs.google.com/spreadsheets/d/1DJWz09Nwi59TtbwRE5r4FPMHWcDsNBTaZWNnx7Ax7_U/edit#gid=289465492 for live version
![image](https://user-images.githubusercontent.com/62879582/214833686-ed7f0ff1-d3fb-4f3a-88bf-b175b4888e8a.png)

carmenfan:
> ![image](https://user-images.githubusercontent.com/11945337/212923241-0be959e8-7607-4b11-ae44-51f6f282759e.png)

@mosment3DRepo I think you might've deleted too many things here

Collaborators of a federation (whatever that means) would have `edit federation` rights, and even viewers would be able to look at the settings, they just can't edit them.


Similar with containers

The-Daniel:
@mosment3DRepo 
1. What should happen when a non-admin user tries to access an authorised page (e.g. by putting the URL for teamspace settings into the browser). Should it redirect to the projects tab of the same teamspace? Should it trigger an error modal?

2. Should there be some restrictions on the Teamspace Users tab?

mosment3DRepo:
> 1. What should happen when a non-admin user tries to access an authorised page (e.g. by putting the URL for teamspace settings into the browser). Should it redirect to the projects tab of the same teamspace? Should it trigger an error modal?
We should redirect them to the project page with a modal saying insufficient access (we can use the API response)


> 2. Should there be some restrictions on the Teamspace Users tab?
"Project Permissions" and "User Permissions" should be hidden from users whihc are not teamspace or project admins

mosment3DRepo:
@The-Daniel  I have updated the issue with a table and a link to google sheets, it should make more sense now

PR #3997, opened 2023-02-21
ISSUE 3990 - allow unity-util to work without index db

### LINKED ISSUE #3990: Updates to unity-util incompatible with PowerBI
### Description

`deleteDatabase` and `indexedDbCache` prevent the viewer from loading on PowerBI. Also, models are not loading as expected.

- This is happening in: ~prod/staging/~ both 

### Steps to replicate
- Use 3D Repo PowerBI custom visual

### Current Behaviour
- Doesn't load
- 
### Expected Behaviour
- Should load as usual

Charence:
This fixes #3990 

#### Description
- Use of IndexDB not supported on PowerBI causing visual to fail
- try/catch indexDB operations

#### Test cases
- viewer should work as usual on web
- viewer should work in Power BI


carmenfan (changes requested):
@Charence  if you merge #3887 in it should contain sebastian's fixes on indexedDB cache.

I think the only one you need to catch is clearIdbfs at this point but it'd be good to check.

carmenfan (approved):

PR #4006, opened 2023-02-23
ISSUE 4005 IndexedDbCache cannot be created by an embedded viewer

### LINKED ISSUE #4005: IndexedDbCache cannot be created by an embedded viewer
### Description

Embedding the viewer into a third party site doesn't work because the IndexedDbCache Web Worker cannot be instantiated due to a security issue.

- This is happening in: both 

The workaround is to disable caching: https://github.com/3drepo/3drepounity/issues/453

### Steps to replicate
- Serve the `3drepo.io\demo` project in a simple Http server. (e.g. open the demo folder in VSCode and use the GoLive plugin)
- Insert the API key for .io (prod)

### Current Behaviour
- See in the console the error:

```
exception thrown: SecurityError: Failed to construct 'Worker': Script at 'https://www.3drepo.io/unity/indexeddbworker.js' cannot be accessed from origin 'http://localhost:5500'.,Error: Failed to construct 'Worker': Script at 'https://www.3drepo.io/unity/indexeddbworker.js' cannot be accessed from origin 'http://localhost:5500'.
```

### Expected Behaviour
- The viewer loads successfully.

image

sebjf:
This fixes #4005

#### Description
This PR changes how `IndexedDbCache` initialises its Worker. It now fetches the script itself and creates a blob URL, respecting the pages CORS settings, rather than asking the Worker to fetch the script (which is always done with "same-origin"). [This](https://stackoverflow.com/questions/58098143/why-are-cross-origin-workers-blocked-and-why-is-the-workaround-ok/60015898#60015898) allows the Worker to be instantiated when the `unity-util` and `indexeddbworker` are hosted on different origins (e.g. when the viewer is embedded).

#### Test Cases
This should be verified on as many browsers as possible to make sure the new way of instantiating the Worker is acceptable.

The new mechanism is the same whether running an embedded viewer, or a regular deployment, so only the normal deployment needs to be tested (not the demo site necessarily).

### DIFF
--- a/frontend/internals/webpack/webpack.dev.config.js
+++ b/frontend/internals/webpack/webpack.dev.config.js
@@ -18,8 +18,14 @@ const statsOptions = {
 module.exports = getWebpackConfig({
   mode: MODES.DEVELOPMENT,
   devtool: 'inline-source-map',
+  entry: {
+    unityutil: { // This entry creates the dev unity-util.js that is usually constructed by webpack.prod.config in production environments
+      import: './src/globals/unity-util-external.ts', 
+			filename: '../unity/unity-util.js' 
+    }
+  },
   output: {
-	clean: true
+    clean: true
   },
   plugins: [
     new LiveReloadPlugin({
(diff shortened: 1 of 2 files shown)


carmenfan (approved):
Embedded viewer now works

PR #4033, opened 2023-03-01
#4032 don't wait for helm to complete during CD

### LINKED ISSUE #4032: Don't wait for the helm chart to completely deploy on CD
### Description
Occasionally we get a timeout on deployment because we're waiting for the containers to spin up and it needed to put extra hardware up and there's a delay. We don't need to wait, as operationally we have a few minutes wait anyway for the certs to fix themselves.

### Suggested solution
Don't wait for the helm chart to deploy when deploying.

### Goal
- [x] faster more reliable deployments

### Tasks
- [x] tweak azure pipeline

keifgwinn:
This fixes #4032

#### Description
- don't wait for the deployment

#### Test cases
- `/azp run` should be faster and not hang about

### DIFF
--- a/.azure/build-and-deploy.yaml
+++ b/.azure/build-and-deploy.yaml
@@ -121,6 +121,7 @@ stages:
           chartType: 'Name'
           chartName: '3drepo/io'
           chartVersion: '4.29.0'
+          waitForExecution: false
           releaseName: '$(branchName)'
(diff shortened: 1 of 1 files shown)


keifgwinn:
previously, 
image

now
image

carmenfan (approved):

PR #4043, opened 2023-03-03
Issue 4042 - log the error when we process event messages

### LINKED ISSUE #4042: More verbose logging around model processing to better diagnose a situation
### Description
We're running into a situation where when a job is being processed by the bouncer (i.e. turning from `queued` to `processing` is not happening, however there is no clear indication as to why and it's not always reproducible. This issue aims to add more logging in that area in help to better understand the situation

carmenfan:
This fixes #4042

#### Description
- removed excessive logging when we have an error in the responder in v4
- add logging if we caught an exception during event processing for model processing signals
image

### DIFF
--- a/backend/src/v4/response_codes.js
+++ b/backend/src/v4/response_codes.js
@@ -338,7 +338,7 @@
 				return this.INCORRECT_USERNAME_OR_PASSWORD;
 			}
 			// other error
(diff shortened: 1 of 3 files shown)


keifgwinn (approved):
lgtm

PR #4048, opened 2023-03-06
ISSUE #3919 - Confirm modal appears when removing an admin user from teamspace

### LINKED ISSUE #3919: Remove teamspace admin user has no confirmation modal
### Description
There used to be a confirmation modal when trying to remove a teamspace admin from a teamspace, this got lost with database schema change.

The confirmation is necessary to avoid accidentally doing something, but we don't need to summarise their permissions so it can be all front end.

If a user (admin or non-admin) has been assigned permissions to a container/federation the old confirmation modal does appear

### Steps to replicate
- In either v4 or v5 go to the teamspace user settings
- Try to remove a user who has 'Teamspace admin' permissions

### Current Behaviour
- The user is instantly removed

### Expected Behaviour
- A modal opens asking if the user is sure they want to remove a teamspace admin
- On clicking 'Confirm' the user is removed
- On clicking cancel the user is not removed and the modal is closed

The old v4 modal 
MicrosoftTeams-image (1)

The-Daniel:
This fixes #3919 
#### Description
- In V5 if you try to remove a teamspace admin from a teamspace you get a confirmation modal
- There are no modals for other permissions

#### Test cases
- Try to remove a teamspace admin
- Try to remove a non-teamspace admin
- Check V4 is unaffected (if a user is a collaborator on a model there will be a modal)


sanmont3drepo:
The confirmation modal should appear regardless of the permissions and version of the app. That means it should be a confimation dialog just by trying to delete a user in v4 and v5.

frontend/src/v4/modules/userManagement/userManagement.sagas.ts
```diff
@@ -18,6 +18,9 @@
 import { isV5 } from '@/v4/helpers/isV5';
 import { DialogsActionsDispatchers } from '@/v5/services/actionsDispatchers';
 import { formatMessage } from '@/v5/services/intl';
+import { UsersHooksSelectors } from '@/v5/services/selectorsHooks';
+import { selectIsTeamspaceAdmin } from '@/v5/store/teamspaces/teamspaces.selectors';
+import { selectUser } from '@/v5/store/users/users.selectors';
```
sanmont3drepo, line 23: These imports are not longer necessary, right?

frontend/src/v4/modules/userManagement/userManagement.sagas.ts
```diff
@@ -169,64 +174,50 @@ export function* removeInvitation({ email }) {
...
-			yield put(DialogActions.showEndpointErrorDialog('remove', 'licence', error));
+			const config = {
+				title: 'Remove User',
+				template: RemoveUserDialog,
+				confirmText: 'Remove',
+				onConfirm: () => UserManagementActions.removeUserCascade(username),
+				data: {
+					// models: errorData.models,
+					// projects: errorData.projects,
+					// teamspacePerms: '',
```
sanmont3drepo, line 213: Instead of commenting, you can delete this.

sanmont3drepo (changes requested):

frontend/src/v4/modules/userManagement/userManagement.sagas.ts
The-Daniel: oops

sanmont3drepo (approved):

sanmont3drepo (approved):

PR #4074, opened 2023-03-10
ISSUE #4070 - Sso changes for plugin compatibility

### LINKED ISSUE #4070: SSO changes for plugin compatibility
### Description

Small SSO changes need to happen in order for the plugins to be able to use the API endpoints.

More specifically, sso middleware needs to change so that the sessions coming from the plugin are not stored as web sessions.

ChristosTsiotsias:
This fixes #4070 

#### Description
Small SSO changes in order for plugin SSO sessions to be stored properly in the database (ie correct userAgent, referer).

#### Test cases
Log in via SSO from the plugin and inspect the session stored in the database
The session should not be a web session and have no referer

### DIFF
--- a/backend/src/v5/middleware/sso/index.js
+++ b/backend/src/v5/middleware/sso/index.js
@@ -31,10 +31,13 @@ Sso.redirectWithError = (res, url, errorCode) => {
 	res.redirect(urlRedirect.href);
 };
 
-Sso.setSessionReferer = async (req, res, next) => {
-	if (req.headers.referer) {
-		req.session.referer = getURLDomain(req.headers.referer);
-	}
+Sso.setSsoInfo = async (req, res, next) => {
+	const ssoInfo = {
+		userAgent: req.headers['user-agent'],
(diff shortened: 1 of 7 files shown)


backend/src/v5/middleware/sso/index.js
```diff
@@ -31,10 +31,13 @@ Sso.redirectWithError = (res, url, errorCode) => {
 	res.redirect(urlRedirect.href);
 };
 
-Sso.setSessionReferer = async (req, res, next) => {
-	if (req.headers.referer) {
-		req.session.referer = getURLDomain(req.headers.referer);
-	}
+Sso.setSsoInfo = async (req, res, next) => {
```
carmenfan, line 34: probably better to call this set sessionInfo as the context we're in is SSO (you could argue every function here is setting sso Info 😆 )

carmenfan (changes requested):

carmenfan:
@ChristosTsiotsias  backend linter failed

carmenfan (approved):

PR #4079, opened 2023-03-13
ISSUE #3884 - Custom Tickets Required number inputs that are unset do not disable submit button

### LINKED ISSUE #3884: In Custom Tickets required toggle and number inputs that are unset do not disable the submit button
### Description

If you are creating a new custom ticket that includes either a required Toggle or a required Number Input you are able to leave these unset and the submit button will become enabled. Clicking submit does nothing however

- This is happening in: both 

### Steps to replicate
- Create a Ticket Template that has both a 'boolean' and 'number' property that are required.
- Create a Custom Ticket with this template
- Fill in all the required fields except for these two properties
- You will notice that the submit button becomes enabled

### Current Behaviour
- The submit button becomes enabled when you have required toggles and/or required number fields

### Expected Behaviour
- The submit button only becomes enabled once all required fields have been set

image
image
(NOTE: I believe the tooltips you see in this picture are from the browser. Fixing the schema so the submit button isn't enabled will stop these from appearing)

The-Daniel:
This fixes #3884 

#### Description
- Required number inputs that are unset were not disabling the submit button. Upon clicking submit nothing would happen (or the browser would indicate that an input is required for submission)
- This was due to the validation for numbers requiring that the value was not null, so undefined values slipped through

#### Test cases
On a ticket that has a required number field try various different inputs. E.g. 14, 0, -11

#### Questions to ponder
Should there maybe be options when creating a template for if the number input should be integer/decimal/positive/non-zero?

### DIFF
--- a/frontend/src/v5/validation/shared/validators.ts
+++ b/frontend/src/v5/validation/shared/validators.ts
@@ -15,7 +15,7 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 import { formatMessage } from '@/v5/services/intl';
-import { isNull } from 'lodash';
+import { isNull, isNumber } from 'lodash';
 import * as Yup from 'yup';
 
 export const trimmedString = Yup.string().transform((value) => value.trim());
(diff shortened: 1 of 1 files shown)


The-Daniel:
Converted to draft because I need to rebase to staging

Amantini1997 (approved):

sanmont3drepo (approved):

Issue #3915, opened 2023-01-26
All containers and federations stats are fetched on viewer opening


Amantini1997:
### Description
When a container or federation is opened in the viewer, all the stats for all the containers and federations are fetched

- This is happening in: staging
- This requires a hotfix to production: no

### Steps to replicate
- Given a project with at least one federation and one container, open either in the viewer

### Current Behaviour
- If I open a container (or federation), the stats for all the federations and containers are being fetched

### Expected Behaviour
- If I open a container, only the stats for the containers should be fetched (and _vice versa_ with a federation)

### Pictures/ Gifs of the error
![image](https://user-images.githubusercontent.com/43547703/214885176-1f337cc9-28aa-44a3-a1e3-b443d12c5533.png)

PR #4084, opened 2023-03-14
ISSUE #4072 - Upload and open a revision crashes

### LINKED ISSUE #4072: Upload and open a revision crashes
### Description
Uploading a revision and immediately opening it has the app crashing

### Steps to replicate
- In the containers tab, click on "Upload Files";
- Select a revision and upload it to a container that you create on the spot;
> The revision I used was the LegoRoundTree
- Wait for the revision to process and the open it;
A bunch of errors should be triggered

### Current Behaviour
- A bunch of errors are triggered following an api call that goes "/undefined/undefined/sequences/?rev_id=<id>"
- The app crashes

### Expected Behaviour
- The app should not crash

### Notes
The behaviour seems to be affected by time. The longer the time passed since the processing is complete, the lower the amount of errors (I know, it sounds weird)

The-Daniel:
This fixes #4072

#### Description
Uploading a revision to a new container and immediately opening it when ready would cause an error modal
This error modal was because the fetchs in `src\v5\ui\routes\viewer\viewer.tsx` caused the viewer to re render when new data was retrieved. I added a new saga for fetching only the necessary data for the viewer while setting a fetching state so it doesnt get to render the viewer until it finishes fetching everything.

Because it only fetches the necessary data, this also addresses #3915

#### Test cases
Upload several models whilst simultaneously creating new containers. Immediately open them (in the same tab). There should be no errors


sanmont3drepo (dismissed):

sanmont3drepo:
(this was converted to a draft because Im still researching the origin of the bug)

Amantini1997 (dismissed):

Amantini1997 (approved):

sanmont3drepo (approved):

PR #4121, opened 2023-03-22
Issue 4107 - Changing tickets correctly sets the value of properties

### LINKED ISSUE #4107: Unrequired text property in custom tickets gets it content ovewritten
### Description
When navigating through tickets of the same type, the non required text is being overwritten with the previous ticket value.

### Steps to replicate
- Create a ticket template with this json:
```
{
  "properties": [
    {
      "name": "A textfield",
      "type": "text"
    }
  ],
  "config": {},
  "code": "TTT",
  "name": "Text ticket"
}
```
- Create a ticket with something in the 'A textfield' field
- Create another ticket without filling the "A textfield" field
- Go to the tickets list
- Click on the textfield with a value in the textfield
- Click on the arrow to navigate to the other ticket

### Current Behaviour
- The current ticket still has the value in the textfield from the previous ticket

### Expected Behaviour
- The textfield should be empty in the ticket without a value.

error

Amantini1997:
This fixes #4107

#### Test cases
Test all the properties, in particular date, text, and numberfield

### DIFF
--- a/frontend/src/v5/ui/controls/inputs/datePicker/baseCalendarPicker/baseCalendarPicker.component.tsx
+++ b/frontend/src/v5/ui/controls/inputs/datePicker/baseCalendarPicker/baseCalendarPicker.component.tsx
@@ -33,7 +33,7 @@ export const BaseCalendarPicker = ({
 	helperText,
 	error,
 	required,
-	value = null,
+	value,
 	...props
 }: BaseCalendarPickerProps) => {
 	const [open, setOpen] = useState(false);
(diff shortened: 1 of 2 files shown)


frontend/src/v5/ui/controls/inputs/datePicker/baseCalendarPicker/baseCalendarPicker.component.tsx
```diff
@@ -46,7 +46,7 @@ export const BaseCalendarPicker = ({
 	return (
 		<PickerComponent
 			{...props}
-			value={value}
+			value={value || (defaultValue ? dayjs(defaultValue) : null)}
```
sanmont3drepo, line 49: I dont think is right to assume that everything with a default value is a date. Isnt there another place we can take this decision where we safely know is a date?

sanmont3drepo (changes requested):

The-Daniel (approved):

sanmont3drepo (approved):

sanmont3drepo (approved):

PR #4129, opened 2023-03-23
ISSUE #4101 - Add project modal should not allow the user to change teamspace

### LINKED ISSUE #4101: Add project modal should not allow the user to change teamspace
### Description
Not sure if this is facelifted or not, but it appears that we are using the same drop down as v4 on add project
image

And you can select a different teamspace than the one you're currently scoped in. This is a bit strange as, unlike v4, we no longer show all data from all teamspace, so the action of `Add project` should really only be in the context of the teamspace we're in.

We should remove the teamspace drop down.

### Steps to replicate
Try to add project inside a teamspace

### Current Behaviour
You can select a different teamspace entirely

### Expected Behaviour
You should only be able to create project in the teamspace you're currently looking at

The-Daniel:
This fixes #4101 

The user can no longer change the target teamspace when creating a project.
The teamspace is displayed in a disabled textfield as the user needs context of which teamspace it is being added to

#### Test cases
- create a new project


george-hadfield:
image

george-hadfield (approved):

PR #4131, opened 2023-03-23
Issue 4114 - Improved saga testing and added tests for viewer (v5)

### LINKED ISSUE #4114: Improve on saga testing
### Description
redux-saga-test-plan has some limitations that doesnt allow the sagas to be tested properly. It works on a vacuum so you cant test it properly with other sagas, it also works on timeout, that means that the test files using redux-saga-test-plan have a minimum waiting time and it doesnt test for final states in the redux store.

### Suggested solution
Create a small framework to test it like we do with other redux functions

### Goal
- [ ] Have a more useful and reliable way to test saga functions

### Tasks
- [ ] Create a small framework to test sagas
- [ ] Add viewer testing

sanmont3drepo:
This fixes #4114 

#### Description
- Added a new saga testing framework
- Added tests for viewer
- Fixed a bug in viewer sagas

### DIFF
--- a/frontend/src/v5/services/actionsDispatchers.ts
+++ b/frontend/src/v5/services/actionsDispatchers.ts
@@ -29,7 +29,7 @@ import { TicketsActions, ITicketsActionCreators } from '@/v5/store/tickets/ticke
 import { TicketsCardActions, ITicketsCardActionCreators } from '@/v5/store/tickets/card/ticketsCard.redux';
 import { ITeamspacesActionCreators, TeamspacesActions } from '@/v5/store/teamspaces/teamspaces.redux';
 import { IUsersActions, UsersActions } from '@/v5/store/users/users.redux';
-import { ViewerActionsCreators, ViewerActionsCreatorsType } from '@/v5/store/viewer/viewer.redux';
+import { ViewerActions, ViewerActionsCreators } from '@/v5/store/viewer/viewer.redux';
 
 import { Action } from 'redux';
 
@@ -53,4 +53,4 @@ export const TicketsActionsDispatchers = createActionsDispatchers<ITicketsAction
 export const TicketsCardActionsDispatchers = createActionsDispatchers<ITicketsCardActionCreators>(TicketsCardActions);
 export const TicketCommentsActionsDispatchers = createActionsDispatchers<ITicketCommentsActionCreators>(TicketCommentsActions);
 export const UsersActionsDispatchers = createActionsDispatchers<IUsersActions>(UsersActions);
-export const ViewerActionsDispatchers = createActionsDispatchers<ViewerActionsCreatorsType>(ViewerActionsCreators);
+export const ViewerActionsDispatchers = createActionsDispatchers<ViewerActionsCreators>(ViewerActions);
--- a/frontend/src/v5/services/actionsDispatchers.ts
+++ b/frontend/src/v5/services/actionsDispatchers.ts
@@ -29,7 +29,7 @@ import { TicketsActions, ITicketsActionCreators } from '@/v5/store/tickets/ticke
 import { TicketsCardActions, ITicketsCardActionCreators } from '@/v5/store/tickets/card/ticketsCard.redux';
 import { ITeamspacesActionCreators, TeamspacesActions } from '@/v5/store/teamspaces/teamspaces.redux';
 import { IUsersActions, UsersActions } from '@/v5/store/users/users.redux';
-import { ViewerActions, ViewerActionCreators } from '@/v5/store/viewer/viewer.redux';
+import { ViewerActionsCreators, ViewerActionsCreatorsType } from '@/v5/store/viewer/viewer.redux';
 
 import { Action } from 'redux';
 
@@ -53,4 +53,4 @@ export const TicketsActionsDispatchers = createActionsDispatchers<ITicketsAction
 export const TicketsCardActionsDispatchers = createActionsDispatchers<ITicketsCardActionCreators>(TicketsCardActions);
 export const TicketCommentsActionsDispatchers = createActionsDispatchers<ITicketCommentsActionCreators>(TicketCommentsActions);
 export const UsersActionsDispatchers = createActionsDispatchers<IUsersActions>(UsersActions);
-export const ViewerActionsDispatchers = createActionsDispatchers<ViewerActionCreators>(ViewerActions);
+export const ViewerActionsDispatchers = createActionsDispatchers<ViewerActionsCreatorsType>(ViewerActionsCreators);
--- a/frontend/test/test.helpers.ts
+++ b/frontend/test/test.helpers.ts
@@ -43,26 +43,19 @@ export const spyOnAxiosApiCallWithFile = (api, method) => {
 	});
 };
 
-
-// let sagaPromise = null;
-// let waitPatterns = [];
-// export const getSagaPromise = (...wait) => {
-// 	waitPatterns = wait;
-// 	sagaPromise;
-// }
-
 export const createTestStore = () => {
 	let middlewares = undefined;
 	let waitingActions: Action[] | string[] = [];
-	let resolvePromise: (value: unknown) => void | null = null;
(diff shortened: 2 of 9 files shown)


frontend/src/v5/services/actionsDispatchers.ts
```diff
@@ -53,4 +53,4 @@ export const TicketsActionsDispatchers = createActionsDispatchers<ITicketsAction
 export const TicketsCardActionsDispatchers = createActionsDispatchers<ITicketsCardActionCreators>(TicketsCardActions);
 export const TicketCommentsActionsDispatchers = createActionsDispatchers<ITicketCommentsActionCreators>(TicketCommentsActions);
 export const UsersActionsDispatchers = createActionsDispatchers<IUsersActions>(UsersActions);
-export const ViewerActionsDispatchers = createActionsDispatchers<ViewerActionCreators>(ViewerActions);
+export const ViewerActionsDispatchers = createActionsDispatchers<ViewerActionsCreatorsType>(ViewerActionsCreators);
```
The-Daniel, line 56: Why were these renamed in this way? Would it not be best to be consistent with the naming convention in this file of IViewerActionCreators and ViewerActions respectfully? Although I can see that not all the other dispatchers are consistent with this

frontend/test/viewer/viewer.sagas.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+				.get(`/teamspaces/${teamspace}/projects/${projectId}/containers/${baseContainers[2]._id}/stats`)
+				.reply(200, containersStats[2]);			
+				
+			await waitForActions(() => {
+				dispatch(ViewerActionsCreators.fetchData(teamspace, projectId, containerOrFederationId));
+			}, [ViewerActionsCreators.setFetching(false)]);
+
+
+			const federation = selectFederationById(getState(), containerOrFederationId);
+			const containersId = [...federation.containers].sort();
```
The-Daniel, line 109: I'd probably call this containersIds. containersId akes it sond like there's just one Id in total

frontend/test/viewer/viewer.store.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+		({ dispatch, getState } = createTestStore());
+	});
+
+	it('should change the state of fetching to true when setFetching(true) hass been dispatched' , async () => {
+		dispatch(ViewerActionsCreators.setFetching(true));
+		const fetching = selectIsFetching(getState());
+		expect(fetching).toBeTruthy();
+	});
+
+	it('should change the state of fetching to false when setFetching(true) has been dispatched' , async () => {
```
The-Daniel, line 37: I'm not sure how useful this test actually is. If setFetching is broken and does nothing this will still pass since isFetching is false by default. I wonder if it's best to combine it with the previous test. (i.e. dispatch setFetching(false) and do the falsy check after doing the truthy check.)

Also the 'it' message is slightly wrong, it should be setFetching(false) in this instance.

frontend/test/viewer/viewer.store.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+	const teamspace = 'myteamspace';
+	const projectId = 'myprojectid';
+	
+	let dispatch, getState;
+
+	beforeEach(() => {
+		({ dispatch, getState } = createTestStore());
+	});
+
+	it('should change the state of fetching to true when setFetching(true) hass been dispatched' , async () => {
```
The-Daniel, line 31: Not really important but small typo with hass => has

frontend/test/viewer/viewer.store.spec.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ *  Copyright (C) 2022 3D Repo Ltd
```
The-Daniel, line 2: Year

frontend/test/viewer/viewer.store.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+import { ViewerActionsCreators } from '@/v5/store/viewer/viewer.redux';
+import { selectIsFetching } from '@/v5/store/viewer/viewer.selectors';
+import { createTestStore } from '../test.helpers';
+
+describe('Viewer: redux', () => {
```
The-Daniel, line 21: This should be 'Viewer: store'

The-Daniel (changes requested):

frontend/test/test.helpers.ts
```diff
@@ -40,8 +43,53 @@ export const spyOnAxiosApiCallWithFile = (api, method) => {
...
+	let resolvePromiseObj = { resolvePromise: null };
+
+	const sagaMiddleware = createSagaMiddleware();
+	
+	middlewares = applyMiddleware(sagaMiddleware);
+
+	const discountMatchingActions =  (action: Action) => {
+		const waitingAction = waitingActions[0];
+	
+		if (action.type === waitingAction || isEqual(waitingAction, action)) {
```
Amantini1997, line 58: It's just an idea, but this could be re-written as 
```ts
if (_.isMatch(action, waitingAction)) { ... }
```
This way, line 139 of  `viewer.sagas.spec.ts` can be re-written as 
```ts
}, [DialogsActions.open('alert')]); // instead of [DialogsTypes.OPEN]
```
and be more consistent as we only pass actionCreators as waitingActions

frontend/test/viewer/viewer.sagas.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+		});
+
+		it('should fetch the containers, the federations, the federation data and the containers data for a particular federation', async () => {
+			const containers = times(6, () => containerMockFactory());
+			const federations = times(3, () => federationMockFactory());
+			const baseFederations = federations.map(prepareMockBaseFederation);
+			const containersStats = containers.map(prepareMockStatsReply);
+			const baseContainers = containers.map(prepareMockBasecontainer);
+			const containersInState = baseContainers.map((base, index) => prepareSingleContainerData(base, containersStats[index]));
+			const theFederation = federations[1];
```
Amantini1997, line 79: image

Amantini1997 (changes requested):

frontend/test/test.helpers.ts
sanmont3drepo: No, because dialog actions have more properties, like the props and I think is really annoying to have to match the exact error in here.

frontend/test/viewer/viewer.store.spec.ts
sanmont3drepo: The setFetching(true) is tested in the other test. So even if this one passes the other one would still fail.

frontend/test/test.helpers.ts
```diff
@@ -40,8 +43,53 @@ export const spyOnAxiosApiCallWithFile = (api, method) => {
...
+					resolvePromiseObj.resolvePromise = null;
+				}
+				return {};
+			}
+		}
+	), middlewares);
+	
+	const waitForActions = (func, waitActions) =>  { 
+		waitingActions = waitActions;
+		var promise = new Promise((resolve) => {resolvePromiseObj.resolvePromise = resolve;});
```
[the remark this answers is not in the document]
sanmont3drepo: Adding const to javascript was a mistake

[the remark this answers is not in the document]
Amantini1997: All the variables should be only const, following a pure functional approach and disallowing mutable variables
image

Amantini1997 (approved):

frontend/test/viewer/viewer.store.spec.ts
```diff
@@ -1,5 +1,5 @@
 /**
- *  Copyright (C) 2022 3D Repo Ltd
+ *  Copyright (C) 2023 3D Repo Ltd
```
The-Daniel, line 2: image

PR #4160, opened 2023-04-05
Super mesh bounding box should not be visible if all the objects are hidden by choice

### LINKED ISSUE #4159: Super mesh bounding box should not be visible if all the objects are hidden by choice
This branch is to test the fix to https://github.com/3drepo/3drepounity/issues/458

sebjf:
This fixes #4159 - it is to test a new Unity viewer that resolves the issue where bounding boxes are shown for bundles that are unloaded because they have been deliberately hidden.


carmenfan:
image

PR #4161, opened 2023-04-06
Issue 4146 - Project page does not show any data if we enter it in an error state

### LINKED ISSUE #4146: Project page does not show any data if we enter it in an error state
### Description
When you try to go into a container which you have no access and get bumped back out, the project data doesn't seem to populate correctly

### Steps to replicate
- Get a link to a container that you don't have access to
- Follow the modal and redirect to the project page

### Current Behaviour
- The user doesn't have access to the project either, seems to be bumped to the first available project
- Project ID on the URL doesn't match the redirected project, it's still the id of the project I have no access to
- Notice how there's no federations or containers, even after refresh. I don't seem to get the data until I go back to the teamspace list and go in from there (which rectifies the project ID situation)

### Expected Behaviour
- If the user doesn't have access to the project either, we should redirect them to the teamspace page.

img

Amantini1997:
This fixes #4146

#### Description
A project that does not grant access to the user is now correctly detected and the, on reaching an invalid model in the viewer, the user is redirected accordingly.

#### Test cases
Within a shared teamspace where no model is accessible (unauthorised) and try to access one => the user should be redirected to the dashboard
Within a shared teamspace where at least one model is accessible, try to access an unaccessible one (unauthorised) => the user should be redirected to the project list

### DIFF
--- a/frontend/src/v5/store/projects/projects.selectors.ts
+++ b/frontend/src/v5/store/projects/projects.selectors.ts
@@ -45,5 +45,5 @@ export const selectCurrentProjectName = createSelector(
 
 export const selectIsProjectAdmin = createSelector(
 	selectCurrentProjectDetails,
-	(project): boolean => project?.isAdmin || null,
+	(project): boolean => project?.isAdmin ?? null,
 );
--- a/frontend/src/v5/store/projects/projects.selectors.ts
+++ b/frontend/src/v5/store/projects/projects.selectors.ts
@@ -45,5 +45,5 @@ export const selectCurrentProjectName = createSelector(
 
 export const selectIsProjectAdmin = createSelector(
 	selectCurrentProjectDetails,
-	(project): boolean => project?.isAdmin ?? null,
+	(project): boolean => !!project?.isAdmin,
 );
--- a/frontend/src/v5/store/teamspaces/teamspaces.selectors.ts
+++ b/frontend/src/v5/store/teamspaces/teamspaces.selectors.ts
@@ -40,5 +40,5 @@ export const selectCurrentQuota = createSelector(
 
 export const selectIsTeamspaceAdmin = createSelector(
 	selectCurrentTeamspaceDetails,
-	(teamspace): boolean => teamspace?.isAdmin || null,
+	(teamspace): boolean => teamspace?.isAdmin ?? null,
 );
--- a/frontend/src/v5/store/teamspaces/teamspaces.selectors.ts
+++ b/frontend/src/v5/store/teamspaces/teamspaces.selectors.ts
@@ -40,5 +40,5 @@ export const selectCurrentQuota = createSelector(
 
 export const selectIsTeamspaceAdmin = createSelector(
 	selectCurrentTeamspaceDetails,
-	(teamspace): boolean => teamspace?.isAdmin ?? null,
+	(teamspace): boolean => !!teamspace?.isAdmin,
 );
--- a/frontend/src/v5/store/viewer/viewer.sagas.ts
+++ b/frontend/src/v5/store/viewer/viewer.sagas.ts
@@ -18,10 +18,10 @@ import * as API from '@/v5/services/api';
 import { FetchContainersResponse } from '@/v5/services/api/containers';
 import { FetchFederationsResponse } from '@/v5/services/api/federations';
 import { formatMessage } from '@/v5/services/intl';
-import { put, select, take, takeLatest } from 'redux-saga/effects';
+import { cancel, put, race, select, take, takeLatest } from 'redux-saga/effects';
 import { prepareContainersData } from '../containers/containers.helpers';
 import { ContainersActions, ContainersTypes } from '../containers/containers.redux';
-import { DialogsActions } from '../dialogs/dialogs.redux';
+import { DialogsActions, DialogsTypes } from '../dialogs/dialogs.redux';
 import { prepareFederationsData } from '../federations/federations.helpers';
 import { FederationsActions, FederationsTypes } from '../federations/federations.redux';
 import { selectFederationById } from '../federations/federations.selectors';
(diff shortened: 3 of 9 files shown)


frontend/src/v5/store/projects/projects.selectors.ts
```diff
@@ -45,5 +45,5 @@ export const selectCurrentProjectName = createSelector(
 
 export const selectIsProjectAdmin = createSelector(
 	selectCurrentProjectDetails,
-	(project): boolean => project?.isAdmin || null,
+	(project): boolean => project?.isAdmin ?? null,
```
sanmont3drepo, line 48: Why this change? in any case I would return false instead of null as this selector returns a boolean in theory

frontend/src/v5/store/teamspaces/teamspaces.selectors.ts
```diff
@@ -40,5 +40,5 @@ export const selectCurrentQuota = createSelector(
 
 export const selectIsTeamspaceAdmin = createSelector(
 	selectCurrentTeamspaceDetails,
-	(teamspace): boolean => teamspace?.isAdmin || null,
+	(teamspace): boolean => teamspace?.isAdmin ?? null,
```
sanmont3drepo, line 43: same comment as before

frontend/src/v5/store/viewer/viewer.sagas.ts
```diff
@@ -42,19 +42,36 @@ function* fetchData({ teamspace, containerOrFederation, project }: FetchDataActi
 
 		const isFederation = federationsWithoutStats.some(({ _id }) => _id === containerOrFederation);
 
+		const fetchContainerStatsHasError = { value: false };
```
sanmont3drepo, line 45: I think is not necesarry to make this an object

frontend/src/v5/store/viewer/viewer.sagas.ts
```diff
@@ -42,19 +42,36 @@ function* fetchData({ teamspace, containerOrFederation, project }: FetchDataActi
...
 			const federation: IFederation = yield select(selectFederationById, containerOrFederation);
 
 			for (const containerId of federation.containers) {
-				yield put(ContainersActions.fetchContainerStats(teamspace, project, containerId));
+				yield put(ContainersActions.fetchContainerStats(
+					teamspace,
+					project,
+					containerId,
+					null,
+					() => { fetchContainerStatsHasError.value = true; },
```
sanmont3drepo, line 58: this whole thing is kind of weird, the saga is waiting for a FETCH_CONTAINER_STATS_SUCCESS to be dispatched. In the case of a an error is that event dispaatched?

frontend/src/v5/store/viewer/viewer.sagas.ts
```diff
@@ -42,19 +42,36 @@ function* fetchData({ teamspace, containerOrFederation, project }: FetchDataActi
...
+				));
 				yield take(ContainersTypes.FETCH_CONTAINER_STATS_SUCCESS);
 			}
 		} else {
-			yield put(ContainersActions.fetchContainerStats(teamspace, project, containerOrFederation));
+			yield put(ContainersActions.fetchContainerStats(
```
sanmont3drepo, line 63: same comment as on top

frontend/src/v5/ui/components/shared/modalsDispatcher/templates/alertModal/alertModal.component.tsx
```diff
@@ -37,30 +37,32 @@ interface IAlertModal {
...
+	const projectNotFound = isProjectNotFound(code);
 	const unauthorized = isPathNotAuthorized(error);
-	const unauthInTeamspace = unauthorized && teamspace;
-	const unauthInProject = unauthorized && project;
 
 	const getSafePath = () => {
 		// eslint-disable-next-line max-len
-		if (isResourceNotFound(code) || (unauthInProject)) return generatePath(PROJECT_ROUTE_BASE, { teamspace, project });
-		if (isProjectNotFound(code) || (unauthInTeamspace)) return generatePath(TEAMSPACE_ROUTE_BASE, { teamspace });
+		if ((modelNotFound || unauthorized) && hasAccessToProject) return generatePath(PROJECT_ROUTE_BASE, { teamspace, project });
```
sanmont3drepo, line 55: this logic seems a bit complex here.  What does it mean?

sanmont3drepo (changes requested):

frontend/src/v5/ui/components/shared/modalsDispatcher/templates/alertModal/alertModal.component.tsx
Amantini1997: It basically checks whether something (a model/a project, etc) is not reachable and where the user can be redirected to.

For instance, in the line above, if the user does not have access to a model (lack of authorisation or model not found), but they have access to the project, this is where they get redirected

The-Daniel:
A related bug I've found. (This wasn't introduced with this issue)

The redirect only happens if an alert modal is called. It is actually possible to access an unauthorised location and not get an alert modal if there is no data to fetch. (I think this is how it works anyway)

For example as **User A** create a project with no containers or federations. Have a **User B** who is not project Admin and so should not be able to access project or user permissions.

Try to access one of these pages as **User B** (this is most easily done by navigating to these pages as **User A**, logging out, and then logging in as **User B**)

You see a blank project page. The browser tab says _'Loading...'_. The breadcrumbs say you're in a different project but this is not true. (It just selects the first project you should have permission to. Empty projects are invalid even if you have access rights to them)
image

Note: It doesn't necessarily have to be an empty project. Just a project that **User B** has access to no containers/feds. i.e. all the models have the 'unassigned' permission level.

This might be a problem for a new issue - I don't know what you think

frontend/src/v5/ui/components/shared/modalsDispatcher/templates/alertModal/alertModal.component.tsx
```diff
@@ -37,30 +37,32 @@ interface IAlertModal {
 export const AlertModal: FC<IAlertModal> = ({ onClickClose, currentActions = '', error, details, open }) => {
 	const teamspace = TeamspacesHooksSelectors.selectCurrentTeamspace();
 	const project = ProjectsHooksSelectors.selectCurrentProject();
+	const accessibleProjects = ProjectsHooksSelectors.selectProjects()[teamspace] || [];
+	const hasAccessToProject = accessibleProjects.findIndex(({ _id }) => _id === project) >= 0;
```
The-Daniel, line 41: Couldn't this just be a `.some()`?

Amantini1997:
> A related bug I've found. (This wasn't introduced with this issue)
> 
> The redirect only happens if an alert modal is called. It is actually possible to access an unauthorised location and not get an alert modal if there is no data to fetch. (I think this is how it works anyway)
> 
> For example as **User A** create a project with no containers or federations. Have a **User B** who is not project Admin and so should not be able to access project or user permissions.
> 
> Try to access one of these pages as **User B** (this is most easily done by navigating to these pages as **User A**, logging out, and then logging in as **User B**)
> 
> You see a blank project page. The browser tab says _'Loading...'_. The breadcrumbs say you're in a different project but this is not true. (It just selects the first project you should have permission to. Empty projects are invalid even if you have access rights to them) image
> 
> Note: It doesn't necessarily have to be an empty project. Just a project that **User B** has access to no containers/feds. i.e. all the models have the 'unassigned' permission level.
> 
> This might be a problem for a new issue - I don't know what you think

I see what you mean, even if you enter a fake name project an try to access the board tab content, no error is thrown. This is due to the alert modal relying on `4xx` responses from the backend, which is correct. One thing that we can do, is in the `projectNavigation` do something like this 
```ts
export const ProjectNavigation = (): JSX.Element => {
	let { url } = useRouteMatch();
	url = discardSlash(url);
	const isProjectAdmin = ProjectsHooksSelectors.selectIsProjectAdmin();
	const currentProject = ProjectsHooksSelectors.selectCurrentProject();
	const projects = ProjectsHooksSelectors.selectCurrentProjects();

	useEffect(() => {
		if (!currentProject) return;
		if (!(currentProject in projects)) {
			DialogsActionsDispatchers.open('alert', { error: { response: { data: { code: 'PROJECT_NOT_FOUND' } } } });
		}
	}, [currentProject]);
```

But, this means we have to artificially create the error ourselves, which does not seem right.

The-Daniel (approved):

PR #4163, opened 2023-04-06
ISSUE #4144 - clear SSO params on normal login

### LINKED ISSUE #4144: login related query strings should reset after login
### Description
When you try to log into 3drepo via SSO and failed, login related query strings lingers and you can end up with some pretty weird UX

### Steps to replicate
- login with  SSO on an email that isn't linked with SSO
- see the error message, login with username and password
- go to profile

### Current Behaviour
- the URL after login still contains `loginPost=1&error=3`
image
- Going to profile modal will show an error message for an action that was not performed
-image

### Expected Behaviour
- `loginPost/error` should reset
- Profile modal should not show the error

img

Amantini1997:
This fixes #4144

#### Description
On normal login, SSO params are cleared out

### DIFF
--- a/frontend/src/v5/ui/routes/login/login.component.tsx
+++ b/frontend/src/v5/ui/routes/login/login.component.tsx
@@ -27,7 +27,7 @@ import { AuthTemplate } from '@components/authTemplate';
 import { AuthHooksSelectors } from '@/v5/services/selectorsHooks';
 import { SubmitButton } from '@controls/submitButton/submitButton.component';
 import { AuthSubHeader, Divider } from '@components/authTemplate/authTemplate.styles';
-import { useSSOLogin } from '@/v5/services/sso.hooks';
+import { useSSOLogin, useSSOParams } from '@/v5/services/sso.hooks';
 import { Gap } from '@controls/gap';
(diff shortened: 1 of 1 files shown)


frontend/src/v5/ui/routes/login/login.component.tsx
```diff
@@ -45,12 +48,13 @@ export const Login = () => {
...
 		resolver: yupResolver(LoginSchema),
 	});
-	const isPending = AuthHooksSelectors.selectIsPending();
 
+	const isPending = AuthHooksSelectors.selectIsPending();
 	const errorMessage: string = AuthHooksSelectors.selectLoginError();
 
 	const onSubmit = ({ username, password }) => {
 		AuthActionsDispatchers.login(username, password);
+		resetSSOParams();
```
sanmont3drepo, line 57: Instead of doing this I think is better to get reset the ssoParams on unmount of the component so we can tackle all other links and cases (like signup, forget password, etc)

sanmont3drepo (changes requested):

sanmont3drepo (approved):

PR #4164, opened 2023-04-06
ISSUE #4102 - Make project name validation case insensitive

### LINKED ISSUE #4102: Project name restriction should be case insensitive
### Description
WE do not allow users to create a project of the same name exactly, but it can be created if it is in a slightly different case

We should make sure it's restricted on both front end and backend

### Steps to replicate
- try to create a project with a name of a different project that already exists, but enter it in a slightly different case

### Current Behaviour
- It is allowed and the project gets created

### Expected Behaviour
- The project should not be created and an error should show up

img

Amantini1997:
This fixes #4102

#### Description
Project name validation is now case insesitive

#### Test cases
Try to create (or edit) a project using the same name as an existing one, but with different casing

#### NOTES:
This issue requires a fix on the backend too, as such, it can't be tested correctly until that

### DIFF
--- a/frontend/src/v5/ui/routes/dashboard/projects/projectSettings/projectSettings.component.tsx
+++ b/frontend/src/v5/ui/routes/dashboard/projects/projectSettings/projectSettings.component.tsx
@@ -83,7 +83,7 @@ export const ProjectSettings = () => {
 
 	useEffect(() => {
 		if (existingNames.length) trigger('projectName');
-	}, [existingNames.length]);
+	}, [existingNames]);
 
 	if (_.isEmpty(currentProject)) return (<></>);
 


frontend/src/v5/validation/shared/validators.ts
```diff
@@ -20,6 +20,8 @@ import * as Yup from 'yup';
 
 export const trimmedString = Yup.string().transform((value) => value.trim());
 
+export const trimmedLowercaseString = trimmedString.transform((value) => value.toLocaleLowerCase());
```
The-Daniel, line 23: This means that project names are now all lowercase. Is this what we want?

The-Daniel: This change also means that the project settings behaves strangely.
If you change the name to be the same but with a different case then you get a success message although nothing actually changes
image
(image shows trying to rename 'all cap' -> 'ALL CAP'. The project name continues being 'all cap')

The-Daniel (changes requested):

The-Daniel (approved):

The-Daniel (dismissed):

frontend/src/v5/ui/routes/dashboard/projects/projectsList/createProjectModal/createProjectModal.component.tsx
```diff
@@ -54,15 +54,14 @@ export const CreateProjectModal = ({ open, onClickClose }: CreateProjectModalPro
...
+		resolver: yupResolver(ProjectSchema),
+		context: { existingNames },
 		defaultValues: DEFAULT_VALUES,
 	});
 
 	const onSubmissionError = (error) => {
 		if (projectAlreadyExists(error)) {
-			const { projectName } = getValues();
-			setExistingProjects((currentValue) => [...currentValue, projectName]);
+			setExistingNames((currentValue) => [...currentValue, getValues('projectName').toLocaleLowerCase()]);
```
sanmont3drepo, line 64: This is not necessary, the comparison is already changing the case projectSchemes

frontend/src/v5/ui/routes/dashboard/projects/settingsModal/settingsModal.component.tsx
```diff
@@ -171,6 +171,8 @@ export const SettingsModal = ({
 	const containerOrFederationName = isContainer ? 'Container' : 'Federation';
 	const EMPTY_GIS_VALUES = { latitude: 0, longitude: 0, angleFromNorth: 0 };
 
+	const nameWasChanged = () => watch('name')?.trim() !== containerOrFederation.name;
```
sanmont3drepo, line 174: nameWaschanged is not necessary.

sanmont3drepo (changes requested):

sanmont3drepo (approved):

PR #4175, opened 2023-04-12
Issue 4167 - Better support for users without teamspace

### LINKED ISSUE #4167: Better support for users without teamspaces
### Description
We have a script that can remove user's teamspace. so it is possible for a user to log and see a teamspace screen when the teamspace list is empty

### Steps to replicate
- remove a teamspace via the utility script:
> yarn run-script removeTeamspaceAndOwner --accounts {name of that teamspace}
- login as that teamspace's owner

### Current Behaviour
image

### Expected Behaviour
UI shouldn't look broken (maybe just create teamspace button? or a prompt? consult @george-hadfield )

Amantini1997:
This fixes #4167 

#### Description
A user that has not teamspace (because it was deleted) will not see the loading screen, but will be prompted to create a new teamspace

#### Test cases
Delete the teamspace from the cmd and go the teamspace selection page.
To delete the teamspace of a user, you can:
- open the cmd and navigate to the `local` folder;
- call `.\run-io-backend-scripts.ps1 removeTeamspaceAndOwner --accounts <user>` where `<user>` is the user whose teamspace should be deleted

### DIFF
--- a/frontend/src/v5/store/teamspaces/teamspaces.redux.ts
+++ b/frontend/src/v5/store/teamspaces/teamspaces.redux.ts
@@ -26,16 +26,22 @@ export const { Types: TeamspacesTypes, Creators: TeamspacesActions } = createAct
 	fetchQuota: ['teamspace'],
 	fetchQuotaSuccess: ['teamspace', 'quota'],
 	setCurrentTeamspace: ['currentTeamspace'],
+	setTeamspacesArePending: ['teamspacesArePending'],
 }, { prefix: 'TEAMSPACES2/' }) as { Types: Constants<ITeamspacesActionCreators>; Creators: ITeamspacesActionCreators };
 
 export const INITIAL_STATE: ITeamspacesState = {
 	teamspaces: [],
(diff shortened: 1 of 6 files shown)


carmenfan:
> To delete the teamspace of a user, you can:
>
> comment out line 56 in the script ~\3drepo.io\backend\src\scripts\utility\teamspaces\removeTeamspaceAndOwner.js

@Amantini1997  you shouldn't need to do this, `removeOwners` should not be set to true unless you flagged it via `--removeOwners true` as a parameter, so L56 should never run.

Amantini1997:
> > To delete the teamspace of a user, you can:
> > comment out line 56 in the script ~\3drepo.io\backend\src\scripts\utility\teamspaces\removeTeamspaceAndOwner.js
> 
> @Amantini1997 you shouldn't need to do this, `removeOwners` should not be set to true unless you flagged it via `passing --removeOwners true` as a parameter, so L56 should never run.

I updated the description, thanks!

PR #4183, opened 2023-04-18
Issue 2734 - Update packages to fix warnings

### LINKED ISSUE #2734: Research warnings with packages
### Description
When installing modules a bunch of "unmet peer dependency" and "deprecate" warnings shows up. We need to investigate what it means, if they can be a problem and in that case 

### Steps to replicate
- remove node_modules
- npm install or yarn install

Amantini1997:
This fixes #2734 

#### Description
After deleting the node modules and calling `yarn install`, several warnings were triggered in the console.
Some of these were also observable without deleting the node_modules by calling `yarn install --check-files`.

To briefly understand what is going on, it is important to understand how node packages work and what the warning were saying.

In `package.json` there are different properties to tell npm (or yarn) what packages are required to work.
The 2 most common properties are `dependencies` and `dev-dependencies`.
Then, we have something called `peer-dependencies`. A peer dependency is required by a module (e.g. cssnano) in order for the latter to work; however, when we call `yarn install cssnano`, none of the peer dependency is installed.
Instead, we manually have to install them.
The warnings about these dependencies are of 2 kind (following are examples, not necessarily correct versions-wise):
- "> cssnano@4.2.3" has unmet peer dependency "postcss@^10.0.0 || ^11.0.0" -> this means that `postcss` is a required dependency and the version must be `10.0.0 or higher` or `11.0.0 or higher` (Note that v11 is not higher than v10, rather, they are treated as different packages)
What this warning is telling us is that we are missing such a dependency and we have to install it manually;
- " > react-event-listener@0.6.6" has incorrect peer dependency "react@^16.3.0". -> In this case, we already have `react` installed, but the version is not correct, so we have to either update the direct dependency (react-event-listener) or the peer dependency (react) so that the 2 are compatible.

#### Test cases
Delete the `yarn.lock` file as well as all the node_modules, then call `yarn install`.
Now test the app, test storybook, test the linting and building


sanmont3drepo (approved):

sanmont3drepo:
> https://github.com/3drepo/3drepo.io/actions/runs/4733037273/jobs/8399975209#step:5:11 is showing warnings still?

Not every warning has been addressed. Some will need more time to fix, but we have other priorities.

The-Daniel (approved):

PR #4196, opened 2023-04-24
Issue 3933 - Custom Ticket: Groups

### LINKED ISSUE #3933: Custom Tickets: Groups
### Description
Add  full `state` support in views on 3drepo.io UI. When we display a custom ticket view, it should reset the object status on the viewer and apply what is saved inside the state object.

#### Data schema
See #3969 for the full schema of this object:
- `showDefaultHidden`: a boolean that depicts if hidden by default objects should be shown. This essentially affects if the `Show hidden geometry` option inside the tree: 
image
- `colored`: specifies a list of groups that needs to have colour override applied (Note: it is possible where a single object is coloured by multiple groups - in that situation apply the same logic as v4 groups: apply the last applied override, if that is removed, apply the 2nd from last applied override). Unlike v4, it is possible to only apply a transparency override without applying a colour
- `hidden`: specifies a list of groups that needs to be hidden 
- `transformed`: This is a hidden list of groups (not shown to the UI) that specifies the list of objects with a 4x4 matrix that it is being applied. This is very specific to views saved when viewing a 4D Sequence and the user can only `edit` this by snapshotting the current view.

##### Prefix
The `prefix` specifies the hierachy of the group. This is a proper support of Group of groups in v4. e.g. in V4 you would define a hierachy of `Level 1` > `Kitchen` onto a group by prefixing the group name with `Level 1:: Kitchen`. In this implementation you should leave the group name as it is, and provide a prefix value of `["Level 1", "Kitchen"]`

##### View property
- The view property card has been updated to facilitate the groups support.
- Clicking on to the `groups` button will now lead you to the new groups UI

##### Groups list UI
- Groups should be collapsed by default
- When entering the groups card, we should reset the view to the selected view again to ensure the states are in sync
- Coloured groups:
  - Colour override should be on by default (i.e. check box ticked)
  - Colour override status is a temporary thing and does not get saved to the view - if the user wishes to no longer colour override a certain group they should just remove the group from the list.
  - Smart group should have a different icon to manual group, and the colour of the icon should match the colour/transparency override of that group (e.g. just like v4 groups)
  - Clicking on that group should highlight the objects in the specified colour (or default if only opacity is overriden)
  - Opacity cannot be set to 0 - they should be using hidden groups instead.
- Hidden groups
  - clicking on a group should unhide the group and highlight it with the default colour
  - The user can also show/hide the object by toggling the show/hide icon
    - This is a temporary viewer state that is not saved into the view - if the user wishes to show a certain hidden group, they should just remove the group
- Deleting a collection of groups should delete all groups within that collection 

##### Group edit/new view
- Smart group should be the default option
- Colour picker should only be available for coloured groups edit
- You can only save a group if it is a manual group with at least 1 object, or a smart group with at least 1 filter

### Goals
- [ ] As a ticket creator/editor (everyone but viewer):
  - [ ] I should be able to "snapshot" my current view into a saved view by a simple click of a button. It should preserve my camera settings, clipping planes and object states (hidden, colour override, transforms). Any highlighted objects should be grouped together by their highlighted colours
  - [ ] I should be able to specify what the view's state should be by creating groups (smart or manual), declaring them to be overridden by certain colour and/or opacity, or whether they should be hidden
  - [ ] I should NOT be able to declare a transformed group - these should only be possible by snapshotting my current view
  - [ ] I should be able to edit the group listings without affecting my camera/clipping point positions
  - [ ] I should be able to edit the group hierachy to introduce group collections the group should be under
  - [ ] I should be able to update the rules/objects of a group without changing anything else on the view
  - [ ] I should be able to delete the states without deleting the rest of the view
  - [ ] I should be able to delete the whole view - which would delete the states

- [ ] As a ticket viewer
  - [ ] When I'm being shown the viewpoint (e.g. clicking on it on the ticket list, entering the ticket, clicking on the show viewpoint option), it should reset my current object states and show me the object states that the view was saved with
  - [ ] I should be able to interact with the groups:
    - [ ] clicking on a colour overridden group should highlight the group with that colour
    - [ ] toggling the check box on a colour overridden group should toggle on/off colour override of that group
    - [ ] clicking on a hidden group should show the group and highlight the objects in default colour
    - [ ] I should be able to toggle the hidden group on or off
    - [ ] I should NOT be able to interact with hidden groups

<details> <summary> Original user journeys (outdated) </summary>

- [ ] As an admin I want to be able to add a groups module to a custom ticket template
- [ ] As a user I want to ignore any preexisting colour overrides in the 3D view when creating a custom ticket viewpoint
- [ ] As a user I want to automatically apply all the group colour overrides in order for a custom ticket when I enter the ticket details panel
- [ ] As a user I want all existing colour overrides to be removed when I open a custom ticket so that any groups within that ticket are applied exclusively
- [ ] As a user I want to changes to groups to be included in the activity log
- [ ] As a user I do not want all the groups to load when I load the model so that the initial loading time is shorter
- [ ] As a user I want to be able to create smart groups using multiple rule sets
- [ ] As a user I want to be able to create normal group using the selection method in v4
- [ ] As a user I want smart groups to be the default
- [ ] As a user I want to be able to edit the colour of the group and have the model update immediately with the new colours
- [ ] As a user I want the groups to stay active even after I have left the custom ticket details page
- [ ] As a user I want to be able to display a hierarchy of groups in the same way as v4
- [ ] As a user I want to be able to isolate the objects within a group or group collection
- [ ] As a user I want to be able to edit a group name
- [ ] As a user I want to be able to set a group opacity without a colour override so that all objects in that group retain their source colours
- [ ] As a user I want to be able to see the number of object included in a group
- [ ] As a user I want the number of objects in the group to update every time I make changes to the group
- [ ] As a user I want to see the number of groups displayed in the "Groups" accordion header
- [ ] As a user I would like to be able to use regular expressions on the field name as well as the value so that I be more flexible with my groups
- [ ] As a user I would like to have the count of objects to be included in the API response so that it is easier to display totals in external applications such as PowerBI (due to the complexity of the tree calculation)

</details>

### Related Resources
[Design](https://www.figma.com/file/3FjNEOdQbcHwzquVZxUqMl/3D-Repo?node-id=3251-29959&t=UIQkeOgJ6s8RRcSO-0)

Amantini1997:
This fixes #3933 

**NOTE:** this is the **base** branch for Custom Tickets: Group

#### Description
The basic custom ticket group structure has been added. 
So far, only the UI has been implemented.
The groups are mocked and none of the actions linked to them (edit/toggle show, delete, or even check a single group or a group of groups) will work.
Clicking on the delete button, however, triggers the modal.
> Update the description as new stuff is added

#### Test cases
In the viewer, open any custom ticket with a view and click on **groups**

#### Bugs
- [x] [Required view validation doesn't seem to go away even if i've filled in everything.](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630684612) (Santiago)
- [x] [Groups is enabled even when there's no groups inside](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630693585)
- [x] [Delete view is not deleting the image](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630711772)
- [x] [Smart group should be the default option](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630715514)
- [x] [Wrong blurb is shown on manual groups](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630724991)
- [x] [Imposing colour on the groups panel is removing them else where](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630747181) (Santiago)
- [x] [Clear all overrides does not update the panel state](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630749763) (Santiago)
- [x] [Icons do not match the design](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630752599) (Need @george-hadfield as the current icons seem corrupted)
- [x] [Opacity value is obscured](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630757709)
- [x] [Confirmation button in the colour picker](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630784493)(Alessandro)
- [x] [Missing rules pasting feature](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630935696) (Alessandro in a separate issue #4329)
- [x] [Smart groups doesn't work until you refresh](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630945778) (Alessandro)
- [x] [num objects is not correct](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630952944) (Alessandro)
- [x] [use same syntax as v4 for smart rules](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630954705) (Alessandro)
- [x] [Hover buttons on groupings](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1630972305)
- [x] [V4 override all button is effecting the custom tickets groups](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1640051628)
- [x] [User should not be able to delete a required view](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1640053905)
- [x] [Missing UI stuff for manual group](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1640056594) 
- [x] [Hidden groups not behaving correctly](https://github.com/3drepo/3drepo.io/pull/4196#issuecomment-1640439007)(Santiago)


frontend/src/v5/ui/routes/viewer/tickets/ticketGroupsCard/ticketGroupsCard.component.tsx
```diff
@@ -0,0 +1,66 @@
...
+						/>
+					</NewGroupButton>
+				</GroupsAccordion>
+				<GroupsAccordion
+					title={formatMessage({ id: 'ticketCard.groups.hidden', defaultMessage: 'hidden' })}
+					groups={groups.hidden}
+				>
+					<NewGroupButton startIcon={<AddCircleIcon />}>
+						<FormattedMessage
+							id="ticketCard.groups.coloured.addGroup"
```
The-Daniel, line 58: The ID says coloured when it is hidden. It might be possible to reduce code duplication here

frontend/src/v5/ui/routes/viewer/tickets/ticketGroupsCard/groups/groups.component.tsx
```diff
@@ -0,0 +1,135 @@
...
+	GroupItemContainer,
+	GroupItemButtons,
+	GroupCollectionTitle,
+	GroupCollectionAccordion,
+	GroupCollectionContainer,
+} from './groups.styles';
+import { GroupToggle } from '../groupToggle/groupToggle.component';
+
+type GroupProps = { group: IGroupFromApi, color?: [number, number, number], opacity?: number, colored: boolean };
+const GroupItem = ({ group, color, opacity, colored }: GroupProps) => {
```
The-Daniel, line 45: I think this component should be in its own file

frontend/src/v5/ui/routes/viewer/tickets/ticketGroupsCard/groups/groups.styles.ts
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		.MuiAccordionDetails-root {
+			padding: 9px 0 0 16px;
+			overflow: hidden;
+		}
+	}
+`;
+
+export const GroupTitle = styled.span`
```
The-Daniel, line 104: You might need to add some truncation styling here. Unless you can be sure that group/collection titles can't be that long
image

frontend/src/v5/ui/routes/viewer/tickets/ticketGroupsCard/groupsAccordion/groupsAccordion.component.tsx
```diff
@@ -0,0 +1,67 @@
...
+	colored?: boolean;
+};
+export const GroupsAccordion = ({ title, groups = [], children, colored }: GroupsAccordionProps) => {
+	const [checked, setChecked] = useState(false);
+
+	const groupsCount = groups.length;
+
+	const toggleCheckbox = (e) => {
+		e.stopPropagation();
+		setChecked(!checked);
```
The-Daniel, line 39: I think this should be done using a[ callback function with the previous value.](https://legacy.reactjs.org/docs/hooks-reference.html#functional-updates)

I don't think that it's a problem at the moment since the state change is instantaneous. But once the logic is added that connects this to the api it might cause a delay that can cause race conditions

The-Daniel (changes requested):

frontend/src/v5/ui/routes/viewer/tickets/ticketGroupsCard/ticketGroupsCard.component.tsx
sanmont3drepo: I think both should be `ticketCard.groups.addGroup`

sanmont3drepo: I think both should be `ticketCard.groups.addGroup`

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketGroups/groups/groups.component.tsx
```diff
@@ -0,0 +1,63 @@
...
+	const collections = values(collectionsDict);
+
+	const collectionToGroup = (collection) => collection.map(({ prefix, ...group }) => ({
+		...group,
+		prefix: prefix.slice(1),
+	}));
+
+	return (
+		<>
+			{groupItems.map((group) => (<GroupItem {...group} colored={colored} />))}
```
The-Daniel, line 42: You need to add keys to these maps

The-Daniel (approved):

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketForm.component.tsx
```diff
@@ -32,9 +32,15 @@ interface ModulePanelProps {
 
 const ModulePanel = ({ module, scrollPanelIntoView, defaultExpanded, ...rest }: ModulePanelProps) => {
 	const required = some(module.properties, 'required');
+	const { title, ...accordionProps } = getModulePanelProps(module);
 	return (
-		// eslint-disable-next-line max-len
-		<Accordion {...getModulePanelTitle(module)} onChange={scrollPanelIntoView} defaultExpanded={defaultExpanded} required={required}>
+		<Accordion
```
sanmont3drepo, line 37: why this change?

sanmont3drepo (approved):

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketForm.component.tsx
Amantini1997: We want to indicate when a module has at least a required field. This was previously achieved passing a "required" property that added a red asterisk next the title.

However, this forced the accordion header to be a `FormLabel`, which was bad as, on click (expand/collapse accordion), a checkbox inside it would also get the click, and so toggles its state.

This change was to make the header a span, so it's down to the parent to decide what the title should look like, without affecting the rest of the content
(I have problems posting pictures, but checkout [frontend/src/v5/ui/controls/accordion/accordion.component.tsx](https://github.com/3drepo/3drepo.io/pull/4196/files/db1d82ac58fcf5619c69b4bd6c348612de8c8f74#diff-85f1acf0bb813aba2edad551be532ac7690ad44123825dc11d17da5df75831c1) and [frontend/src/v5/ui/controls/accordion/accordion.styles.tsx](https://github.com/3drepo/3drepo.io/pull/4196/files/db1d82ac58fcf5619c69b4bd6c348612de8c8f74#diff-138913dbc3a8af50b5bb822b9cb4cbaeb6bd4975dabf6be8bfc48f7b892537b7))

P.S. I forgot to remove the prop required which is still being passed to the Accordion for no reason

The-Daniel (dismissed):

sanmont3drepo:
- [ ] Update the camera in a viewpoint with a smart group -> go to view no longer works for that smart group

carmenfan (changes requested):
2 things I spotted whilst I used this to debug #4296:

1. image
Can we label the rules the same way as we do in v4? This is too wordy and unnecessary.
@george-hadfield  The formatting on the rules chip doesn't look too great, can you take a look?

2. smart group should be default when you try to create a new group
image

carmenfan:
### Required view validation doesn't seem to go away even if i've filled in everything.

#### Steps to replicate
- try to create a ticket with the template with a required view (you can use `Many Views` in `devTeamWorkspace`)
- in the required view property, add an image, then camera, then group separately

#### Current behaviour
- The error message is shown even when i've filled in everything

#### Expected behaviour
- A view should be considered filled in if any one of the sub property has data.

carmenfan:
### Groups is enabled even when there's no groups inside
When you go into the groups panel and exit without creating any groups, you can see the group section stays disabled looking. but if you saved a view that has no groups, it's enabled

#### Steps to replicate
- Save a view that has no groups configuration via `save current view`

#### Current behaviour
Groups button is enabled implying there's groups inside, but it's empty if you inspect it

#### Expected behaviour
It should be disabled as there's no groups.

carmenfan:
### Delete view is not deleting the image

#### Steps to replicate
- Save the current view
- delete view

#### Current behaviour
Groups and views are removed, image remains

#### Expected behaviour
everything should be wiped.
img

carmenfan:
### Smart group should be the default option

image

carmenfan:
#### Wrong blurb is shown on manual groups
It's showing the smart groups instructions when manual group is toggled
image

carmenfan:
### Imposing colour on the groups panel is removing them else where

Enabling colour override on the tickets panel is clearing colour override else where. We expect this to happen when we enter the groups panel (where we reset the state to the state specified in the UI, but we don't expect that with interaction on the check boxes

img

carmenfan:
#### Clear all overrides does not update the panel state

If I click clear all overrides on the bottom tool bar it should update the tickets groups UI panel (same with show all with hidden)
img

carmenfan:
### Icons do  not match the design

image
image

carmenfan:
### Opacity value is obscured

image

carmenfan:
### Confirmation button in the colour picker

Apparently X is to cancel, and clicking away from the window is to commit.

If I can't get my head around this our users won't be able to. We need a OK button.

carmenfan:
#### Missing rules pasting feature
image
image

carmenfan:
### Smart groups doesn't work until you refresh

Add or edit a smart group, upon update, the UI shows no object and will not highlight/colour them until you refresh.

img

carmenfan:
### use same syntax as v4 for smart rules 
Can we not reinvent the wheel and use the v4 syntax for rules.


Also drop the bold as it just makes things more unreadable at that font size.

image

carmenfan:
### Hover buttons on groupings

There should be hover buttons here too. I should be able to isolate/delete all the groups underneath a grouping. Ideally we should be able to edit the name of the grouping (but nice to have)

image

george-hadfield:
Colour windows should align

image

george-hadfield:
Filter chip text too small and is being cut off
Increase text size and expand width to accommodate text

image

Amantini1997:
> Filter chip text too small and is being cut off Increase text size and expand width to accommodate text
> 
> image

Not sure how you got the text being cut off, it seems to contain the whole text for me 
image

carmenfan:
#### V4 override all button is effecting the custom tickets groups

The button should only apply to the v4 groups (unlike the bottom tool bar one)
img

carmenfan:
#### User should not be able to delete a required view
We know the backend won't allow it, may as well remove that option if it's required.
img

carmenfan:
> > Filter chip text too small and is being cut off Increase text size and expand width to accommodate text
> > image
> 
> Not sure how you got the text being cut off, it seems to contain the whole text for me image

also not seeing the issue here 
image

Amantini1997:
**hidden groups are not correctly handled**
Hidden groups in default view do not hide when focusing/entering a ticket and do not unhide when exiting

https://github.com/3drepo/3drepo.io/assets/43547703/1137aaff-43ea-4187-aad3-874e92629abe

carmenfan:
> Hidden groups in default view do not hide when focusing/entering a ticket and do not unhide when exiting

@Amantini1997 @sanmont3drepo  they don't need to unhide when exiting. We don't reset the view when the user exit a ticket. we should only be resetting a view when a new view is imposed

carmenfan:
#### Highlighted objects should turn into coloured groups when we save current view

If there are highlighted objects when we save the current view, they should turn into coloured groups (just like currently coloured objects).

different colours should turn into different groups, i.e. in this situation:
image


If I save the view now (and there are no colour overrides enabled), it should create 2 coloured groups, one for the purple highlights, and one for the yellow highlights.

carmenfan (approved):

PR #4202, opened 2023-04-28
Issue 4201 - modified zombie script

### LINKED ISSUE #4201: Missing file detection script
A missing file detection script in utils to ensure all refs are catered for by the fs of choice.

If any files are missing, list out the entry and their expected location

keifgwinn:
This fixes #4201

#### Description
New version of zombie file detection script

#### Test cases
- should be more robust and doesn't fail when there is a large amount of files in the FS
- should output the list of zombie files after check
- should remove zombies reliably if delete is flag


carmenfan:
@Charence  tagging you as this is related to https://github.com/3drepo/DevOps/issues/582

Charence (approved):

PR #4203, opened 2023-05-02
ISSUE #4198 - Mesh Optimisation

### LINKED ISSUE #4198: Mesh optimisation
### Description
Changes required for mesh optimisation testing

The-Daniel:
This fixes #4198

This PR is to test the [Mesh Optimisation Bouncer](https://github.com/3drepo/3drepobouncer/issues/599). It has a couple of parameters added to the Create Revision dialog in order to control the optimisation.


carmenfan:
closing this PR as the feature is now on hold

sebjf:
Reopening for new ODA parameters

carmenfan:
closing the PR again to save deployment cost as we're not actively doing naything here.

reopen when necessary

sebjf:
Reopening now we have lod settings and the config needs to be changed https://github.com/3drepo/3drepobouncer/issues/632

sebjf:
Hi @carmenfan @The-Daniel  the config has been updated to use a new (integer) parameter, `lod`, instead of the surface and normal tolerances.

0 is the 'default' (whatever the importer considers that to be - basically the behaviour we have now if the user doesn't choose anything), and 1-6 correspond to the Plugin's LODs (Very Low, Low, Medium, High, Very High, Maximum), which map to the parameters we've found in 632 (inside FileProcessorRvt).

Typing one of those numbers works through to bouncer now, so I think all that needs to happen is the sidebar form be updated to replace the `FormNumberField` with a drop down.

carmenfan (changes requested):
@The-Daniel 

Can we make it so the LoD option is only available if the extension is rvt or rfa?

Essentially we don't currently support LoD variation for any other file format so it will be misleading

carmenfan:
@The-Daniel  your changes look fine to me (thank you!)

Not approving this just yet as I need to delve deeper to see if  the actual LOD level is doing things right as I'm not 100% convince it's correct.

carmenfan (dismissed):
image

image

PR #4215, opened 2023-05-09
Issue 3624 - Assignees Select

### LINKED ISSUE #3624: Custom tickets - Assignees list
### Description
Right now the assignees modal is very basic. In order to provide a bit more of information we are pushing a new design.
- The list is separated in two lists:  roles and users with a divider between them
- The list is ordered by first name
- The part with users have the avatar, fullname and role in the company
- On hover in the users the tooltip with more information for the user will appear (reuse the same tooltip we have) 
- The list should maintain the search capabilities (including highlighting)

### Goals
- [ ] As a user I want to have a friendlier way to search the users  

### Tasks
- [ ] Create the Roles item component
- [ ] Create the user item component
- [ ] Add hover to user item

### Related Resources
[Mock up](https://www.figma.com/file/3FjNEOdQbcHwzquVZxUqMl/3D-Repo?node-id=2724%3A22061&t=9fVWe6lH0lpqQNe9-0)

The-Daniel:
This fixes #3624 
#### Description
- Add new select menu when adding/removing custom tickets assignees
- Refactored code for popover circles and assignees circle lists
- Updated job circle to include job colour

#### Test cases
Try out the assignees select in both the tickets list and the details view


Amantini1997:
When you toggle any checkbox, the images for Users refresh 
image

frontend/src/v5/ui/controls/assigneesSelect/assigneesSelectMenu/assigneesSelectMenu.component.tsx
```diff
@@ -0,0 +1,131 @@
...
+					const [filteredUsers, filteredJobs] = partition(filteredItems, isUser);
+					return (
+						<HiddenSelect
+							open={open}
+							value={value || []}
+							multiple
+							MenuProps={{
+								disableAutoFocusItem: true,
+								PaperProps: {
+									style: { maxHeight: 531, maxWidth: 218 },
```
Amantini1997, line 72: Just as a suggestion, when I need to style nested Mui components and I can only do it passing props, like in this case, I use the styled.attrs({ }) method, so that the styling is kept confined in the `.styles.ts` file.

Amantini1997 (changes requested):

george-hadfield:
### Change the admin job colour to #EB5281

image

george-hadfield:
### Add 3px margin right to list items in assignee list

The scrollbar is too tight to the checkboxes

image

george-hadfield:
### Make assignees dropdown appear to the right side of the assignees

image

george-hadfield:
### Make assignees dropdown appear to the right side of the assignees

image

george-hadfield:
### Change Job predefined colours

I've created a new artboard displaying which colours to use for each of the jobs. Each colour used is from the 'User library' that was implemented by Alessandro for 'Groups' in custom tickets.

https://www.figma.com/file/bRqIl0g9psqD3ISkmxwcNp/3D-Repo?node-id=80%3A31260&t=5DbwJMXqLTYmQCQC-1

image

george-hadfield:
### Reduce height of assignees drop down so it appears below the assignee circles

the other dropdown clickable elements have their menus appear below so this needs to function in the same way. It doesn't matter if the list is smaller in heigh than it is now as its an ordered list with a search bar.

image
image

george-hadfield (changes requested):
A few small changes

frontend/src/v5/ui/controls/assigneesSelect/controlledAssigneesSelect.component.tsx
```diff
@@ -0,0 +1,40 @@
+/**
+ *  Copyright (C) 2022 3D Repo Ltd
```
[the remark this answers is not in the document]
sanmont3drepo: 🤦‍♂️

sanmont3drepo:
Its not saving the assignees:
assignees

frontend/src/v5/ui/controls/assigneesSelect/assigneesSelectMenu/assigneesSelectMenu.component.tsx
```diff
@@ -44,88 +42,76 @@ export const AssigneesSelectMenu = ({
...
+				/>
+			</SearchInputContainer>
+			<ListSubheader>
+				<FormattedMessage id="assigneesSelectMenu.jobsHeading" defaultMessage="Jobs" />
+			</ListSubheader>
+			{jobs.length > 0 && jobs.map(({ _id }) => (
+				<AssigneesSelectMenuItem
+					key={_id}
+					assignee={_id}
+					value={_id}
```
Amantini1997, line 80: Where `AssigneesSelectMenuItem` is used, `assignee` and `value` are the same (i.e. `_id` here and `user.user`) later. Is there a reason why they are two separate props?

Amantini1997:
When passing from nothing selected to 1 job or 1 user, the position of the element goes either up or down  

https://github.com/3drepo/3drepo.io/assets/43547703/6732955d-83ca-4e36-961f-ff2deed0fa42

Amantini1997:
The click to select the assignees happens even if you are not hovering the button (notice not even the tooltip is triggered)

Uploading Registrazione 2023-05-17 161123.mp4…

frontend/src/v5/ui/controls/assigneesSelect/assigneesSelectMenu/assigneesSelectMenu.component.tsx
```diff
@@ -0,0 +1,117 @@
...
```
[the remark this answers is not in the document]
The-Daniel: Because of the way MUI Selects work the value prop doesn't actually reach the custom component. The select also uses the value prop of its immediate children, which in this case is <AssigneeSelectMenuItem /> and not the <ListItemContainer /> inside AssigneeSelectMenuItem, so you can't set the key or value in there.

From some cursory googling it seems possible to pass value to custom components by using it like a function. Something like this...
image

But personally I find that more confusing than just passing in a duplicate prop. I don't know what you think or if you can think of a better work around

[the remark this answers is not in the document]
Amantini1997: Ahh ok, yeah. I bumped into this issue before, you are right. value gets mapped to `data-value`.
You could potentially access it doing props['data-value'], but I am not sure it's a good idea.
I will mark this as resolved

frontend/src/v5/ui/controls/assigneesSelect/assigneesSelect.styles.ts
```diff
@@ -58,3 +65,23 @@ export const AssigneesListContainer = styled.div`
...
+		outline: none;
+		margin: 0;
+		&::before {
+			background-color: transparent;
+		}
+	}
+`;
+
+export const Tooltip = styled(TooltipBase)`
+	margin: 0 0 0 13px;
```
Amantini1997, line 86: image
The component is now misplaced and you can click on the red cross and it will still open.
This is probably due to the fact that you would be clicking on the margin, which is part of the component, even though you are not hovering it.

You can fix that by replacing the style of the tooltip to 
```ts
margin: 0;
width: 24px; // so that the tooltip has the same width of the button
```

Amantini1997 (changes requested):

frontend/src/v5/store/users/users.sagas.ts
```diff
@@ -22,15 +22,16 @@ import { DialogsActions } from '@/v5/store/dialogs/dialogs.redux';
...
 
 export function* fetchUsers({ teamspace }: FetchUsersAction) {
 	try {
 		const users = yield select(selectUsersByTeamspace, teamspace);
 
 		if (isEmpty(users)) {
 			const { data: { members } } = yield API.Users.fetchTeamspaceMembers(teamspace);
-
-			yield put(UsersActions.fetchUsersSuccess(teamspace, members));
+			const membersWithImg = members.map((member) => ({ ...member, avatarUrl: getMemberImgSrc(teamspace, member.user), hasAvatar: true }));
```
sanmont3drepo, line 33: I think in general this is not needed. It should be done only in the avatar component files.
I created an issue to tackle this in the future
https://github.com/3drepo/3drepo.io/issues/4232

sanmont3drepo (approved):

Amantini1997 (changes requested):

Amantini1997 (approved):

Issue #4250, opened 2023-05-31
App crashes when trying to edit a federation where some of the containers have insufficient permissions 


The-Daniel:
### Description

For a federation that contains at least one model where the user has 'unassigned' permissions clicking edit federation briefly opens the edit federation modal before crashing.

Seems to be because the edit federation modal loops through the containers to get the _ids of the constituent containers, but the unassigned containers are undefined.

NOTE: This bug is similar but different to #4248. Changes made for that issue might also affect this one

- This is happening in: both 
- This requires a hotfix to production: no 

### Steps to replicate
- Create a federation with containers A and B
- Set the permissions for a user so that they can view container A, but container B is 'unassigned'
- As that user attempt to edit the federation

### Current Behaviour
- The modal briefly opens when the containers load the app crashes

### Expected Behaviour
- The modal successfully opens and displays the containers the user has permissions to

### Pictures/ Gifs of the error

![image](https://github.com/3drepo/3drepo.io/assets/62481250/6337d5c9-8081-41c9-ad4d-cfdb445fa030)

PR #4255, opened 2023-06-05
ISSUE #4248 - Fix crashes when a Federation contains an 'unassigned' container

### LINKED ISSUE #4248: Receiving 401 error when  viewing a federation where some of the containers have insufficient permissions
### Description

You get a 401 error message when you open up a federation where one or more constituent containers have 'unassigned' as the user permission.
This is occurring because there is a `fetchContainerStats` call being made (in viewer.sagas.ts). This is done for a check (in viewer.tsx) to see if all the containers have no revisions. If all the containers are empty the "Federation is empty" message gets called (in viewer.tsx).
There should be a different way of checking if the federation has no valid models to show that doesn't use `fetchContainerStats`

- This is happening in: both 

### Steps to replicate
- Create a federation with containers A and B
- Set the permissions for a user so that they can view container A, but container B is 'unassigned'
- As that user attempt to view the federation

### Current Behaviour
- You get a 401- NOT_AUTHORIZED error and are redirected to the project page

### Expected Behaviour
- You can access the model and only see the containers you have permissions to

MicrosoftTeams-image (2)

The-Daniel:
This fixes #4248  and #4250 

#### Description
- Fixed a crash that occurred when opening a mixed permissions federation. When the containers stats are fetched only containers which the user has permissions to are used
-  Fixed a crash that occurred when editing a mixed permissions federation. Unassigned containers were showing up as 'undefined'. These are now filtered out

#### Test cases
-
    - Create a federation with containers A and B
    - Set the permissions for a user so that they can view container A, but container B is 'unassigned'
    - As that user attempt to edit the federation
    - As that user attempt to view the federation

- Opening a federation with no containers will still show the 'Add containers' screen

- Opening a federation that contains only empty containers will still show the 'Upload revisions' screen

### DIFF
--- a/frontend/src/v5/ui/routes/dashboard/projects/federations/editFederationModal/editFederation/editFederation.component.tsx
+++ b/frontend/src/v5/ui/routes/dashboard/projects/federations/editFederationModal/editFederation/editFederation.component.tsx
@@ -26,6 +26,7 @@ import { SearchContextComponent } from '@controls/search/searchContext';
 import { Tooltip } from '@mui/material';
 import { useCallback, useEffect, useState } from 'react';
 import { FormattedMessage } from 'react-intl';
+import { identity, pickBy } from 'lodash';
 import { useContainersData } from '../../../containers/containers.hooks';
 import { IconContainer, IncludeIcon, RemoveIcon } from './editFederation.styles';
 import { ActionButtonProps, EditFederationContainers } from './editFederationContainersList/editFederationContainersList.component';
@@ -52,7 +53,7 @@ export const EditFederation = ({ federation, onContainersChange }: EditFederatio
 	}, [containers]);
 
 	useEffect(() => {
-		onContainersChange(includedContainers);
+		onContainersChange(pickBy(includedContainers, identity)); // filter out 'undefined' which are unassigned containers
 		setAvailableContainers(containers.filter((container) => !includedContainers.includes(container)));
 	}, [includedContainers]);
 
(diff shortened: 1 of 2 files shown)


Amantini1997:
Trying to create a federation seems to not work anymore. The submit button is stuck in a loading state

Amantini1997:
Trying to edit a federation, the submit button is always available and onSubmit the page refreshes

frontend/src/v5/ui/routes/dashboard/projects/federations/editFederationModal/editFederation/editFederation.component.tsx
```diff
@@ -52,7 +53,7 @@ export const EditFederation = ({ federation, onContainersChange }: EditFederatio
 	}, [containers]);
 
 	useEffect(() => {
-		onContainersChange(includedContainers);
+		onContainersChange(pickBy(includedContainers, identity)); // filter out 'undefined' which are unassigned containers
```
Amantini1997, line 56: `pickBy` seems to return an object rather than an array, maybe you could use `includedContainers.filter(Boolean)`?

I think this should also fix the problems I mentioned early in the main feed

Amantini1997 (changes requested):

Amantini1997:
If I try to edit a federation that contains a container I am unassigned to, the app crashes

frontend/src/v5/ui/routes/dashboard/projects/federations/editFederationModal/editFederation/editFederation.component.tsx
Amantini1997: Should the `unassigned` container be displayed though when a collaborator is editing the federation?

Amantini1997 (approved):

sanmont3drepo (approved):
I created a separate issue to treat the edit federation modal problems #4263

PR #4313, opened 2023-07-06
ISSUE #4308 - Now in tickets is sending a timestamp when the date fields are changed

### LINKED ISSUE #4308: Dates in custom tickets  are being updated/created using the wrong format
### Description
In  custom tickets when a date property is being updated  a string is being sent instead of a number

- This is happening in: both 

### Steps to replicate
- Open the developers tools
- Update a date property of a custom ticket 

### Current Behaviour
- It sends a string

### Expected Behaviour
- It should send a timestamp

image

sanmont3drepo:
This fixes #4308

#### Description
- Changed due date to use timestamp
- Changed date property fields to use timestamp
- Should sent a timestamp instead of a string for dates

#### Test cases
- update a due date (in existing and a new ticket)
- update a due date in the details ticket view  (in existing and a new ticket)
- update another type of date in the details ticket view  (in existing and a new ticket)

### DIFF
--- a/frontend/src/v5/ui/controls/dueDate/baseDueDate.component.tsx
+++ b/frontend/src/v5/ui/controls/dueDate/baseDueDate.component.tsx
@@ -18,6 +18,7 @@
 import { formatDayOfWeek } from '@controls/inputs/datePicker/dateFormatHelper';
 import { DatePicker } from '@mui/x-date-pickers';
 import { ReactElement, useState } from 'react';
+import dayjs from 'dayjs';
 import { StopBackgroundInteraction } from './dueDate.styles';
 
 type IBaseDueDate = {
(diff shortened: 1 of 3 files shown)


frontend/src/v5/ui/controls/dueDate/baseDueDate.component.tsx
```diff
@@ -38,17 +39,18 @@ export const BaseDueDate = ({ value: initialValue, disabled, onBlur, ...props }:
 		if (!open) setOpen(!disabled);
 	};
 	const onDateChange = (newValue) => {
+		const timestamp = !newValue ? null : newValue.toDate().getTime();
```
The-Daniel, line 42: I find the use of '!' here unusual, but I don't care too much

The-Daniel (dismissed):

frontend/src/v5/ui/controls/dueDate/baseDueDate.component.tsx
sanmont3drepo: Maybe thiw would be better?
```suggestion
		const timestamp = newValue?.toDate()?.getTime() || null;
```

The-Daniel (approved):

PR #4316, opened 2023-07-07
Issue 4299 - Custom Ticket: Groups -> fix performance issue

### LINKED ISSUE #4299: Custom Ticket: Groups - Bug fixes
### Description
- [x] Update a smart group in a viewpoint -> Update the camera -> go to view no longer works for that viewpoint; (mentioned in #3933 but cannot reproduce)
- [x] Viewpoint should be valid if it has at least one of the following: camera, clippingPlanes, state;
- [x] In a new ticket, `save current view` attempts to save the ticket;
- [x] (this may depend on the bug above) In a new ticket, with some groups selected in the groups panel, `save current view` does not save the viewpoint
- [x] selecting the no-color from the palette for a colored group triggers error modal
> This may not be a quick fix. It may need to tweak modest amount of code 
- [x] Hover state of "add to favourites" icon (in containers/fed list) has wrong background color

<i>The following bug has been fixed in #4285</i>
- [ ] Change the camera position and the checked overrides, then update/create a group. The camera gets repositioned and the overrides are wrong (if this is done in a view different from the `ticket default view`, the overrides do not even match the groups)
- [x] Select groups override in v4 and try to create a viewpoint . The app crashes. It should create the new viewpoint whith the overrides a groups.

Amantini1997:
This fixes #4299

#### Description
The performance issue was related to the way groups were edited/added/removed.
> **Important** In the frontend, we have a datatype called "groupOverride" which follows this structure
```ts
interface GroupOverride { 
    color?: [number, number, number],
    opacity?: number,
    group: Group | string, // <-- independent entity in the backend
}
```
> Although we visualise color and opacity as part of a group in the frontend, in the backend, a group is a independent entity. 

In the backend, there is an endpoint to **edit** a ticket group, but no endpoints to **add/remove** a ticket group, so the way to go is to send the whole ticket:
- to DELETE a group, we shall send all the groups but the one to delete.
- to CREATE a group, we shall send all the groups plus the new one.

Since the other groups already exist in the database, we can send the groups that were not modified as ids (if we were to send the whole group, the backend would create a brand new group, which is what caused the performance issue).
> We will still send the color and the opacity for those tickets though as they are part of the ticket itself

As such, when we want to update a group, if we changed the color and/or the opacity, we shall still use the update ticket endpoint and pass the ticket as a string (its id), but shall also call the update ticket group endpoint.

#### Test cases
Test the performance when editing/adding/deleting a group. Test changes in the color/opacity only, in the group only, and in both.
Test smart groups.
Test also the realtime events (play around with groups which may or may not be checked and see if they update correctly when a group is edited/deleted/created in realtime)

### DIFF
--- a/frontend/src/v5/ui/routes/viewer/tickets/ticketDetails/ticketsDetailsCard.component.tsx
+++ b/frontend/src/v5/ui/routes/viewer/tickets/ticketDetails/ticketsDetailsCard.component.tsx
@@ -20,7 +20,7 @@ import { useContext, useEffect } from 'react';
 import { useParams } from 'react-router-dom';
 import { TicketsHooksSelectors, TicketsCardHooksSelectors } from '@/v5/services/selectorsHooks';
 import { TicketsCardActionsDispatchers, TicketsActionsDispatchers } from '@/v5/services/actionsDispatchers';
-import { modelIsFederation, sanitizeViewVals, templateAlreadyFetched } from '@/v5/store/tickets/tickets.helpers';
+import { findEditedGroup, modelIsFederation, sanitizeViewVals, templateAlreadyFetched } from '@/v5/store/tickets/tickets.helpers';
 import { getValidators } from '@/v5/store/tickets/tickets.validators';
 import { FormProvider, useForm } from 'react-hook-form';
 import { CircleButton } from '@controls/circleButton';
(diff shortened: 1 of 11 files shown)


frontend/src/v5/ui/routes/viewer/tickets/ticketDetails/ticketsDetailsCard.component.tsx
```diff
@@ -73,12 +73,22 @@ export const TicketDetailsCard = () => {
...
+		const editedGroup = findEditedGroup(validVals, ticket, template);
+		if (editedGroup) {
+			TicketsActionsDispatchers.updateTicketGroup(
+				teamspace,
+				project,
+				containerOrFederation,
+				ticket._id,
+				editedGroup,
+				isFederation,
+			);
```
sanmont3drepo, line 87: Is not necessary to continuing doing stuff. Is one endpoint or the other.
```suggestion
			);
			return;
```

Amantini1997: The color is a ticket property in the backend, but in the frontend is part of a group. If the user updates the color plus anything else, both the endpoints must be called

PR #4337, opened 2023-07-14
ISSUE #4233 fix email formatting

### LINKED ISSUE #4233: Fix layout on emails
### Description
image

layout is severely squished when it is viewed inside outlook, we need to tweak it 

BaseTemplate and SystemTemplate are the 2 templates that drives this UI: https://github.com/3drepo/3drepo.io/tree/master/backend/src/v5/services/mailer/templates/html

### Goal
- Ensure it looks readable and sane in MS Outlook
- Ensure it still looks decent on mail providers like Gmail

carmenfan:
This fixes #4233

#### Description
turns out the logoDiv wasn't closed properly so the whole content div is being seen as a child of the logo div, hence the restricted width formatting

#### Test cases
email should render correctly now

### DIFF
--- a/backend/src/v5/services/mailer/templates/html/baseTemplate.html
+++ b/backend/src/v5/services/mailer/templates/html/baseTemplate.html
@@ -170,7 +170,7 @@
 </head>
 <body>
     <div class="content">
-        <div class="logoDiv" />
+		<div class="logoDiv"></div>
         <div class="logoDiv">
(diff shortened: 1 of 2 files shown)


Charence (approved):
Kind of weird as I thought `<something />` was supposed to be shorthand for `<something></something>`. Did a test and this makes a difference in Outlook.

PR #4414, opened 2023-08-22
ISSUE #4383 - Group field name rules addition

### LINKED ISSUE #4383: Supporting more options on field name variations on smart rules
### Description
See [scoping document](https://github.com/3drepo/3D-Repo-Product-Team/issues/308) for full context.

**This is a sub task of #4386. Please branch/merge with that branch instead of staging.**

We need to augment the [rules schema](https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/schemas/rules.js) to support varied field names, similar to values.

The following operators need to be supported:
- `IS` : field name is an exact match to the string
- `CONTAINS`: the string is a substring within the field name i.e `/{string}/i`
- `STARTS_WITH`: field name begins with the string i.e. `/^{string}/i`
- `ENDS_WITH`: field name ends with the string i.e. `/{string}$/i`
- `REGEX`: user provides a regex to specify what they would like.

#### Schema changes
We would expect the rule to look something like this

````js
{
  field: "fruits",
  operator: "EQUALS",
  values: ["apples", "oranges"]
}
````

````js
{
  name: "rule name"
  field: { operator: "IS", values: ["fruits", "items"]},
  operator: "EQUALS",
  values: ["apples", "oranges"]
}
````
##### Notes
- All operators support 1 or more values, with the exception of regex.
- We need to ensure we still support the old schema to ensure compatibility for any incoming requests, we need a migration for existing data
- name is a new field and is required. for existing data, we need a database migration to add a name based on the logic from the front end.

### Goals
- [ ] To be able to specify rules with variation on field names
- [ ] Old schema should still function correctly.
- [ ] Ensure this function is supported in both v5 ticket groups and v4 groups

### Tasks
- Schema updates
  - [ ] Update [rules schema](https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/schemas/rules.js#L79) to support the new data structure
  - [ ] Ensure old schema is still understood (I would do this in rules schema and have a cast function)
    - [ ] we should only output the rules under new schema (i.e. convert it to new schema if we're returning a smart group with old data
    - [ ] old schema coming in from a network request will be understood and converted
    - [ ] old schema in existing groups will be understood and converted
- [ ] [Translates](https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/models/metadata.rules.js#L32 ) the new operators to mongo database queries 
- [ ] Ensure same logic is applied for v4 groups (make v4 codebase use v5 functions if appropriate)
- Unit Tests
  - [ ] Ensure all operators have tests
  - [ ] Ensure old schema have tests
- e2e Tests
  - [ ] Ensure all operators have tests
  - [ ] Ensure old schema have tests
- [ ] Data migration
  - [ ] Add a name to all rules, based on the logic currently on the front end
  - [ ] Change all `{ field: string }` to `{ field: {operator: "IS", value: [string]}}` (make it separate, we will run it one release later)

### Related Resources
Figma: https://www.figma.com/file/bRqIl0g9psqD3ISkmxwcNp/3D-Repo?type=design&node-id=715-17375&mode=design&t=QUnvuU28wOnbtypc-0
Scope: https://github.com/3drepo/3D-Repo-Product-Team/issues/308

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #4383

#### Description
Option for the user to add field name rules in groups and ticket groups for v4 and v5. 
Migration script added to change existing data.
Tests and swagger schema adjusted.

#### Test cases
Group rules should now require a name and field property should be an object with operator and values properties.
Old schema should still be accepted but should be converter to the new schema before added in the database.
Groups that still have old schema should be converted to the new schema when displayed.
Running the migration script should convert all old-schema groups to new-schema.

- [x] Should be able to create a v5 ticket manual group as before
- [x] Should be able to create a v5 ticket smart group with the old schema
  - [x] Should fail without name
- [x] Should be able to create a v5 ticket smart group ticket with new schema
  - [x] Should fail without name
  - [x] Should fail with unknown operator
  - [x] Should fail with invalid values for the given operator
- [x] Should be able to create a v4 manual group as before
- [x] Should be able to create a v4 smart group with the old schema
  - [x] Should fail without name
- [x] Should be able to create a v4 smart group ticket with new schema
  - [x] Should fail without name
  - [x] Should fail with unknown operator
  - [x] Should fail with invalid values for the given operator

### CODE REVIEW

**Comment by ChristosTsiotsias** [comment]
azp-run

**Review by carmenfan — CHANGES_REQUESTED** [review]
@ChristosTsiotsias  We should also have e2e tests for both a new version of the example and old? or is that already included by the virtue of you changing the service helper?

File: backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/commons/groups.js [hunk 1]
```diff
@@ -47,7 +60,10 @@ Groups.validateGroupsImportData = async (req, res, next) => {
 	try {
 		const { groups } = req.body;
 		if (!groups?.length) throw createResponseCode(templates.invalidArguments, 'Groups array cannot be empty');
-		await Promise.all(groups.map(validateSchema));
+		await Promise.all(groups.map((g) => {
```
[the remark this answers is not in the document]
ChristosTsiotsias: when you pass strict = true the transformations are not executed. But if I pass strict = false it accepts different data types that required ... for example it would accept values : [1,2,3] in STARTS_WITH and convert them to strings thats why I thought this is the only solution

backend/src/v5/models/tickets.groups.js
```diff
@@ -33,8 +44,13 @@ Groups.deleteGroups = async (teamspace, project, model, ticket, groupIds) => {
 	await deleteMany(teamspace, GROUPS_COL, { teamspace, project, model, ticket, _id: { $in: groupIds } });
 };
 
-Groups.getGroupsByIds = (teamspace, project, model, ticket, groupIds, projection) => find(
-	teamspace, GROUPS_COL, { teamspace, project, model, ticket, _id: { $in: groupIds } }, projection);
+Groups.getGroupsByIds = async (teamspace, project, model, ticket, groupIds, projection) => {
+	let groups = await find(teamspace, GROUPS_COL,
+		{ teamspace, project, model, ticket, _id: { $in: groupIds } }, projection);
+
+	groups = await Promise.all(groups.map(convertGroupRules));
```
carmenfan, line 51: again, no reason to assign groups as a `let` just to assign it then immediately return it.

backend/src/v5/models/tickets.groups.js
```diff
@@ -52,13 +68,15 @@ Groups.updateGroup = async (teamspace, project, model, ticket, groupId, data, au
...
 	projection = { teamspace: 0, project: 0, model: 0, ticket: 0 }) => {
-	const group = await findOne(
+	let group = await findOne(
 		teamspace, GROUPS_COL, { teamspace, project, model, ticket, _id: groupId }, projection);
 
 	if (!group) {
 		throw templates.groupNotFound;
 	}
 
+	group = await convertGroupRules(group);
```
carmenfan, line 78: similar here

backend/src/v5/models/groups.constants.js
```diff
@@ -0,0 +1,53 @@
+/**
```
carmenfan, line 1: Can we rename this to  metadata.rules.constants.js

The operators are on the rules afterall 😆

backend/src/v5/schemas/rules.js
```diff
@@ -15,99 +15,68 @@
...
-	case 'NOT_IN_RANGE':
-		return true;
-	default:
-		return false;
+const Rules = {};
+
+const validateValuesArray = (operatorName, values) => {
+	const operator = OPERATORS[operatorName];
+	if (operator) {
+		const { minValues } = operator;
```
carmenfan, line 36:
```suggestion
		const { minValues, maxValues } = operator;
```

backend/src/v5/schemas/rules.js
```diff
@@ -15,99 +15,68 @@
...
+		operator: Yup.string().uppercase().oneOf(Object.keys(FIELD_NAME_OPERATORS)).required(),
+		values: Yup.array().required()
+			.when('operator', (operator, schema) => schema
+				.test('Field rules validation', 'field values field is not valid with the field operator selected',
+					(value) => validateValuesArray(operator, value))),
+	}).transform((val, oldVal) => (isString(oldVal)
+		? { operator: OPERATORS.IS.name, values: [oldVal] } : val)).required(),
+	operator: Yup.string().uppercase().oneOf(Object.keys(FIELD_VALUE_OPERATORS)).required(),
+	values: Yup.array()
+		.when('operator', (operator, schema) => schema
```
carmenfan, line 64: same here

backend/tests/v5/helper/services.js
```diff
@@ -566,8 +567,17 @@ ServiceHelper.generateGroup = (isSmart = false, {
 	if (isSmart) {
 		group.rules = [
 			{
+				name: ServiceHelper.generateRandomString(),
 				field: 'IFC GUID',
-				operator: 'IS',
+				operator: OPERATORS.IS.name,
+				values: [
+					'1rbbJcnUDEEA_ArpSqk3B7',
```
carmenfan, line 574: you didn't crate it but we should probably just do generateRandomString() here (and the one below)

backend/tests/v5/unit/middleware/dataConverter/inputs/teamspaces/projects/models/commons/groups.test.js
```diff
@@ -79,6 +80,13 @@ const deserialiseGroup = (group) => {
 		}
 	}
 
+	if (output?.rules) {
+		output.rules = output.rules.map((r) => ({
```
carmenfan, line 84: Since you don't want to copy the field, split it with the spread operator and you can address it directly as well.
```suggestion
		output.rules = output.rules.map(({field, ...rest}) => ({
		   ...rest
```

carmenfan (changes requested):

carmenfan:
@ChristosTsiotsias  also need an answer regarding this: 
image

ChristosTsiotsias:
> @ChristosTsiotsias also need an answer regarding this: image

even though there is not a defined test for it that should be covered because I changed the service helper to generate a group of mixed rules 

image

are we ok with that or should I change make service helper to return new schema groups and create a test specifically for old schema ?

carmenfan:
> > @ChristosTsiotsias also need an answer regarding this: image
> 
> even though there is not a defined test for it that should be covered because I changed the service helper to generate a group of mixed rules
> 
> image
> 
> are we ok with that or should I change make service helper to return new schema groups and create a test specifically for old schema ?
@ChristosTsiotsias 
I think it'd be helpful to in general make anything generate the new schema, and have explicit test for old schema so it's obvious what's broken when things break.

backend/src/v5/models/tickets.groups.js
```diff
@@ -25,11 +25,11 @@ const { templates } = require('../utils/responseCodes');
 
 const Groups = {};
 
-const convertGroupRules = async (group) => {
+const convertGroupRules = (group) => {
 	const convertedGroup = { ...group };
 
-	if (convertedGroup.rules) {
-		convertedGroup.rules = await Promise.all(convertedGroup.rules.map(castSchema));
+	if (convertedGroup?.rules) {
```
carmenfan, line 31: similar here

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/groups.test.js
```diff
@@ -136,6 +149,16 @@ const testImportGroups = () => {
 		ServiceHelper.generateLegacyGroup(teamspace, container._id, true, false),
 	];
 
+	const groupWithStringField = {
```
carmenfan, line 152: Can we call this legacy field schema

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/groups.test.js
```diff
@@ -161,6 +184,7 @@ const testImportGroups = () => {
 			['the groups are valid', createRoute(), true, { post: { groups: postData.groups.map(({ _id }) => _id) }, data: groups }],
 			['the same groups are being imported again', createRoute(), true, { post: { groups: postData.groups.map(({ _id }) => _id) }, data: groups }],
 			['the user is updating only some of the groups', createRoute(), true, { post: { groups: [...groups.map(({ _id }) => _id), newGroup._id] }, data: expect.arrayContaining([changedGroup, ...groups.slice(1), newGroup]) }, { groups: partialGroupUpdateTestData }],
+			['the groups have string field property', createRoute(), true, { post: { groups: [groupWithStringField._id] }, data: [{ ...groupWithStringField, rules: groupWithStringField.rules.map(castSchema) }] }, { groups: [groupWithStringField] }],
```
carmenfan, line 187: same with the message

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/tickets.groups.test.js
```diff
@@ -19,7 +19,7 @@ const { cloneDeep } = require('lodash');
 const SuperTest = require('supertest');
 const ServiceHelper = require('../../../../../../helper/services');
 const { src } = require('../../../../../../helper/path');
-const { OPERATORS } = require('../../../../../../../../src/v5/models/groups.constants');
+const { FIELD_VALUE_OPERATORS } = require('../../../../../../../../src/v5/models/metadata.rules.constants');
```
carmenfan, line 22:
```suggestion
const { FIELD_VALUE_OPERATORS } = require(`${src}/models/metadata.rules.constants`);
```

backend/tests/v5/helper/services.js
```diff
@@ -43,7 +43,7 @@ const { generateUUID, UUIDToString, stringToUUID } = require(`${src}/utils/helpe
 const { PROJECT_ADMIN } = require(`${src}/utils/permissions/permissions.constants`);
 const { deleteIfUndefined } = require(`${src}/utils/helper/objects`);
 const FilesManager = require('../../../src/v5/services/filesManager');
-const { OPERATORS } = require('../../../src/v5/models/groups.constants');
+const { FIELD_NAME_OPERATORS, FIELD_VALUE_OPERATORS } = require('../../../src/v5/models/metadata.rules.constants');
```
carmenfan, line 46: this and the line above shoudl be referencing `src` instead

backend/tests/v5/unit/middleware/dataConverter/inputs/teamspaces/projects/models/commons/groups.test.js
```diff
@@ -18,7 +18,7 @@
 const _ = require('lodash');
 const { src } = require('../../../../../../../../helper/path');
 const { generateLegacyGroup, generateUUIDString, generateRandomString } = require('../../../../../../../../helper/services');
-const { OPERATORS } = require('../../../../../../../../../../src/v5/models/groups.constants');
+const { FIELD_NAME_OPERATORS, FIELD_VALUE_OPERATORS } = require('../../../../../../../../../../src/v5/models/metadata.rules.constants');
```
carmenfan, line 21: same here

carmenfan: NOTE: we only need the relative path without variables if it's going into a `jest.mock()`

backend/tests/v5/unit/middleware/dataConverter/outputs/teamspaces/projects/models/commons/groups.test.js
```diff
@@ -15,7 +15,7 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-const { OPERATORS } = require('../../../../../../../../../../src/v5/models/groups.constants');
+const { FIELD_NAME_OPERATORS, FIELD_VALUE_OPERATORS } = require('../../../../../../../../../../src/v5/models/metadata.rules.constants');
```
carmenfan, line 18: same here, use the src path in L21 (and the line below this too)

backend/tests/v5/unit/models/metadata.rules.test.js
```diff
@@ -15,7 +15,7 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-const { OPERATORS } = require('../../../../src/v5/models/groups.constants');
+const { FIELD_NAME_OPERATORS, FIELD_VALUE_OPERATORS } = require('../../../../src/v5/models/metadata.rules.constants');
```
carmenfan, line 18: ditto

backend/tests/v5/unit/models/tickets.groups.test.js
```diff
@@ -16,8 +16,9 @@
  */
 
 const { times } = require('lodash');
-const { generateRandomString, generateRandomObject } = require('../../helper/services');
+const { generateRandomString, generateRandomObject, generateGroup } = require('../../helper/services');
 const { src } = require('../../helper/path');
+const { castSchema } = require('../../../../src/v5/schemas/rules');
```
carmenfan, line 21: this too

carmenfan (changes requested):

carmenfan (approved):

PR #4419, opened 2023-08-25
ISSUE #4418 Support for Unity 2023 and allow swapping between them with viewer flag

### LINKED ISSUE #4418: Upgrade Unity to 2023.2/Option to switch viewer to 2023.2 (2023 Update)
### Description
New versions of Unity support more recent versions of emscripten, which support for 4GB address space (1.39+).

We would like to allow specific customers to get access to a version of the viewer built with 2023 and the enhanced memory. 

The new version of Unity does not change how the viewer is loaded. There is a minor change to how API calls are chunked, which has been added to support loading the models the 2023 upgrade has been tested for. (It is quite likely this would be required for those same models on the older version of the viewer too.)

This issue is to (1) properly implement the new chunking behaviour and (2) provide support for switching between viewers in `UnityUtil`.

### Goals
- [ ] Support for loading the existing viewer (3drepounity staging) or a version built with 2023
- [ ] `multipleCallInChunks` operates in smaller chunks (5000) and distributes calls over sequential unity frames (allowing the gc to run between them)

https://docs.unity3d.com/2023.2/Documentation/Manual/webgl-native-plugins-with-emscripten.html

carmenfan:
This fixes #4418

#### Description
Minor adjustments to the front-end to support swapping between viewers based on 2021.2.3 and 2023.x.x, and improving performance on the large models that the 2023 viewer is intended for.

Specifically, this PR:

1. Introduces the UnityUtil.onUpdate method, which should be called from inside the viewer, in order to dispatch chunks from `multipleCallInChunks` across frames. Calling chunks across frames is important because it gives the GC a chance to run.
2. Introduces the `unityBuildSubdirectory` member, and updates UnityUtil and viewer.tsx to use this to work out which version of Unity to load. 
3. Adds a method to swap to a version of Unity built into `frontend/unity/2023/unity`.

Do not merge.


carmenfan:
@sebjf I've altered it so instead of making the frontend call `useUnity2023()` separately, it just takes in a boolean at init to determine the build directory. Let me know if that makes sense to you?

the other thing is that can we make the default viewer also a sub directory within `frontend/unity`? at the moment it looks like this:
image

I think it'll be better if we do this 
image


So it's better organised, and easier to rebuild in the future!

sebjf:
Hi @carmenfan,

Yes I am happy with both a boolean and a dedicated subdirectory for the default! 

The reason I didn't do it like that to start with was because I thought it might break white-label viewers: that they assume that unity-loader.js is found in /unity, whereas it would end up in /unity/default/unity/. Is this not a concern, or have I misunderstood how the white-label viewers work? (I was going by demo.js).

carmenfan:
> Hi @carmenfan,
> 
> The reason I didn't do it like that to start with was because I thought it might break white-label viewers: that they assume that unity-loader.js is found in /unity, whereas it would end up in /unity/default/unity/. Is this not a concern, or have I misunderstood how the white-label viewers work? (I was going by demo.js).

Good point, but I think we can probably overcome that if we tweak the [webpack ](https://github.com/3drepo/3drepo.io/blob/master/frontend/internals/webpack/webpack.common.config.js#L42) a bit. So we get that structure in the public folder but not in the source code bit.

sebjf:
Hi @carmenfan, 

I have updated 4418. 

Now, when building unity the viewers should go into:

`frontend/unity/default/unity`
or
`frontend/unity/beta/unity`

(The final /unity is still needed to get Unity's toolchain to give the right filenames)

When built, the default viewer goes into

`public/unity'

And the beta one into

`public/unity/beta`


(I've also renamed it beta so it isn't outdated in just 3 months!)

PS. do you recall what this line is for?

https://github.com/3drepo/3drepo.io/blob/72cd5a276a488e8a78aa7393f0addf80ce86c87c/frontend/internals/webpack/webpack.common.config.js#L44

I can't find any reference to it in .io and all the loading code uses the proper name of the loader script now.

I have updated webpack for that too so it still works, but it always assumes its the default loader.

carmenfan:
@sebjf The render looks a bit different (left is old, right is unity 2023):
image
image

Issue #4421, opened 2023-08-29
Revision description and format are not supplied to the containerNewRevision realtime event


The-Daniel:
### Description

Some properties are not being supplied to the containerNewRevision realtime event which results in the revision list initially showing nothing when a new revision is uploaded.

The same is also probably true for the update revision realtime event, but currently I believe you can only update the void status so it's not a problem yet

- This is happening in: both 
- This requires a hotfix to production: no 

### Steps to replicate
- Open the revision details for a container
- Upload a new revision to that container with a revision description
- Observe the revisions list

### Current Behaviour
- The description (and format if #3652 is merged) are blank

### Expected Behaviour
- The description (and format) are correctly displayed in the table

### Pictures/ Gifs of the error
![image](https://github.com/3drepo/3drepo.io/assets/62481250/2551d8c0-f029-46b8-a033-afe30e067683)

PR #4422, opened 2023-08-29
ISSUE #3652 - Add file format to revisions details

### LINKED ISSUE #3652: V5 - Cannot tell what format a model was uploaded in
### Description
Currently in V5, there is no clear indication in the UI as to what file format a revision is uploaded in unless we download the revision to see the source filename. In V4, we are able to see the format within the revision information.
This was raised by the commercial team suggesting it should appear again. 

This requires #4375 for an updated endpoint.

### Goals
- As I user, I want to know which file format was the model uploaded in.

### Tasks
- [ ] Add `Format` column to the revision listing and display the file extension provided by the api
- [ ] Some aesthetic changes
  - [ ] background colour of each row
  - [ ] addition vertical strip on the left of each row
  - [ ] Void button style
  - [ ] `added by` is no longer in caps, and spacing between firstname and surname has changed
  - [ ] download icon update
  - [ ] no bold on revision code
  - [ ] new on hover behaviour (detailed in figma)

### Related Resources
Figma: https://www.figma.com/file/bRqIl0g9psqD3ISkmxwcNp/3D-Repo?type=design&node-id=743%3A43993&mode=design&t=gm8B8UbEuif1hEin-1
**Currently in V4:** 
image
**Currently in V5:**
image

The-Daniel:
This fixes #3652

#### Description
- Add column in the revisions details that displays the file format
- Gave the revision details a makeover based on [this ](https://www.figma.com/file/bRqIl0g9psqD3ISkmxwcNp/3D-Repo?type=design&node-id=743-43993&mode=design&t=9umD5BwG1NFbeTVg-0)mockup 

Note: There is a bug where the format (as well as the description) are initially empty when a new revision gets added to the list.
An issue has been made for it here: #4421

#### Test cases
Look at different containers' revisions details containing different file formats


Amantini1997:
Hovering the "added by" column renders both a tooltip and a popover

Amantini1997:
`isVoid` on hover has the same style as `active`, but `active` on hover does not have the same style as `void`. Is this intended?
image

Amantini1997:
Not introduced in this PR, just mentioning it

The "revisions" count actually only refers to the number of active ones.
When switching a revision to void, the count in the item does not update
image

Amantini1997 (changes requested):

The-Daniel:
> Not introduced in this PR, just mentioning it
> 
> The "revisions" count actually only refers to the number of active ones. When switching a revision to void, the count in the item does not update image

I set it so that the revision related things (revision count, latest revision, last updated) update when a void status changes, however this brought with it a problem in that the list item moves up and down the list when you're editing the void statuses (i.e. if the list is sorted by last updated and you void the latest revision then the last updated value goes back in time and therefore the item moves further down the list). I'm not sure how we should handle this... @carmenfan ?

Amantini1997 (approved):

Amantini1997 (approved):

george-hadfield (approved):

PR #4441, opened 2023-09-07
Issue 4219 - Sort tree elements alphabetically

### LINKED ISSUE #4219: Sort tree siblings alphabetically
### Description
siblings within a tree is currently not sorted, so the order is non deterministic.

We should display all siblings in alphabetical order so the user can navigate around to search for the node they're interested in

Amantini1997:
This fixes #4219

#### Description
Tree Sibling are sorted alphabetically (when the search mode has some filters on, the tree is flattened and all the elements are considered siblings)

#### Test cases
Open different containers and federations in the viewer and then expand the tree card.
- Check the tree is sorted;
- Expand some parent nodes and check that children are recursively sorted;
- Use the search tool and confirm that the all the displayed elements are sorted.


frontend/src/v4/modules/tree/tree.selectors.ts
```diff
@@ -59,6 +58,11 @@ export const selectTreeNodesList = createSelector(
 	(treeProcessingData) => treeProcessingData.nodesList || []
 );
 
+export const selectSortedTreeNodesList = createSelector(
+	selectTreeNodesList,
+	(nodesList) => orderBy(nodesList, 'name')
```
The-Daniel, line 63: I think there needs to be a toLowerCase() somewhere here because currently lowercase nodes are always ordered after uppercase nodes

The-Daniel: And perhaps in the transforming,ts file too

The-Daniel (changes requested):

The-Daniel (approved):

PR #4450, opened 2023-09-13
Issue 4448 - Add option to run Unity2023

### LINKED ISSUE #4448: Add an option to run with unity 2023
### Description
We want to offer unity 2023 build of our game as an experimental feature to a selection of our clients, whilst keeping everyone else on the current viewer.

A change has been made in #4418 for the `UnityUtils.init` function to take an optional argument to determine which viewer to load.

We need to wire this to the viewer settings so the user can switch this feature on and off.

#### Changes in viewer settings
image
- Add a toggle button on here for `Use 4GB viewer (Experimental)`
- If this data has been changed, warn the user that they will need to refresh the page in order for this change to take effect
- if this is flagged, pass true to `useBetaViewer` in `UnityUtils.init`.

### Goals
- As a standard user, nothing should change and I should be able to use 3drepo as before.
- As a user looking to evaluate the latest viewer, I should be able to change to the experimental viewer and try out this feature.

### Tasks
TBC by the developer

Amantini1997:
This fixes #4448 

#### Description
Unity 2023 has been added as an experimental feature.
The logic to run it is already in place, but the final user needed an option to decide whether or not to use it.
In Visual settings, a toggle was added to decide which unity version to use.

The setting is stored in the local storage, but not as part of the currentUser. Instead, the same value is used by all the users accessing the platform from the same machine, meaning User1 changes will reflect on User2 account as well

#### Test cases
Open visual settings and set "use beta viewer" to true. Closing and re-opening the modal, refreshing the page, and even logging in with another account should still show the setting set to true

### DIFF
--- a/frontend/src/v4/modules/viewer/betaViewer.helpers.ts
+++ b/frontend/src/v4/modules/viewer/betaViewer.helpers.ts
@@ -1,0 +1,20 @@
+/**
+ *  Copyright (C) 2023 3D Repo Ltd
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Affero General Public License as
+ *  published by the Free Software Foundation, either version 3 of the
+ *  License, or (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export const USE_BETA_VIEWER = 'useBetaViewer';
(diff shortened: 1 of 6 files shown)


frontend/src/v4/modules/viewer/betaViewer.helpers.ts
```diff
@@ -0,0 +1,20 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export const USE_BETA_VIEWER = 'useBetaViewer';
+export const setUseBetaViewer = (useBetaViewer = false) => localStorage.setItem(USE_BETA_VIEWER, JSON.stringify(useBetaViewer));
```
The-Daniel, line 19: I noticed that initially the experimental feature was turned on by default. I'm guessing this is to do with the browser even though I don't think I would have set it before. Is this okay?

carmenfan: by default it should be false, we want people to use the non beta viewer unless they've opted to do so

The-Daniel (approved):

PR #4459, opened 2023-09-20
Issue 4384 - Augment field definition in ticket groups

### LINKED ISSUE #4384: Augment field definition in ticket groups
### Description
We are adding support in smart rules to search for substring or a user specified regex. This requires changes on the Add/Edit group panel within the ticket groups, and the update the logic

**This is a sub task of #4386. Please branch/merge with that branch instead of staging.**

Figma design: https://www.figma.com/file/bRqIl0g9psqD3ISkmxwcNp/3D-Repo?type=design&node-id=715-17375&mode=design&t=QUnvuU28wOnbtypc-0

Backend/schema specification: https://github.com/3drepo/3drepo.io/issues/4383

#### Notes:
- By default, Field operator should set to `IS`
- There needs to be at least one value
- if operator is `REGEX`, there can only be one value thus `+` should be disabled.
- We should keep the suggested drop down and should be available regardless of operator (even regex, so the user can then trim it to their liking and add syntax to prevent typos)
- after https://github.com/3drepo/3drepo.io/issues/4383, the backend will always output the new schema, so you do not need to expect field to be a string.
- Title is a new field for the rule and is required. the name of the rule will then be displayed as a chip
- We also need to do this for V4 groups - if there is any common code for consideration.

### Goals
- [ ] The user should be able to create rules where they specify a subset/regex matching string for the field name
- [ ] The user should be able to edit those rules as they wish.

### Tasks
Developer: please fill this in as appropriate

### Related Resources
figma: https://www.figma.com/file/bRqIl0g9psqD3ISkmxwcNp/3D-Repo?type=design&node-id=715-17375&mode=design&t=QUnvuU28wOnbtypc-0
scoping document: https://github.com/3drepo/3D-Repo-Product-Team/issues/308
backend issue: https://github.com/3drepo/3drepo.io/issues/4383

Amantini1997:
This fixes #4384 

#### Description
Ticket groups use the new rule schema

#### Test cases
Open a ticket with groups and create/edit the rules.
The new interface should not allow to select the field, but that should be set with a rule (similar to the value)


The-Daniel:
There seems to be a few things missing that are listed in the 'notes' section of this issue. For example default operator being 'IS' and disabling the use of `+` (as well as `*` maybe?)

These also need to apply to the V4 PR

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketGroups/groups/groupRulesForm/groupRulesInputs/ruleFieldValues/fieldValueInput/fieldValueInput.component.tsx
The-Daniel, line 1: I'm not sure you want the autosuggest list when the type is 'regex' (maybe only for the 'is' type, but I don't know)

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketGroups/groups/groupRulesForm/groupRulesInputs/ruleOperator/ruleOperator.component.tsx
The-Daniel, line 1: image
Adding multiple invalid fields before switching to regex move makes the form invalid although all the visible input fields are valid
image

The-Daniel, line 1: image
Cancelling and reopening the form does not seem to reset the field values
image
Everything else is reset

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketGroups/groups/groupRulesForm/groupRulesInputs/ruleValues/ruleValues.component.tsx
The-Daniel, line 1: I still seem able to use the invalid regex values `+` (and maybe* too, maybe ask Carmen?)

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketGroups/groups/groupRulesForm/groupRulesForm.component.tsx
The-Daniel, line 1: image
There's an unnecessary scrollbar

The-Daniel (changes requested):

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketGroups/groups/groupRulesForm/groupRulesInputs/ruleFieldValues/fieldValueInput/fieldValueInput.component.tsx
Amantini1997: It is actually specified in the issue _We should keep the suggested drop down and should be available regardless of operator (even regex, so the user can then trim it to their liking and add syntax to prevent typos)_

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketGroups/groups/groupRulesForm/groupRulesInputs/ruleValues/ruleValues.component.tsx
Amantini1997: Also specified in the issue _if operator is REGEX, there can only be one value thus + should be disabled._

carmenfan: I'm a bit confused about the meaning behind this question. but to be clear:

the +/- button on the side should be disabled for regex, there should be 1 input given for regex

syntax within the regex `*`, `+` should be supported. i.e. I should be able to do something like `.*is.+`

The-Daniel: Ahh right. Yes, I misunderstood

The-Daniel (approved):

PR #4464, opened 2023-09-21
Issue 4421 - New revision events now include format and description

### LINKED ISSUE #4386: Adding rules to support a variation of field names
### Description
This is a parent issue for many sub tasks around supporting regexed field names within smart rules.

See scoping issue for details https://github.com/3drepo/3D-Repo-Product-Team/issues/308

### Goals
- As a user I want to be able to create smart group rules for a certain field regardless of the parent category the field belongs in
- As a user I want to be able to create a smart group rule where for multiple fields, and consider an object a match if it contains matching data for any of those fields
- As a user I want to be able to create smart group rules for multiple fields regardless of the parent category (i.e. the combination of the top 2 stories)
- As a power user that understands how to create a regex, I want to be able to just provide a regex to explain the field name I want to match so I can maximise the flexibility.
### Tasks
- [x] https://github.com/3drepo/3drepo.io/issues/4383
- [x] https://github.com/3drepo/3drepo.io/issues/4385
- [x] https://github.com/3drepo/3drepo.io/issues/4384

ChristosTsiotsias:
This fixes #4386

#### Description
containerNewRevision and federationNewRevision now include format and description properties in the event message data

#### Test cases
When a new revision event is fired the message data should include format and description

### DIFF
--- a/backend/src/v5/processors/teamspaces/projects/models/containers.js
+++ b/backend/src/v5/processors/teamspaces/projects/models/containers.js
@@ -80,17 +80,15 @@ Containers.getContainerStats = async (teamspace, container) => {
 	return stats;
 };
 
+Containers.getRevisionFormat = (rFile) => (rFile ? '.'.concat(rFile[0].split('_').pop()) : undefined);
+
 Containers.getRevisions = async (teamspace, container, showVoid) => {
 	const revisions = await getRevisions(teamspace,
 		container, showVoid, { _id: 1, author: 1, timestamp: 1, tag: 1, void: 1, desc: 1, rFile: 1 });
 
 	return revisions.map(({ rFile, ...r }) => {
-		if (rFile) {
-			const format = '.'.concat(rFile[0].split('_').pop());
-			return { ...r, format };
-		}
-
-		return r;
+		const format = Containers.getRevisionFormat(rFile);
+		return { ...r, ...(format ? { format } : {}) };
 	});
 };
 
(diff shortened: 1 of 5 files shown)


backend/src/v5/processors/teamspaces/projects/models/containers.js
```diff
@@ -80,17 +80,15 @@ Containers.getContainerStats = async (teamspace, container) => {
 	return stats;
 };
 
+Containers.getRevisionFormat = (rFile) => (rFile ? '.'.concat(rFile[0].split('_').pop()) : undefined);
```
carmenfan, line 83: This should live in `models/revisions.js`, we're manipulating a data that belongs to revision rather than general container

backend/src/v5/services/eventsListener/components/modelEvents.js
```diff
@@ -66,12 +67,20 @@ const revisionUpdated = async ({ teamspace, project, model, data, sender }) => {
...
-		await createModelMessage(event, { _id: revision, tag, author, timestamp: timestamp.getTime() },
-			teamspace, project, model);
+		const format = getRevisionFormat(rFile);
+
+		await createModelMessage(event, { _id: revision,
+			tag,
+			author,
+			timestamp: timestamp.getTime(),
+			description,
+			...(format ? { format } : undefined) }, teamspace, project, model);
```
carmenfan, line 83: use `deleteIfUndefined` from object helper instead
```suggestion
			...(format ? { format } : undefined) }, teamspace, project, model);
```

carmenfan (changes requested):

carmenfan (approved):

PR #4470, opened 2023-09-27
ISSUE #4451 - Session Ids are now UUID strings

### LINKED ISSUE #4451: Session ID is not unique enough
### Description
Very occasionally we see a session ID clash when we try to save the login record. We should ensure that does not happen

image

ChristosTsiotsias:
This fixes #4451

#### Description
Session ID generation functionality modified to generate session IDs with UUID format

#### Test cases
Session Id should now be of a UUID form


backend/src/v5/services/sessions.js
```diff
@@ -31,6 +32,7 @@ const initialiseSession = async () => {
 
 	// istanbul ignore next
 	const middleware = expressSession({
+		genid: () => generateUUIDString(),
```
carmenfan, line 35: Can we generate a uuid (not string), then represent it in base64? so it's shorter

carmenfan (changes requested):

carmenfan (approved):

PR #4471, opened 2023-09-27
ISSUE #4369 - add autosuggest strong password

### LINKED ISSUE #4369: Suggest password doesn't come up on the sign up page
### Description
It seems to think it is a existing password field instead of a new password field
image

### Steps to replicate
- Open sign up page
- right click on password drop down

### Current Behaviour
Suggest password is not an available option, instead the user can check saved passwords

### Expected Behaviour
It should prompt the user to suggest password, the suggested password should also populate the confirm password field

Amantini1997:
This fixes #4369

#### Description
New passwords input have a "strong password" suggestion from the browser

#### Test cases
In the signup form, click on the password field, you should find the autosuggestion option from the browser.
The same logic was applied in the edit profile modal to replace the current password with a new one

#### Notes
In Chrome, the autosuggestion option does not work out of the box. When the browser tries to autofill the password fields with the suggested passwords, the following popup is triggered.
image
If the user does not click on "Update password", the input does not get populated and it stays empty from the code perspective (you will notice that the value cannot be copied or edited).
This is quite an annoying caveat as the browser is basically forcing you to save that password in order to use it


The-Daniel (dismissed):

carmenfan:
The suggested password is displayed in plain text - is this normal? I don't think I've ever seen this happening elsewhere, it should be masked?

img-02-33-23-05-33
image
image

carmenfan (changes requested):
See comment above

sanmont3drepo:
> The suggested password is displayed in plain text - is this normal? I don't think I've ever seen this happening elsewhere, it should be masked?
> 
> img-02-33-23-05-33 [ img-02-33-23-05-33 ](https://user-images.githubusercontent.com/11945337/272030278-0071f177-7e53-408d-9d84-ec18b3fae541.gif) [ ](https://user-images.githubusercontent.com/11945337/272030278-0071f177-7e53-408d-9d84-ec18b3fae541.gif) image image


The confirmation field is the one that stays in plain all the time. Lets take that one out @Amantini1997

carmenfan:
@Amantini1997  as discussed with @sanmont3drepo, we'll follow the modern trend and remove the confirm password field, in favour of having an eye toggle to show masked password. 

To be super clear we would do the following:
- remove confirm password input box
- add eye toggle to password box to show a masked password

This need to be done in all places where we ask for a password:
- sign up
- forgot password landing page
- change password from profile modal
- sign in just adding the eye toggle for the login password

carmenfan (approved):

sanmont3drepo (approved):

The-Daniel (approved):

PR #4476, opened 2023-09-28
ISSUE #4376 - Container name visibility in tree

### LINKED ISSUE #4376: Container name visibility on tree
### Description
TBC

### Motivation
Currently we truncate with `...` the end of the container name if it becomes too long to be displayed, but the end of the container is often the part that contains the most meaningful information.

We should just display it the container name in multiple lines.

### Goals
- For long container names, we should see the full name, even if it takes multiple lines.

Amantini1997:
This fixes #4376 

#### Description
Containers names in Tree are cut at the start instead of the end

#### Test cases
Open a container in tree (viewer) with a name long enough;
Do the same with a federation with a container with a name long enough

### DIFF
--- a/frontend/src/v4/routes/viewerGui/components/tree/components/treeNode/treeNode.styles.ts
+++ b/frontend/src/v4/routes/viewerGui/components/tree/components/treeNode/treeNode.styles.ts
@@ -140,6 +140,7 @@ export const NameWrapper = styled.div`
 	display: flex;
 	overflow: hidden;
 	flex: 1;
+    align-items: center;
 `;
 
 export const ParentOfVisible = styled.span`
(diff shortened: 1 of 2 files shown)


The-Daniel (changes requested):
The issue says the container name should wrap onto multiple lines

frontend/src/v4/routes/viewerGui/components/tree/components/treeNode/treeNode.styles.ts
```diff
@@ -138,6 +159,17 @@ export const Container = styled.li<IContainer>`
 	box-sizing: border-box;
 	cursor: inherit;
 
+	${({ $truncateLeft }) => $truncateLeft && css`
```
The-Daniel, line 162: This doesn't seem to be truncating left
image
I take it the idea is that the root node is displayed in full and children are truncated at the start since the distinguishing information is usually at the end of the name?

The-Daniel (changes requested):

frontend/src/v4/routes/viewerGui/components/tree/components/treeNode/treeNode.styles.ts
Amantini1997: truncateLeft was a leftover name, I renamed it to be more meaningful, i.e. "isContainer".

The idea is to display the full name of the container(s), not the root.
In your case, I assume you are looking at a federation that has at least 2 containers "Cathedral Door" and "incredibly long ... sort of thing". If you open the respective containers in the viewer, those very same name should be truncated

The-Daniel (approved):

PR #4478, opened 2023-09-28
ISSUE #4477 - new Clip tool

### LINKED ISSUE #4477: Clip Tool Improvements
### Description
Rebuild the Clip Tool to the specification here: https://github.com/3drepo/3D-Repo-Product-Team/issues/250

The Unity changes are here: https://github.com/3drepo/3drepounity/issues/479, which include all the logic and the 3D UI. The Unity Issue will also include an updated toolbar for embedded viewers.

This issue should rebuild the toolbar in the frontend however, utilising the public API of the Clip Tool.

sebjf:
This fixes #4477

#### Description
Frontend changes required to support the new clip tool UX. 

This PR is being created early to test out changes in the viewer. Do not merge yet.

Because the frontend UX behaviour is changing, for now **to switch between Six Plane and Single Plane modes, you need to Delete the planes using the Clip Toolbar**, before selecting the mode anew using the Clip Tool Button.

The clip planes are in the old viewer - **turn off the 4GB mode** to see it.


carmenfan (changes requested):
#### Single plane

- Could we default the plane to cut on the up direction? (Z plane... depending on coordinates system) In the past I've been told this is the most likely use case for single plane, which makes sense as you're likely to be trying to cut to a certain floor.
- It's a but confusing to the user if I trigger the single plane whilst zoomed into the model as we don't see the gizmo - it looks broken as nothing is showing up. Could we make it that if we don't see the whole model, the sectioning cuts in to the model by default so it's shown in the camera view? chances are it would be the user intention anyway.
img-06-55-23-11-55

- Arrows should point outwards, currently it's always pointing inwards:
image
image
image
image

#### Box clip
- Rotation works great when you're looking at the "front" of the gizmo, but if you're controlling it from the backside, it's a bit non deterministic and  sometimes goes into the wrong direction

img-06-07-23-12-07

sebjf:
Hi @carmenfan, I have pushed the latest unity from 479 (as the default/2GB viewer). 

This should be functionally complete, barring any future feedback.

The inbuilt toolbar is still available, but hidden by default. `UnityUtil.enableClipToolbar()` can be used to show it for testing until the new one is in place (and from then on by embedded viewers, if we want to give them access to interactive clipping).

For whoever is going to do the frontend toolbar, the APIs for the clip tool begin here: https://github.com/3drepo/3drepo.io/blob/0dd25f45f8ab8a994238312f88d50a5b26032074/frontend/src/globals/unity-util.ts#L841

carmenfan (changes requested):
From design review meeting today:

- rotate/move should be more like toggle buttons as they are representing the states (i.e. if someone clicks on it it should highlight until the user clicks on the other one. I guess this is more for the front end but it'd be good if we make it happen on the hidden toolbar too
- We need a reset button for resetting the clipping plane back to the original position - original being the default when we don't have @apeitup  if you can point @sebjf to an icon to use that'll be great 😄 
- There's arguments over the aesthetics of the gizmo, @apeitup  will work it out with @jozefdobos  and get back to us on a finalised version to implement

frontend/src/globals/unity-util.ts
```diff
@@ -861,6 +874,105 @@ export class UnityUtil {
 		UnityUtil.toUnity('StopClipEdit', undefined, undefined);
 	}
 
+	/**
```
carmenfan, line 877: Can we add the `@category` tag in these function comments? We run this file through typedoc (https://typedoc.org/guides/overview/) so these comments needs to conform to the schema so it shows up nicely [in our documentation](https://3drepo.github.io/3drepo.io/frontend/classes/unity_util.UnityUtil.html)

carmenfan (changes requested):

frontend/src/globals/unity-util.ts
sebjf: Hi @carmenfan, done! I went through the file and added comments to methods from other features that didn't have `@category` and were not `@hidden` too.

carmenfan (approved):
Looks good thanks!

Will close this PR in favour of #4623

PR #4485, opened 2023-10-03
ISSUE #4252 - Logs of import error email are now attached as a log file

### LINKED ISSUE #4252: Asite email has a size limit that we are occasionally hitting
### Description
We're occasionally seeing the following email being triggered:
image

This seems to happen when we upload a model that times out during unity. Google says that this error message is due to the message size is over the limit imposed by the recipient email exchange. Since the timeout email content is relatively small, it is most likely due to us trying to attach log files that are too long.

Could we potentially zip the log files into a zip file?

It may be better to stop attaching the log file, and implement an admin that only system admins have access to to view it

### Steps to replicate
upload one of these models:
https://asitesol.sharepoint.com/:f:/s/3DRepoTeamFolder/EnsGrEPx6INJhGrtkaR8j-sBmc8JVWfcU0g3bvFGt0u36A?e=Wu0m4r

### Current Behaviour
- Eventually it times out and io exist with an unhandled promise rejection like the picture above

### Expected Behaviour
- Should be able to relay the information without crashing the app

ChristosTsiotsias:
This fixes #4252

#### Description
The email that is being sent to the user when a model import fails now includes the logs in a zip file instead of a list of attachments.

#### Test cases
Import a model and make the import fail due to timeout.
An email should be sent that includes the logs as a zip file and not as a list of attachments.


carmenfan (changes requested):

backend/src/v4/models/chatEvent.js
```diff
@@ -217,34 +223,41 @@ const subscribeToV5Events = () => {
...
+					files.forEach((file) => {
+						if (file.endsWith(".log")) {
+							archive.file(path.join(sharedDir, file), { name: file });
+						}
+					});
+
+					const zipPath = path.join(sharedDir, "logs.zip");
+					const output = fs.createWriteStream(zipPath);
+
+					output.on("close", () => {
```
carmenfan, line 244: As it stands it's a callback function so the this function can easily exits before the callback is called. If sendImportError errors out, ti's entering no mans land and we are probably going to crash the server again with unresolved promise rejection.

So I think we need to create a promise that is only resolved in the callback function and wait on it would be a better managed flow.

backend/src/v4/models/chatEvent.js
```diff
@@ -217,34 +223,41 @@ const subscribeToV5Events = () => {
...
-					model,
-					username: user,
-					err: message,
-					corID: corId,
-					bouncerErr: errCode,
-					attachments
-				});
+					});
+
+					archive.pipe(output);
```
carmenfan, line 256: I would create the output stream and pipe  to output before L235, so it can get the stream as soon as it is available.

backend/src/v4/models/chatEvent.js
```diff
@@ -217,34 +223,41 @@ const subscribeToV5Events = () => {
...
+
+					output.on("close", () => {
+						Mailer.sendImportError({
+							account: teamspace,
+							model,
+							username: user,
+							err: message,
+							corID: corId,
+							bouncerErr: errCode,
+							attachments: [{ filename: "logs.zip", path: zipPath }]
```
carmenfan, line 252: since you use `logs.zip` in multiple places you may as well use a variable for it to make sure it's consistent.

backend/src/v4/models/chatEvent.js
carmenfan: It would also help to tap on the errors and reject the promise (same with the archiver stuff)

carmenfan (changes requested):

backend/src/v4/models/chatEvent.js
```diff
@@ -16,11 +16,17 @@
  */
 
 "use strict";
+const archiver = require("archiver");
+const fs = require("fs");
 const { v5Path } = require("../../interop");
 const EventsManager = require(`${v5Path}/services/eventsManager/eventsManager`);
 const EventsV5 = require(`${v5Path}/services/eventsManager/eventsManager.constants`).events;
+const Mailer = require("../mailer/mailer");
```
carmenfan, line 24: would also be safer if you just import the function instead of the whole mailer

(It was probably done inline with the code due to circular dependencies...)

carmenfan (approved):

Issue #4260, opened 2023-06-07
Change the way we read mesh data


carmenfan:
### Description
This is a pairing issue for https://github.com/3drepo/3drepobouncer/issues/626

Since we are changing the way we store the data, we need to make sure we know how to read it on 3drepo.io

### Goals
- [x] Should be able to read mesh data from the older schema
- [x] Should be able to read mesh data from the new schema

PR #4503, opened 2023-10-06
ISSUE #4393 remove _extRef support

### LINKED ISSUE #4393: Remove compatibility code for old mesh binary schema
### Description
#4260 from v5.4 had a migration to the new schema. We're keeping backwards compatibility for one release to allow smooth transition

So for v5.6, we should remove the method to read the old data as it's no longer necessary.

This applies to bouncer as well.

carmenfan:
This fixes #4393

#### Description
Remove the deprecated schema to associate binary to a scene, `_extRef`. All `_extRef` references should have been migrated to `_blobRef` as per 5.4 migration script (instigated by #4260)

- Removed Texture end point - it is referencing gridFS directly, so it would've stopped working as of 4.28. It is an endpoint that isn't used
- Removed the ability to read mesh data from legacy schema
- Removed probing `_extRef` to figure out scene binaries and removing them when we remove incomplete revisions
  - Fixed a bug with this script - we were removing the reference of the new files but were not deleting the actual file!

#### Test cases
Should function as before
pre 5.4 schema will no longer work


ChristosTsiotsias:
travis build is failing

carmenfan:
> travis build is failing

@ChristosTsiotsias  fixed now

ChristosTsiotsias (changes requested):
backend tests seem to be failing

carmenfan:
> backend tests seem to be failing

just needed a rerun it seems, passed now

ChristosTsiotsias (approved):

PR #4514, opened 2023-10-11
ISSUE #4400 - Fix template chip ticket count not updating

### LINKED ISSUE #4400: Custom ticket filters not updated upon Completed filter click
### Description

Filter's label (count) should update when Completed filter is clicked to show the number of tickets that have the filter template but also satisly the Completed filter. 

However, this does not seem to happen as count stays the same regardless of the Complete filter active status.

### Steps to replicate
- Open a model with multiple tickets of multiple templates
- Click Completed filter
- The count in the label of filters does not update

image
image

The-Daniel:
This fixes #4400

#### Description
The ticket count in the template chips now updates to reflect the number of tickets in the respective template that matches the search filters and completed state.
For a better explanation of how the numbers should work look at the comments in the issue

#### Test cases
Check that the numbers are right when selecting different chips and using different/multiple search terms

### DIFF
--- a/frontend/src/v5/ui/routes/viewer/tickets/ticketsList/ticketsList.component.tsx
+++ b/frontend/src/v5/ui/routes/viewer/tickets/ticketsList/ticketsList.component.tsx
@@ -44,7 +44,7 @@ export const TicketsList = ({ tickets }: TicketsListProps) => {
 	const [availableTemplates, setAvailableTemplates] = useState([]);
 	const [showingCompleted, setShowingCompleted] = useState(false);
 	const [selectedTemplates, setSelectedTemplates] = useState<Set<string>>(new Set());
-	const [filteredTickets, setFilteredTickets] = useState<ITicket[]>([]);
+	const [ticketsFilteredByComplete, setTicketsFilteredByComplete] = useState<ITicket[]>([]);
 	const templates = TicketsHooksSelectors.selectTemplates(containerOrFederation);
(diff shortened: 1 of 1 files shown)


frontend/src/v5/ui/routes/viewer/tickets/ticketsList/ticketsList.component.tsx
```diff
@@ -69,22 +69,20 @@ export const TicketsList = ({ tickets }: TicketsListProps) => {
...
 	};
 
-	useEffect(() => {
-		const filtered = tickets.filter((ticket) => {
-			const matchesTemplateFilters = !selectedTemplates.size || selectedTemplates.has(ticket.type);
-			return (matchesTemplateFilters && matchesCompletedState(ticket));
-		});
-		setFilteredTickets(filtered);
-	}, [tickets, selectedTemplates, showingCompleted]);
+	const applyTemplatesFilter = (items) => items.filter((ticket) => !selectedTemplates.size || selectedTemplates.has(ticket.type));
```
Amantini1997, line 72: Can we rename this to something like "filterByTemplate"?

Also, you can check if there are no selected templates just once for all the tickets:

```ts
	const applyTemplatesFilter = (items) => {
		if (!selectedTemplates.size) return items;
		return items.filter((ticket) => selectedTemplates.has(ticket.type));
	};
```

Amantini1997 (changes requested):

Amantini1997 (approved):

carmenfan (approved):

PR #4518, opened 2023-10-13
ISSUE #4517 - fix: reset visual settings crashes the app

### LINKED ISSUE #4517: White screen if the user  resets viewer background colour
### Description
If the user set a background colour of the viewer settings then tries to reset it, this is causing a white screen error.

The error persists even if you refresh, and can only be resolved by deleting local storage

### Steps to replicate
- set a background colour in visual settings
- hit save
- open visual settings again, hit reset, hit save.

### Current Behaviour
front end white screens with error

### Expected Behaviour
shoudl reset the background if the user is on the viewer and it should just continue as normal.

img-13-52-23-02-52

Amantini1997:
This fixes #4517

#### Description
"undefined" was passed as a string and saved in the local storage.
Reading and parsing such a value was the problem since the value was a string

#### Test cases
Change the viewer background color, save, and reset the settings

### DIFF
--- a/frontend/src/v4/modules/viewer/betaViewer.helpers.ts
+++ b/frontend/src/v4/modules/viewer/betaViewer.helpers.ts
@@ -16,5 +16,5 @@
  */
 
 export const USE_BETA_VIEWER = 'useBetaViewer';
-export const setUseBetaViewer = (useBetaViewer) => localStorage.setItem(USE_BETA_VIEWER, JSON.stringify(useBetaViewer));
-export const getUseBetaViewer = () => JSON.parse(localStorage.getItem(USE_BETA_VIEWER) || 'true');
+export const setUseBetaViewer = (useBetaViewer = true) => localStorage.setItem(USE_BETA_VIEWER, JSON.stringify(useBetaViewer));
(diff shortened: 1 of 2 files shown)


frontend/src/v4/modules/viewer/betaViewer.helpers.ts
```diff
@@ -16,5 +16,5 @@
  */
 
 export const USE_BETA_VIEWER = 'useBetaViewer';
-export const setUseBetaViewer = (useBetaViewer) => localStorage.setItem(USE_BETA_VIEWER, JSON.stringify(useBetaViewer));
-export const getUseBetaViewer = () => JSON.parse(localStorage.getItem(USE_BETA_VIEWER) || 'true');
+export const setUseBetaViewer = (useBetaViewer = true) => localStorage.setItem(USE_BETA_VIEWER, JSON.stringify(useBetaViewer));
```
The-Daniel, line 19: The 4GB viewer state acts weird with the reset
image
Pressing reset turns the toggle off (even though on is the default)
image
But then upon pressing save and going back into the settings it is on again
image

The-Daniel (changes requested):

The-Daniel (approved):

carmenfan (approved):

PR #4531, opened 2023-10-18
ISSUE #4519 - Dates appear as timestamps on ticket update event data

### LINKED ISSUE #4519: Custom tickets dates are converted to string
### Description
Given a ticket with dateTime inputs (sequencing will use them, so at the moment, there is none available yet), when the user changes the value, the frontend sends the new value as a number, but then the chat event sends back a string 
image

For existing dates, it's not a big deal perhaps, but when the frontend tries to clear the input sending either the number 0 or null, the backend sends back a stringified date object set to 01/01/1970 
image
- This is happening in: both

### Steps to replicate
- branch off from #3609 (you have to create a ticket with sequencing module)

### Current Behaviour
- The date is sent as a string

### Expected Behaviour
- The date should be sent as a number

ChristosTsiotsias:
This fixes #4519

#### Description
Bug fix: Dates are now serialised into timestamps on UPDATE_TICKET event data

#### Test cases
Update a ticket and change the start or end date on sequencing module
Observe the data in UPDATE_TICKET event
The dates should appear as timestamps or 0 if its set to null


carmenfan (changes requested):
Can we add a test for it?

carmenfan (approved):

Issue #4385, opened 2023-08-04
Augment field definition in v4 Groups


carmenfan:
### Description
We are adding support in smart rules to search for substring or a user specified regex. This requires changes on the Add/Edit group panel, and the update the logic

**This is a sub task of #4386. Please branch/merge with that branch instead of staging.**

This needs to function the same way as v5 ticket groups, so see https://github.com/3drepo/3drepo.io/issues/4384 for details.

### Goals
- [ ] The user should be able to create rules where they specify a subset/regex matching string for the field name
- [ ] The user should be able to edit those rules as they wish.

### Tasks
Developer: please fill this in as appropriate

### Related Resources
figma: https://www.figma.com/file/bRqIl0g9psqD3ISkmxwcNp/3D-Repo?type=design&node-id=715-17375&mode=design&t=QUnvuU28wOnbtypc-0
scoping document: https://github.com/3drepo/3D-Repo-Product-Team/issues/308
backend issue: https://github.com/3drepo/3drepo.io/issues/4383
front end v5 ticket groups issue: https://github.com/3drepo/3drepo.io/issues/4384

PR #4539, opened 2023-10-20
Issue 4386 - Addition of field name rules

### LINKED ISSUE #4386: Adding rules to support a variation of field names

ChristosTsiotsias:
This fixes #4386

#### Description
Parent PR for addition of field name rules.

Main issue: https://github.com/3drepo/3drepo.io/issues/4386
Backend: https://github.com/3drepo/3drepo.io/issues/4383
Frontend: v4 groups https://github.com/3drepo/3drepo.io/issues/4385
v5 ticket groups https://github.com/3drepo/3drepo.io/issues/4384


carmenfan (changes requested):
#### You can't see the full name of the rule if it's too long
We're truncating the end with `...`, which is fine, but I think we need to display a tooltip on hover to show the full name.
image
image


#### copy and paste filter not working
The meta card is still generating the old schema, which the filters part no under understand (interesting enough if I save the group, the backend translates it correctly and will give you the right information again). It should give it a default name and update to new schema.
img-20-05-23-05-05
img-20-12-23-05-12

carmenfan (changes requested):
@Amantini1997  Haven't started the second round yet, but a quick one I found whilst testing backend:

#### Should auto-focus on the first field of the form (name field) instead of the field value field:
img-23-10-23-12-10

#### V5 rule panel doesn't update when rules/add filter is clicked
If the panel is already opened (whether by means of `add filter` or clicking on an existing filter), the data within the panel does not update if you click on another filter or `add filter`
img-23-37-23-01-37

carmenfan (approved):

PR #4597, opened 2023-11-14
Issue 4495 - remove warnings when calling "yarn install"

### LINKED ISSUE #4495: Fix incorrect peer dependencies
### Description
ATM we have some warnings when installing the packages. These warnings regards to different  peer dependencies requirements. ATM they doesn't seem to cause any problems but its better to ensure that every package is compatible

### Goal
- [ ]  ensure compatibility between packages

### Tasks
- [ ] Update the packages that can be updated in order to remove the warnings

Amantini1997:
This fixes #4495

#### Description
Most<sup> 1</sup> of the warnings have been removed. This was done updating/adding/removing some dependencies and rewriting some code (e.g. WindowEventListener).

A new file (package.json.md) has been added. The reason for this is that a json file does not support comments and this makes it hard to keep track of why we have some dependencies/resolutions. So the new .md file aims at providing explanations to them.

**<sup>1</sup> Why most and not all of them?**
Calling `yarn install` on this branch should output the following:
```
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
warning Pattern ["@mui/styled-engine-sc@5.12.0"] is trying to unpack in the same destination "C:\\Users\\artig\\AppData\\Local\\Yarn\\Cache\\v6\\npm-@mui-styled-engine-5.12.0-41040fb27667efab46e9200cec9657e62b544c48-integrity\\node_modules\\@mui\\styled-engine" as pattern ["@mui/styled-engine@npm:@mui/styled-engine-sc@5.12.0","@mui/styled-engine@^5.14.10","@mui/styled-engine@^5.4.1"]. This could result in non-deterministic behavior, skipping.
[4/5] Linking dependencies...
warning "react-trello > react-redux@5.1.2" has incorrect peer dependency "react@^0.14.0 || ^15.0.0-0 || ^16.0.0-0".
warning " > react-truncate@2.4.0" has incorrect peer dependency "react@<= 16.x.x".
[5/5] Building fresh packages...
```

About these warnings:
- `[3/5] Fetching packages...` cannot be fixed. The problem is a clash between the resolution and MUI styled-engine. However, the solution adopted is the same suggested in their own (MUI) website.
- `[4/5] Linking dependencies...`
   - **react-trello**: it looks like the author has not updated the dependencies in years. The version we use is already the latest one and changing it for an alternative library might be too much for this issue.
   - **react-truncate**: it seems that the app is no longer maintained. Other alternatives may provide code which is even older and would trigger more warnings, or they are at an early stage and not fully working. Possibly, a custom component might be created for this, but again, it seems too much for this issue

#### Test cases
Delete all the `node_modules` and call `yarn install`. Do **NOT** delete the lock file.
> **Tip**: select the node_modules folder and click "SHIFT + Delete" to delete all the modules without bining them. This will delete them faster

Check that the warnings are the same as described at<sup> 1</sup>.
Check that everything still works as intended.
Check that storybook runs.


The-Daniel (approved):

frontend/src/v4/routes/components/messagesList/components/message/components/markdownMessage/markdownMessage.component.tsx
```diff
@@ -16,20 +16,15 @@
...
 };
-
 export const MarkdownMessage = ({ className, children }: MarkdownMessageProps) => (
-	<ReactMarkdown
-		source={children}
-		renderers={renderers}
-		plugins={[breaks]}
-		className={className}
-	/>
+	<ReactMarkdown components={components} className={className}>
```
sanmont3drepo, line 27: Quote and reply (marked below) don't seem to be working.

This example is in production

image

sanmont3drepo (changes requested):

sanmont3drepo (approved):

PR #4599, opened 2023-11-16
ISSUE #4577 - Fixed gendoc compile

### LINKED ISSUE #4577: gen docs is broken on the front end
### Description
`yarn docs` doesn't work anymore. maybe because of #2233 and we need to upgrade typedoc?
image

We should add this on the automated testing so we know if it ever breaks again

### Steps to replicate
- run `yarn docs` on the front end folder

### Current Behaviour
- Errors out with an unrecognised function

### Expected Behaviour
- Should generate unity docs

sanmont3drepo:
This fixes #4577

#### Description
- Updated version of typedoc to 0.25.3
- Update config of typedoc

#### Test cases
- run `yarn docs` , it should generate the documentation for unityutil in `/frontend/docs`


sanmont3drepo:
@carmenfan  because I've updated typedoc some old options (most) weren't available. The format of the generated docs is different also. I dont know if this is what we want.

carmenfan (changes requested):
there's a conflict on yarn.lock

PR #4602, opened 2023-11-17
Issue 4579 - Custom Ticket Endpoint updates

### LINKED ISSUE #4579: Custom ticket endpoint updates
### Description
There's 2 enhancements we want to do with custom ticket endpoints:

#### Add a query string on get ticket templates endpoint to return full templates
Endpoint: https://staging.dev.3drepo.io/docs/#/Tickets/getModelTicketTemplates
Changes: Add an optional query string `getDetails` (defaults to false)
Description: When `getDetails` is set to true, the entire template is returned (same level of details as [get template details](https://staging.dev.3drepo.io/docs/#/Tickets/getModelTicketTemplateDetails))

#### Add query string to define the values to return on get ticket list
Endpoint: https://staging.dev.3drepo.io/docs/#/Tickets/GetTicketList
Changes: Add query string `filter` to allow the user to impose what values to receive
Description: By default, it should return what it currently returns, But a filter string should specify any additional field that should be included. e.g. `filter=Description,safetibase.element` should also include Description property and the element property within the safetibase module within the response.

### Goals
- [ ] As a front end developer I want to specify the properties that should be included in the list query so I can perform functions such as smart filter querying and custom property grouping without getting the full details of all the tickets. 
- [ ] As a front end developer I want to see the full details of the template so in situations where I need to know the what custom properties are available, I don't have to do individual calls to get them.

### Tasks
TBC

ChristosTsiotsias:
This fixes #4579

#### Description
Update 1:

getDetails query param added in get model templates endpoint (default false). If set to true, the templates returned include all their modules and properties
endpoint: https://staging.dev.3drepo.io/docs/#/Tickets/getModelTicketTemplates

Update 2:

filter query param added in get tickets. Filter query is a comma sperarated string which allows the user to set which modules and properties are included in the fetched tickets. To include a property pass just the name of the property and to include a module property pass <module name>.<property name>
endpoint: https://staging.dev.3drepo.io/docs/#/Tickets/GetTicketList

#### Test cases
Update 1
Call get templates endpoint using getDetails = true query
The templates returned should include all properties and modules

Update 2
Call get tickets endpoint using a filter which includes both a property and module property
The returned tickets should include these properties among the default ones which are already fetched


carmenfan (changes requested):

carmenfan (changes requested):

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/commons/tickets.js
```diff
@@ -66,6 +66,26 @@ const templateIDToParams = async (req, res, next) => {
 	}
 };
 
+TicketsMiddleware.convertFilterToProjection = async (req, res, next) => {
```
carmenfan, line 69: as discussed, move this back to processor, and call split on the filter before calling the processor function

carmenfan (changes requested):

carmenfan (approved):

Issue #4563, opened 2023-10-31
Only filtered ticket pins should be visible 


The-Daniel:
### Description

This was initially found in 5.6 QA https://github.com/3drepo/3drepo.io/pull/4557#issuecomment-1783156043
The ticket pins should reflect what is shown in the ticket list. If tickets are filtered out of that list their pins should not be visible. If the ticket card is closed the pins should remain as if the card had been left open.
If a user goes into ticket details only the ticket pins for that ticket should be visible

### Goals

- [ ] The visible pins reflect the filtering on tickets
- [ ] Default pins are added/removed when tickets list filtering changes
- [ ] When the user first opens a model all the uncompleted ticket pins should be visible (As in it should match the ticket list when opened)
- [ ] The visible ticket pins should remain constant when the ticket card is closed. Only upon opening it again and the list is reset should the pins also reset

### Tasks

- [ ] Move ticket pin logic out of v4 (`viewerCanvas.component.tsx`)
- [ ] Have ticket pin logic depend on items in the ticket list

### Related Resources
I had an attempt by adding a useEffect to TicketItem
![image](https://github.com/3drepo/3drepo.io/assets/62481250/c2534f86-0871-43e7-a59c-6397cc7cdb6d)
But could not find a way of not removing the pin when closing the card (the ticketsCardOpen technique doesn't work )

PR #4630, opened 2023-12-04
Issue 4585 - Select the next ticket/issue/risk if the user removes the current ticket/issue/risk from the filtered list

### LINKED ISSUE #4585: Select the next ticket/issue/risk if the user closed the ticket/issue/risk.
### Description
At the moment, if you have a long list of risks/issue/ticket, and you have changed a property that makes it so your currently selected ticket is filtered out, you lose your place in the list when you go back to the details, for e.g:
img-10-28-23-05-28

It would be ideal if the next item (or previous if it's easier) is selected in this situation so the user doesn't lose track of where they got to.

### Goals
- [ ] As a user who is going through a list of tickets/issues/risks, I want to be able to close an item out and go back to the item listing without losing my place

### Tasks
- [ ] Modify the ticket list to remember the position
- [ ] When going back keep the position

Amantini1997:
This fixes #4585

#### Description
Select the next ticket/issue/risk if the user removes the current ticket/issue/risk from the filtered list

#### Known bug
Tickets will not work correctly. For instance, if you have 4 tickets (A, B, C, D) and:
- expand C and set it to close/void;
- using the arrow, go from C to B;
- set B to close/void;
- go back to the list.

You would expect D to be selected, but in fact, the next item in the list is "C", so C is selected. This is because, expanding a ticket, the filters are lost. (This PR takes for granted that those filters persist upon expanding a ticket)


frontend/src/v5/ui/routes/viewer/tickets/ticketDetails/ticketsDetailsCard.component.tsx
```diff
@@ -62,6 +58,14 @@ export const TicketDetailsCard = () => {
 	const goPrev = () => changeTicketIndex(-1);
 	const goNext = () => changeTicketIndex(1);
 
+	const goBack = () => {
+		// TODO - replace `getTicketIsCompleted` with <not part of the filtered tickets list>
+		if (getTicketIsCompleted(ticket)) {
```
The-Daniel, line 63: Oh boy this is going to cause conflicts with my pin filtering PR lol. I'll need to make sure my PR does this with the new filtering logic.

One problem with this line in particular is that if you are viewing a completed ticket and exit to the tickets list it will increase the selected index by 1 which means a seemingly random ticket is selected when no values were even changed.

I think this should be easy to fix in my PR (I think this is what you mean with your TODO) because the filtering persists whatever state the tickets card is in. This allows me to use only the filtered tickets when cycling between tickets. So you can just use whether the ticket _id still exists in the filtered tickets to determine whether the index should change.

I'm happy to ignore the problem with getTicketIsCompleted since it will get changed in my PR anyway

frontend/src/v4/routes/viewerGui/components/issues/components/issueDetails/issueDetails.component.tsx
```diff
@@ -229,12 +232,23 @@ export class IssueDetails extends PureComponent<IProps, IState> {
...
-			this.props.setActiveIssue({}, null, true);
+		if (filteredIssues.find(({ _id }) => _id === issue._id)) {
+			// item is part of the filtered items list, no action required
+			return;
 		}
+
+		if (filteredIssues.length) {
+			const { listIndex } = this.state;
+			// set next item in the list as active
+			setActiveIssue(filteredIssues[listIndex % filteredIssues.length], revision, false);
```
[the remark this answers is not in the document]
Amantini1997: I checked with Carmen and she said this behaviour is fine for now

The-Daniel (approved):

frontend/src/v4/routes/viewerGui/components/issues/components/issueDetails/issueDetails.component.tsx
```diff
@@ -34,6 +34,7 @@ import { IssueDetailsForm } from './issueDetailsForm.component';
 interface IProps {
 	viewer: any;
 	jobs: any[];
+	filteredIssues: any[];
```
sanmont3drepo, line 37: Let's wait for #4563 to be merged into staging because in that PR we get the filters to persist throughout the session.

sanmont3drepo (changes requested):

The-Daniel (approved):

frontend/src/v4/routes/viewerGui/components/issues/components/issueDetails/issueDetails.component.tsx
```diff
@@ -34,6 +34,7 @@ import { IssueDetailsForm } from './issueDetailsForm.component';
 interface IProps {
```
sanmont3drepo, line 34: - If I close the last risk (didnt try it with tickets or issues) it loses the position and goes straight to the top
lost position

sanmont3drepo (changes requested):

frontend/src/v4/routes/viewerGui/components/issues/components/issueDetails/issueDetails.component.tsx
sanmont3drepo: So in tickets if I close the first ticket it goes to the last ticket on the list? That's not good either. @carmenfan  what do you think?

carmenfan: @ChristosTsiotsias  am I right? whatever we do it needs to be consistent between the web and the plugin

sanmont3drepo: Ok, so then ticket should work as issues

ChristosTsiotsias: In the plugins if the last issue/risk/custom ticket is filtered out the scroll bar goes to the top but no ticket is selected

sanmont3drepo (approved):

PR #4631, opened 2023-12-04
Issue 4625 - Virtualized list for containers

### LINKED ISSUE #4625: projects with a lot of containers essentially get stuck on loading the container list
### Description
We need to address the performance issue when there is a lot of containers. As it stands the UI is not functional.

3 things we need to address:
1.  Instead of having a placeholder until the stats are loaded. we should display the readily available information (i.e. the container name) and make it interactable with the stats value pending. i.e. show the list like this with the red lined items replaced with some sort of pending/loading icon (consult @apeitup for UI if necessary) 
image
2. instead of creating a component for all the containers in one go, we should limit it to ~20 containers, and only produce the next batch of 20 if the user scrolls. Note we may have to reset if the user does a text filter
3. We should avoid fetching container stats all in one go - do them in batches of ~20 and only dispatch the next 20 when the previous ones have returned.

### Steps to replicate
- go to https://staging.dev.3drepo.io/v5/dashboard/DevTeamWorkspace/499d9c10-7a81-42e7-8d06-9aca39e638f0/t/containers

### Current Behaviour
- It will never load

### Expected Behaviour
- It should provide known data as soon as possible

sanmont3drepo:
This fixes #4625

#### Description
- Can support more than 1000 containers

### DIFF
--- a/frontend/src/v5/store/containers/containers.sagas.ts
+++ b/frontend/src/v5/store/containers/containers.sagas.ts
@@ -99,7 +99,7 @@ export function* fetchContainers({ teamspace, projectId }: FetchContainersAction
 		const storedContainers = yield select(selectContainers);
 		const isPending = yield select(selectIsListPending);
 
-		// Only update if theres is new datad
+		// Only update if theres is new data
 		if (isPending || !isEqualWith(storedContainers, containersWithoutStats, compByColum(['_id', 'name', 'role', 'isFavourite']))) {
 			yield put(ContainersActions.fetchContainersSuccess(projectId, containersWithoutStats));
 		}
--- a/frontend/src/v5/store/containers/containers.sagas.ts
+++ b/frontend/src/v5/store/containers/containers.sagas.ts
@@ -73,7 +73,7 @@ export function* removeFavourites({ containerId, teamspace, projectId }: RemoveF
 export function* fetchContainerStats({ teamspace, projectId, containerId }: FetchContainerStatsAction) {
 	try {
 		const container: IContainer = yield select(selectContainerById, containerId);
-		const stats = yield statsQueue.enqueue(teamspace, projectId, containerId, container.name);
+		const stats = yield statsQueue.enqueue(teamspace, projectId, containerId);
 		
 		const basicDataEqual = compByColum(['unit', 'type'])(container, stats);
 		// eslint-disable-next-line max-len
--- a/frontend/src/v5/store/containers/containers.sagas.ts
+++ b/frontend/src/v5/store/containers/containers.sagas.ts
@@ -195,6 +195,10 @@ export function* deleteContainer({ teamspace, projectId, containerId, onSuccess,
 	}
 }
 
+export function* resetContainerStatsQueue() {
+	statsQueue.resetQueue();
+}
+
 export default function* ContainersSaga() {
 	yield takeLatest(ContainersTypes.ADD_FAVOURITE, addFavourites);
 	yield takeLatest(ContainersTypes.REMOVE_FAVOURITE, removeFavourites);
@@ -205,4 +209,5 @@ export default function* ContainersSaga() {
 	yield takeLatest(ContainersTypes.UPDATE_CONTAINER_SETTINGS, updateContainerSettings);
 	yield takeEvery(ContainersTypes.CREATE_CONTAINER, createContainer);
 	yield takeLatest(ContainersTypes.DELETE_CONTAINER, deleteContainer);
+	yield takeEvery(ContainersTypes.RESET_CONTAINER_STATS_QUEUE, resetContainerStatsQueue);
 }
--- a/frontend/src/v5/store/containers/containers.sagas.ts
+++ b/frontend/src/v5/store/containers/containers.sagas.ts
@@ -40,66 +40,9 @@ import { selectContainerById, selectContainers, selectIsListPending } from './co
 import { compByColum } from '../store.helpers';
 import { getSortingFunction } from '@components/dashboard/dashboardList/useOrderedList/useOrderedList.helpers';
 import { SortingDirection } from '@components/dashboard/dashboardList/dashboardList.types';
+import { LifoQueue } from '@/v5/helpers/functions.helpers';
 
-type QueueItem<T> = { promise: Promise<T>, resolve, args };
-class Queue<T> {
-	private queue: QueueItem<T>[] = [] ;
-
-	private dict = {};
-
-	private batchSize;
-
-	private running = false;
-
-	private func: (...args) => Promise<T>;
-
-	private getPromise(args):QueueItem<T>  {
-		let prom = this.dict[JSON.stringify(args)];
-		if (!prom) {
-			prom = {};
-			prom.promise = new Promise((resolve) => prom.resolve = resolve);
-			prom.args = args;
-		}
-		
-		return prom;
-	}
-
-	private async runqueue() {
-		this.running = true;
-		while (this.queue.length) {
-			const batch =  this.queue.splice(Math.max(this.queue.length - this.batchSize, 0));
-			await Promise.all(batch.reverse().map(async (p) => {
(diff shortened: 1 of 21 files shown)


frontend/src/v5/ui/components/dashboard/dashboardList/dashboardListCollapse/dashboardListCollapse.component.tsx
```diff
@@ -44,7 +44,7 @@ export const DashboardListCollapse = ({
 	const [isExpanded, setIsExpanded] = useState(defaultExpanded);
 
 	return (
-		<Container className={className} isLoading={isLoading}>
+		<div className={className} >
```
The-Daniel, line 47: Getting rid of the `<Container />` here messes up the padding in the tickets list. (Notice the column headers for instance)
image

You can see in `ticketsTableContent.styles` that <Container /> is styled

frontend/src/v5/ui/components/dashboard/dashboardList/dashboardListItem/components/dashboardListItemTitle/dashboardListItemContainerTitle/dashboardListItemContainerTitle.component.tsx
```diff
@@ -65,13 +65,13 @@ export const DashboardListItemContainerTitle = ({
...
 			tooltipTitle={
-				hasRevisions ? (
+				container.hasStatsPending || hasRevisions ? (
 					<FormattedMessage id="containers.list.item.title.tooltip" defaultMessage="Launch latest revision" />
 				) : (
 					<FormattedMessage id="containers.list.item.title.tooltip.empty" defaultMessage="No revisions" />
 				)
 			}
-			disabled={!hasRevisions}
+			disabled={false}
```
The-Daniel, line 74: I don't this should be false. Previously this was disabled when a container was empty.
This Branch: image

Staging:  image

frontend/src/v5/ui/routes/dashboard/projects/containers/containersList/containerListItem/containerEllipsisMenu/containerEllipsisMenu.component.tsx
```diff
@@ -121,12 +121,14 @@ export const ContainerEllipsisMenu = ({
 					container._id,
 				) }}
 			/>
-			<EllipsisMenuItem
-				title={formatMessage(selected
-					? { id: 'containers.ellipsisMenu.hideRevisions', defaultMessage: 'Hide Revisions' }
-					: { id: 'containers.ellipsisMenu.viewRevisions', defaultMessage: 'View Revisions' })}
-				onClick={() => onSelectOrToggleItem(container._id)}
-			/>
+			{onSelectOrToggleItem && (
```
The-Daniel, line 124: Why this change?

frontend/src/v5/ui/routes/dashboard/projects/federations/editFederationModal/editFederation/editFederationContainersList/editFederationContainersListItem/editFederationContainersListItemLoading.component.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { TeamspacesHooksSelectors, ProjectsHooksSelectors } from '@/v5/services/selectorsHooks';
+import { SkeletonListItem } from '@components/shared/revisionDetails/components/skeletonListItem/skeletonListItem.component';
+import { useEffect } from 'react';
+
+export const EditFederationContainersListItemLoading = ({ index, container }) => {
+	const teamspace = TeamspacesHooksSelectors.selectCurrentTeamspace();
+	const project = ProjectsHooksSelectors.selectCurrentProject();
+	
+	useEffect( () => {
+		ContainersActionsDispatchers.fetchContainerStats(teamspace, project, container._id);
```
The-Daniel, line 28: I think this might result in some unnecessary fetchContainerStats calls, but I don't understand why...
Removing this causes the items to be stuck in the loading stage despite the fact the stats are fetched in fetchContainers beforehand.
Also if a container is favourited then this will get called twice. Once for the top list, and once for the bottom list

The-Daniel: It actually seems to get called twice per item in the top list. And then once again per item in bottom list. I think this might be why it's quite slow when opening the edit model for a federations with 1000 containers

frontend/src/v5/ui/routes/dashboard/projects/containers/containersList/containerListItem/containerListItemLoading.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ *  Copyright (C) 2021 3D Repo Ltd
```
The-Daniel, line 19: Too much legal jargon

frontend/src/v5/store/containers/containers.sagas.ts
```diff
@@ -65,49 +70,50 @@ export function* removeFavourites({ containerId, teamspace, projectId }: RemoveF
...
-		const revisionsEqual = (container?.latestRevision === stats?.revisions?.latestRevision && container?.status === stats?.status) // if it has revisions
-							|| (container?.latestRevision === '' && stats?.revisions?.total === 0); // if it doesnt
+		statsQueue.resetQueue();
 
-		// Only update if theres is new data
-		if (!basicDataEqual || !revisionsEqual) {
-			yield put(ContainersActions.fetchContainerStatsSuccess(projectId, containerId, stats));
-		}
+		yield all(
+			containers.sort(getSortingFunction({ column: ['name'], direction:[SortingDirection.DESCENDING] })).map(
```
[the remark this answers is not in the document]
sanmont3drepo: Because we want to fetch the data in the same order the user sees it
If it is not ordered in the ordered list in the containers list it can look like this


| loading container 0 |
| container 1 loaded |
| loading container 2 |
| loading container 3 |
| loading container 4 |
| container 5 loaded |
| container 6 loaded |

this way it looks like this
| container 0 loaded |
| container 1 loaded |
| container 2 loaded |
| loading container3 |
| loading container 4 |
| loading container 5 |
| loading container 6 |

frontend/src/v5/ui/components/dashboard/dashboardList/dashboardListItem/components/dashboardListItemTitle/dashboardListItemContainerTitle/dashboardListItemContainerTitle.component.tsx
sanmont3drepo: While loading you dont know if the container has revisions or not. I'll change it so that while loading is enabled but once loaded if the container has no revisions it will be disabled.

frontend/src/v5/store/containers/containers.sagas.ts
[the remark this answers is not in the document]
The-Daniel: This will only be the case if the alphabetically sorted list matches the lastUpdated sorted list because the containers list is sorted by lastUpdated by default. Is it possible to do the same sort here?

You can see this when loading the containers in a project other than the generated 1000 containers one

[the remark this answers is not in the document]
sanmont3drepo: Is not possible because the date comes with the stats

The-Daniel (approved):

PR #4632, opened 2023-12-04
Issue 4584 - Issues and risks: show default filtering on the search

### LINKED ISSUE #4584: Issues and risks - show default filtering on the search
### Description
By default, we are hiding any resolved issues/risks. This is currently not being shown on the search filter. To some users, it's not obvious that we're hiding closed/resolved issues and risks. For transparency/consistency, we want to show the default filters that are being applied to issues/risks:

In essence, the starting state of issues and risks should be as follows:
image
image

### Goals
- [ ] As a user I want to be clearly shown what filters are being applied on the default issues/risks list.

### Tasks
- [ ] modify selectors to show the default state

Amantini1997:
This fixes #4584 

#### Description
Instead of hiding closed items, set the filters to only show the open ones and make the filter selection immediately visible

#### Test cases
- Open the issues list and see that the items visible at the beginning are the same as before the changes;
- Remove all the filters and you should be able to see all the issues, even the ones "closed" or "void"
- Repeat for risks (a risk is considered "close" depending on the treatment_status)

### DIFF
--- a/frontend/src/v4/modules/issues/issues.selectors.ts
+++ b/frontend/src/v4/modules/issues/issues.selectors.ts
@@ -128,7 +128,7 @@ export const selectAllFilteredIssuesGetter = createSelector(
 );
 
 export const selectFilteredIssues = createSelector(
-	selectAllFilteredIssuesGetter, (allFilteredIssuesGetter) => allFilteredIssuesGetter()
+	selectAllFilteredIssuesGetter, (allFilteredIssuesGetter) => allFilteredIssuesGetter(true)
 );
 
 export const selectShowPins = createSelector(
--- a/frontend/src/v4/modules/issues/issues.selectors.ts
+++ b/frontend/src/v4/modules/issues/issues.selectors.ts
@@ -118,13 +118,16 @@ export const selectSelectedFilters = createSelector(
 
 export const selectAllFilteredIssuesGetter = createSelector(
 	selectIssues, selectSelectedFilters, selectSortOrder, selectSortByField,
-		(issues, selectedFilters, sortOrder, sortByField) => (forceReturnHiddenIssue = false) =>  {
-			const returnHiddenIssue = selectedFilters.length && selectedFilters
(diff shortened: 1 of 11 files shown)


frontend/src/v4/modules/issues/issues.selectors.ts
The-Daniel, line 1: (Same applies to risks). Closed items do not show up in the kanban board.

Changing an issue on the board so that the status becomes Void or Closed removes it from the board entirely instead of moving it to the Void or Closed column

The-Daniel (changes requested):

frontend/src/v4/routes/viewerGui/components/issues/issues.component.tsx
```diff
@@ -18,7 +18,6 @@
 import { PureComponent } from 'react';
```
sanmont3drepo, line 18: Isn't missing "for approval" ? 

This is in the original issue #4584 
image

And this is in the PR:
image

frontend/src/v4/routes/viewerGui/components/risks/risks.component.tsx
```diff
@@ -87,14 +87,6 @@ export class Risks extends PureComponent<IProps, any> {
...
 
-	get showDefaultHiddenItems() {
-		if (this.props.selectedFilters.length) {
-			return this.props.selectedFilters
-				.some(({ value: { value } }) => RISK_DEFAULT_HIDDEN_LEVELS.includes(value));
-		}
-		return false;
-	}
-
 	public renderDetailsView = renderWhenTrue(() => (
```
sanmont3drepo, line 90: I see difference between the default filters in the original issue #4584 
image
and in the PR:

image

sanmont3drepo (changes requested):

frontend/src/v4/routes/viewerGui/components/risks/risks.component.tsx
Amantini1997: same as the previous comment

sanmont3drepo (approved):

PR #4634, opened 2023-12-04
Issue 4427 - Remove backwards compatibility wiith legacy federation/groups schema

### LINKED ISSUE #4427: Remove backwards compatibility code created from #4391
### Description
#4391 changes the schema of modelSettings and we needed backwards compatibility code to ensure the system functions before and after the migration.

We should be able to remove these compatibility code for the next release as the data should be fully migrated.

carmenfan:
This fixes #4427

#### Description
remove lingering backwards compatibility for pre v4.6 schema support.

This was kept as part of the 4.6 release to ensure we have a seamless service pre + post migration script.

#### Test cases
- service should function as normal


backend/testDBVersion
```diff
@@ -1 +1 @@
-896c6802eac730e95176c76a7b6dcf8da31ec69b
+0740cc90d73505896191d1974380de46b6bf1162
```
ChristosTsiotsias, line 1: should that be included ?

backend/unreferenced_files.csv
```diff
@@ -0,0 +1,4 @@
+69/160/9c26f4d2-92c7-11ee-a053-dd84b614f5d7
```
ChristosTsiotsias, line 1: should this file be included ?

ChristosTsiotsias:
- e2e test for export groups still has a test case for legacy groups 
- testSerialiseGroupArray unit test method still has code for testing legacy group conversion

ChristosTsiotsias (changes requested):

backend/src/v4/models/helper/model.js
```diff
@@ -653,11 +653,7 @@ function importModel(account, model, username, modelSetting, source, data) {
 }
 
 async function isSubModel(account, model) {
-
-	// Old schema support, to remove in 5.8
-
 	const query = {$or: [
-		{ subModels: model },
```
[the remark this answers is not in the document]
carmenfan: Yes 😆

backend/testDBVersion
carmenfan: yes, had to do a db migration of the test database. Thsi is why this the hash commit has been updated.

backend/unreferenced_files.csv
carmenfan: nope 😆 But i just realised this file comes from running the script tests. so adding code to remove the file as part of this issue.

ChristosTsiotsias (approved):

PR #4648, opened 2023-12-07
Issue 4578 - Send email to SSO users when they trigger forgot password request

### LINKED ISSUE #4578: Send an email to SSO users when they request to reset their password
### Description
At the moment, SSO linked accounts do not receive any confirmation of their reset password request as there is no password to reset. This sometimes can confuse the user as they may have forgotten that they have linked their account so a notification to their email will be useful to let them know.

Taking the forgot password email template, we want to keep most of it but change the blurb circled:
image

Change it to:

> We received your request to reset your account password. However, I would like to inform you that your account is associated with an `<sso type>` account, and you should log by clicking the `Sign in with <sso type>` button on the sign in page.

### Goals
- [ ] As a SSO linked user I want to be informed that I have linked my 3drpeo account with an SSO provider so that I am not confused about how to access my account.

### Tasks
TBC by the developer

carmenfan:
This fixes #4578

#### Description
- will now send an email to sso users if they triggered a forgot password request.

#### Test cases
- trigger forgot password on an SSO user should now sends an email
- the following behaviour should still stand:
  - triggering forgot password on a non SSO user should continue to receive an email
  - triggering forgot password on an invalid user/email should not receive an email.


backend/src/v5/middleware/dataConverter/inputs/users.js
```diff
@@ -158,15 +161,28 @@ Users.validateForgotPasswordData = async (req, res, next) => {
...
+				'customData.firstName': 1,
+				'customData.email': 1,
+			});
 
 			if (sso) {
+				await sendEmail(mailTemplates.FORGOT_PASSWORD_SSO.name, email,
+					{
+						email,
+						username: user,
+						firstName,
```
ChristosTsiotsias, line 177: do we need to send the firstName and email ?

backend/tests/v5/e2e/routes/users.test.js
```diff
@@ -565,14 +565,14 @@ const testForgotPassword = () => {
...
-		test('should not send email but return ok if user is an SSO user', async () => {
+		test('should send email and return ok if user is an SSO user', async () => {
 			await User.linkToSso(ssoTestUser.user, ssoTestUser.basicData.firstName, ssoTestUser.basicData.lastName,
 				ssoTestUser.basicData.email,
 				{ type: ServiceHelper.generateRandomString(), id: ServiceHelper.generateRandomString() });
 			await agent.post('/v5/user/password').send({ user: ssoTestUser.user })
 				.expect(templates.ok.status);
 			await User.unlinkFromSso(ssoTestUser.user, ssoTestUser.password);
-			expect(Mailer.sendEmail).not.toHaveBeenCalled();
+			expect(Mailer.sendEmail).toHaveBeenCalledTimes(1);
```
ChristosTsiotsias, line 575: we could also check the params (even though we dont do it for any test) ?

ChristosTsiotsias (changes requested):

backend/src/v5/middleware/dataConverter/inputs/users.js
carmenfan: firstName is required as part of the template so we can write the "hello %firstname%" part. https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/services/mailer/templates/baseTemplate.js#L27

Email.. i guess not.

backend/tests/v5/unit/services/sso/sso.constants.test.js
```diff
@@ -0,0 +1,39 @@
...
+const { src } = require('../../../helper/path');
+const { generateRandomString, determineTestGroup } = require('../../../helper/services');
+
+const SSOConstants = require(`${src}/services/sso/sso.constants`);
+
+const testGetProviderLabel = () => {
+	describe('Get Provider label', () => {
+		test('All known providers should return a label', () => {
+			Object.values(SSOConstants.providers).forEach((provider) => {
+				expect(SSOConstants.getProviderLabel(provider)).not.toBeUndefined();
```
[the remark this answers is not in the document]
carmenfan: I don't think there's much point of this.

that suggest we'd hard code the values here to make sure it match.

The alternative would be to expose the value and match it.

If someone changed the label, it would've gone through the code review. We just need to make sure there's always a label for each SSO type

backend/tests/v5/e2e/routes/users.test.js
carmenfan: I don't think it's necessary. it's a e2e test we just need to make sure it's sending an email.

Too much effort 😆

ChristosTsiotsias (approved):

PR #4653, opened 2023-12-08
ISSUE_4563 - Ticket pins get filtered with the list

### LINKED ISSUE #4563: Only filtered ticket pins should be visible

The-Daniel:
This fixes #4563 
#### Description
- In the tickets list filtered pins on the model also get filtered with their respective list items
- Ticket filtering persists when the tickets card is closed and reopened
- Cycling between tickets now uses the filtered tickets list

#### Test cases
- Create tickets with default pins
- Filter out tickets with the completed chip, template chips, search queries, and sequencing dates
- Can view tickets in tabular view with no crashes


Amantini1997 (changes requested):

Amantini1997 (changes requested):

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/coordsProperty/coordsProperty.component.tsx
```diff
@@ -104,14 +109,13 @@ export const CoordsProperty = ({ value, label, onChange, onBlur, required, error
 
 	// Update pin when colour changes
 	useEffect(() => {
+		if (!prevValue.current) return;
```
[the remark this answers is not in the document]
The-Daniel: It should only do the animation when clicking the select pin button, not when creating a pin

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/coordsProperty/coordsProperty.component.tsx
```diff
@@ -26,31 +26,36 @@ import { FormHelperText, Tooltip } from '@mui/material';
...
 import { ViewerParams } from '@/v5/ui/routes/routes.constants';
 import { useParams } from 'react-router-dom';
-import { useFormContext } from 'react-hook-form';
 import { ITicket } from '@/v5/store/tickets/tickets.types';
 import { isEqual } from 'lodash';
+import { useFormContext } from 'react-hook-form';
+
+const NEW_TICKET_ID = 'temporaryIdForNewTickets';
 
 export const CoordsProperty = ({ value, label, onChange, onBlur, required, error, helperText, disabled, name }: FormInputProps) => {
```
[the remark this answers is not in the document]
The-Daniel: This will be addressed here https://github.com/3drepo/3drepo.io/issues/4664

frontend/src/v5/store/tickets/card/ticketsCard.selectors.ts
```diff
@@ -97,23 +99,82 @@ export const selectSelectedTemplate = createSelector(
...
+	selectCurrentTemplates,
+	(tickets, isComplete, queries, templates) => {
+		const filteredByCompleted = tickets.filter((ticket) => getTicketIsCompleted(ticket) === isComplete);
+		if (!queries.length) return filteredByCompleted;
+		return filteredByCompleted.filter((ticket) => {
+			const templateCode = templates.find((template) => template._id === ticket.type).code;
+			const ticketCode = `${templateCode}:${ticket.number}`;
+			return queries.some((q) => [ticketCode, ticket.title].some((str) => str.toLowerCase().includes(q.toLowerCase())));
+		});
+	},
```
sanmont3drepo, line 130: To avoid filtering twice the tickets we can use just one loop:

```suggestion
	(tickets, isComplete, queries, templates) => tickets.filter((ticket) => {
		const templateCode = templates.find((template) => template._id === ticket.type).code;
		const ticketCode = `${templateCode}:${ticket.number}`;
		return getTicketIsCompleted(ticket) === isComplete && 
		(!queries.length || queries.some((q) => [ticketCode, ticket.title].some((str) => str.toLowerCase().includes(q.toLowerCase()))));
	}),
```

sanmont3drepo: @carmenfan As it is now the tickets that has no issuproperties or safetibaseproperties are filetered out when filtering by complete. Is ok this or those tickets should appear regardless the value pf **completed** filter?

sanmont3drepo (changes requested):

frontend/src/v5/store/tickets/card/ticketsCard.selectors.ts
carmenfan: yes this is fine for now

frontend/src/v5/ui/routes/viewer/tickets/ticketDetails/ticketsDetailsCard.component.tsx
```diff
@@ -51,16 +56,22 @@ export const TicketDetailsCard = () => {
...
+		let index = currentIndex === -1 ? initialIndex.current : currentIndex;
+		if (currentIndex === -1 && delta === IndexChange.UP) index -= 1;
+		return (index + delta) % filteredTickets.length;
+	};
 
-	const changeTicketIndex = (delta: number) => {
-		const index = currentIndex === -1 ? initialIndex.current : currentIndex;
-		const updatedId = filteredTickets.slice((index + delta) % filteredTickets.length)[0]._id;
+	const changeTicketIndex = (delta: IndexChange) => {
+		const updatedId = filteredTickets.slice(getUpdatedIndex(delta))[0]._id;
```
Amantini1997, line 68: `getUpdatedIndex(delta)` returns the index of the ticket, so I was wondering why not just writing `filteredTickets[getUpdatedIndex(delta)]` instead of calling `slice` and get the first element.

Then I found out that, when you go to the last ticket from the first one, `getUpdatedIndex` returns `-1`. (Supposing this is why you call `slice`) I think that, if you want to keep `getUpdatedIndex` as a separate method, the latter should never return `-1` as that's not a real index. Otherwise, if you merge the 2 methods, there is an array method called `at` which is similar to the `[<index>]` operator, but accepts negative numbers. Negative numbers act the same way as python, moving backwards e.g. `[0,1,2,3,4].at(-1) => 4 `

frontend/src/v5/ui/routes/viewer/tickets/ticketDetails/ticketsDetailsCard.component.tsx
```diff
@@ -51,16 +56,22 @@ export const TicketDetailsCard = () => {
 	const defaultView = ticket?.properties?.[AdditionalProperties.DEFAULT_VIEW];
 	const currentIndex = filteredTickets.findIndex((tckt) => tckt._id === ticket._id);
 	const initialIndex = useRef(currentIndex);
+	const disableCycleButtons = currentIndex > -1 ? filteredTickets.length < 2 : filteredTickets.length < 1;
+
+	const getUpdatedIndex = (delta: number) => {
+		let index = currentIndex === -1 ? initialIndex.current : currentIndex;
```
Amantini1997, line 62: for this to work, you need to update the `initialIndex.current` everytime the ticket changes, like:
```ts
useEffect(() => {
	initialIndex.current = currentIndex;
}, [ticket?._id]);
```

Otherwise, reproduce this scenario:
- You have 4 tickets, all named "AAA";
- In the search input, type in "AAA";
- Open up the first ticket (initialIndex.current === 0);
- Using the arrows, reach the third ticket
- Change the name to "BBB"
- Using the arrows, go to the previous ticket

As a result, you will reach the fourth ticket as opposed to the second

frontend/test/tickets/card/ticketsCard.store.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { TicketsCardActions } from "@/v5/store/tickets/card/ticketsCard.redux";
```
Amantini1997, line 1: Not sure whether this is going to be a problem, but I keep seeing these 2 errors printed in the console

image

Amantini1997 (changes requested):

Amantini1997 (approved):

carmenfan (approved):

carmenfan:
Merging without @sanmont3drepo 's final review as he's on leave for another week and this issue has dependencies.

Issue #4641, opened 2023-12-06
v4 Groups - full IFC GUID and revit ID support


carmenfan:
### Description
We're introducing Revit ID support and also better IFC Guid support in https://github.com/3drepo/3drepo.io/issues/4552 for ticket groups.

We also need this on v4 groups.

### Goals
- [ ] As a user I want my groups to work across revisions if I am providing a BIM model that has persistent IFC guids or Revit IDs

### Tasks
TBC

carmenfan:
Merged into #4552

PR #4662, opened 2023-12-15
Issue 4552 - Revit Ids and Ifc Guids in ticket groups

### LINKED ISSUE #4552: Ticket groups: IFC guids and revit Ids support
### Description
in #3933 we added support for groups within the ticket. But we implemented the minimal scope were we take in either unique IDs (manual groups) or a smart rule.

In v4 groups, we also support IFC Guids and we now want to introduce this to ticket groups.

Additionally, we would like to start supporting revit Ids.

#### Adding or updating a manual group
The user should be able to specify either the `_ids`, `ifcGuids`, `revitIds`. For e.g. when creating a group I can instead provide ifcGuids:
````js
//when updating or creating a ticket
{
  "View Name": {
     state: {
       // Keep using UUID to reference existing groups. You can even reference groups from different state arrays (i.e. move a colour override group to hidden
       hidden: [ {group: UUID}, {group: UUID}, {prefix: ["a", "b"], group:  {
         name: "A new group to be hidden",
         objects: [
           { container: "container id", ifcGuids: ["3saVcraO9CJRqkF0xoe8Ya", "1kWP76DUH9Re1av3vI20e7"] }
         ]
        }}]
     }
   }
}

// or provide revitIds:
//when updating or creating a ticket
{
  "View Name": {
     state: {
       // Keep using UUID to reference existing groups. You can even reference groups from different state arrays (i.e. move a colour override group to hidden
       hidden: [ {group: UUID}, {group: UUID}, {prefix: ["a", "b"], group:  {
         name: "A new group to be hidden",
         objects: [
           { container: "container id", revitIds: ["3saVcraO9CJRqkF0xoe8Ya", "1kWP76DUH9Re1av3vI20e7"] }
         ]
        }}]
     }
   }
}

````
##### NOTES:
- we should be storing the data as revit Ids or ifc guids - do not convert them to 3drepo ids.
- If 3drepo ids are provided, we should try to store ifc guids, or revit Ids where possible (in that preferential order)
- it should be mutually exclusive - within the same container, the user cannot provide a combination of revitIds/ifcguids/_ids
- However, in a federation, the user should be able to provide objects from 1 container in a different format to another (e.g. container A with 3drepo ids, container B with ifc guids

#### getting a group
The user can tell us if they want the data in the original format, or be converted to 3drepo guids. This should be indicated by a flag on the query string 

````js

with flag ?convertIds=true (or not specified)
{
  _id: UUID,
   name: "Group name",
   description: "Description of the group",
  objects: [
    { container: "container id", _ids: [] } 
   ]
}

with flag ?convertIds=false
{
  _id: UUID,
   name: "Group name",
   description: "Description of the group",
  objects: [
    { container: "container id", ifcGuids: [] } 
   ]
}

````
##### NOTES:
- This should also apply to smart groups, when converting rules, provide ifcGuids/revitIds (in that preferential order) if `convertIds` is set to false.

#### Id namings
IFC IDs can be found in the following formats, and we need to make sure that we are capturing the metadata fields are all of them:
- Ifc
- Revit (native)
- Revit (via .bim)
- Navis (native)
- Navis (via .bim)

Revit IDs can only be found in the following formats
- Revit (native)
- Revit (via .bim)
- Navis (native)
- Navis (via .bim)

### Goals
- [ ] As a user I want to be able to see my groups across revisions if my model is uploaded in a BIM file format
- [ ]  As an integration developer I want to be able to retrieve IFC GUIDs or revit Ids so I can apply them in my application.

### Tasks

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #4552

#### Description
Revit Id and Ifc Guid support for ticket groups:
Ticket groups now accept revit_ids and ifc_guids for object ids.
If object ids are provided they are stored as external ids if possible.
When retrieved the user has the option to convert them to 3d repo ids or keep them as external ids.
Unit tests added.

#### Test cases
revit_ids and ifc_guids should be acceptable in ticket groups objects
when retrieving an object you could receive ifc_guids or revit_ids if convertIds is set to false

### CODE REVIEW

File: backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js [hunk 1]
```diff
@@ -40,14 +57,23 @@ const getObjectArrayFromRules = async (teamspace, project, model, revId, rules)
...
-	const [
-		{ matched, unwanted },
-		idToMeshes,
-	] = await Promise.all([
-		getMetadataByRules(teamspace, project, model, revision, rules, { parents: 1 }),
-		getIdToMeshesMapping(teamspace, model, revision),
-	]);
+	const projection = { parents: 1,
+		...(convertTo3dRepoIds ? {} : { metadata: {
+			$elemMatch: { $or: Object.values(externalIdNamesToKeys).flat().map((n) => ({ key: n })) } } }) };
```
carmenfan, line 62: probably easier to read if we have the `Object.values` part be in a different line and just put in a variable here.

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -40,14 +57,23 @@ const getObjectArrayFromRules = async (teamspace, project, model, revId, rules)
...
 
+	const { matched, unwanted } = await getMetadataByRules(teamspace, project, model, revision, rules, projection);
+
+	if (!convertTo3dRepoIds && matched.some((m) => m.metadata)) {
```
carmenfan, line 66: Bit confused with this logic.

so if we don't want 3drepo Ids, and there are no match, it will fall out of this function and the IdToMeshes stuff?

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -40,14 +57,23 @@ const getObjectArrayFromRules = async (teamspace, project, model, revId, rules)
...
+		const wantedIds = [...new Set(matched.map(({ metadata }) => metadata[0].value))];
+		const unwantedIds = [...new Set(unwanted.map(({ metadata }) => metadata[0].value))];
+
+		unwantedIds.filter((id) => (wantedIds.includes(id))).forEach((id) => delete wantedIds[id]);
+
+		const externalIdName = getExteralidNameFromMetadata(matched);
+		return { container: model, [externalIdName]: wantedIds };
+	}
+
+	const idToMeshes = await getIdToMeshesMapping(teamspace, model, revision);
```
carmenfan, line 76: Use helper function that I've suggested previously [get the difference](https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/utils/helper/arrays.js#L23)

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -79,22 +105,126 @@ const getObjectArrayFromRules = async (teamspace, project, model, revId, rules)
 	return { container: model, _ids: Object.values(matchedMeshes) };
 };
 
-TicketGroups.updateTicketGroup = updateGroup;
+const convert3dRepoIdsToExternalIds = async (teamspace, project, objects) => {
```
carmenfan, line 108:
```suggestion
const convertToExternalIds = async (teamspace, project, objects) => {
```

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -79,22 +105,126 @@ const getObjectArrayFromRules = async (teamspace, project, model, revId, rules)
...
+	const convertedObjects = await Promise.all(objects.map(async (obj) => {
+		if (!obj._ids) {
+			return obj;
+		}
+
+		const convertedObject = { ...obj };
```
carmenfan, line 114: create this when you need it  (around L122)

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -79,22 +105,126 @@ const getObjectArrayFromRules = async (teamspace, project, model, revId, rules)
...
+
+		const shared_ids = await getNodesByIds(teamspace, project, obj.container, obj._ids,
```
carmenfan, line 116: camel case (this is why we should not be turning off a linter rule on the whole file!)

carmenfan, line 116: the context of this variable is an array of  node objects,  so `shared_ids` being the variable name is quite confusing. Should rename to `nodes` or something

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -79,22 +105,126 @@ const getObjectArrayFromRules = async (teamspace, project, model, revId, rules)
...
+			{ _id: 0, shared_id: 1 });
+
+		const externalIdKeys = Object.values(externalIdNamesToKeys).flat();
+		const query = { parents: { $in: shared_ids.map((s) => s.shared_id) }, 'metadata.key': { $in: externalIdKeys } };
```
carmenfan, line 120: this will work, but in the principle that shared_ids can be across revisions, we should really put rev_id in this query.

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.js
```diff
@@ -26,9 +25,11 @@ const {
 	viewGroups,
 } = require('../../../../../schemas/tickets/templates.constants');
 const { createResponseCode, templates } = require('../../../../../utils/responseCodes');
+const { deleteGroups, getGroupsByIds } = require('../../../../../models/tickets.groups');
```
carmenfan, line 28: this starts to  look confusing as to why we're using model here but processor for addGroups. I think we should just have  the functions (just a passthrough, assign the model functions to the processor object) in processors so it's getting all the group functions in one file.

backend/src/v5/processors/teamspaces/projects/models/federations.js
```diff
@@ -32,10 +32,11 @@ const { queueFederationUpdate } = require('../../../../services/modelProcessing'
 const Federations = { ...Groups, ...Views, ...Tickets, ...Comments, ...TicketGroups };
 
 // Override
-Federations.getTicketGroupById = async (teamspace, project, federation, revId, ticket, groupId) => {
+Federations.getTicketGroupById = async (teamspace, project, federation, revId, ticket, groupId,
+	convertTo3DRepoGuids) => {
```
carmenfan, line 36: rename `convertTo3DRepoGuids` to something similar in the ticket groups processor suggestion

backend/src/v5/routes/teamspaces/projects/models/common/tickets.groups.js
```diff
@@ -44,8 +44,10 @@ const getGroup = (isFed) => async (req, res, next) => {
 	const { teamspace, project, model, ticket, group } = params;
 
 	try {
+		const convertTo3DRepoGuids = query.convertIds !== 'false';
```
carmenfan, line 47: rename `convertTo3DRepoGuids` to something similar in the ticket groups processor suggestion

backend/src/v5/schemas/tickets/tickets.groups.js
```diff
@@ -25,19 +25,39 @@ const Groups = {};
...
+const objectEntryValidator = Yup.object().shape({
+	container: Yup.string().test('Container id', 'Container ID must be an UUID string', isUUIDString).required(),
+	_ids: Yup.array().of(types.id).min(1),
+	ifc_guids: Yup.array().of(Yup.string().length(22)).min(1),
+	revit_ids: Yup.array().of(Yup.number()).min(1),
+}).test(
+	'Object item check',
+	'Can only contain either _ids or ifc_guids or revit_ids',
+	/* eslint-disable no-underscore-dangle */
+	(value) => (value._ids && !(value.ifc_guids || value.revit_ids))
```
carmenfan, line 37: Unless i'm missing something, this logic would allow me to have no ids? I think `{container: "<some uuid>"}` would pass here.

If this is really true then you're missing a unit test...

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/tickets.groups.test.js
```diff
@@ -37,244 +38,342 @@ const ScenesModel = require(`${src}/models/scenes`);
...
-		const project = generateRandomString();
-		const model = generateRandomString();
-		const revId = generateRandomString();
-		const ticket = generateRandomString();
-		const group = generateRandomString();
-
-		test('[Normal group] should return the group found', async () => {
-			const expectedData = generateRandomObject();
-			GroupsModel.getGroupById.mockResolvedValueOnce(expectedData);
+const externalIdNamesToKeys = {
```
carmenfan, line 41: you shoulnd't be declaring constants that is declared in  source here. use an exposed constant object from source.

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/tickets.groups.test.js
```diff
@@ -37,244 +38,342 @@ const ScenesModel = require(`${src}/models/scenes`);
...
+		group: smartGroup,
+		matchedMeta: getMeta('IFC GUID'),
+		unwantedMeta: [],
+		revision: generateRandomString(),
+		convertTo3dRepoIds: true,
+		latestRevisionFail: false,
+	};
+
+	describe.each([
+		['when its a container group', defaultOptions],
```
carmenfan, line 73: it is, or it\'s
```suggestion
		['when it is a container group', defaultOptions],
```

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/tickets.groups.test.js
```diff
@@ -37,244 +38,342 @@ const ScenesModel = require(`${src}/models/scenes`);
...
+		['when its a federation group', { ...defaultOptions, containers: [container] }],
```
carmenfan, line 74: it is, or it's

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/tickets.groups.test.js
```diff
@@ -37,244 +38,342 @@ const ScenesModel = require(`${src}/models/scenes`);
...
+
+	const getMeta = (externalIdName) => times(10, () => ({ parents: times(2, () => generateRandomString()),
+		metadata: [{ key: externalIdName, value: generateRandomString() }] }));
+
+	const defaultOptions = {
+		group: smartGroup,
+		matchedMeta: getMeta('IFC GUID'),
+		unwantedMeta: [],
+		revision: generateRandomString(),
+		convertTo3dRepoIds: true,
```
carmenfan, line 68: rename this

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/tickets.groups.test.js
```diff
@@ -37,244 +38,342 @@ const ScenesModel = require(`${src}/models/scenes`);
...
-
-			const expectedData = cloneDeep(groupData);
-			expectedData.objects = [{
-				container: model,
-				_ids: nodeRes.map(({ _id }) => _id),
-			}];
-
-			await expect(Groups.getTicketGroupById(teamspace, project, model, revId, ticket, group))
-				.resolves.toEqual(expectedData);
+	describe.each([
```
carmenfan, line 163: This is too confusing to read. split the test function into 2

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/tickets.groups.test.js
```diff
@@ -37,244 +38,342 @@ const ScenesModel = require(`${src}/models/scenes`);
...
+		['when its a federation group', { ...defaultOptions, containers: [container], group: ifcGuidGroup, externalIdName: 'ifc_guids' }],
```
carmenfan, line 164: should not be hard coding `ifc_guids`

carmenfan: similar comments with the other test cases

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/tickets.groups.test.js
```diff
@@ -37,244 +38,342 @@ const ScenesModel = require(`${src}/models/scenes`);
...
-
-			expect(MetaModel.getMetadataByRules).toHaveBeenCalledTimes(1);
-			expect(MetaModel.getMetadataByRules).toHaveBeenCalledWith(teamspace, project, model, revId,
-				groupData.rules, { parents: 1 });
-
-			expect(ScenesModel.getNodesBySharedIds).toHaveBeenCalledTimes(2);
-			expect(ScenesModel.getNodesBySharedIds).toHaveBeenCalledWith(teamspace, project, model,
-				revId, metaRes.flatMap(({ parents }) => parents), { _id: 1 });
+		describe.each([
+			['preserve oject as it is (no metadata found)', undefined, []],
```
carmenfan, line 255:
```suggestion
			['preserve object as it is (no metadata found)', undefined, []],
```

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/tickets.groups.test.js
```diff
@@ -37,244 +38,342 @@ const ScenesModel = require(`${src}/models/scenes`);
...
-				revId, metaRes.flatMap(({ parents }) => parents), { _id: 1 });
+const testUpdateGroup = () => {
+	const teamspace = generateRandomString();
+	const project = generateRandomString();
+	const container = generateRandomString();
+	const ticket = generateRandomString();
+	const groupId = generateRandomString();
+	const author = generateRandomString();
+
+	describe('Update groups', () => {
```
carmenfan, line 311: what's the significant difference between update group and add group use cases? i feel like it's pretty much copy and pasted with different function calls. Should there be a helper function that generates the test cases for both?

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/tickets.groups.test.js
```diff
@@ -37,244 +38,342 @@ const ScenesModel = require(`${src}/models/scenes`);
...
+		test('should update a group', async () => {
```
carmenfan, line 312: this should be a smart group test?

Otherwise i'm confused as to why this exists, if it's just a ticket group with no metadata match then it's just `preserve object as it is (no metadata found)` tetst in the loop

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/tickets.groups.test.js
```diff
@@ -37,244 +38,342 @@ const ScenesModel = require(`${src}/models/scenes`);
...
-				groupData.rules, { parents: 1 });
+			expect(GroupsModel.updateGroup).toHaveBeenCalledTimes(1);
+			expect(GroupsModel.updateGroup).toHaveBeenCalledWith(teamspace, project, container, ticket, groupId,
+				data, author);
+		});
 
-			expect(ScenesModel.getNodesBySharedIds).not.toHaveBeenCalled();
+		describe.each([
+			['preserve oject as it is (no metadata found)', undefined, []],
+			['convert object ids to ifc_guids', 'ifc_guids', times(10, () => ({ metadata: [{ key: 'IFC GUID', value: generateRandomString() }] }))],
```
carmenfan, line 332: again, we should not be hard coding constants

backend/tests/v5/unit/processors/teamspaces/projects/models/federations.test.js
```diff
@@ -340,15 +340,15 @@ const testGetTicketGroupById = () => {
 			const expectedData = generateRandomObject();
 			const fn = jest.spyOn(TicketGroup, 'getTicketGroupById').mockResolvedValueOnce(expectedData);
 
-			await expect(Federations.getTicketGroupById(teamspace, projectId, federation, revId, ticket, groupId))
+			await expect(Federations.getTicketGroupById(teamspace, projectId, federation, revId, ticket, groupId, true))
```
carmenfan, line 343: we should probably put true as default on the function definition to avoid unexpected situations

backend/tests/v5/unit/schemas/tickets/tickets.groups.test.js
```diff
@@ -70,11 +71,27 @@ const testSchema = () => {
...
+			objects: [{ ifc_guids: [generateRandomString(22)], container: generateUUIDString() }],
+		}, true],
+		['data has objects with both ifc_guids and revit_ids', false, false, {
+			name: generateRandomString(),
+			objects: [{ ifc_guids: [generateRandomString(22)], revit_ids: [], container: generateUUIDString() }],
+		}, false],
+		['data has objects with both _ids and revit_ids', false, false, {
+			name: generateRandomString(),
+			objects: [{ _ids: [generateUUID()], revit_ids: [], container: generateUUIDString() }],
+		}, false],
```
carmenfan, line 94: add test for no ids on the schema

carmenfan (changes requested):

backend/src/v5/routes/teamspaces/projects/models/common/tickets.groups.js
carmenfan: Also missing swagger update for this additional flag

backend/src/v5/schemas/tickets/tickets.groups.js
```diff
@@ -25,19 +25,39 @@ const Groups = {};
 
 const uuidString = Yup.string().transform((val, orgVal) => UUIDToString(orgVal));
 
+const objectEntryValidator = Yup.object().shape({
+	container: Yup.string().test('Container id', 'Container ID must be an UUID string', isUUIDString).required(),
+	_ids: Yup.array().of(types.id).min(1),
+	ifc_guids: Yup.array().of(Yup.string().length(22)).min(1),
```
carmenfan, line 31: Can we also use the labelling from constants here

backend/tests/v5/unit/schemas/tickets/tickets.groups.test.js
carmenfan: also need tests for wrong type on revit/ifc guids

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -14,22 +14,39 @@
...
+const { getNodesByIds, getNodesBySharedIds } = require('../../../../../models/scenes');
 const { getFile } = require('../../../../../services/filesManager');
 const { getLatestRevision } = require('../../../../../models/revisions');
-const { getMetadataByRules } = require('../../../../../models/metadata');
-const { getNodesBySharedIds } = require('../../../../../models/scenes');
 
 const TicketGroups = {};
 
+const externalIdNamesToKeys = {
+	ifc_guids: ['IFC GUID', 'Ifc::IfcGUID', 'Element::IfcGUID'],
```
carmenfan, line 29: I think you're missing IFC Guids from RVT file 
image
https://staging.dev.3drepo.io/v5/viewer/DevTeamWorkspace/aa721c11-4f87-491c-b18e-002a8505cf9d/72d992ea-026d-4312-8bbc-c80c658e033f

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -14,22 +14,39 @@
...
+	revit_ids: ['Element ID', 'Element ID::Value'],
```
carmenfan, line 30: Missing Revit ID on IFC
image
https://staging.dev.3drepo.io/v5/viewer/DevTeamWorkspace/aa721c11-4f87-491c-b18e-002a8505cf9d/e0dafa06-8033-4d44-acda-4a3b781b93b6

also Revit ID on IFC through Navis
image

carmenfan (changes requested):

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -79,22 +105,126 @@ const getObjectArrayFromRules = async (teamspace, project, model, revId, rules)
...
-				const objs = await getObjectArrayFromRules(teamspace, project, con, rev, group.rules);
+				const objs = await getObjectArrayFromRules(teamspace, project, con, rev,
+					group.rules, convertTo3dRepoIds);
 				// eslint-disable-next-line no-underscore-dangle
-				return objs._ids.length ? objs : [];
+				return objs._ids?.length || objs.revit_ids?.length || objs.ifc_guids?.length ? objs : [];
 			}),
 		)).flat();
+	} else if (convertTo3dRepoIds) {
+		group.objects = await convertExternalIdsTo3dRepoIds(teamspace, project, rev, group.objects);
```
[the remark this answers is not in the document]
carmenfan: We also need to check the container listed in the groups is a submodel of the federation (i.e. check if it's in the `con` array

If it's no longer a submodel then we shouldn't doing all the work to get the translation, just remove it from the results.

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -79,22 +105,126 @@ const getObjectArrayFromRules = async (teamspace, project, model, revId, rules)
...
 
-TicketGroups.getTicketGroupById = async (teamspace, project, model, revId, ticket, groupId, containers) => {
+	if (data.objects) {
+		convertedData.objects = await convert3dRepoIdsToExternalIds(teamspace, project, data.objects);
+	}
+
+	await updateGroup(teamspace, project, model, ticket, groupId, convertedData, author);
+};
+
+TicketGroups.getTicketGroupById = async (teamspace, project, model, revId, ticket, groupId, convertTo3dRepoIds,
```
carmenfan, line 209: I swapped the order between `convertTo3DRepoGuids` and containers, this broke federations previously (discovered whilst writing e2e tests).

I probably broke unit tests with the change, please fix

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -14,22 +14,39 @@
...
 const getIdToMeshesMapping = async (teamspace, model, revId) => {
 	const fileData = await getFile(teamspace, `${model}.stash.json_mpc`, `${UUIDToString(revId)}/idToMeshes.json`);
 	return JSON.parse(fileData);
 };
 
-const getObjectArrayFromRules = async (teamspace, project, model, revId, rules) => {
+const getExteralidNameFromMetadata = (metadata) => {
+	let externalIdName;
+	Object.keys(externalIdNamesToKeys).forEach((name) => {
+		if (externalIdNamesToKeys[name].some((m) => m === metadata[0].metadata[0].key)) {
```
carmenfan, line 41: what if we only have partial match? i.e. we have 10 3drepo ids, 5 of which have IFC guids, 5 of which don't.

This is possible from Navisworks upload, where it is a federation of multiple files, and the files could be of different file type. E.g. if i upload a nwd that has a DGN and an IFC file. With this assumption,  if i create a manually selected group that contains DGN and IFC objects, i'm going to lose all the DGN records from the group?

I think we need to ensure there's a 100% match for a particular type of ID before we declare a match.

carmenfan: `metadata[0].metadata[0]` is weird.

What if I am using an IFC file that has both revit and IFC guids? can you guarantee metadata[0] will always be the IFC GUID?

Or am I missing something?

carmenfan (changes requested):
A few more bugs/changes needed

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/tickets.groups.test.js
```diff
@@ -85,25 +89,171 @@ const setupBasicData = async ({ users, teamspace, project, fed, con, template, t
...
+		modules: [],
+	};
+	const ticket = ServiceHelper.generateTicket(template);
+
+	const rev = ServiceHelper.generateRevisionEntry();
+	const revId = rev._id;
+
+	const rootNode = ServiceHelper.generateBasicNode('transformation', revId);
+
+	// FIXME: constant reference
```
carmenfan, line 117: @ChristosTsiotsias  please go through the FIXMEs in this file and replace the hard coded naming with a constant from the previous review.

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -14,22 +14,39 @@
  *  You should have received a copy of the GNU Affero General Public License
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+/* eslint-disable no-underscore-dangle */
```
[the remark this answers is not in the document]
ChristosTsiotsias: In this case I added it because of _ids property of groups, so alternatively I have to add the linter disable above each line in the file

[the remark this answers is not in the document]
carmenfan: > I have to add the linter disable above each line in the file

Do that, this is what you did on the unit test anyway 😆

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -14,22 +14,39 @@
...
+
 const getIdToMeshesMapping = async (teamspace, model, revId) => {
 	const fileData = await getFile(teamspace, `${model}.stash.json_mpc`, `${UUIDToString(revId)}/idToMeshes.json`);
 	return JSON.parse(fileData);
 };
 
-const getObjectArrayFromRules = async (teamspace, project, model, revId, rules) => {
+const getExteralidNameFromMetadata = (metadata) => {
+	let externalIdName;
+	Object.keys(externalIdNamesToKeys).forEach((name) => {
```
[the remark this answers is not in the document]
ChristosTsiotsias: but I want to assign it to the key of externalIdNamesToKeys. ie if it metadata[0].metadata[0].key is Element ID I want externalIdName to be revit_ids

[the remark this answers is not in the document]
carmenfan: ah true, fair enough

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
ChristosTsiotsias: its basically saying that if convertIds is false and there are any matched metadata return the external Ids otherwise return 3d repo Ids

carmenfan: @ChristosTsiotsias  this is my issue.

If the user has asked for ifc guids etc, the schema we're returning shouldn't contain 3drepo ids just because we couldn't find any matches. if should return an empty array for ifc_guids or something.

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
ChristosTsiotsias: even without the linter rule disabling shared_ids is fine, it only complains when the _ is in the beggining, but I can change it to sharedIds regardless

carmenfan: as per the comment below, change it to nodes or something.

sharedIds is not a good representative for what it is holding

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
ChristosTsiotsias: this rev_id should I get it from nodes though ? 🤔

carmenfan: or yes, you can just take it from the nodes. they should be coming from the same revision

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
ChristosTsiotsias: something like this or not at all ? 🤔 

image

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/tickets.groups.test.js
ChristosTsiotsias: ah it was just to cover the scenario that data do not update objects

image

I could just rename the description to 'should update a group without updating objects' ?

carmenfan: right, in which case we should be checking convertToExternalIds is not being called.

I know you can't do it here as it's an internal function, but i've had to expose it in #4641 

Leave this one open, i'll add it on that branch

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
carmenfan: errr, just leave it, i'll have another look on #4641

carmenfan (approved):

carmenfan (approved):

PR #4666, opened 2024-01-04
ISSUE #4623 - New clipping box toolbar items

### LINKED ISSUE #4623: New clipping box
### Description
Because of several complaints by clients we decided to overhaul the clipping box functionality . For doing so we are adding a user interface with buttons to set the differents modes of the clipping tool.

NOTE: this will need to branch off https://github.com/3drepo/3drepo.io/issues/4477 to get the gizmo. Make sure you are using the 2GB viewer to see the latest clipping plane functionality from unity.
 
### Goals
- [ ] Users can create clipping boxes through the toolbar
- [ ] Users can change the mode of the clipping box to resize, move, or rotate through buttons in the toolbar **(scale shouldn't be available in plane mode**)
- [ ] Users can select between box or plane through a button in the toolbar
- [ ] Users can cancel the clipping creation through a button in the toolbar
- [ ] Users can shrink/expand the clipping box/plane based on what objects have been highlighted through a button in the toolbar
- [ ] Users can shrink/expand the clipping box/plane based on what objects have been highlighted through a button in the toolbar
- [ ] Users can flip the clipping plane. (**this shouldn't be available on cube mode**)

### Tasks
- [ ] Add callbacks to unity to set the mode

### Related Resources

- Fixed buttons in the toolbar:
image

- Contextual buttons that appears when in clipping mode
image

**IMPORTANT** 
- when is in plane mode the rotate gizmo shouldnt appear.
- when is in cube mode the flip button shouldnt appear

General design:
image

[MockUp](https://www.figma.com/proto/bRqIl0g9psqD3ISkmxwcNp/3D-Repo?type=design&node-id=1900-51556&t=1kPdVdpeEHrt2P06-0&scaling=min-zoom&page-id=182%3A30991&starting-point-node-id=2210%3A51164&show-proto-sidebar=1)

The-Daniel:
This fixes #4623

#### Description
- The clipping gizmo can now be controlled by buttons in the toolbar
- The plane gizmo has a flip button but no scale gizmo mode
- The box mode has no flip button but does have a scale gizmo mode

#### Test cases
Try all the new buttons related to the clipping gizmo


The-Daniel:
(Personally I think the disabled state needs to be darker on the toolbar buttons)

carmenfan:
> (Personally I think the disabled state needs to be darker on the toolbar buttons)

I suggest doing a azp build and show @apeitup just to confirm he's happy with the design

frontend/assets/icons/viewer/align.svg.tsx
```diff
@@ -0,0 +1,28 @@
+/**
```
carmenfan, line 1: Some tool tip changes

image

Can we call this `Move`?

image

This should be `Remove` or `Delete`.

image
`Align to surface`

image

`Resize`


Can we call this `Sectioning`, `Section by plane`, `Section by Box`

carmenfan: Agreed the disable look isn't prominent enough. we should look into making it more obvious

carmenfan, line 1: Clicking on a view with clipping plane should then allow me to click on the clip tool to edit the plane - at the moment the bottom tool bar doesn't seems to think there's a clip on the model

carmenfan, line 1: Clip icon should update to clip box if i selected a viewpoint with a clip box in view:

carmenfan, line 1: Clip should reset if I click on a view that does not have a clip configuration

carmenfan, line 1: Clicking the delete/Remove/Cancel (whatever we call it) should hide the Clip tool sub menu

carmenfan (changes requested):

frontend/assets/icons/viewer/align.svg.tsx
Amantini1997, line 1: After clicking the "Align" button, the viewer is waiting for a click on a mesh, but the button itself doesn't look "active". Should it change the visible state?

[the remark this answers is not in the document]
carmenfan: > click on "align" (at this point the plane become hidden, is this correct?);

No it should behave the same as clicking on this button on the old clip 
image


i.e. Which is nothing until the user clicks on a surface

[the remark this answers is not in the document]
The-Daniel: This might be an issue for @sebjf as all I'm doing is calling `clipToolRealign()`. I'm not doing anything in the frontend to hide the gizmo gui here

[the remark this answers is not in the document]
carmenfan: I wouldn't worry about this.

But it would be good if we make the align button a toggle. which toggles off when the user clicks on a surface (maybe via the point info callback? https://github.com/3drepo/3drepo.io/blob/master/frontend/src/globals/unity-util.ts#L496)

Ideally we should have instruction displayed too until it's done (bit like the pin component on tickets) to tell the user to select a surface to re-align the sectioning), again should disappear when the user clicked on a surface (may want to consult @apeitup for this)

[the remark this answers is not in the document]
carmenfan: Can we make the left panel end at the red line?

image

carmenfan:
Changing the merge target to staging.

I've pulled the latest of #4477 in here now, and there isn't anything else that needs changing there.

frontend/assets/icons/viewer/align.svg.tsx
carmenfan: > Agreed the disable look isn't prominent enough. we should look into making it more obvious

@The-Daniel don't think there was any changes for this?

carmenfan: @The-Daniel  should not automatically enable the edit mode. It should just set the tool to be the state where there;s a clipping plane, but not in edit mode (i.e. same as you've set the clip, but clicked on the main clipping button to disable the editing)

frontend/assets/icons/viewer/align.svg.tsx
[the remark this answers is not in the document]
carmenfan: > But it would be good if we make the align button a toggle. which toggles off when the user clicks on a surface (maybe via the point info callback? https://github.com/3drepo/3drepo.io/blob/master/frontend/src/globals/unity-util.ts#L496)

@The-Daniel  this one is also unresolved.

carmenfan (changes requested):
@The-Daniel i've tagged you on the unresolved issues from the previous reviews

frontend/assets/icons/viewer/align.svg.tsx
The-Daniel: I think I might need @sebjf for this. I might need one or two UnityUtil tools.

Firstly I'm not sure there's currently a way to cancel the UnityUtil.clipToolRealign() function which would be used when clicking the icon again.

Secondly I might need something for when a point is clicked that I can listen to so that I can disable the selected state and listener

carmenfan: as mentioend in my comment: https://github.com/3drepo/3drepo.io/pull/4666#discussion_r1443176110

You should be able to untoggle it as soon as you detect a currentPointInfo callback, as it only happens when someone clicks. (unless we surppress it during that mode)

I don't think we need a cancel button, the user can just click - they never had it and it's triggered by their actions.

@sebjf  if you disagree let me know

Amantini1997 (changes requested):

frontend/src/v5/ui/routes/viewer/toolbar/selectionToolbar/selectionToolbar.component.tsx
```diff
@@ -14,27 +14,86 @@
...
+					selected={alignActive}
+					title={formatMessage({ id: 'viewer.toolbar.icon.alignToSurface', defaultMessage: 'Align To Surface' })}
+				/>
+				<ToolbarButton
+					Icon={ClipSelectionIcon}
+					hidden={!clippingSectionOpen}
+					onClick={Viewer.clipToolClipToSelection}
+					title={formatMessage({ id: 'viewer.toolbar.icon.clipToSelection', defaultMessage: 'Clip To Selection' })}
+					disabled={!hasHighlightedObjects}
+				/>
```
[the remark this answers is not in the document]
The-Daniel: No, I already asked this. We want to show that it is a possibility. Hovering over shows what it is called and suggests that the user needs to make a selection before they can use it

frontend/src/v4/services/viewer/viewer.tsx
```diff
@@ -1242,18 +1242,51 @@ export class ViewerService {
 
 	public startBoxClip() {
 		UnityUtil.startBoxClip();
+		this.emit(VIEWER_EVENTS.UPDATE_CLIP_MODE, 'BOX');
```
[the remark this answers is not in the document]
The-Daniel: There are some places where the clip mode gets changed that aren't triggered in the frontend. For instance when selecting a view there is nothing in the frontend that actually sets the mode.

Sure, you could check the clipping planes and set the mode with that, but then if there are similar other ways to change the clipping mode that I've missed then they won't be accounted for at the moment, and you'll need to manually set the mode each time. This is my attempt of having the viewer state be the one source of truth

frontend/src/v4/modules/viewerGui/viewerGui.redux.ts
```diff
@@ -44,11 +45,14 @@ export const { Types: ViewerGuiTypes, Creators: ViewerGuiActions } = createActio
...
+	updateClipModeSuccess: ['mode'],
 	setClipEdit: ['isClipEdit'],
-	setClipEditSuccess: ['isClipEdit'],
+	updateClipEditSuccess: ['isClipEdit'],
 	setClipNumber: ['clipNumber'],
-	updateClipState: ['clipNumber'],
+	setGizmoMode: ['mode'],
+	setGizmoModeSuccess: ['mode'],
+	updateClipMode: ['clipMode'],
+	updateClipEdit: ['clipEdit'],
```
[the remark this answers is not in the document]
The-Daniel: The update methods just update the frontend state whereas the set methods change the state of the actual viewer. This was done because I needed away to listen to the viewer and update the frontend state accordingly.

Nevermind, I just tried again and found a way of making it work with just one method for each

frontend/src/v4/modules/viewerGui/viewerGui.sagas.ts
```diff
@@ -311,21 +316,37 @@ function* setClippingMode({mode}) {
 		if (mode) {
 			const isSingle = mode === VIEWER_CLIP_MODES.SINGLE;
 			yield Viewer.startClip(isSingle);
+			yield put(ViewerGuiActions.setGizmoMode(VIEWER_GIZMO_MODES.TRANSLATE));
```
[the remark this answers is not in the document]
The-Daniel: Yeah. There was a bug where if you're in a cube clip with the resize gizmo and then change to a viewpoint with a single clip the gizmo mode wouldn't change, so you have a single clip with the resize gizmo which shouldn't be allowed. I could add a more specific check that only resets it if this event occurs perhaps. What do you think @carmenfan

The-Daniel:
I think I've found a bug for @sebjf (not sure if it should be made into a new issue)

When you select 'Align to surface' and then click the background it cancels the 'align' tool, but it also resets the clipping tool.
image
image

carmenfan:
> I think I've found a bug for @sebjf (not sure if it should be made into a new issue)
> 
> When you select 'Align to surface' and then click the background it cancels the 'align' tool, but it also resets the clipping tool. image image

This is the same behaviour as the old clip plane. I wouldn't consider this as a bug (at least until we've had customers played around with it and got feedback)

frontend/assets/icons/viewer/align.svg.tsx
carmenfan: @The-Daniel  this was not addressed

> @The-Daniel should not automatically enable the edit mode. It should just set the tool to be the state where there;s a clipping plane, but not in edit mode (i.e. same as you've set the clip, but clicked on the main clipping button to disable the editing)

frontend/src/v4/modules/viewerGui/viewerGui.sagas.ts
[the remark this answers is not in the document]
carmenfan: I think the clip tool should reset (which is what it's doing now), this is fine.

It'll be less confusing if this been addressed:
https://github.com/3drepo/3drepo.io/pull/4666#discussion_r1448696879

carmenfan (changes requested):
One remaining comment unaddressed (tagged you)

frontend/assets/icons/viewer/align.svg.tsx
The-Daniel: Oops, I did fix that at some point but I think my other changes must have reverted it

carmenfan (changes requested):
@The-Daniel i don't know if we're just not understanding each other or what, but it's still happening 😆 

So i'm just going to repeat here with more details:

If i click on a view with a clipping plane, it's automatically going into edit mode:


Whereas on staging, I don't see this:


When a user clicks on a view that imposes/changes the clipping plane, it should impose the clip with the **toggle set to off** i.e. the geometry should be clipped, but I don't want to see the gizmo or the sub toolbar.

The-Daniel:
@carmenfan Sorry! I had stupidly forgotten to push my changes! There might be 1 small change I have to make that Santiago has just pointed out to me. But I'll make sure I have pushed before I request a review again

carmenfan (dismissed):

sanmont3drepo (approved):

Amantini1997 (dismissed):

sanmont3drepo (approved):

PR #4668, opened 2024-01-04
ISSUE #4621 - leaving a ticket saves unsaved changes

### LINKED ISSUE #4621: Tickets Toggle property's onBlur can cause 'INVALID_ARGUMENTS' error
### Description
The onBlur for a toggle property gets called as the user clicks away from the toggle (this prevents API spam if the user were to rapidly flick the toggle on and off). If the user changes a different property value with this 'click away'<sup>*</sup> then the toggle's onBlur will get called at the same time as this other property's. This results in the change of toggle value being posted twice giving an INVALID_ARGUMENTS error (No valid properties to update).

A similar thing seems to happen if user clicks a toggle an even number of times (so the value is unchanged) and then  cycles between tickets (it doesn't seem to happen when returning to list view)

\* By clicking the 'Set current view' button on a viewpoint, or clicking delete on a pin.

- This is happening in: both 

### Steps to replicate
- Open a ticket with both a toggle and a viewpoint
- Have the ticket in a state where both the ticket and viewpoint are unset (delete both camera and image).
- Change the toggle from off to on
- Without clicking anywhere else so as to not trigger the onBlur click the 'Save current view' teal tick button

### Current Behaviour
- An onBlur gets called twice with the same payload
- An error modal appears saying the arguments are invalid

### Expected Behaviour
- All properties successfully change
- No error modal

1) image
2) image
3) image
Payload 1 (200)- image
Payload 2 (ERROR 400) -  image

Amantini1997:
This fixes #4621

#### Description
Imagine this scenario:
- you edit the title of ticket;
- as the ticket is still focused, you change the ticket using the arrows

The ticket data is updated, but onBlur has not been called yet. By the time onBlur is called, the data is compromised as it refers to the new ticket, so the saving actions is triggered against the new ticket.

What I have done is to make sure that, before changing (or closing by any means) the current ticket, the data is saved

#### Test cases
1. Edit the title -> as it is still focused, change ticket;
2. Edit the title -> as it is still focused, go back to the ticket list;
3. Edit the title -> as it is still focused, close the ticket card.

Repeat using other inputs.

The ticket should always be saved correctly. In case of `scenario 1`, make sure the next ticket is not overridden

### DIFF
--- a/frontend/src/v5/ui/routes/dashboard/projects/tickets/ticketsList/slides/ticketSlide.component.tsx
+++ b/frontend/src/v5/ui/routes/dashboard/projects/tickets/ticketsList/slides/ticketSlide.component.tsx
@@ -48,9 +48,6 @@ export const TicketSlide = ({ template, ticketId }: TicketSlideProps) => {
 	});
 
 	const onBlurHandler = async () => {
-		const dirtyFields = formData.formState.dirtyFields;
-		if (isEmpty(dirtyFields)) return;
-
 		const formValues = formData.getValues();
 		let errors = {};
 		try {
@@ -61,7 +58,7 @@ export const TicketSlide = ({ template, ticketId }: TicketSlideProps) => {
 		} catch (yupError) {
 			(yupError?.inner || []).forEach(({ path, message }) => set(errors, path, { message }));
 		}
-		const values = dirtyValues(formValues, dirtyFields);
+		const values = dirtyValues(formValues, formData.formState.dirtyFields);
 		const validVals = removeEmptyObjects(nullifyEmptyObjects(filterErrors(values, errors)));
 		sanitizeViewVals(validVals, ticket, template);
 		if (isEmpty(validVals)) return;
--- a/frontend/src/v5/ui/routes/dashboard/projects/tickets/ticketsList/slides/ticketSlide.component.tsx
+++ b/frontend/src/v5/ui/routes/dashboard/projects/tickets/ticketsList/slides/ticketSlide.component.tsx
@@ -26,7 +26,7 @@ import { getValidators } from '@/v5/store/tickets/tickets.validators';
 import { DashboardTicketsParams } from '@/v5/ui/routes/routes.constants';
 import { TicketForm } from '@/v5/ui/routes/viewer/tickets/ticketsForm/ticketForm.component';
 import { yupResolver } from '@hookform/resolvers/yup';
-import { isEmpty } from 'lodash';
+import { isEmpty, set } from 'lodash';
 import { useEffect } from 'react';
 import { FormProvider, useForm } from 'react-hook-form';
 import { useParams } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


frontend/src/v5/ui/routes/dashboard/projects/tickets/ticketsList/slides/ticketSlide.component.tsx
```diff
@@ -39,16 +39,27 @@ export const TicketSlide = ({ template, ticketId }: TicketSlideProps) => {
...
+		try {
+			// cannot use formState.errors because the validation for complex objects is completed after
+			// onBlur gets called, so formState.errors for those objects is updated to the previous
+			// onBlur call instead and might claim there are no errors when it's not the case
+			await templateValidationSchema.validateSync(formValues, { abortEarly: false });
+		} catch (yupError) {
+			(yupError?.inner || []).forEach(({ path, message }) => set(errors, path, { message }));
+		}
+		const values = dirtyValues(formValues, formData.formState.dirtyFields);
+		const validVals = removeEmptyObjects(nullifyEmptyObjects(filterErrors(values, errors)));
```
The-Daniel, line 62: Is the `removeEmptyObjects` here doing anything? Surely there aren't any empty values after `nullifyEmptyObjects` is called.

As a side note (I know it's not part of this issue or anything you've done), but I don't think those two functions have accurate names if I'm understanding them correctly.

The-Daniel (changes requested):

frontend/src/v5/ui/routes/dashboard/projects/tickets/ticketsList/slides/ticketSlide.component.tsx
Amantini1997: > Is the `removeEmptyObjects` here doing anything? Surely there aren't any empty values after `nullifyEmptyObjects` is called.

I feel you are right, I checked and tested the app a bit to find out a possible scenario where `removeEmptyObjects` is being useful, but I couldn't find anything. Let's keep this as is for now and let's ask Santiago when he gets back as he's the one who added it

> As a side note (I know it's not part of this issue or anything you've done), but I don't think those two functions have accurate names if I'm understanding them correctly.

Why not? `nullifyEmptyObjects` claims to convert `{}` to `null`, whereas `removeEmptyObjects` claims to remove them, and so they do
image

The-Daniel: Ah, if that's the case then the description provided above `nullifyEmptyObjects` is wrong
image

The-Daniel: And either way, one acts on empty strings and the other doesn't although they are similarly named?

Amantini1997: The description looks right to me, but I guess you are right, it could be renamed to make it clear that empty strings are treated as empty objects. Let's wait for Santiago to see if we still need the `removeEmptyObjcets` first, and then we deal with namings as well, ok?

The-Daniel (approved):

The-Daniel (approved):

PR #4679, opened 2024-01-10
Issue 4636 - Toolbar: transform reset

### LINKED ISSUE #4636: Toolbar Transform reset
### Description
It's possible for us to move geometry via a sequence and even save it as part of a issues/risks viewpoint, you can find a sample model with such sequence here https://issue-4608.dev.3drepo.io/v5/viewer/DevTeamWorkspace/72d935e5-d15f-4bc4-8433-fb6d79d57301/bc8f82ef-db39-4cb0-a4f9-9bb47aa14dc9
img-04-36-23-06-36

image

At the moment besides manipulating the sequence card or imposing a different view, there's no way for us to reset the transform, so it would be good if we can add a button for that like we do for colour overrides etc.

@apeitup  has designed an icon for this : [transform-reset-o.svg.zip](https://asitesol-my.sharepoint.com/:u:/g/personal/thalliday_asite_com/EWwQRgJhXltPvsugh82ZYIMBs_sopUAVGMbCdvRDv4jCRw?e=c01KJu)
Screenshot 2023-12-12 at 14 30 52

### Task

Identify a way to determine whether some geometry is hidden and if so, add a section in the `SectionToolbar` with a button ("Reset Transformation"  or something similar).

Branch off from #4608

Amantini1997:
This fixes #4636 

#### Description
Added the logic to clear transformations (not just from sequences)

#### Test cases
- Select a view, open an issue/risk/ticket with default view, or select the date of a sequence that has transformations: The "reset transformations" buttons should show up in the toolbar;
- Click that button: the transformations should reset and the button should disappear


carmenfan (approved):

carmenfan (approved):

The-Daniel (dismissed):

The-Daniel (approved):

carmenfan (approved):

frontend/src/v4/routes/viewerGui/components/sequences/sequences.component.tsx
```diff
@@ -126,21 +122,6 @@ export class Sequences extends PureComponent<IProps, {}> {
 		return <SequencesIcon />;
 	}
 
-	public onPlayStarted = () => {
```
sanmont3drepo, line 130: Why did you get rid of this?

Amantini1997: Because that logic was being called only when the play started, which was not quite accurate.
Imagine this scenario: you have just selected a view that has some transformations and some colour overrides. You then open a sequence and move the slider up and down. The colour overrides from the sequence are visible, however, the transformations and the transparencies don't, at least, until you hit play the first time.

That's not correct, as every time you change the sequence using the slider, you should see only the current view from the sequence. So I moved it in `sequences.sagas.tsx`, inside  `updateSelectedStateDefinition`. But what's really needed, is to clear the viewpoint, so what I put there is simply
```ts
if (!viewpoint) {
	yield put(ViewpointsActions.setActiveViewpoint(null));
}
```

sanmont3drepo: Its a different code is `deselectViewsAndLeaveClipping` I dont remember why it was using that (you'll need to check the  commits) but there was a reason for it

sanmont3drepo (changes requested):

frontend/src/v4/routes/viewerGui/components/sequences/sequences.component.tsx
sanmont3drepo: Its a different code is deselectViewsAndLeaveClipping I dont remember why it was using that (you'll need to check the commits) but there was a reason for it

Amantini1997: It was introduced here https://github.com/3drepo/3drepo.io/pull/2247/commits/70bfc4bd1e7ebea5f6da68af48a63c1dd9342885 for this reason: image

It seems correct then, what do you think?

sanmont3drepo: Yeah but take a look that it says leave clipping while the other one wont

Amantini1997: Yes, but resetting the viewpoint that way doesn't get rid of the clipping planes, so the final outcome in that sense is the same

The-Daniel (dismissed):

carmenfan (dismissed):

frontend/src/v4/modules/sequences/sequences.sagas.tsx
```diff
@@ -35,6 +35,7 @@ import {
 	selectActivitiesDefinitions, selectFrames, selectNextKeyFramesDates, selectSelectedDate, selectSelectedFrameViewpoint,
```
sanmont3drepo, line 35: can we change the extension back to ".ts"?

sanmont3drepo (changes requested):

The-Daniel (approved):

PR #4680, opened 2024-01-10
Issue 4663 - Due Date can overflow

### LINKED ISSUE #4663: Due date can overflow in ticket details
### Description

Due date text overflows in ticket details if assignees are set. This bug is a result of dates now including times

Revised mock up can be found here: https://www.figma.com/file/bRqIl0g9psqD3ISkmxwcNp/3D-Repo?node-id=4265%3A70717&mode=dev

### Steps to replicate
- Open a ticket which has issue properties
- Set the due date
- Assign 3+ users/jobs

### Current Behaviour
- The due date will spill over 2 lines crossing over the horizontal rule

### Expected Behaviour
- The due date does not overlap other elements

image

Amantini1997:
This fixes #4663 

#### Description
Since the addition of the time to date across the app, due date was sometimes breaking in two lines.
This was fixed in the standard ticket using a new design and in tabular view allowing more room.
Note that "created at" in tabular view was also overflowing sometimes, so this was also addressed.

#### Test cases
- Play around with the new ticket-issue-properties design.
- In tabular view, change the due date (and possibly the "created at" using the inspector) to see if it can overflow


carmenfan (changes requested):
image
Unless @apeitup  severely rejects to it, can we move this to a new line and aligned to the left? Essentially pretty much the same design as the standard ManyOf with Jobs and Users as the selection

It's very out of place as none of our property design have the data on the same line as the title


Also, is it just me do these 2 have different formatting?
image

The-Daniel (approved):

carmenfan (changes requested):
> Also, is it just me do these 2 have different formatting? image

@Amantini1997  still looks off to me, did we verify if the formatting is the same?

image

PR #4682, opened 2024-01-11
ISSUE #4664 - User can lose form data if they close tickets card whilst creating a new ticket

### LINKED ISSUE #4664: User can lose form data if they close tickets card whilst creating a new ticket
### Description

**This depends on #4563**

With recent ticket changes the user can close and reopen the tickets card and not lose the view they're on. This might make some users think that closing the card when creating a new ticket will retain all data but it doesn't.

There is an additional bug that if the user creates a pin and selects it, when they close and reopen the card the pin input will show selected but there is no pin set.

This was highlighted [here](https://github.com/3drepo/3drepo.io/pull/4653#discussion_r1425528329) 

### Steps to replicate
- Try to create a new ticket (fill in data)
- close the ticket form
- reopen the ticket form

### Current Behaviour
- The data is lost.

### Expected Behaviour
- <del> Perhaps a warning should be given that data will get lost if the card is closed </del>
- Or perhaps we can find a way to maintain the form data

image
image
(notice the 'selected' pin)

Amantini1997:
This fixes #4664

#### Description
When exiting a new ticket, if it is not because the user is returning to the tickets list, the data is stored in redux and re-used upon opening the ticket

#### Test cases
Edit a new ticket, then:

##### Data persistence scenario
1. click the ticket icon twice (to close and re-open the tickets card)

##### Data loss scenarios
1. click the back arrow and then try to create a new ticket with any templates;
2. change con/fed and then try to create a new ticket with any templates;


frontend/src/v5/store/tickets/card/ticketsCard.redux.ts
```diff
@@ -59,6 +61,7 @@ export const INITIAL_STATE: ITicketsCardState = {
 	view: TicketsCardViews.List,
 	overrides: null,
 	readOnly: false,
+	unsavedTicket: null,
```
The-Daniel, line 64: You need to revert to this state after a ticket is created otherwise if the user tries to create another ticket of the same template directly afterwards all the fields will be prefilled.

The-Daniel (changes requested):

frontend/src/v5/store/tickets/card/ticketsCard.redux.ts
Amantini1997: I do at [newTicket.component.tsx](https://github.com/3drepo/3drepo.io/pull/4682/files#diff-f102d917f230d34018d1a0f60090970fb2427e2c74401ce19011e8f4411169dd), unless I misunderstood what you're saying. In that case, could you add the reproduction steps?

The-Daniel (dismissed):

frontend/src/v5/store/tickets/card/ticketsCard.redux.ts
```diff
@@ -45,6 +46,7 @@ export interface ITicketsCardState {
 	view: TicketsCardViews,
 	readOnly: boolean,
 	overrides: OverridesDicts | null,
+	unsavedTicket: IUnsavedTicket | null,
```
sanmont3drepo, line 49: is **IUnsavedTicket** type necessary? cant we we just blank the ticket when is saved and use the passed **templateId** and **containerOrFederation** for saving the ticket? I find that duplicating known data always leads to errors in a near future

frontend/src/v5/ui/routes/viewer/tickets/newTicket/newTicket.component.tsx
```diff
@@ -30,23 +30,33 @@ import { yupResolver } from '@hookform/resolvers/yup';
...
+	const templateId = template._id;
 
-	const defaultTicket = getDefaultTicket(template);
+	let defaultTicket = getDefaultTicket(template);
+	if (
+		unsavedTicket?.containerOrFederation === containerOrFederation
+		&& unsavedTicket?.templateId === templateId
+	) {
+		defaultTicket = merge(defaultTicket, unsavedTicket.data);
+	}
```
sanmont3drepo, line 59: The data still persists after I explicitly cancelled the edition

### Steps to replicate
- I fill in a form to create a ticket
- I click on X
- I click on create with the same template as before

### Current Behaviour
- The previously filled ticket appears again

### Expected Behaviour
- The form should get blanked

over presistence

sanmont3drepo (changes requested):

frontend/src/v5/store/tickets/card/ticketsCard.redux.ts
Amantini1997: I'm not sure I understand what you mean

sanmont3drepo: You could just use an unsavedTicket: ITicket | null. Why do you have to keep the templateId and containerOrFederation in unSavedTicket?

sanmont3drepo (changes requested):

sanmont3drepo (approved):

The-Daniel (approved):

sanmont3drepo (approved):

PR #4687, opened 2024-01-15
ISSUE 4659 - Support Model Caching

### LINKED ISSUE #4659: ModelCacher should manage the web requests
### Description

This is the .io counterpart of https://github.com/3drepo/3drepounity/issues/463

The plan is that in the browser, web requests will be handled entirely in JavaScript, and the responses passed to unity as close to complete as possible (i.e. byte[] arrays for AssetBundle requests, and strings for Json requests).

To do this, the ModelCacher will be re-written to offload more functionality into JavaScript. A class in addition to the IndexedDbCache should be added to interface with the ModelCacher, which will turn into a general request manager.

### Goals

- [x] The viewer issues all web requests through our own web request handler
- [x] The handler can use IndexedDb as a cache
- [x] The cache is configurable

### Tasks

- [x] Write a counterpart to the Unity WebRequestHandler2 that can return fetch responses to Unity
- [x] Update IndexedDbCache to use a Promise-based API
- [x] Update WebRequestHandler2 to use the cache 
- [x] Create all the endpoints to configure the request manager properly
- [x] Add logic to control which calls are cached
- [x] Make sure that ArrayBuffers are moved to the cache worker as [Transferable Objects](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects)

sebjf:
This fixes #4659 

#### Description
This PR implements the support for WebRequestHandler2, a Unity Component that offloads Get requests to the browser for use by BundleLoader.

This PR adds a new class to the Unity Utils set that acts as a direct counterpart to new handler, and also refactors IndexedDbCache to make it usable from JavaScript instead of Unity. This is done by wrapping the message based API with Promises.

Finally, this PR changes the use of `setAPIHost` to match what is described by TypeScript.

**This PR has been created in order to get a deployment for testing and should not yet be merged.**


carmenfan (changes requested):
rewrsdfdsf

frontend/src/globals/unity-webrequesthandler2.ts
```diff
@@ -0,0 +1,210 @@
...
+				new Int8Array(this.requests[id].data), // Specify the Type to interpet the ArrayBuffer as when copying
+
+				// This 'shift' converts offset from a signed to unsigned
+				// integer, which is necessary to support the WASM 4GB address
+				// space https://v8.dev/blog/4gb-wasm-memory
+				// eslint-disable-next-line no-bitwise
+				offset >>> 0,
+			);
+		} catch (error) {
+			console.log(error);
```
carmenfan, line 179:
```suggestion
			console.error(error);
```

frontend/src/globals/unity-webrequesthandler2.ts
```diff
@@ -0,0 +1,210 @@
...
+		}
+	}
+
+	/** Converts the data returned from a get request into a string and returns
+	 * it. The plugin will allocate an unmanaged buffer for the string and copy
+	 * it into it.
+	 */
+	getResponseString(id: number): string {
+		const body = new TextDecoder().decode(this.requests[id].data);
```
carmenfan, line 188: Tiny thing, but we can just return since we don't use the body here.

```suggestion
return new TextDecoder().decode(this.requests[id].data);
```

carmenfan: We're sure that there's no situation where we'd go index out of bounds?

Just quickly checked the unity code we also don't catch any error.

If we're confident it's all good, just playing devil's advocate here!

carmenfan (changes requested):
Some minor comments

frontend/src/globals/unity-webrequesthandler2.ts
```diff
@@ -0,0 +1,210 @@
...
+	/** Initialises this handler for an instance of Unity. Should be called by
+	 * the viewer once it is fully initialised. */
+	setUnityInstance(unityInstance: any, gameObjectName: string) {
+		this.unityInstance = unityInstance;
+		this.gameObjectName = gameObjectName;
+	}
+
+	setAPIHost(hostNames: [string]) { // An array for backwards compatability reasons
+		// eslint-disable-next-line prefer-destructuring
+		this.apiHost = hostNames[0];
```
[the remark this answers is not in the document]
sebjf: That worked!

frontend/src/globals/unity-webrequesthandler2.ts
sebjf: Hi @carmenfan, 

Not that I can think of - the member is set in the `createGetRequest` method, which is synchronous, and only deleted by `releaseRequest`, which is also a synchronous call from Unity (i.e. the js side doesn't get any say over what this is or when its updated). Additionally, the control flow is implemented as a co-routine, which only suspends execution, between frames, so there is no way to change the order of operations.

My thinking is more towards the fail fast, fail early approach for this: if we ever went out of bounds here it would not only be an unrecoverable error, but might also call into question whether we've actually been returning the right data for other requests, so probably not something we want to try and recover silently from (and that's if corrupted data hasn't already resulted in a load of failures elsewhere).

sebjf:
Hi @carmenfan, this is ready for another look!

carmenfan (approved):
Looks good.

As discussed, we should look into caching the jsons as well after #4734 where the URL will show clearly whether a model is a federation or container.

PR #4704, opened 2024-01-19
ISSUE #4699 - Frontend doesn't read ifc/revit ids in groups

### LINKED ISSUE #4699: Frontend doesn't read ifc/revit ids in groups
### Description
When the frontend receives a chat event that a manual group has been updated, if the revision is IFC or Revit type, the objects ids are missing from the update group in the chat data and the `ifc_guids` or revit equivalent are received instead. 

### Steps to replicate
- Open an IFC revision;
- Create a ticket with a view;
- Inside the view create a manual group;
- Update the manual group by including another mesh

### Current Behaviour
- The group count says "No objects"

### Expected Behaviour
- The group count should update correctly

### Suggested approach
The chat event listener should send a GET request to the backend and fetch the object ids that can be read by the frontend

Amantini1997:
This fixes #4699 

#### Description
When the frontend receives a chat event that a ticket was updated, if the objects does not have standard _ids (i.e. revit or ifc ids) refetch the group

#### Test cases
Create a container with an ifc (or revit) revision.
Create a manual group and then update it.
The object count should be correct.

Repeat with a federation that contains the aforementioned container

### DIFF
--- a/frontend/src/v5/services/realtime/ticket.events.ts
+++ b/frontend/src/v5/services/realtime/ticket.events.ts
@@ -22,6 +22,7 @@ import { getMeshIDsByQuery } from '@/v4/services/api';
 import { meshObjectsToV5GroupNode } from '@/v5/helpers/viewpoint.helpers';
 import { subscribeToRoomEvent } from './realtime.service';
 import { TicketsActionsDispatchers } from '../actionsDispatchers';
+import { fetchTicketGroup } from '../api/tickets';
 
 // Container ticket
 export const enableRealtimeContainerUpdateTicket = (teamspace: string, project: string, containerId: string) => (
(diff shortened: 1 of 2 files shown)


The-Daniel (approved):

sanmont3drepo (changes requested):
Is this a bug or a quirk of the model @carmenfan ?
If I select the floor when it saves the group I get all meshes

selects all

frontend/src/v5/services/realtime/ticket.events.ts
```diff
@@ -22,6 +22,7 @@ import { getMeshIDsByQuery } from '@/v4/services/api';
 import { meshObjectsToV5GroupNode } from '@/v5/helpers/viewpoint.helpers';
 import { subscribeToRoomEvent } from './realtime.service';
 import { TicketsActionsDispatchers } from '../actionsDispatchers';
+import { fetchTicketGroup } from '../api/tickets';
```
sanmont3drepo, line 25: Is this a bug or a quirk of the model @carmenfan ?
If I select the floor when it saves the group I get all meshes

selects all

sanmont3drepo (changes requested):

frontend/src/v5/services/realtime/ticket.events.ts
Amantini1997: I had the same issue. It's a problem specific to this model apparently 
image

carmenfan: as @Amantini1997  said.

Use something in the building

sanmont3drepo (approved):

carmenfan (approved):

Issue #4710, opened 2024-01-19
Meshes of an edited ticket group are not being unhighlighted with the toolbar 'Clear Overrides' button


The-Daniel:
### Description
After editing a group the 'Clear overrides' toolbar option does not work on the meshes highlighted by the edited group.
Therefore it is possible to have meshes highlighted and not have the toolbar options to clear overrides by editing a ticket's groups.

This only happens on custom tickets groups, not v4 groups.

- This is happening in: both 
- This requires a hotfix to production: no 

### Steps to replicate
- Open a ticket that has a view.
- Create a group with objects. (Can be smart or manual, but manual is easier)
- Click edit group
- Cancel the edit
![image](https://github.com/3drepo/3drepo.io/assets/62481250/112e258a-2a9d-4695-94de-bbe3340e2e73)
- Click 'Clear overrides in the toolbar
- The meshes of the edited group are not cleared and the toolbar option to clear overrides disappears. The group is 'unticked' in the group list
![image](https://github.com/3drepo/3drepo.io/assets/62481250/bd29dc06-c31a-488a-b72d-cb4a7bf6ad95)

### Current Behaviour
- The meshes of edited groups are not unhighlighted when clicking 'clear overrides'

### Expected Behaviour
- All highlighted meshes are cleared

carmenfan:
Pretty sure we've been through this previously.

Highlight and colour override are 2 different concepts. If you're highlighting something you have the object selected - you click away, the highlight disappears. It's a bit like in word, highlighting a bunch of text but selecting it with your mouse so it's rendered in a different colour with background

Colour override is imposing a colour on an object. if you click away, it stays, it is not highlighted (i..e if it's obscured by something else, you won't see the object) - think actually imposing a colour/colour background on text.

Unless i've mistaken, this is not a bug, so closing. Reopen with more details if I understood this incorrectly!

The-Daniel:
> Pretty sure we've been through this previously.
> 
> Highlight and colour override are 2 different concepts. If you're highlighting something you have the object selected - you click away, the highlight disappears. It's a bit like in word, highlighting a bunch of text but selecting it with your mouse so it's rendered in a different colour with background
> 
> Colour override is imposing a colour on an object. if you click away, it stays, it is not highlighted (i..e if it's obscured by something else, you won't see the object) - think actually imposing a colour/colour background on text.
> 
> Unless i've mistaken, this is not a bug, so closing. Reopen with more details if I understood this incorrectly!

[...]

PR #4718, opened 2024-01-23
Issue 4712 - ticket groups bugs

### LINKED ISSUE #4712: Ticket Groups: bugs
### Bugs
1. When all the meshes are hidden, save to current view creates a manual coloured group with no objects
2. On a new ticket, calling "save to current view" creates some groups cannot be edited. Editing them creates a different group instead

### Description
- These are happening in: both
- These require a hotfix to production: no

### Steps to replicate
- In the viewer, open a new ticket;
- Hide all the meshes;
- Save to current view;
> Bug 1: a colored **manual** group is created with no objects

- Try to edit either the hidden or the colored group
> Bug 2: instead of editing the group, a new one is created

https://github.com/3drepo/3drepo.io/assets/43547703/e59c69b0-0a87-4cd3-95e5-4c7d7ac35fb3

Amantini1997:
This fixes #4712

#### Description
- The highlighted objects can be hidden. If that's the case, the equivalent group is not created on save/update current view
- Groups via save/update current view in a new ticket were missing the key

#### Test cases
Click on "new ticket" and select a template with view and:
1. Highlight a mesh
2. Hide that mesh
3. Save the current view
> No error should be triggered
4. Open the groups (there should be a hidden group)
5. Edit the hidden group
> The group should be edited correctly and no other groups should be created


frontend/src/v5/helpers/viewpoint.helpers.ts
The-Daniel, line 1: I think this might be the same bug as #4710
Sometimes when clicking 'edit' on a hidden group it is saying there are 0 objects when you can see in the viewer there are some highlighted objects. Sometimes the number is accurate. I'm not entirely sure how to reproduce every time
image

The-Daniel, line 1: Sometimes just making the form dirty validates it.
image
image

The-Daniel (changes requested):
I know a lot of these are unrelated to the original issue, but since they fit in to the vague PR title I thought I'd post here. You might want to create new issues for them though

frontend/src/v5/helpers/viewpoint.helpers.ts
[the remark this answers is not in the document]
Amantini1997: I noticed it this morning, but I was only able to reproduce it on a particular model, so I thought the problem was in the model. Btw, as I was debugging it, I added a breakpoint right before the line making the API call (and so slowing down the flow in the frontend), the problem was not happening anymore. I suspect it could be a backend bug, but I am not sure

The-Daniel, line 1: I have a model with 8000 objects. When editing a group with many selected objects the object count doesn't get higher than 5000 even though the correct number appear highlighted in the viewer.
image

frontend/src/v5/helpers/viewpoint.helpers.ts
```diff
@@ -86,7 +86,10 @@ export const getViewerState = async () => {
 	}
 
 	if (viewpointV4.highlighted_group) {
-		state.colored = (state.colored || []).concat(convertToV5GroupOverride(viewpointV4.highlighted_group, ViewpointGroupOverrideType.COLORED));
+		state.colored = (state.colored || [])
+			.concat(convertToV5GroupOverride(viewpointV4.highlighted_group, ViewpointGroupOverrideType.COLORED))
+			// eslint-disable-next-line no-underscore-dangle
+			.filter(({ group }) => !!(group as Group).objects.filter((obj) => obj._ids.length).length);
```
sanmont3drepo, line 92: Why did you change this?

Amantini1997: To fix this bug
image

sanmont3drepo (approved):

sanmont3drepo (approved):

The-Daniel (approved):

PR #4727, opened 2024-01-23
ISSUE #4548 - Smart group value fields only get autofocus when adding a new field

### LINKED ISSUE #4548: Auto focus is going to the field value property on rules panel
### Description
in both v4 groups and v5 ticket groups, when you edit or create a new rule, the auto focus is going straight to the field value, which is very odd as the Name is the first item on the list and is a required field.

The same behaviour can be observed when you try to edit a rule

img-23-10-23-12-10
image

### Steps to replicate
- Either in v4 groups or v5 ticket groups, try to create/edit a rule in a smart group

### Current Behaviour
- the 2nd input box is currently on auto focus when the panel appears

### Expected Behaviour
- The first input box (i.e. Name) should be on auto focus

The-Daniel:
This fixes #4548

#### Description

When you create a filter for a smart group the autofocus is now on the name field
When adding a value field the input is focused

#### Test cases
Create a new smart group filter
- The autofocus should initially be on the name
- Adding new value fields should set the focus to the new field
Edit an existing smart group filter
- The autofocus should still be on the name no matter how many value fields exist
- Upon adding new value fields the focus switches to the new field input

### DIFF
--- a/frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketGroups/groups/groupRulesForm/groupRulesInputs/ruleFieldValues/fieldValueInput/fieldValueInput.component.tsx
+++ b/frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketGroups/groups/groupRulesForm/groupRulesInputs/ruleFieldValues/fieldValueInput/fieldValueInput.component.tsx
@@ -24,7 +24,7 @@ import { get } from 'lodash';
 import { ListboxComponent } from './listboxComponent/listboxComponent.component';
 import { Autocomplete } from './fieldValueInput.styles';
 
-export const FieldValueInput = ({ name, disabled }) => {
+export const FieldValueInput = ({ name, autoFocus = false, disabled }) => {
 	const { formState: { errors } } = useFormContext();
 	const fields = useSelector(selectMetaKeys);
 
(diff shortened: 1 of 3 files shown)


frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketGroups/groups/groupRulesForm/groupRulesInputs/ruleFieldValues/fieldValueInput/fieldValueInput.component.tsx
```diff
@@ -24,7 +24,7 @@ import { get } from 'lodash';
 import { ListboxComponent } from './listboxComponent/listboxComponent.component';
```
Amantini1997, line 24: In custom ticket groups:
- Open a smart group;
- Click on a filter and edit the title;
- Click on "update filter";
- Click on that filter again

The "update filter" button is enabled but it shouldn't

P.S. Something analogous happens with v4 groups (same steps), but that's already in staging, so not sure if you want to/can fix it here

Amantini1997 (changes requested):

sanmont3drepo (approved):

Amantini1997 (approved):

Issue #4716, opened 2024-01-22
Release Candidate v5.8


carmenfan:
QA branch for 5.8

PR #4738, opened 2024-01-31
Issue 4719 - Cannot delete ticket group description

### LINKED ISSUE #4719: Cannot delete the description from a ticket group
### Description

Although it is possible to create a ticket group with no description the user cannot edit an existing group and delete the description. An error modal appears stating that the description must be at least 1 character long

- This is happening in: both 

### Steps to replicate
- Create a ticket group and give it a description
- Edit the group and delete all characters from the description input
- Save the updated group

### Current Behaviour
- An error modal appears stating the description must be at least 1 char long
- The group is not saved

### Expected Behaviour
- The user can update a ticket group to have no description

image

Amantini1997:
This fixes #4719

#### Description
Empty ticket group description is being sent as `null`, which means we want to delete it

#### Test cases
Create a group with a description;
Edit the group deleting the description;
Save


backend/src/v5/schemas/tickets/tickets.groups.js
```diff
@@ -43,7 +43,7 @@ const objectEntryValidator = Yup.object().shape({
 Groups.schema = (allowIds, fieldsOptional) => {
 	let group = Yup.object({
 		name: types.strings.title,
-		description: types.strings.longDescription,
+		description: types.strings.longDescription.nullable(),
```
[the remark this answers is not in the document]
carmenfan: Also by the looks of the code, we're doing an update to `{$set: { description: null }}`. Which would be incorrect - we don't want to start storing nulls everywhere...

We want to `$unset` the description if it's null. so I think you need to change the logic on `models/tickets.groups` (and add tests)

sanmont3drepo (dismissed):

backend/src/v5/models/tickets.groups.js
```diff
@@ -37,15 +38,29 @@ Groups.getGroupsByIds = (teamspace, project, model, ticket, groupIds, projection
...
-	const updates = { $set: data };
+	const toUpdate = { };
+	const toUnset = { };
 
-	if (data.rules) {
-		updates.$unset = { objects: 1 };
-	} else if (data.objects) {
-		updates.$unset = { rules: 1 };
+	Object.keys(data).forEach((key) => {
+		if (data[key]) {
```
carmenfan, line 45: Can we specifically check for null please?

we just had a bug in #4716 that was because we're trying to set something to 0 :'(

So it'd be much easier to check for what we're actually after.

```suggestion
		if (data[key] !== null) {
```

backend/src/v5/models/tickets.groups.js
```diff
@@ -37,15 +38,29 @@ Groups.getGroupsByIds = (teamspace, project, model, ticket, groupIds, projection
...
+	});
+
+	if (toUpdate.rules) {
+		toUnset.objects = 1;
+	} else if (toUpdate.objects) {
+		toUnset.rules = 1;
 	}
+
+	const actions = {};
+	if (!isEqual(toUpdate, {})) actions.$set = toUpdate;
```
carmenfan, line 59: Probably more of a readibility thing here, but I think it's better to do `Object.keys(...).length`

carmenfan (changes requested):

PR #4743, opened 2024-02-01
Issue 4731 - Status should be shown at all times

### LINKED ISSUE #4731: Status should be shown at all times
### Description
With https://github.com/3drepo/3drepo.io/issues/4713, status is now a default property of all custom tickets.

However, it seems like the front end is relying on issue properties to render it instead of taking it from the property list. Therefore currently they are not showing:
image

We need all tickets to show status in accordance to the template. We would also like status to be moved to the left of the issue property strip so it's consistent across all tickets (with and without other issue properties). 

For details see scoping issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/331

### Goals
- [ ] As a user I want to status value of all tickets
- [ ] As a user I want to be able to set an ticket as closed or voided

### Tasks
- [ ] Change ui and move the status to the left in tickets in the details form
- [ ] Show the status in the list regardless that the ticket has issuproperties in config or not
- [ ] Make all tickets use the default property for status 

### Note
Ensure that the tickets uses the same status UI. Atm its probably checking that it has issue properties to render the pill in the list and the the details. so just make every ticket use that ui.

### Images
image

Amantini1997:
This fixes #4731 

#### Description
Templates get always fetched before tickets so that, when fetching the tickets, all the ones that are missing the status use the default one from the template

#### Test cases
Access old tickets that are not supposed to have a status and check that the default one from template is applied.
In the viewer:
- open the tickets list and expand one of those tickets
- open the viewer directly to that ticket expanded (using the search param)

In tabular view:
- access the table via ticket selection;
- access the table via refresh from the table view;
- expand a ticket from the table view


frontend/src/v5/store/tickets/tickets.sagas.ts
```diff
@@ -37,17 +37,23 @@ import {
...
 			: API.Tickets.fetchContainerTickets;
-		const tickets = yield fetchModelTickets(teamspace, projectId, modelId);
-		yield put(TicketsActions.fetchTicketsSuccess(modelId, tickets));
+		const tickets: ITicket[] = yield fetchModelTickets(teamspace, projectId, modelId);
+		const fullTickets = [];
+		for (const ticket of tickets) {
+			const template = yield select(selectTemplateById, modelId, ticket.type);
+			fullTickets.push(getTicketWithStatus(ticket, template));
+		}
+		yield put(TicketsActions.fetchTicketsSuccess(modelId, fullTickets));
```
sanmont3drepo, line 56: I think a better strategy is to (if it can be done) move the `getticketwithstatus` bit to ticket selectors. In 'selectTickets'. In that case  if the data changes  (which I agree can be very rarely) is always updated. it doesnt depend on this particular bit. It also makes it just one place to look for the processed data.

frontend/src/v5/ui/routes/dashboard/projects/tickets/ticketsTable/ticketsTable.component.tsx
```diff
@@ -200,6 +201,21 @@ export const TicketsTable = () => {
...
 
+	useEffect(() => {
+		ticketsFilteredByTemplate.forEach(({ modelId, modelName, ...ticket }) => {
+			if (!ticket.properties[BaseProperties.STATUS]) {
+				TicketsActionsDispatchers.updateTicket(
+					teamspace, 
+					project,
+					modelId,
+					ticket._id,
+					getTicketWithStatus(ticket, selectedTemplate),
```
sanmont3drepo, line 212: This is necessary because getticketWithStatus is done in the fetch instead of the selector.

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketsTopPanel/IssuePropertiesInputs/issuePropertiesInputs.component.tsx
```diff
@@ -23,95 +23,105 @@ import { InputController } from '@controls/inputs/inputController.component';
...
+				)}
+			</FlexContainer>
+			<Gap $height="10px" />
+			<AssigneesContainer>
+				<Typography variant="h5">
 					<FormattedMessage
-						id="ticketTopPanel.status.label"
-						defaultMessage="Status"
+						id="customTicket.topPanel.assignees.label"
+						defaultMessage="Assignees"
```
The-Daniel, line 111: Should Assignees not be in the hasIssueProperties condition?

The-Daniel (changes requested):

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketsTopPanel/IssuePropertiesInputs/issuePropertiesInputs.component.tsx
sanmont3drepo: No, because this PR changes the concept of status being a issue properties only into being a property for all tickets.

The-Daniel: Yes, but this isn't status it's assignees. This causes errors when I try to update the assignee value in the tickets table

The-Daniel (approved):

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketsTopPanel/ticketsTopPanel.component.tsx
```diff
@@ -94,7 +93,7 @@ export const TicketsTopPanel = ({
 					/>
 					{_.get(formState.errors, `properties.${BaseProperties.DESCRIPTION}`) && <ErrorTextGap />}
 				</DescriptionProperty>
-				{hasIssueProperties && <IssuePropertiesInputs onBlur={onPropertyBlur} readOnly={readOnly} />}
+				<IssuePropertiesInputs properties={properties} onBlur={onPropertyBlur} readOnly={readOnly} />
```
sanmont3drepo, line 96: Im not sure about the naming of `IssuePropertiesInputs` because now is not only for issue properties.
I see two options.
- rename it to another thing that is more representative (sorry I cant imagine of another name)
- refactor `Issuepropertisinputs` to reuse a `statusinput` component and keep the 'hasissueProperties' here to render that one.  and have a plain `statusPropertyBar` that would reuse `statusinput` ( `<FlexContainer> <StatusInput /> </FlexContainer>`  )  that would render if hasissueProperties is false

frontend/src/v5/store/tickets/tickets.types.ts
```diff
@@ -195,7 +195,7 @@ export type OverridesDicts = {
 	transparencies: MeshIdTransparencyDict,
 };
 
-export type TicketWithModelIdAndName = ITicket & { modelId: string; modelName: string };
+export type TicketWithModelId = ITicket & { modelId: string };
```
sanmont3drepo, line 198: you can change the original type of ITicket to have `modeld?: string`

sanmont3drepo (changes requested):

frontend/src/v5/store/tickets/tickets.selectors.ts
```diff
@@ -19,10 +19,20 @@ import { createSelector } from 'reselect';
...
+	if (ticket.properties[BaseProperties.STATUS] || !template) return ticket;
+	const statusProperty = template.properties?.find(({ name }) => name === BaseProperties.STATUS);
+	return {
+		...ticket,
+		properties: {
+			...ticket.properties,
+			[BaseProperties.STATUS]: statusProperty?.default,
+		},
+	};
+};
```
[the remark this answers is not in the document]
Amantini1997: Placing it there causes a circular dependency:
<code>src\v5\store\tickets\tickets.selectors.ts
&nbsp;     -> src\v5\store\tickets\tickets.helpers.ts
&nbsp;     -> src\v5\services\selectorsHooks.ts
</code>

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketsTopPanel/statusProperty/statusProperty.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import styled from 'styled-components';
+
+export const Container = styled.div`
+	.MuiChip-root {
+		padding: 0;
+	}
+`;
+
+export const Title = styled.div`
```
The-Daniel, line 26: Do you need to define this here when it is the same as PropertyTitle from issuePropertiesInputs.styles

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketsTopPanel/statusProperty/statusProperty.component.tsx
```diff
@@ -0,0 +1,50 @@
...
+		<Title>
+			<FormattedMessage
+				id="ticketTopPanel.status.label"
+				defaultMessage="Status"
+			/>
+		</Title>
+		<FormChipSelect
+			variant="text"
+			tooltip={formatMessage({
+				id: 'customTicket.topPanel.status.tooltip',
```
The-Daniel, line 40: This id is inconsistent with the id above it

The-Daniel (changes requested):

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketsTopPanel/ticketsTopPanel.component.tsx
```diff
@@ -28,9 +28,11 @@ import _ from 'lodash';
 import { BaseProperties, IssueProperties } from '../../tickets.constants';
 import { TitleProperty } from '../properties/titleProperty.component';
 import { PropertiesList } from '../propertiesList.component';
-import { IssuePropertiesInputs } from './IssuePropertiesInputs/issuePropertiesInputs.component';
-import { BaseTicketInfo, DescriptionProperty, TopPanel } from './ticketsTopPanel.styles';
+import { BaseTicketInfo, DescriptionProperty, TopPanel, FlexContainer } from './ticketsTopPanel.styles';
+import { IssuePropertiesInputs } from './issuePropertiesInputs/issuePropertiesInputs.component';
```
sanmont3drepo, line 32: I think the case of the directory is wrong. It starts with uppercase I
image

I'd suggest to rename the directory, following the camelcase naming we currently have.

sanmont3drepo (changes requested):

sanmont3drepo (approved):

The-Daniel (approved):

PR #4749, opened 2024-02-02
ISSUE #4748 -Fix crash when editing groups in new ticket

### LINKED ISSUE #4748: Entering groups in a new ticket crashes the app
### Description
When creating a new ticket if you try to add a group without saving first it crashes the app

- This is happening in: prod
- requires a hotfix to production: yes/no @carmenfan ?

### Steps to replicate
- Create a ticket  from a with a default view
- Without saving the new ticket enter groups

### Current Behaviour
- The app crashes

### Expected Behaviour
- Groups should be able to be created without problems

sanmont3drepo:
This fixes #4748 

#### Description
- Rollback to how ticket groups worked.
- Added isEditting group to avoid clashes when editing the group and showing the viewpoint.

#### Test cases
- Creating a new ticket and editing the groups in the moment should work

**No regression bugs:**
- Selecting a ticket in the list should show the view, closing and opening the ticket card should not reset the view.
- Editing groups should work as expected

### DIFF
--- a/frontend/src/v4/routes/viewerGui/viewerGui.component.tsx
+++ b/frontend/src/v4/routes/viewerGui/viewerGui.component.tsx
@@ -73,6 +73,7 @@ interface IProps {
 	selectedTicket: ITicket | undefined;
 	treeNodesList: any;
 	ticketsCardView: TicketsCardViews;
+	isEditingGroups: boolean;
 	stopListenOnSelections: () => void;
 	stopListenOnModelLoaded: () => void;
 	stopListenOnClickPin: () => void;
@@ -188,7 +189,7 @@ export class ViewerGui extends PureComponent<IProps, IState> {
 			this.props.resetCompareComponent();
 		}
 
-		if (this.props.ticketsCardView === TicketsCardViews.DetailsGroups) {
+		if (this.props.isEditingGroups) {
 			return;
 		}
 
--- a/frontend/src/v5/ui/routes/viewer/tickets/ticket.context.tsx
+++ b/frontend/src/v5/ui/routes/viewer/tickets/ticket.context.tsx
@@ -14,7 +14,6 @@
  *  You should have received a copy of the GNU Affero General Public License
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-import { TicketsCardActionsDispatchers } from '@/v5/services/actionsDispatchers';
 import { createContext, useState } from 'react';
 
 export enum TicketDetailsView {
--- a/frontend/src/v5/ui/routes/viewer/tickets/ticket.context.tsx
+++ b/frontend/src/v5/ui/routes/viewer/tickets/ticket.context.tsx
@@ -23,22 +23,22 @@ export enum TicketDetailsView {
 
 export interface TicketContextType {
 	isViewer?: boolean;
-	view: TicketDetailsView;
-	viewProps?: any;
+	detailsView: TicketDetailsView;
+	detailsViewProps?: any;
 	setDetailViewAndProps: (view: TicketDetailsView, props?: any) => void;
 }
 
 const defaultValue: TicketContextType = {
 	isViewer: false,
-	view: TicketDetailsView.Form,
+	detailsView: TicketDetailsView.Form,
 	setDetailViewAndProps: () => {},
 };
 export const TicketContext = createContext(defaultValue);
 TicketContext.displayName = 'TicketContext';
 
 export const TicketContextComponent = ({ children, isViewer }) => {
-	const [view, setView] = useState(TicketDetailsView.Form);
-	const [viewProps, setViewProps] = useState();
+	const [detailsView, setView] = useState(TicketDetailsView.Form);
+	const [detailsViewProps, setViewProps] = useState();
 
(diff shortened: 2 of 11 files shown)


frontend/src/v4/routes/viewerGui/viewerGui.component.tsx
```diff
@@ -73,6 +73,7 @@ interface IProps {
 	selectedTicket: ITicket | undefined;
 	treeNodesList: any;
 	ticketsCardView: TicketsCardViews;
+	isEditingGroups: boolean;
```
Amantini1997, line 76: I can't see any component using this

Amantini1997 (changes requested):

frontend/src/v4/routes/viewerGui/viewerGui.component.tsx
sanmont3drepo: in line 192

Amantini1997: Oh, right, my bad.

In that case, I saw that the `view` object that you get from Ticket context is either `Form` or `Groups`; but:

- Using `TicketsCardHooksSelectors.selectView()` is possible to determine if you are looking at the `Form` (if the value is not `List`, it's either the new ticket or the details of an existing one, so the form view).
- With the new `isEditingGroups` is also possible to determine if the form is group mode, equivalent to `Groups` from ticketContext.

As such, do we still need to keep the `view` in ticketContext?

sanmont3drepo: Yes. If you take a look the PR undoes that idea because that approach makes it prone to errors. Because the ticketview property is agnostic to wether is a new ticket or you are updating an existing ticket, it is better that navigation (switching between groups and no groups) is also agnostic. Otherwise it has to know if you are going from a new ticket or an already existing one. Bottom line, that is an idea that already failed.

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketGroups/ticketGroups.component.tsx
```diff
@@ -197,7 +197,12 @@ export const TicketGroups = ({ value, onChange, onBlur }: TicketGroupsProps) =>
 		dispatch(ViewpointsActions.clearColorOverrides());
 
 		ViewerService.on(VIEWER_EVENTS.BACKGROUND_SELECTED, clearHighlightedIndex);
-		return () => ViewerService.off(VIEWER_EVENTS.BACKGROUND_SELECTED, clearHighlightedIndex);
+
+		TicketsCardActionsDispatchers.setEditingGroups(true);
+		return () => {
+			ViewerService.off(VIEWER_EVENTS.BACKGROUND_SELECTED, clearHighlightedIndex);
+			TicketsCardActionsDispatchers.setEditingGroups(false);
```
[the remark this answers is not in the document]
sanmont3drepo: If it is a whitescreen that appears in this branch and not in prod then I should fix it.

frontend/src/v4/routes/viewerGui/viewerGui.component.tsx
The-Daniel: I'll create a ticket for the pin bug

Amantini1997 (approved):

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/ticketGroups/ticketGroups.component.tsx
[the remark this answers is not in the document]
sanmont3drepo: Fixed the crash, but the original bug still remains.

frontend/src/v5/ui/routes/viewer/tickets/ticket.context.tsx
```diff
@@ -23,22 +23,22 @@ export enum TicketDetailsView {
...
+	detailsView: TicketDetailsView;
+	detailsViewProps?: any;
 	setDetailViewAndProps: (view: TicketDetailsView, props?: any) => void;
 }
 
 const defaultValue: TicketContextType = {
 	isViewer: false,
-	view: TicketDetailsView.Form,
+	detailsView: TicketDetailsView.Form,
 	setDetailViewAndProps: () => {},
```
The-Daniel, line 34: Sorry. One last thing. Should this not be `setDetailsViewAndProps` instead of `setDetailViewAndProps`

carmenfan (approved):

frontend/src/v5/ui/routes/viewer/tickets/ticket.context.tsx
sanmont3drepo: there are two fields that are set by that function `detailsView` and `detailsViewProps`

The-Daniel: Yes, so why is it detail and not details in this case?

sanmont3drepo: Ahh ok. cool. I'll rename that in another PR, is not worth to get another round of reviews just for an s.

The-Daniel (approved):

PR #4759, opened 2024-02-06
Issue 4755 - fix date overflowing with am/pm approach

### LINKED ISSUE #4755: Date is overflowing
### Description
We need to fix the date so it doesn't overflow

- On new groups (only new, existing seems fine):
image

- For @AsiteZane - this is showing AM/PM which is overflowing. All dates shoudl be in 24hr clock format
image

Amantini1997:
This fixes #4755

#### Description
- due date uses the 24h format
- v4 groups "updated at" uses less margin on the right to allow more text;
> The problem was not related to the group status (new vs. edit), but it was probably a coincidence.

#### Test cases
You need to first [switch your browser settings](https://stackoverflow.com/a/41190151/10536697) to use the American system, then check that:
- tickets due date uses the 24h format
- v4 groups "updated at" does not break (you may want to use the inspector to change the date-time with something longer; I tested against `00/00/0000 00:00` which should be the longest string that uses digits)

### DIFF
--- a/frontend/src/v5/helpers/intl.helper.ts
+++ b/frontend/src/v5/helpers/intl.helper.ts
@@ -88,12 +88,13 @@ export const formatShortDate = (date) => formatDate(date, { // DD/MM/YY
 	year: '2-digit',
 });
 
-export const formatShortDateTime = (date) => formatDate(date, { // DD MM YYYY hh:mm
+export const formatShortDateTime = (date, overrides = {}) => formatDate(date, { // DD MM YYYY hh:mm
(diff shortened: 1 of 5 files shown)


carmenfan (approved):

Amantini1997:
@carmenfan I have noticed a bug in ticket date property, so I trigger a rebuilt and retagged you

sanmont3drepo (approved):

carmenfan (approved):
All working and no overflowing

@AsiteZane also confirmed it's working for him now 🎊

PR #4773, opened 2024-02-09
Issue 4733 - Custom status options in custom tickets

### LINKED ISSUE #4733: Custom tickets: Customisable status configuration
### Description
We want the ability for a user who's creating a template to be able to customise their status.

See details and schema proposal in https://github.com/3drepo/3D-Repo-Product-Team/issues/370

### Goals
- [ ] As an admin of a project, I want to be able to freely define what are my custom ticket status for each ticket template
- [ ] As an admin of a project, I want to be able to define what the custom status means so it can be integrated to existing filters

### Tasks
- [x] Create a statusTypes list that contains the possible status types (0.1 hrs)
- [x] Modify  addTicketTemplate endpoint /teamspaces/{teamspace}/settings/tickets/templates/{template}: payload should accept and validate status property (2 hrs)
- [x] Swagger update for addTicketTemplate (0.1 hrs)
- [x] Modify  updateTicketTemplate endpoint /teamspaces/{teamspace}/settings/tickets/templates/{template}: payload should accept and validate status property (1 hrs)
- [x] Swagger update for updateTicketTemplate (0.1 hrs)
- [x] Modify  getModelTemplates endpoint /teamspaces/{teamspace}/projects/{project}/{type}/{model}/tickets/templates: should include status property in the response in case getDetails query is set to true (3 hrs)
- [x] Swagger update for getModelTemplates (0.1 hrs)
- [x] Modify  getTemplateDetails endpoint /teamspaces/{teamspace}/projects/{project}/{type}/{model}/tickets/templates/{template}: should include status property in the response (1 hrs)
- [x] Swagger update for getTemplateDetails  (0.1 hrs)
- [x] Modify  addTicket endpoint /teamspaces/{teamspace}/projects/{project}/{type}/{model}/tickets: payload validation should ensure status type is valid (3 hrs)
- [x] Modify  updateTicket endpoint /teamspaces/{teamspace}/projects/{project}/{type}/{model}/tickets: payload validation should ensure status type is valid (1)
- [x] Add unit tests (4 hrs)
- [x] Add e2e tests (4 hrs)

ChristosTsiotsias:
This fixes #4733

#### Description
Users can now create templates that define custom status options in thein config property
Tickets with custom statuses can now created based on these templates
A default property is defined in the template in case the user does not provide a value
Unit and E2E tests added
Swagger documenation updated

#### Test cases

Test 1
Create a template that has a status with some values defined in its config property 
Create a ticket based on the template
You should be able to create it provided you pass one of the required values.
If no value is provided the default value should be taken

Test 2
Create a template that has no status defined in its config property 
Create a ticket based on the template
You should be able to create it provided you pass one of the default status values
If no value is provided the default value (Open) should be taken


backend/src/v5/schemas/tickets/templates.constants.js
```diff
@@ -99,22 +101,23 @@ TemplateConstants.presetModulesProperties = {
 
 };
 
-TemplateConstants.defaultProperties = [
+TemplateConstants.defaultProperties = (config) => [
```
carmenfan, line 104: No I don't mean it like this, this is too clunky 😢 

What I mean is it constructures a function like the `availableIf` callback you see in `createPropertyEntry`

backend/src/v5/schemas/tickets/templates.constants.js
```diff
@@ -99,22 +101,23 @@ TemplateConstants.presetModulesProperties = {
...
 	createPropertyEntry('Description', propTypes.LONG_TEXT),
 	createPropertyEntry('Owner', propTypes.TEXT, undefined, undefined, true),
 	createPropertyEntry('Created at', propTypes.DATE, undefined, undefined, true),
 	createPropertyEntry('Updated at', propTypes.DATE, undefined, undefined, true),
-	createPropertyEntry('Status', propTypes.ONE_OF, ['Open', 'In Progress', 'For Approval', 'Closed', 'Void'], 'Open'),
+	createPropertyEntry('Status', propTypes.ONE_OF,
+		config?.status ? config?.status.values.map((v) => v.name) : ['Open', 'In Progress', 'For Approval', 'Closed', 'Void'],
+		config?.status ? config?.status.default : 'Open'),
```
carmenfan, line 111: This can be simplified
```suggestion
		config?.status?.default || 'Open'),
```

backend/src/v5/schemas/tickets/templates.constants.js
carmenfan: Actually I think I have a better way of refactoring this. Let me have a think and get back to this.

backend/src/v5/schemas/tickets/templates.js
```diff
@@ -145,6 +146,11 @@ const moduleSchema = Yup.object().shape({
 }).test('Name and type', 'Only provide a name or a type for a module, not both',
 	({ name, type }) => (name && !type) || (!name && type));
 
+const customStatus = Yup.object({
+	name: Yup.string().min(1).required(),
```
carmenfan, line 150: this should be string.title from the yup helper

backend/src/v5/schemas/tickets/templates.js
```diff
@@ -153,9 +159,13 @@ const configSchema = Yup.object().shape({
 	defaultView: defaultFalse,
 	defaultImage: Yup.boolean().when('defaultView', (defaultView, schema) => (defaultView ? schema.strip() : defaultFalse)),
 	pin: Yup.lazy((val) => (val?.color ? Yup.object({ color: pinColSchema }) : defaultFalse)),
+	status: Yup.lazy((val) => (val ? Yup.object({
+		values: Yup.array().of(customStatus).min(1).required(),
+		default: Yup.mixed().when('values', (values) => Yup.string().oneOf(values.map((v) => v.name)).required()),
```
carmenfan, line 164: We use spread operator when possible so we should do it here too for consistency
```suggestion
		default: Yup.mixed().when('values', (values) => Yup.string().oneOf(values.map(({name}) => name)).required()),
```

backend/src/v5/schemas/tickets/templates.js
```diff
@@ -153,9 +159,13 @@ const configSchema = Yup.object().shape({
 	defaultView: defaultFalse,
 	defaultImage: Yup.boolean().when('defaultView', (defaultView, schema) => (defaultView ? schema.strip() : defaultFalse)),
 	pin: Yup.lazy((val) => (val?.color ? Yup.object({ color: pinColSchema }) : defaultFalse)),
+	status: Yup.lazy((val) => (val ? Yup.object({
```
carmenfan, line 162: why do we need lazy here?

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/tickets.test.js
```diff
@@ -204,6 +205,7 @@ const testAddTicket = () => {
 	describe('Add ticket', () => {
 		const { users, teamspace, project, con, fed } = generateBasicData();
 		const template = ServiceHelper.generateTemplate();
+		const statusValues = times(5, () => ({ name: ServiceHelper.generateRandomString(), type: statusTypes[0] }));
```
carmenfan, line 208: maybe make the type variable too so we're sure all the cases are covered?

```suggestion
		const statusValues = times(statusTypes.length, (i) => ({ name: ServiceHelper.generateRandomString(), type: statusTypes[i] }));
```

backend/tests/v5/unit/schemas/tickets/templates.constants.test.js
```diff
@@ -15,12 +15,18 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+const { times, cloneDeep } = require('lodash');
 const { src } = require('../../../helper/path');
+const { generateRandomString } = require('../../../helper/services');
+const { statusTypes } = require('../../../../../src/v5/schemas/tickets/templates.constants');
```
carmenfan, line 21: We should not be using relative path in the test files to reference source (i've pointed this out a few times before! 😆 )

backend/tests/v5/unit/schemas/tickets/templates.constants.test.js
```diff
@@ -15,12 +15,18 @@
...
 
 const TemplateConstants = require(`${src}/schemas/tickets/templates.constants`);
 
 const testGetApplicableDefaultProperties = () => {
 	describe('Get applicable default properties', () => {
+		const customStatusValues = times(5, () => ({ name: generateRandomString(), type: statusTypes[0] }));
```
carmenfan, line 27: same here, make your test case use all possible status types

carmenfan: Since you've need this in multiple place you may as well put it in the service helper?

carmenfan (changes requested):

carmenfan (changes requested):
Missed something in the last PR!

backend/src/v5/schemas/tickets/templates.js
ChristosTsiotsias: we do that for pin so I think we have to do something similar here 🤔

carmenfan: that's because a pin could either be an object or a boolean (due to old schema)

status either is an object or it doesn't exist so I don't think we need a lazy evaluation here...

backend/src/v5/schemas/tickets/templates.constants.js
carmenfan: @ChristosTsiotsias  I've changed this in the following way:

- `defaultProperties` is no longer exposed, no one outside of the constant file should need it. They either need to get applicable properties which calls the function, or they just want the label which should be using `basePropertyLabels`.
- This is now split into 2, `defaultProperties` - things that are not customisable and `customisableProperties` - i.e. status and pins - they take a default property object, and a function (with a config) that alters it. 


Which i believe is cleaner, if you have any thoughts, let me know. https://github.com/3drepo/3drepo.io/pull/4773/commits/0833cffeb1325186aad8601680b9297c4ead4c6c

carmenfan (approved):

carmenfan:
@The-Daniel the endpoints for customisable status are now in staging

PR #4778, opened 2024-02-14
Issue 4744 - Immutable properties in custom tickets

### LINKED ISSUE #4744: Custom tickets: Add ability to declare a property as immutable
### Description
To add the ability to define a property as immutable.

For details, see https://github.com/3drepo/3D-Repo-Product-Team/issues/375

### Goals
- [ ] As a template creator, I want to ensure that a field that can only be written once cannot be altered
- [ ] As a template creator, I want to be notified if the property field I create is not viable, for example, I should not be able to declare something as both `read only` and `immutable`

### Tasks
Template endpoints
- [x] Amend template create / edit endpoints (teamspaces/{teamspace}/settings/tickets/templates) so that immutable field is supported on properties
- [x] Amend get endpoint  (teamspaces/{teamspace}/settings/tickets/templates/{template}) and make sure it returns immutable fields 
- [x] Amend unit and e2e tests to cover the new cases
- [x] Update swagger on all endpoints

Ticket endpoints
- [x] Amend ticket create / edit endpoints (/teamspaces/{teamspace}/projects/{project}/{type}/{model}/tickets/{ticket}) so that immutable properties can only be set once
- [x] Amend unit and e2e tests to cover the new cases

ChristosTsiotsias:
This fixes #4744

#### Description
Immutable properties can now be created in custom ticket templates
Ticket with immutable properties can be created
Immutable properties in a ticket can only be set and cannot be altered afterwards
Unit and E2E tests added
Swagger documentation updated

#### Test cases
Create a template that includes an immutable property
Create a ticket based on this template and set the property
Try to update the property. The an error should be returned
If the property is not set upon creation the update should be allowed but an error should be returned for every update afterwards


backend/src/v5/schemas/tickets/index.js
```diff
@@ -90,6 +90,9 @@ const generatePropertiesValidator = async (teamspace, properties, oldProperties,
 				}
 
 				const oldValue = oldProperties?.[prop.name];
+
+				validator = validator.test('Immutable property', `Immutable property ${prop.name} cannot be edited`, (value) => !(prop.immutable && oldValue && value));
```
carmenfan, line 94: Can we wrap it around with ifs like the other stuff?

also we should check for undefined as it's possible for oldValue to be `false` or `0` if the type is boolean/number

```suggestion
if(prop.immutable && oldValue !==undefined){
				validator = validator.test('Immutable property', `Immutable property ${prop.name} cannot be edited`, (value) => value !== undefined);
}
```

backend/src/v5/services/swagger/swagger.schemas.js
```diff
@@ -59,6 +59,7 @@ const ticketTemplatePropSchema = {
 			name: helpers.stringDef('Name of the prop', 'Floor'),
 			type: helpers.stringDef('Property type', propTypes.ONE_OF, Object.values(propTypes)),
 			deprecated: helpers.boolDef('Denotes if this prop is no longer in use', false),
+			immutable: helpers.boolDef('Denotes if this prop can be altered once it is set', false),
```
carmenfan, line 62: The description is misleading, as it's implying if it's true it can be altered. Immutable is a valid word in the dictionary, so let's just use that.
```suggestion
			immutable: helpers.boolDef('Denotes if this prop is immutable', false),
```

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/tickets.test.js
```diff
@@ -606,6 +606,7 @@ const testUpdateTicket = () => {
 	describe('Update ticket', () => {
 		const { users, teamspace, project, con, fed } = generateBasicData();
 		const requiredPropName = ServiceHelper.generateRandomString();
+		const immutableProperty = ServiceHelper.generateRandomString();
```
carmenfan, line 609: Can we add a test for (not sure where abouts in the file, i'm just tagging it here, doesn't mean we should do it here):

- A template with an immutable property that has a default value
- Create a ticket without filling in the property
- try to update the timmutable property - should fail (as it should've taken the default value)

carmenfan (changes requested):

carmenfan (approved):

carmenfan:
@The-Daniel immutable properties backend code is now on staging

PR #4783, opened 2024-02-15
Issue 4734 - Repo Bundle testing branch

### LINKED ISSUE #4734: Add support for our own Bundles
### Description

We want to replace Unity Asset Bundles with our own Bundles that contain pre-processed graphics Mesh Assets. This will require frontend APIs to:

1. Return the actual RepoBundles on request
2. Return the mappings for them
3. Return textures for materials separately
4. List the available bundles in the assets lists

### Goals

- [ ] TDB

### Tasks

- [ ] TBD

### Related Resources

Bouncer Issue: `https://github.com/3drepo/3drepobouncer/issues/666`
Coda doc: `link to coda`

sebjf:
This fixes #4734 - Support for RepoBundles

#### Description
This PR adds or adjusts three APIs in support of `RepoBundles`, and adds new members to `unity-util`. Specifically:

1. Add an API `/:uid.repobundle`, which requests a stream from the `.stash.repobundles.ref` collection of a container.
2. Add an API `/:uid.texture`, which gets a node of the type Texture from the `.scene` collection of a container, and uses it to return a texture.
3. Additionally, two new routes are added for the `supermeshes_unity.json` "file".

These routes are used were the viewer is operating in forced-legacy mode, to make .io reference the backup `supermeshes.json` cache constructed by `AssetModelExport`, instead of the one built by `AssetGenerator`. 
In normal case (for both pre & post RepoBundles imports), this should not be needed, as using the existing `supermeshes.json` request will return the correct cache. It is only when RepoBundles have written the supermeshes.json cache, and it is *not* desired, is a special request needed.

### v4, v5 & decommissioning AssetBundleCreator 

* The two new APIs are v4-style, as the Viewer is not currently aware of the Project id
    * [Unity Issue 463](https://github.com/3drepo/3drepounity/issues/463) should be merged first, then the `WebRequestHandler`s in Unity upgraded to use V5 endpoints for `RepoBundles` and `Textures`
    * Currently the individual supermesh json mpc APIs are not called anywhere in the viewer.
* When `AssetBundleCreator` is decommissioned, the `supermeshes_unity.json` should be removed also.

### Legacy Mode

* For the next release, new uploads will continue to generate Unity AssetBundles and associated mappings, exactly as before.
* The Viewer can be forced to use AssetBundles even if RepoBundles are available, by calling `UnityUtil.setLegacyAssetMode()` during `init()`. This must be called before `unityAssets.json` is downloaded, so is only useful at start-up. (This is not exposed in a dialog.)

### Textures

* Only JPG and PNG are supported with RepoBundles. Textures for now are retrieved from the `.scene` collection.
* Textures can be turned on and off at Runtime using the `UnityUtil.enableTextures/disableTextures` calls. (These are not exposed in any dialogs currently.)


carmenfan:
@sebjf Since the other PRs are ready and assigned to me to have a look I'm guessing I should look at this one as well

Let me know if it's not ready yet!

sebjf:
> @sebjf Since the other PRs are ready and assigned to me to have a look I'm guessing I should look at this one as well

Yes, thank you it's ready for a look!

backend/src/v4/models/unityAssets.js
```diff
@@ -76,4 +76,40 @@ UnityAssets.getUnityBundle = function(account, model, id) {
...
+	const textureFilename = `${id}`;
+	const collection = `${model}.scene`;
+
+	const node = await db.findOne(account, collection, { _id: utils.stringToUUID(textureFilename) }, {
+		_id: 1,
+		_blobRef: 1,
+		extension: 1
+	});
+
+	const {elements, buffer} = node._blobRef;
```
carmenfan, line 95: the server is going to crash if node is undefined here.

We need an if statement to make sure we found a matching node before we proceed, and respond accordingly if we failed to find a matching texture with that id.

carmenfan: it will also do weird thigns if we found a matching node but it doesn't have a blobRef (e.g. if someone for some reason plugged in a transformation id here.

So perhaps we should filter for `type: texture` in the query as well

carmenfan (changes requested):

backend/src/v4/models/unityAssets.js
sebjf: Done! (It now queries only for textures and returns a 404 if a node isn't found.)

carmenfan (approved):

Issue #4857, opened 2024-04-02
[2D MVP] Milestone 3: Unity changes to support manual Calibration


sebjf:
### Description

Support Manual Calibration as described in this ticket: https://github.com/3drepo/3D-Repo-Product-Team/issues/379

This PR for now is mainly to test the viewer modifications. It will be extended with the frontend changes as necessary.

### Goals

- TBD

### Tasks

- TBD

### Related Resources

Product Team: https://github.com/3drepo/3D-Repo-Product-Team/issues/379
Unity: https://github.com/3drepo/3drepounity/issues/518

carmenfan:
merged into #4897

PR #4858, opened 2024-04-02
ISSUE #4857 [2D MVP]: Manual Calibration - Unity changes

sebjf:
This addresses #4857

#### Description
Right now, this PR exists to test the behaviour of the calibration tools in the viewer.

The calibration mode can be controlled via the console, using the same APIs that the frontend will use. These APIs can be found in `unity-uti`l under the _Calibration_ category. **The comments in unity-util.ts should always be considered the latest documentation, superseding any in the linked tickets.**

The APIs currently supported are:

`setCalibrationToolMode`
`setCalibrationToolDrawing`
`selectCalibrationToolLowerPlane`
`selectCalibrationToolUpperPlane`
`setCalibrationToolVerticalPlanes`
`setCalibrationToolFloorToObject`
`setCalibrationToolVector`

The tool does not need to be explicitly enabled or disabled. Modes are turned on and off by calling `setCalibrationToolMode` with the mode name or "none".

### Vector Mode

To start the tool in Vector Mode, open the console and give the command:

`UnityUtil.setCalibrationToolMode("Vector")`

The tool will not automatically set up the cursor or snapping. This must be done explicitly:

`UnityUtil.enableSnapping()`

Immediately the user can place two points between which the vector will be drawn.

`setCalibrationToolVector` can be used to completely or partially remove or restore the vector, for example,

1. Reset the vector, pointing from the Project origin to Project [1,1,1]:

`UnityUtil.setCalibrationToolVector([0,0,0], [1,1,1])`

2. Remove the end point, allowing the user to click again to set it to another location:

`UnityUtil.setCalibrationToolVector([0,0,0], null)`

3. Remove the start point, allowing the user to click again to set it to another location:

`UnityUtil.setCalibrationToolVector(null, [1,1,1])`

4. Remove both points, the equivalent to resetting the tool:

`UnityUtil.setCalibrationToolVector(null, null)`

### Planes

To use the Vertical Calibration Mode...

1. Start the tool in vertical/floor mode by giving the command:

`UnityUtil.setCalibrationToolMode("Vertical")`

2. Provide a drawing to be shown on the planes. 

For this a DOM object is needed. To test this API before the actual drawing is available, create a placeholder, such as an  node, like so:

```
var i = new Image()
i.crossOrigin = "anonymous"
i.src = "https://as2.ftcdn.net/v2/jpg/02/77/71/45/1000_F_277714513_fQ0akmI3TQxa0wkPCLeO12Rx3cL2AuIf.jpg"
```

(Make sure the image has fully loaded before handing it to the API. This is because the call will immediately read a couple of properties of the image, such as the width, which cannot be changed later.)

Then call `setCalibrationToolDrawing`, like so:

`UnityUtil.setCalibrationToolDrawing(i, [0, 0,  1, 0,  0, 1])`

The array defines the rectangle of the image in the ZX plane as the positions of three corners in Project coordinates.

The above command requests that the  node `i` be shown in a rectangle with the bottom left corner at `[0,0]`, the bottom right corner at `[1,0]` and the top left corner at `[0,1]` - that is, a unit square starting at the origin. (As such, that exact command should only be used with very small models, e.g. one unit cube).

3. Set the initial heights for the tool. If these are not set the drawing will appear at Height 0.

To change between the active planes, give the command:

`UnityUtil.selectCalibrationToolLowerPlane()`

or

`UnityUtil.selectCalibrationToolUpperPlane()`

For convenience, the `unity-util` on this branch prints `currentPointInfo` to the console, in order to help test the `setCalibrationToolFloorToObject` function.

For example, after clicking an object and seeing the log entry

```
{"id":"f0c58283-34b0-4200-a1cb-04b9c4af061b","database":"sebjf","model":"59596f30-4b00-11ed-ac46-93641d60e710","pin":false,"mousePos":[583.0,382.0],"position":[],"normal":[]}
```

The following command could be given:

```
UnityUtil.setCalibrationToolFloorToObject("sebjf","59596f30-4b00-11ed-ac46-93641d60e710","f0c58283-34b0-4200-a1cb-04b9c4af061b")
```

The upper and lower planes can also be set. For example:

`UnityUtil.setCalibrationToolVerticalPlanes(0, 1)`

Any changes should take effect immediately if the tool is visible.

### Events

The calibration tool can raise one event, `calibrationPlanesChanged`. The current implementation state in this PR is that the viewer will call the hidden method in `UnityUtil`, and the arguments are printed to the console. **That is, they do not exist as actual events in Viewer.tsx yet.**

#### Planes

Every frame the user adjusts either of the planes using the gizmo, the viewer will call `UnityUtil.calibrationPlanesChanged`. This method will be passed an object containing the height of each plane in Project coordinates.

```
{
    "lower": -1.4685417413711548,
    "upper": 0
}
```

#### Vector

Vector mode does not raise events. Instead the frontend should rely on the pickPointAlert event.

For example, execute the following snippet in the console to override the behaviour of pickPointAlert to draw a vector between two points:

```
UnityUtil.setCalibrationToolMode("Vector");
UnityUtil.enableSnapping();
UnityUtil._points = [];
UnityUtil.viewer.pickPointEvent = (ev) => { 
UnityUtil._points.push(ev.position); 
if(UnityUtil._points.length == 1){ UnityUtil.setCalibrationToolVector(UnityUtil._points[0], null); }
if(UnityUtil._points.length == 2){ UnityUtil.setCalibrationToolVector(UnityUtil._points[0], UnityUtil._points[1]); UnityUtil._points = []; }
}
```


sanmont3drepo (changes requested):

frontend/src/globals/unity-util.ts
```diff
@@ -25,6 +25,8 @@ import { ExternalWebRequestHandler } from './unity-externalwebrequesthandler';
 declare let SendMessage;
 declare let createUnityInstance;
 
+type DrawingImageSource = ImageBitmap | ImageData | HTMLImageElement | HTMLCanvasElement | OffscreenCanvas;
```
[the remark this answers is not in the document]
sanmont3drepo: This is only a suggestion, but I think that having something like this would look nicer. I dont know if a box per se, but something that acompasses the whole bounding box.

image

sebjf:
From the meeting so I don't forget: after design review, we will either:

1. Rework the Preview Tool so that it doesn't emit events, and handles the change in vertical location internally - i.e. it is only a confirmation and cannot fine tune the calibration.
2. Merge the Preview and Vertical Tools, so that the Vertical Tool shows the actual image with the horizontal calibration (as per the Preview tool), but on the selected plane, and moving the planes updates the Calibration (as per the vertical tool).

carmenfan:
@sebjf 
1. yea that's fine.

> (@carmenfan, which parameters, if any, should be exposed in the settings?)

Not sure, maybe wait for @sanmont3drepo  to get his take on this. I'm thinking this is what I want to see if you guys agree:

unselected surfaces
- Low opacity and just hovering around as a reference

selected surface
- High opacity (but still has a slight transparency so we can the model below for the user to reference)
- OutLined with call to action colour (ideally sent from the frontend so if theme changes it naturally changes)

Could we maybe see it with a floor plane so we can have a better feel of what it looks like? You can find some here https://asitesol.sharepoint.com/sites/3DRepoTeamFolder/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2F3DRepoTeamFolder%2FShared%20Documents%2FDevelopment%2FFeature%20Resources%2F2D%20Viewer&viewid=60884782%2D12a7%2D4f6f%2Db426%2Debefd8aef345

frontend/src/globals/unity-util.ts
```diff
@@ -674,6 +691,54 @@ export class UnityUtil {
...
+		}
+	}
+
+	public static calibrationVectorPoints = [];
+
+	public static calibrationVectorTestReset() {
+		this.calibrationVectorPoints = [];
+	}
+
+	public static calibrationVectorTestGetCoords() {
```
sanmont3drepo, line 717: -  The plane is being cut off if its sticking out of the current model 

image

sanmont3drepo (changes requested):

frontend/src/globals/unity-util.ts
sebjf: Hi @sanmont3drepo, this should be fixed now.

nkuberaf:
azp /run

carmenfan:
@sebjf  can you make a PR on the unity branch as well please? @sanmont3drepo is happy with the functionality so we need to review the code and merge it in

sebjf:
> @sebjf can you make a PR on the unity branch as well please? @sanmont3drepo is happy with the functionality so we need to review the code and merge it in

PR is here: https://github.com/3drepo/3drepounity/pull/540

frontend/src/globals/unity-util.ts
```diff
@@ -674,6 +691,54 @@ export class UnityUtil {
...
+	/**
+	 * Called by the Calibration Tool when a user action changes the state of the Vector.
+	 * The line is given as a start position and end position in Project coordinates.
+	 */
+	/** @hidden */
+	public static calibrationVectorChanged(lineJson: string) {
+		const line = JSON.parse(lineJson);
+
+		// eslint-disable-next-line no-console
+		console.log(line);
```
carmenfan, line 703: do we still need this?

If we want to keep this for debugging purpose I suggest we wrap it around `UnityUtil.verbose` like we do in other places

frontend/src/globals/unity-util.ts
```diff
@@ -674,6 +691,54 @@ export class UnityUtil {
...
+
+		if (line.end && line.end.length > 0) {
```
carmenfan, line 705:
```suggestion
		if (line?.end?.length > 0) {
```

frontend/src/globals/unity-util.ts
```diff
@@ -674,6 +691,54 @@ export class UnityUtil {
...
+	}
+
+	/**
+	 * Called by the Calibration Tool when a user action changes the heights of the vertical planes.
+	 * Heights are given in Project coordinates from the origin.
+	 */
+	/** @hidden */
+	public static calibrationPlanesChanged(planes) {
+		// eslint-disable-next-line no-console
+		console.log(JSON.parse(planes));
```
carmenfan, line 739: similar here, should wrap it around UnityUtil.verbose if it's a debugging thing

All this does is call console.log - do we need this at all?

carmenfan (changes requested):

carmenfan:
@sebjf the reason I'm reviewing this is because @sanmont3drepo wants it merged.

Can you 2 decide what you want to do ? If we're merging this then I'm only fine with having temporary code if have a process to rememebr to remove it somehow (e.g. another child issue to address it)

Or we can remove it now if @sanmont3drepo  don't need it, or put it into UnityUtils

sebjf:
Hi @carmenfan @Amantini1997, this branch (and https://github.com/3drepo/3drepounity/pull/540) have been updated, as discussed yesterday. @Amantini1997 please see the revised [top comment](https://github.com/3drepo/3drepo.io/pull/4858#issue-2220725431).

sanmont3drepo (approved):

sanmont3drepo (approved):

PR #4863, opened 2024-04-03
ISSUE #4702 - New group objects counts reflects selected objects

### LINKED ISSUE #4702: Trying to create a new group with already selected meshes does not detect them
### Description

- This is happening in: both

### Steps to replicate
- In the viewer, select some meshes
- open the groups(v4) card and click the new button

### Current Behaviour
- The save button is disabled

### Expected Behaviour
- The save button should be enabled

Amantini1997:
This fixes #4702

#### Description
New group (v4) objects counts matches the number of elements selected in the viewer and if, upon opening the group card, there are some objects selected, the group can be saved straight away

#### Test cases
- Select some elements in the viewer
- Click on "new group"
- The save button should be enabled

### DIFF
--- a/frontend/src/v4/routes/viewerGui/components/groups/components/groupDetails/groupDetails.component.tsx
+++ b/frontend/src/v4/routes/viewerGui/components/groups/components/groupDetails/groupDetails.component.tsx
@@ -57,6 +57,7 @@ interface IProps {
 	isPending: boolean;
 	deleteGroup: (id: string | null) => void;
 	isReadOnly: boolean;
+	getObjectsCount: (meshesIds: any[]) => number;
 }
 
 interface IState {
@@ -85,6 +86,14 @@ export class GroupDetails extends PureComponent<IProps, IState> {
 
 	get editingGroup() {
 		return this.props.editingGroup;
+	}
+
+	get objectsCount() {
+		if (!this.isNewGroup) {
+			return this.editingGroup.totalSavedMeshes;
+		}
+		const sharedIds = (this.props.selectedNodes || []).flatMap((node) => node.shared_ids);
+		return this.props.getObjectsCount(sharedIds);
 	}
 
 	get isFormValid() {
@@ -134,7 +143,7 @@ export class GroupDetails extends PureComponent<IProps, IState> {
 	));
 
 	public componentDidMount() {
-		this.setState({ isFormDirty: this.isNewGroup, isFormValid: false });
+		this.setState({ isFormDirty: this.isNewGroup, isFormValid: this.isNewGroup && !!this.objectsCount });
 	}
 
 	public componentDidUpdate(prevProps: Readonly<PropsWithChildren<IProps>>) {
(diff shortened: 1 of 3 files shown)


frontend/src/v4/routes/viewerGui/components/groups/components/groupDetails/groupDetails.component.tsx
```diff
@@ -87,6 +88,14 @@ export class GroupDetails extends PureComponent<IProps, IState> {
 		return this.props.editingGroup;
 	}
 
+	get objectsCount() {
+		if (!this.isNewGroup) {
```
The-Daniel, line 92: This means that if I go to edit an existing group the mesh count is stuck at the riginal number. So if I edit the selected meshes this number is wrong
image

frontend/src/v4/routes/viewerGui/components/groups/components/groupDetails/groupDetails.component.tsx
The-Daniel, line 1: Select some objects like so
image
In the toolbar select hide. The object count stays the same
image
Upon saving the mesh count changes to 0
image

The-Daniel (changes requested):

frontend/src/v4/routes/viewerGui/components/groups/components/groupDetails/groupDetails.component.tsx
Amantini1997: That seems correct to me tbh. When you have a new group and save, the number of objects will be the same as the selected objects, so you can see the number changing as you select/deselect some.
Opposite to that, already existing groups shouldn't update the number until you hit save as the number in the groupsCard should indicate the total number of objects belonging to the group rather than then ones currently selected in the viewer.
@sanmont3drepo what's your opinion about this?

frontend/src/v4/routes/viewerGui/components/groups/components/groupDetails/groupDetails.component.tsx
Amantini1997: I can see this sort of already being a problem in staging, where, if you select some meshes and then hide them, you can still create the group which will eventually count 0 objects

frontend/src/v4/routes/viewerGui/components/groups/components/groupDetails/groupDetails.component.tsx
sanmont3drepo: If it was already working this way we can ignore it and create a new issue @The-Daniel

frontend/src/v4/routes/viewerGui/components/groups/components/groupDetails/groupDetails.component.tsx
sanmont3drepo: yup, create a new issue.

Amantini1997: https://github.com/3drepo/3drepo.io/issues/4867

The-Daniel (approved):

sanmont3drepo (approved):

PR #4880, opened 2024-04-10
ISSUE #4789 - Drawings Permissions

### LINKED ISSUE #4789: [2D MVP] Milestone 1: Drawing permissions
### Description
For milestone 1 we want to re-use the permissions from federations/containers. This means we have to add the drawing list in the list together with containers and federations

### Goals
- As an admin, I want to specify who has the right to upload new drawings (collaborators)
- As an admin, I want to specify who has the right to view a drawing. (viewers)

### Tasks
- [ ] Modify the user permissions to have drawings also
- [ ] Add API to get the users permissions for drawings
- [ ] Add API to modify the user permissions for drawings
- [ ] Modify user permissions to post the changes for drawings

### Related Resources
image

The-Daniel:
This fixes #4789

#### Description
- Prevents users with insufficient permissions from certain actions
- Adds drawings to the user permissions table

##### NOTE:
In order to mock the drawings into the permissions API requests I had to mock ALL model permissions api requests. This means that on this branch you cannot change the permissions of federations and containers. This will be reverted once the backend has been added. I have tried to keep the hacky spoofing functions separate and have added some comments where things should be reverted.

#### Test cases
- On a project **where you are not admin** access the drawings list and check whether you can perfom actions you think should not be allowed for the respective permissions labels. (The permissions for a drawing have been added to the title for clarity).
- Open the 'Users permissions' page and change the permissions (These should look like they change successfully but actually the change is not saved due to the whole api mocking thing that has driven me up the wall)


frontend/src/v5/store/projects/projects.helpers.ts
```diff
@@ -27,3 +27,16 @@ export const getProjectImgSrc = (teamspace: string, project: string) => (
 
 export const projectImageFileIsTooBig = (file): boolean => (file.size > clientConfigService.projectImageSizeLimit);
 export const PROJECT_IMAGE_MAX_SIZE_MESSAGE = formatInfoUnit(clientConfigService.projectImageSizeLimit);
+
+export enum ModelTypes {
```
Amantini1997, line 31: I think enums' name should be singular

frontend/src/v5/store/projects/projects.helpers.ts
```diff
@@ -27,3 +27,16 @@ export const getProjectImgSrc = (teamspace: string, project: string) => (
...
+	CONTAINER,
+	FEDERATION,
+	DRAWING,
+}
+
+export const getModelType = (model) => {
+	if (model.federate) return ModelTypes.FEDERATION;
+	if (model.units) return ModelTypes.CONTAINER;
```
Amantini1997, line 39: federations also have units, so this can be misleading. Also, if it's not a federation and it's not a drawing, it's always a container, right? If so, could that be the default return value?

Amantini1997 (changes requested):

Amantini1997 (approved):

sanmont3drepo (approved):
This branch in particular is very dependent on backend.

PR #4883, opened 2024-04-11
ISSUE #4877 - fix final frame temporarily appearing when incrementing sequence

### LINKED ISSUE #4877: 4D model assembling and disassembling itself when sweeping across a time span with many frame
### Description

When cycling through a time range that consists of many frames, the entire model is temporarily displayed and then transitions back to the correct frame in time.  During the transitioning, all activities are shown in one go (temporary works and object status changes.) 

- This is happening in: **production**
- This requires a hotfix to production: **TBD by product manager** 

### Steps to replicate
- Open [link](https://staging.dev.3drepo.io/v5/viewer/SupportTeamWorkspace/abc05930-40fc-11ee-adc2-030e77f6faaf/9ba66107-5d41-4409-baef-001271a34161) to the model and click on sequences tab 
- Provide a 2 month time span and click the fast forward button 

### Current Behaviour

Loads the entire model when fast forwarding by a time span with lots of frames before showing the correct frame 

### Expected Behaviour

The correct frame should be initially visible at a specific point in time without cycling through different frames.

Image 1 - First displays end of sequence model 

image

Image 2 - Then Transitions to showing all activities in one go (including temporary works and object status changes) 

image

Image 3 - Then displays the correct frame in time.

image

Video 
https://asitesol.sharepoint.com/:v:/s/3DRepoTeamFolder/EY5_H-F01m9IqGRRyc4RB7oBijBHmSi-bmP1FGwYjQzf-g?e=rZv9N6

The-Daniel:
This fixes #4877

#### Description
Fixes bug where the final frame of a sequence would temporarily appear when incrementing the frame
This was caused because an undefined value was getting used.

#### Test cases
On a 4D model open up a sequence.
Set the timestamp to the start
Press the forward increment button
The final frame should not appear


Amantini1997 (approved):

frontend/src/v4/modules/sequences/sequences.sagas.ts
```diff
@@ -179,7 +179,11 @@ function * prefetchFrames() {
 function* updateSelectedStateDefinition() {
 	const selectedFrame = yield select(selectSelectedFrame);
 	const stateDefinitions = yield select(selectStateDefinitions);
-	yield put(SequencesActions.setSelectedStateDefinition(stateDefinitions[selectedFrame?.state]));
+	const selectedStateDefinition = stateDefinitions[selectedFrame?.state];
```
sanmont3drepo, line 182: When I enter the sequence it shows everything until the frame is showed. 
showing-everything

Is it possible to show the same state until the frame is loaded?

sanmont3drepo (changes requested):

frontend/src/v4/modules/sequences/sequences.sagas.ts
The-Daniel: To be fixed in #4926

sanmont3drepo (approved):

PR #4884, opened 2024-04-12
Issue 4867 - V4 manual group validation broken with hidden meshes

### LINKED ISSUE #4867: V4 manual group validation broken with hidden meshes
### Description
Editing a v4 group set to manual, it is possible to select some meshes (so that the group is valid) and then hide them using the bottom toolbar. Saving the group will save it with 0 objects included

- This is happening in: both 

Also, number of objects and lastUpdated do not make sense to be there for a new group, so we shall remove them

### Steps to replicate
- Open the groups card in the viewer
- Expand an existing manual group (if none is present, simply create one)
- Select some meshes in the viewer
- Click on "hide" from the bottom toolbar

### Current Behaviour
- The group can be saved and it will include 0 objects

### Expected Behaviour
- The group should not be saveable

After selecting meshes
319654979-03c5fd86-e28f-4e0f-8c41-6d3d696c83e3

After clicking "hide"
319655408-15c7d6c4-3a8e-4102-a695-b558a3d94017

After saving
319656161-86e28575-e491-4d9f-832d-b0a22312fd8e

Amantini1997:
This fixes #4867 

#### Description
- New groups do not show "last upated" or "objects count"
- Selected nodes (in redux state) should always reflect what is effectively selected. As such, selecting all the objects and the hiding them all will make a new group invalid

#### Test cases
- Select all the objects;
- Click on "new group" (the group should be saveable)
- Hide all the objects; (the groups should not be saveable)


frontend/src/v4/helpers/groups.ts
The-Daniel, line 1: image
Tickets groups is now broken. Object count is stuck at 0 so you can't create a manual group

The-Daniel (changes requested):

frontend/src/v4/modules/tree/tree.sagas.ts
The-Daniel, line 1: The isolate and hide functions in the toolbar are broken.
## Isolate
image

## Hide
image

frontend/src/v4/modules/tree/tree.sagas.ts
```diff
@@ -592,8 +597,8 @@ function* setTreeNodesVisibility({ nodesIds, visibility }) {
 function* setSelectedNodesVisibility({ nodeId, visibility }) {
 	yield waitForTreeToBeReady();
 
-	const fullySelectedNodes = yield select(selectFullySelectedNodesIds);
-	const hasSelectedNodes = !!fullySelectedNodes.length;
+	const delectedNodes = yield select(selectSelectedNodes);
```
The-Daniel, line 600: Delectable

The-Daniel (changes requested):

frontend/src/v4/modules/tree/tree.sagas.ts
The-Daniel: It's delectable again

The-Daniel (dismissed):

frontend/src/v4/helpers/groups.ts
```diff
@@ -18,7 +18,7 @@
 import {  isEqual, omit, pick } from 'lodash';
 import { select } from 'redux-saga/effects';
 import { GROUP_TYPES_ICONS, GROUPS_TYPES } from '../constants/groups';
-import { selectGetNumNodesByMeshSharedIdsArray } from '../modules/tree';
+import { selectNumNodesByMeshSharedIdsArray } from '../modules/tree';
```
sanmont3drepo, line 21: I still can reproduce the original bug

sanmont3drepo (changes requested):

carmenfan (approved):

The-Daniel (approved):

sanmont3drepo (approved):

PR #4925, opened 2024-04-24
ISSUE #4889 when deleting a model, remove ticket counter for tickets within the model

### LINKED ISSUE #4889: Remove ticket counter entry if the container has been removed
### Description
With #4851, we've introduced a new collection within a teamspace to deal with ticket counters. But we didn't add logic to remove the entry if the container has been removed

### Steps to replicate
- create a container and generate some tickets
- remove the container
- inspect `tickets.counter` collection
### Current Behaviour
- counters for the remove container remains

### Expected Behaviour
- counters associated with the removed container should also be removed.

carmenfan:
This fixes #4889

#### Description
When removing a model (or project), remove ticket counter entries associated with the model

#### Test cases
- counter entries associated with the model should be removed from the database
- other counters should remain

### DIFF
--- a/backend/src/v5/models/tickets.js
+++ b/backend/src/v5/models/tickets.js
@@ -133,8 +133,13 @@ Tickets.updateTickets = async (teamspace, project, model, oldTickets, data, auth
 	return changeSet;
 };
 
-Tickets.removeAllTicketsInModel = async (teamspace, project, model) => {
-	await DbHandler.deleteMany(teamspace, TICKETS_COL, { teamspace, project, model });
+Tickets.removeAllTicketsInModel = (teamspace, project, model) => {
+	// eslint-disable-next-line security/detect-non-literal-regexp
+	const counterRegex = new RegExp(`^${UUIDToString(project)}_${model}_.*`);
+	return Promise.all([
+		DbHandler.deleteMany(teamspace, TICKETS_COL, { teamspace, project, model }),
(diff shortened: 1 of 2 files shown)


backend/src/v5/models/tickets.js
```diff
@@ -133,8 +133,13 @@ Tickets.updateTickets = async (teamspace, project, model, oldTickets, data, auth
...
 };
 
-Tickets.removeAllTicketsInModel = async (teamspace, project, model) => {
-	await DbHandler.deleteMany(teamspace, TICKETS_COL, { teamspace, project, model });
+Tickets.removeAllTicketsInModel = (teamspace, project, model) => {
+	// eslint-disable-next-line security/detect-non-literal-regexp
+	const counterRegex = new RegExp(`^${UUIDToString(project)}_${model}_.*`);
+	return Promise.all([
+		DbHandler.deleteMany(teamspace, TICKETS_COL, { teamspace, project, model }),
+		DbHandler.deleteMany(teamspace, TICKETS_COUNTER_COL, { _id: counterRegex }),
```
ChristosTsiotsias, line 141: we are only removing one record in the collection so maybe we could use deleteOne but this works too

ChristosTsiotsias (changes requested):

ChristosTsiotsias (approved):

PR #4927, opened 2024-04-24
ISSUE #4881 Fix error on updating tickets with groups

### LINKED ISSUE #4881: Updating ticket with view-groups throws error
### Description
Updating any property in a ticket that has a ticket view property populated with groups throws an error.

The frontend seems to be correctly sending the update data for the ticket and the backend replies with the error in the picture at the bottom.
This suggests the problem is in the backend

### Steps to replicate
- Open a ticket in the viewer that has default view (or any view porperty) with at least one group
- Try to edit and save any other property in the ticket

### Current Behaviour
- An error modal is triggered

### Expected Behaviour
- The ticket should update successfully

image

carmenfan:
This fixes #4881

#### Description
The logic that checks the groups that are still being used still exist is querying the database with group ids as string instead of UUIDs. Resulting in the groups not being found.

#### Test cases
Updating tickets with groups should now work.


backend/src/v5/processors/teamspaces/projects/models/commons/tickets.js
```diff
@@ -157,7 +157,7 @@ const processExternalData = async (teamspace, project, model, ticketIds, data) =
 		if (groups.stillUsed.size) {
 			const stillUsed = Array.from(groups.stillUsed);
 			const existingGroups = await getGroupsByIds(teamspace, project, model, ticketId,
-				stillUsed, { _id: 1 });
+				stillUsed.map(stringToUUID), { _id: 1 });
```
ChristosTsiotsias, line 160: it works, but isnt there a unit test that has to be updated ?

ChristosTsiotsias (changes requested):

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.js
carmenfan: I guess there isn't because there's no failed test cases. but yes we should probably make sure there's a check in place for that..

I'll have a look 😢

ChristosTsiotsias (approved):

ChristosTsiotsias:
looks good but there are some tests failing

carmenfan:
Re-ran the test and it passed, seems like it was just a hiccup on Github.

Merging

PR #4934, opened 2024-04-25
Issue 4917 - When dates are set to null, they shoudl be removed instead of being set to 0

### LINKED ISSUE #4917: Custom tickets: sending a null on a date property is being treated as a timestamp of the value 0
### Description
Sending an update request with a date value set to null (in anticipation of dropping the value), is being converted to a timestamp value of 0 (i.e. being trated as 1970/07/01 UTC)

### Steps to replicate
- Create a ticket with an optional date field
- Enter some values, save the ticket
- update the ticket with date value set to null

### Current Behaviour
- is being transformed to 0

### Expected Behaviour
- should unset the date field

carmenfan:
This fixes #4917
#### Description
Avoid transforming nulls to a date in the middleware

#### Test cases
- when updating a date to null (i.e. removing the property, the property shoudl be removed now.

### DIFF
--- a/backend/src/v5/schemas/tickets/index.js
+++ b/backend/src/v5/schemas/tickets/index.js
@@ -269,7 +269,7 @@ const generateCastObject = ({ properties, modules }, stripDeprecated) => {
 			if (stripDeprecated && deprecated) {
 				res[name] = Yup.mixed().strip();
 			} else if (type === propTypes.DATE || type === propTypes.PAST_DATE) {
-				res[name] = Yup.number().transform((_, val) => val.getTime());
+				res[name] = Yup.number().transform((_, val) => (val === null ? val : val?.getTime())).nullable();
 			} else if (type === propTypes.VIEW) {
 				res[name] = Yup.object({
 					screenshot: uuidString,
--- a/backend/tests/v5/unit/schemas/tickets/index.test.js
+++ b/backend/tests/v5/unit/schemas/tickets/index.test.js
@@ -18,6 +18,7 @@ const { times, cloneDeep } = require('lodash');
 
 const { src, image } = require('../../../helper/path');
 const {
+	determineTestGroup,
 	generateGroup,
 	generateRandomString,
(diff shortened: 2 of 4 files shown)


backend/src/v5/schemas/tickets/index.js
```diff
@@ -269,7 +269,7 @@ const generateCastObject = ({ properties, modules }, stripDeprecated) => {
 			if (stripDeprecated && deprecated) {
 				res[name] = Yup.mixed().strip();
 			} else if (type === propTypes.DATE || type === propTypes.PAST_DATE) {
-				res[name] = Yup.number().transform((_, val) => val.getTime());
+				res[name] = Yup.number().transform((_, val) => (val === null ? val : val?.getTime())).nullable();
```
ChristosTsiotsias, line 272: theoretically we cannot end up with a null date in the database isnt it ?

backend/tests/v5/unit/schemas/tickets/index.test.js
```diff
@@ -1333,8 +1334,34 @@ const testDeserialiseUUIDsInTicket = () => {
...
+			expect(TicketSchema.serialiseTicket(ticket, template)).toEqual(ticket);
+		});
+	});
+};
+
+describe(determineTestGroup(__filename), () => {
 	testValidateTicket();
 	testProcessReadOnlyValues();
 	testDeserialiseUUIDsInTicket();
+	testDeserialiseUUIDsInTicket();
```
ChristosTsiotsias, line 1365: we are calling this twice

ChristosTsiotsias (changes requested):

backend/src/v5/schemas/tickets/index.js
carmenfan: yes, but this is also called by eventManager to serialise an update json for the chat server, so we will get situations where this comes in as null

ChristosTsiotsias (approved):

Issue #536, opened 2017-10-19
Measuring tool button still shown as toggled on even though the functionality has been disabled


carmenfan:
* Measure something 
* Go to some other model
* Measuring tool toggle is still on even though you can't measure.

JamesMilnerUK:
Fixed in #533

PR #4986, opened 2024-05-30
ISSUE #4977 Isolate model from tree card error

### LINKED ISSUE #4977: Isolate model from tree card error
### Description
Issue: In federation if we try to isolate one model from Tree card, its giving below error and 3DRepo restarting.

### Steps to replicate
- Isolate one model from tree card

### Current Behaviour
- In federation if we try to isolate one model from Tree card, its giving below error and 3DRepo restarting.

### Expected Behaviour
- No error

image

FJThiel:
This fixes #4977 

#### Description
Switched centreToPoint and zoomToObjects over to sending Mesh IDs in batches to avoid crashing Unity.

Needs to be merged together with [PR #536](https://github.com/3drepo/3drepounity/pull/536) of the viewer to work

#### Test cases
Load large federations and isolate a high-level node. The viewer may slow down, but should not crash.


sebjf (changes requested):
I think if you make it so `Centre/ZoomToObjectEnd` doesn't take an argument, this can be simplified quite a bit, to something like this:

```
public static zoomToObjects(meshEntries: object[]) {
	meshEntries.forEach((entry) => {
		UnityUtil.multipleCallInChunks(entry.meshes.length, (start,end) => {
			const params = {
				teamspace: entry.teamspace,
				modelId: entry.modelId,
				meshes: entry.meshes.slice(start, end),
			};
			UnityUtil.toUnity('ZoomToObjectsAppend', UnityUtil.LoadingState.MODEL_LOADED, JSON.stringify(params));
		});
	}
	this.unityOnUpdateActions.push(() => {
		UnityUtil.toUnity('ZoomToObjectsEnd', UnityUtil.LoadingState.MODEL_LOADED);
	});
}
```

(With any necessary tweaks to make the closures work correctly, obviously!)

`multipleCallInChunks` itself executes synchronously, so it should be safe to call multiple times on batches of each entry in `meshEntries`, and then append a call to the end of the array it updates (`unityOnUpdateActions`) after the loop.

`unityOnUpdateActions` is not exclusively maintained by `multipleCallInChunks`. For now I think its OK for methods to append it directly, especially because `multipleCallInChunks` doesn't return a promise or anything like that (not that that would help in this case, because it will be called once for each unique teamspace/modelid).

FJThiel:
I simplified it as suggested in [54344d4](https://github.com/3drepo/3drepo.io/pull/4986/commits/54344d4d1956a6448902d510853b2fb35f8c0ea6)

sebjf (dismissed):

sebjf:
Hi @carmenfan, this is approved, but needs a rebuild of the staging viewer now that 535 has been merged.

carmenfan:
@sebjf  updated.

If you can just sanity check that it works as intended, then please approve again and we're good to merge!

sebjf (approved):

PR #4988, opened 2024-05-31
ISSUE #4987 - Fix viewer becoming unresponsive when renavigated

### LINKED ISSUE #4987: [2D MVP] Milestone 2: The viewer becomes non-interactive when navigating
### Description
If you navigate to  the viewer the model loads and you can interact with it normally, after navigating back and forth you can no longer interact with it

- This is happening in: ISSUE_4833

### Steps to replicate
- Navigate to the viewer
- Click on the 3d home icon in breadcrumbs bar
- Click back in the browser

### Current Behaviour
- The viewer is non-interactable

### Expected Behaviour
- should be able to interact with the viewer normally

https://github.com/3drepo/3drepo.io/assets/42032433/89e7a343-5aea-4147-bdb4-ab7628c15b63

The-Daniel:
This fixes #4987

#### Description
This was a result of moving the viewer canvas further down the DOM so that it can easily be resized (e.g. with the 2D splitter).

Before the canvas was always present in the background and so never re-rendered. This changed caused the canvas component to be re-rendered whenever entering the viewer. This would cause the exact same canvas component to be re-added to the DOM.

Now the unity component is only added to the DOM if it doesn't already exist

#### Test cases

Navigate to and from the viewer. See if it remains responsive. Look in the DOM and there should be just one `<canvas />` component

### DIFF
--- a/frontend/src/v4/services/viewer/viewer.tsx
+++ b/frontend/src/v4/services/viewer/viewer.tsx
@@ -115,7 +115,10 @@ export class ViewerService {
 		unityHolder.style['pointer-events'] = 'all';
 
 		this.element.appendChild(this.viewer);
-		this.viewer.appendChild(unityHolder);
+
+		if (!this.viewer.hasChildNodes()) {
+			this.viewer.appendChild(unityHolder);
+		}
 		this.canvas = unityHolder;
 
 		this.unityLoaderScript = document.createElement('script');


frontend/src/v4/services/viewer/viewer.tsx
```diff
@@ -115,7 +115,10 @@ export class ViewerService {
 		unityHolder.style['pointer-events'] = 'all';
 
 		this.element.appendChild(this.viewer);
-		this.viewer.appendChild(unityHolder);
+
+		if (!this.viewer.hasChildNodes()) {
```
Amantini1997, line 119: You can probably move this check at the top of the method body and terminate the method if the condition holds true

Also, there seem to exists a method called `hasInstance` right before this method, probably you can that one

sanmont3drepo (approved):

Amantini1997 (changes requested):

PR #5002, opened 2024-06-06
ISSUE 4991 - Send io-activity logs to Elastic

### LINKED ISSUE #4991: Send io-activity from 3drepo.io instead of determining from logs
### Description
Send io-activity from 3drepo.io instead of determining from logs

### Goals
- [x] Create new index `io-activity` in Elastic
- [x] Send network logs to Elastic

### Tasks
- [x] Send network logs to Elastic

### Related Resources

MockUp : `link to mockup`
Coda doc: `link to coda`

Charence:
This fixes #4991 

#### Description
- NET logs get sent to Elastic if configured

#### Test cases
- Use 3drepo.io with Elastic setup
- Check Elastic for `io-activity` index and records


carmenfan (changes requested):
super quick look, few things:

1. <del>there are 2 loggers - check v5 as well</del> scrap this - I'm thinking responder!
2. this is high coupling.  Can we do it so elastic functionality is its own file (like in bouncerworkerhttps://github.com/3drepo/3drepobouncer/blob/master/tools/bouncer_worker/src/lib/elastic.js) - it should probably live in the service folder in v5 
3. As mentioned in devops meeting, please edit the issue to have a link to the existing script so I can have a read and understand what it does and verify it's doing that here!

Charence:
> 1. I'm thinking responder! But I think this should live in the responder, not the logger.

@carmenfan The v4 and v5 responders seem to be separate (v5 doesn't go through v4). We will need to add this to both v4 and v5 responders?

carmenfan:
> > 1. I'm thinking responder! But I think this should live in the responder, not the logger.
> 
> @carmenfan The v4 and v5 responders seem to be separate (v5 doesn't go through v4). We will need to add this to both v4 and v5 responders?

@Charence correct - hence elastic should be a service that both responder calls

backend/src/v5/services/elastic.js
```diff
@@ -0,0 +1,46 @@
...
+	const elasticBody = {
+		status: parseInt(status, 10),
+		code,
+		latency: parseInt(latency, 10),
+		contentLength: parseInt(contentLength, 10),
+		user,
+		method,
+		originalUrl,
+		timestamp,
+		host,
```
carmenfan, line 40: Host may be annoying since that would mean we'd get ISSUE-xxx.dev. etc for dev records. We may want to keep this, but we certainly want to have the ability to group all dev deployments into one

Can we make a namespace inside the elastic section of the config?

backend/tests/v5/unit/services/elastic.test.js
```diff
@@ -0,0 +1,82 @@
...
+				'io-activity', elasticBody, id);
+		});
+	});
+
+	afterAll(() => {
+		jest.useRealTimers();
+	});
+};
+
+describe('services/elastic', () => {
```
carmenfan, line 80: This should use[ the helper function](https://github.com/3drepo/3drepo.io/blob/master/backend/tests/v5/helper/services.js#L385) in services to determine the naming

carmenfan (changes requested):

backend/src/v4/handler/elastic.js
```diff
@@ -94,10 +94,14 @@ const establishIndices = async (client)=>{
...
 const elasticClientPromise = createElasticClient();
 
-const createElasticRecord = async (index, body, id) => {
+Elastic.createElasticRecord = async (index, body, id) => {
 	try {
 		const elasticClient = await elasticClientPromise;
 		if (elasticClient && body) {
+			const { namespace } = elasticConfig;
+			if (namespace) {
+				body.namespace = namespace;
```
carmenfan, line 103: do we need to do anything to `establishIndices`?

We seem to do that to ensure the indices are created, but that object doesnt' have the additional `namespace` value.

We also don't do this for the io-activity stuff...

carmenfan: also can we call it `Namespace` since everything seems to begin with a capital letter (other than license...)

carmenfan: Can we also add the hostname by default too? Instead of just adding it to io-activity.

carmenfan (changes requested):

backend/src/v4/handler/elastic.js
Charence: On the existing `prod` `io-activity` all field names are lowercase.

Index mapping for `io-activity` added. By default, it will create the mapping as data is added, but by enforcing a mapping we can "define how each field should be treated by Elasticsearch, whether it should be tokenized and filtered, or whether it should be a specific type like integer, date, etc."

`host` added by default on record creation if it exists in `config`.

image

carmenfan: since w'ere adding host and namespace to login activity as well can we update the ensure index fit to match?

carmenfan (changes requested):

carmenfan (approved):

PR #5008, opened 2024-06-11
Issue 4875 - Changing v4 groups too fast can overlap colour overrides

### LINKED ISSUE #4875: Changing (v4) group too fast can overlap colour overrides
### Description
Using the arrows at the top of the card, it is possible to change groups pretty fast. If this is too fast, it might happen that the colours of one group are not cleared and the new colour overrides are "coloured" on top of the previous ones

### Steps to replicate
- Open a model in the viewer and expand the groups card (make sure to have at least 2 manual groups and that they have different meshes)
- Expand either group
- using the arrows, switch groups at least twice really fast

### Current Behaviour
- It is possible to see the colour overrides from both the current and the previous group

### Expected Behaviour
- Only the colour overrides from the current group should be visible

(for the sake of simplicity, I used the console to fastly switch between the groups)
https://github.com/3drepo/3drepo.io/assets/43547703/41b4e5a2-89eb-4bd5-9cad-d8396d1b80ce

Amantini1997:
This fixes #4875

#### Description
The saga responsible for the main call ("showDetails") now acts as a queue leveraging the actionChannel method from redux-saga. On top of that, some other calls have replace the "put" with a "call" to guarantee the order of the calls.

#### Test cases
- Open a model in the viewer and expand the groups card (make sure to have at least 2 manual groups and that they have different meshes)
- Expand either group
- Using the arrows, switch groups at least twice really fast (or use the code below while looking at an expanded group)

```js
const arrow = document.querySelector("#groups-card-details > div > div > div > div > button:nth-child(3)");
arrow.click();
arrow.click();
```


frontend/src/v4/modules/groups/groups.sagas.ts
```diff
@@ -93,12 +93,13 @@ function* resetActiveGroup() {
 
 function* highlightGroup({ group }) {
 	try {
-		const color = group.color ? hexToGLColor(group.color) : Viewer.getDefaultHighlightColor();
+		const colour = group.color ? hexToGLColor(group.color) : Viewer.getDefaultHighlightColor();
```
The-Daniel, line 96: Nice!

The-Daniel (dismissed):

sanmont3drepo (changes requested):

frontend/src/globals/unity-util.ts
```diff
@@ -20,6 +20,7 @@
 
 import { IndexedDbCache } from './unity-indexedbcache';
 import { ExternalWebRequestHandler } from './unity-externalwebrequesthandler';
+import { getWaitablePromise } from '@/v5/helpers/async.helpers';
```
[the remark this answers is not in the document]
carmenfan: what are you guys trying to do here?

If you are trying to figure out if the operation has completed, this doesn't really do it. All this guarantee is that we've sent the message to unity, whether the operation has been processed is still non deterministic at the point of this promise resolving.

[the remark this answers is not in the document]
Amantini1997: @Carmen Is there a way to tell when unity completes the execution?

frontend/src/globals/unity-util.ts
```diff
@@ -2112,14 +2113,19 @@ export class UnityUtil {
 	 */
 	public static multipleCallInChunks(arrLength: number, func:(start: number, end: number) => any, chunkSize = 5000) {
 		let index = 0;
+		const { promiseToResolve, resolve } = getWaitablePromise();
```
sanmont3drepo, line 2116: I think you can just use a plain Promise for this.

sanmont3drepo (changes requested):

sanmont3drepo (approved):

The-Daniel (dismissed):

frontend/src/globals/unity-util.ts
```diff
@@ -1543,7 +1543,7 @@ export class UnityUtil {
 		toggleMode: boolean,
 		forceReHighlight: boolean,
 	) {
-		UnityUtil.multipleCallInChunks(idArr.length, (start, end) => {
+		return UnityUtil.multipleCallInChunks(idArr.length, (start, end) => {
```
carmenfan, line 1546: how does changing the code in highlight fix an issue with colour overrides?

If we do this then shouldn't we do this on everything that is related? (e.g. visibility, colour override, transparency)

frontend/src/globals/unity-util.ts
```diff
@@ -2111,15 +2111,18 @@ export class UnityUtil {
...
+			let index = 0;
+			while (index < arrLength) {
+				const end = index + chunkSize >= arrLength ? undefined : index + chunkSize;
+				const i = index; // For the closure
+				this.unityOnUpdateActions.push(() => {
+					func(i, end);
+				});
+				index += chunkSize;
+			}
+			this.unityOnUpdateActions.push(resolve);
```
carmenfan, line 2124: @sebjf  can you check if the change in this loop makes any sense?

carmenfan (changes requested):

frontend/src/globals/unity-util.ts
sebjf: Hi @carmenfan the loop looks sensible to me, and should do what was discussed in the thread before (resolve a promise when all the chunks have been sent to unity (and its up to the caller to interpret the significance of that, based on whether the calls inside unity are synchronous or not).

frontend/src/globals/unity-util.ts
sanmont3drepo: I think the issue title is a bit wrong. I thought the same thing when Alessandro talked to me about it, but he was refering to the hghlights that are shown when selecting a group, not the the actual mesh override. 

image

carmenfan: but i'm thinking if this is an issue for highlight, woudl it not be an issue for others? (e.g. if you flip through issues/risks/tickets)

should we not be changing the behaviour for all?

It's a bit weird that highglith returns a promise but not override/visibilty etc

sanmont3drepo: My guess is that the problem is there but we haven't found a consistent way of reproducing it for the rest. Or maybe the delay when loading is preventing the thing to manifest. Again, these are all guesses.

carmenfan: either way I think however way we deal with highlights should be however way we deal with other actions that are in the same category.

if a highlight request is now being waited on, then colour overrides/inivisbile should be the same. Or at least, the unity utils should return a promise in similar manner even if you completely ignore it on the front end.

We have to consider that this is a third party facing API here as well, it'll be difficult to explain to customers why this one has a promise you can wait on to ensure the action has completed but the other 2 doesn't

which reminds me, can we also update the documentation comments accordingly please?

sanmont3drepo: @Amantini1997  so, update unity  util and the comment docs within the file to use promises and have a consistent api.

sanmont3drepo (approved):

carmenfan (approved):

PR #5032, opened 2024-06-26
Issue 4983 - Drawing revision endpoints

### LINKED ISSUE #4983: [2D MVP] Milestone 1: Endpoints for revision management
### Description
This is part of https://github.com/3drepo/3D-Repo-Product-Team/issues/377.

Create endpoints to process drawings revisions.

##### New revision of drawings
endpoint: `POST /teamspaces/{teamspace}/projects/{project}/drawings/{drawing}/revisions`
description: Post a new file (pdf or dwg) as a new revision to the drawing
permissions: collaborator+
payload: 
````js
// Multipart form with:
statusCode: Should be one of the known status codes in the get status code endpoint, required
revCode: string, max length 10, alphanumeric ,fullstop,hypens and underscores
desc: string (shortDescription)
file: <dwg or pdf file via multipart>
````

NOTE:
- maximum file size supported should be 500MB
- The combination of StatusCode + RevCode needs to be unique (This is the equivalent of `tag` in container)
- Processing of the drawing should follow the same process as 3d models (i.e. we don't immediately denote the processing is complete, it goes through the processing status that gets stored and send back by socket.io.
  - for the purpose of this issue - **processing is out of scope** - so just pretend it's completed as soon as we've stored the file.
- This should contribute to the quota calculation (just like containers)
- This should send notification of the status through socket.io just like 3D model processing

##### Get revision list
endpoint: `GET /teamspaces/{teamspace}/projects/{project}/drawings/{drawing}/revisions`
description: provide the full list of revisions uploaded within this drawing container
permissions: viewer+
sample response:
````js
{
  "revisions": [
    {
      "_id": "ef0855b6-4cc7-4be1-b2d6-c032dce7806a",
      "author": "someUser",
      "timestamp": 12345678,
      "format": ".dwg",
      "statusCode": "s1",
      "revCode": "P01,
      "description": "Level 2 floor plan",
      "void": true //omitted when false
    }
  ]
}
````

##### Update revision
endpoint: `PATCH /teamspaces/{teamspace}/projects/{project}/drawings/{drawing}/revisions/{revision}`
description: Update a revision - just like containers, you should only be able to void or unvoid a revision.
permissions: collaborator+
payload:
````js
{
  "void": true
}
````

##### Get original files
endpoint: `GET /teamspaces/{teamspace}/projects/{project}/drawings/{drawing}/revisions/{revision}/files/original`
description: Get the original file (the PDF/DWG) for that particular revision
permissions: collaborator+
response:  binary of the original file

##### Get status codes list
see https://github.com/3drepo/3drepo.io/issues/5023

###### Revision records
Revision records should be stored in `drawings.history`:
````js
// Example drawing revision
{
    "_id" : LUUID("cad0c3fe-dd1d-4844-ad04-cfb75df26a63"),
    "project": LUUID("bfe0c3fe-dd1d-4844-ad04-cfb75df26a63"),
    "model": LUUID("aeb0c3fe-dd1d-4844-ad04-cfb75df26a63"),
    "author" : "carmen",
    "statusCode" : "S1",
    "revCode": "abc,
    "timestamp" : ISODate("2023-11-28T19:11:55.000Z"),
    "format": "dwg",
    "rFile" : [ 
        LUUID("0ac71fb2-a18d-4ef8-b5a8-b87879326300")
    ]
}
````
NOTE:
- Unlike containers, every drawing within a teamspace will be stored in the same collection, with project/model properties to denote which project/model they belong to.
- Ensure we add to the delete drawing function to remove revisions
- `rFile` should be links to original files, which should be a reference to a file. The reference should be stored in `drawings.history.ref`
  - the original filename should be a metadata within the reference, instead of the `<UUID>_original_file_name` approach taken in containers.
  - We will record the file extension in the object instead of having to regex it out of the rFile string like in containers.
  - This will also contain the `incomplete` and `void` property like container to denote the state of the model processing.

###### Original file
As stated above, the original file will be stored in `drawings.history.ref` with the original file name as part of the metadata:
````
// drawings.history.ref
{
    "_id" : LUUID("0ac71fb2-a18d-4ef8-b5a8-b87879326300"),
    "type" : "fs",
    "link" : "197/204/f959eda6-3fde-436a-ba0e-b1aa777a413d",
    "size" : 2419670,
    "name": "Floor1.dwg",
    "project": LUUID("bfe0c3fe-dd1d-4844-ad04-cfb75df26a63"),
    "model": LUUID("aeb0c3fe-dd1d-4844-ad04-cfb75df26a63"),
    "rid" : LUUID("cad0c3fe-dd1d-4844-ad04-cfb75df26a63"),
}
````

### Goals
-  As a collaborator, I want to be able to declare a revision as void if I have made a mistake
-  As a collaborator, I want to be able to un-void a revision if I mis-clicked.
-  As an admin, I want to quickly create a drawing container whilst I am uploading a drawing
-  As a collaborator, I want to be able import a 2D drawing (PDFs, DWGs) into 3D Repo
-  As a collaborator, I want to be stopped from uploading the drawing if there is already an existing revision with the same status code and revision code
- As a collaborator, I want to be able to download the original drawing file

##### New revision

- [x] Add validation
- [x] Create route method
- [x] Create processor method
- [x] Create model method
- [x] Update quota to include drawings
- [x]  Create unit & e2e tests
- [x] Update swagger doc

##### Update revision

- [x] Add validation
- [x] Create route method
- [x] Create processor method
- [x] Create model method
- [x]  Create unit & e2e tests
- [x] Update swagger doc

##### Get revision list

- [x] Create route method
- [x] Create processor method
- [x] Create model method
- [x]  Create unit & e2e tests
- [x] Update swagger doc

##### Get original files

- [x] Create route method
- [x] Create processor method
- [x] Create model method
- [x]  Create unit & e2e tests
- [x] Update swagger doc

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #4983

#### Description

Endpoints added:

- New drawing revision
- Update drawing revision status
- Get drawing revision list
- Get original drawing revision file
- Get revision status codes

Swagger documentation, unit & e2e tests added for all endpoints

#### Test cases
All endpoints should be correctly documented and work as intended

### CODE REVIEW

File: backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/drawings.js [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+			Yup.string().oneOf(statusCodes.map(({ code }) => code)).required(),
+		).required(),
+		revCode: Yup.string().min(1).max(10).matches(/^[\w|_|-|.]*$/,
+			// eslint-disable-next-line no-template-curly-in-string
+			'${path} can only contain alpha-numeric characters, full stops, hyphens or underscores')
+			.required(),
+		desc: YupHelper.types.strings.shortDescription,
+	};
+
+	const schema = Yup.object().noUnknown().required().shape(schemaBase)
```
carmenfan, line 45:
```suggestion
	const schema = Yup.object(schemaBase).noUnknown().required()
```

You could just do this

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/drawings.js
```diff
@@ -0,0 +1,65 @@
...
+		.test('check-status-rev-code-uniqueness', 'The combination of statusCode and revCode needs to be unique', ({ revCode, statusCode }) => {
+			if (revCode && statusCode) {
+				const { teamspace, drawing } = req.params;
+				return isRevAndStatusCodeUnique(teamspace, drawing, revCode, statusCode);
+			}
+
+			return true;
```
carmenfan, line 52: I guess this never happens, but if it for some reason does it should probably return false?

backend/src/v5/models/revisions.js
```diff
@@ -25,13 +28,13 @@ const Revisions = {};
 const excludeVoids = { void: { $ne: true } };
 const excludeIncomplete = { incomplete: { $exists: false } };
 
-const collectionName = (model) => `${model}.history`;
+const collectionName = (modelType, model) => (modelType === modelTypes.DRAWING ? DRAWINGS_HISTORY_COL : `${model}.history`);
```
carmenfan, line 31: Can we scrap the constants declaration and if it's DRAWING then just do `{modelTypes}.history`

This is probably the approach we'll move towards on containers/fed

backend/src/v5/models/revisions.js
```diff
@@ -42,43 +45,81 @@ const findOneRevisionByQuery = async (teamspace, model, query, projection, sort)
...
+		await findOneRevisionByQuery(teamspace, model, undefined, { tag });
+		return false;
+	} catch {
+		return true;
+	}
+};
+
+Revisions.isRevAndStatusCodeUnique = async (teamspace, model, revCode, statusCode) => {
+	try {
+		await findOneRevisionByQuery(teamspace, undefined, modelTypes.DRAWING, { revCode, statusCode });
```
carmenfan, line 122: we need model to be part of the query.

We only want it to be unique within the same drawing

backend/src/v5/models/revisions.js
```diff
@@ -42,43 +45,81 @@ const findOneRevisionByQuery = async (teamspace, model, query, projection, sort)
...
+		_id: generateUUID(),
+		project,
+		model,
+		timestamp: new Date(),
+		...data,
+	};
+
+	await db.insertOne(teamspace, collectionName(modelType, model), newRevision);
+
+	publish(events.NEW_REVISION, { teamspace, project, model, modelType, revision: newRevision._id });
```
carmenfan, line 80: This doesn't feel like is the right time for `NEW_REVISION`

This should only trigger when the processing has completed

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -34,7 +35,7 @@ const getObjectArrayFromRules = async (teamspace, project, model, revId, rules,
 	let revision = revId;
 
 	if (revision) {
-		const rev = await getRevisionByIdOrTag(teamspace, model, revision, { _id: 1 });
+		const rev = await getRevisionByIdOrTag(teamspace, model, modelTypes.CONTAINER, revision, { _id: 1 });
```
carmenfan, line 38: This is going to be very bug prone as it is either container or federation

carmenfan: I think you'll need  to pass the model type in htis function

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.groups.js
```diff
@@ -123,7 +124,7 @@ const convertToMeshIds = async (teamspace, project, revId, containerEntry) => {
 
 	let revision = revId;
 	if (revision) {
-		const rev = await getRevisionByIdOrTag(teamspace, container, revision, { _id: 1 });
+		const rev = await getRevisionByIdOrTag(teamspace, container, modelTypes.CONTAINER, revision, { _id: 1 });
```
carmenfan, line 127: same with this one

backend/src/v5/models/revisions.js
```diff
@@ -25,13 +28,13 @@ const Revisions = {};
 const excludeVoids = { void: { $ne: true } };
 const excludeIncomplete = { incomplete: { $exists: false } };
 
-const collectionName = (model) => `${model}.history`;
+const collectionName = (modelType, model) => (modelType === modelTypes.DRAWING ? DRAWINGS_HISTORY_COL : `${model}.history`);
 
-const findRevisionsByQuery = (teamspace, model, query, projection, sort) => db.find(teamspace,
-	collectionName(model), query, projection, sort);
+const findRevisionsByQuery = (teamspace, model, modelType, query, projection, sort) => db.find(teamspace,
```
carmenfan, line 33: Can we review all these functions....

Drawings version will not work well if we don't pass model id into the query.

carmenfan: These should start working a bit like custom tickets functions where we pass in the project/model etc into the query if the model type is drawings

backend/src/v5/processors/teamspaces/projects/models/drawings.js
```diff
@@ -35,7 +40,48 @@ Drawings.addDrawing = (teamspace, project, data) => addModel(teamspace, project,
...
+
+	await Promise.all([
+		deleteModelRevisions(teamspace, project, drawing, modelTypes.DRAWING),
+		deleteModel(teamspace, project, drawing),
+		removeModelFromProject(teamspace, project, drawing),
+	]);
+};
+
+Drawings.getRevisions = async (teamspace, drawing, showVoid) => {
+	const revisions = await getRevisions(teamspace, drawing, modelTypes.DRAWING, showVoid,
```
carmenfan, line 54: Can just remove async from function declaration and do this
```suggestion
	return getRevisions(teamspace, drawing, modelTypes.DRAWING, showVoid,
```

backend/src/v5/processors/teamspaces/projects/models/drawings.js
```diff
@@ -35,7 +40,48 @@ Drawings.addDrawing = (teamspace, project, data) => addModel(teamspace, project,
...
+		{ _id: 1, author: 1, format: 1, timestamp: 1, statusCode: 1, revCode: 1, void: 1, desc: 1 });
+
+	return revisions;
+};
+
+Drawings.newRevision = async (teamspace, project, drawing, data, file) => {
+	const format = `.${file.originalname.split('.').splice(-1)[0].toLowerCase()}`;
```
carmenfan, line 61: just use `path` 
https://www.w3schools.com/nodejs/met_path_extname.asp

backend/src/v5/processors/teamspaces/projects/models/drawings.js
```diff
@@ -35,7 +40,48 @@ Drawings.addDrawing = (teamspace, project, data) => addModel(teamspace, project,
...
+
+	const fileId = generateUUID();
+	const revId = await addRevision(teamspace, project, drawing, modelTypes.DRAWING,
+		{ ...data, format, rFile: [fileId], status: STATUSES.PROCESSING });
+
+	const fileMeta = { name: file.originalname, rid: revId, project, model: drawing };
```
carmenfan, line 67: @sebjf  are you calling it `rev_id`?

If so then @ChristosTsiotsias  please call this `rev_id` as well for consistency

backend/src/v5/processors/teamspaces/projects/models/drawings.js
```diff
@@ -35,7 +40,48 @@ Drawings.addDrawing = (teamspace, project, data) => addModel(teamspace, project,
...
+
+	return revisions;
+};
+
+Drawings.newRevision = async (teamspace, project, drawing, data, file) => {
+	const format = `.${file.originalname.split('.').splice(-1)[0].toLowerCase()}`;
+
+	const fileId = generateUUID();
+	const revId = await addRevision(teamspace, project, drawing, modelTypes.DRAWING,
+		{ ...data, format, rFile: [fileId], status: STATUSES.PROCESSING });
```
carmenfan, line 65: Unless i'm missing something we're not firing the processing signal to chat.

We need to do this

backend/src/v5/processors/teamspaces/projects/models/drawings.js
```diff
@@ -35,7 +40,48 @@ Drawings.addDrawing = (teamspace, project, data) => addModel(teamspace, project,
...
+
+	const fileMeta = { name: file.originalname, rid: revId, project, model: drawing };
+	await storeFile(teamspace, DRAWINGS_HISTORY_REF_COL, fileId, file.buffer, fileMeta);
+
+	await updateRevision(teamspace, drawing, modelTypes.DRAWING, revId, { status: STATUSES.OK });
```
carmenfan, line 70: same here,  I don't think your update Revision fire a the right status update? have a sanity check there

backend/src/v5/routes/teamspaces/projects/models/drawings/revisions.js
```diff
@@ -0,0 +1,326 @@
+/**
```
carmenfan, line 1: how different is this file vs the container one?

Should we be moving this to common (and just don't cater for federations?)

backend/tests/v5/e2e/services/chat/modelEvents/revisions.test.js
```diff
@@ -99,6 +109,32 @@ const revisionUpdateTest = () => {
```
carmenfan, line 109: I think we're missing a new revision test here?

backend/tests/v5/scripts/scene/removeIncompleteRevisions.test.js
```diff
@@ -38,6 +38,8 @@ const { stringToUUID, UUIDToString, generateUUID } = require(`${src}/utils/helpe
 const { count, findOne, insertMany } = require(`${src}/handler/db`);
 const Path = require('path');
 
+const { modelTypes } = require(`${src}/models/modelSettings.constants`);
```
carmenfan, line 41: I think this test (and the associated script it's testing) needs modifying to remove incomplete drawings as well.

carmenfan (changes requested):

backend/src/v5/processors/teamspaces/projects/models/drawings.js
sebjf: Hi @carmenfan, @ChristosTsiotsias, yes it is being called `rev_id` to match the existing definitions.

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/drawings.js
```diff
@@ -0,0 +1,65 @@
...
+const { isRevAndStatusCodeUnique } = require('../../../../../../models/revisions');
+const { respond } = require('../../../../../../utils/responder');
+const { singleFileUpload } = require('../../../../multer');
+const { statusCodes } = require('../../../../../../models/modelSettings.constants');
+const { validateMany } = require('../../../../../common');
+
+const Drawings = {};
+
+const ACCEPTED_DRAWING_EXT = ['.dwg', '.pdf'];
+const DRAWING_MAX_FILE_SIZE = 500000000;
```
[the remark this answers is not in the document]
ChristosTsiotsias: multer website says that its in bytes so the conversion gives me 500000000

image

the other file sizes in config have similar numbers 🤔 
image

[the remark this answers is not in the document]
carmenfan: I don't mean pass in MB, I mean it when I said 500MB i mean 500MiB

so it should be 500 * 1024 * 1024 instead of 500 * 1000 * 1000 (hence those existing numbers are not rounded)

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/drawings.js
ChristosTsiotsias: tests are executed first so if revCode or statusCode is missing and I return false in the test then the error thrown is going to be "The combination of statusCode and revCode needs to be unique". So I decided to return true and then the payload will be rejected due to the required()

backend/src/v5/routes/teamspaces/projects/models/drawings/revisions.js
ChristosTsiotsias: I've done this but I've had to change the downloadOriginalFiles drawing endpoint from 

/teamspaces/{teamspace}/projects/{project}/drawings/{drawing}/revisions/{revision}/files/original
to
/teamspaces/{teamspace}/projects/{project}/drawings/{drawing}/revisions/{revision}/files

to match the containers one... are we ok with this ?

carmenfan: Can we keep them different?

Reason being, the svgs are going to get pulled from `/teamspaces/{teamspace}/projects/{project}/drawings/{drawing}/revisions/{revision}/files/image`

So it's going to be very ambiguous

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/drawings.js
carmenfan: fair enough 👍 

Can we make a comment here so if I forget and look into this I don't get confused again?

backend/src/v5/utils/helper/yup.js
```diff
@@ -26,9 +26,9 @@ const YupHelper = { validators: {}, types: { strings: {} }, utils: {} };
 
 YupHelper.utils.stripWhen = (schema, cond) => Yup.lazy((value) => (cond(value) ? schema.strip() : schema));
 
-YupHelper.validators.alphanumeric = (yupObj) => yupObj.matches(/^[\w|_|-]*$/,
+YupHelper.validators.alphanumeric = (yupObj) => yupObj.matches(/^[\w|_|.|-]*$/,
```
carmenfan, line 29: alphanumeric is used in other places, we can't just add fullstop here, use a boolean

backend/src/v5/processors/teamspaces/projects/models/drawings.js
```diff
@@ -35,13 +37,41 @@ Drawings.getDrawingList = async (teamspace, project, user) => {
...
+};
+
 Drawings.addDrawing = (teamspace, project, data) => addModel(teamspace, project,
 	{ ...data, modelType: modelTypes.DRAWING });
 
 Drawings.updateSettings = updateModelSettings;
 
 Drawings.deleteDrawing = async (teamspace, project, drawing) => {
-	await removeFilesWithMeta(teamspace, DRAWINGS_HISTORY_REF_COL, { model: drawing });
+	await removeFilesWithMeta(teamspace, `${modelTypes.DRAWING}s.history.ref`, { model: drawing });
```
carmenfan, line 74: you don't need to append `.ref` to the collection name

carmenfan: should probably declare a constant in the models constant file like you had before...

carmenfan (changes requested):

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/revisions.test.js
```diff
@@ -0,0 +1,478 @@
...
+				['the user is viewer', { ...generateParams(), key: users.viewer.apiKey }, false, templates.notAuthorized],
+				['the user is commenter', { ...generateParams(), key: users.commenter.apiKey }, false, templates.notAuthorized],
+				['the project does not exist', { ...generateParams(), projectId: ServiceHelper.generateRandomString() }, false, templates.projectNotFound],
+				['the model does not exist', { ...generateParams(), modelId: ServiceHelper.generateRandomString() }, false, modelNotFound],
+				['the model is of wrong type', { ...generateParams(), modelId: models.federation._id }, false, modelNotFound],
+				['the file is missing', { ...generateParams(), file: undefined }, false, templates.invalidArguments],
+				['the file has incorrect format', { ...generateParams(), file: image }, false, templates.unsupportedFileFormat],
+				['the file is valid', generateParams(), true],
+				['the file is valid with uppercase extension', { ...generateParams(), file: dwgModelUppercaseExt, modelId: modelWithNoRev._id }, true],
+				...(modelType === modelTypes.DRAWING ? [
```
carmenfan, line 265: Can we add a few more tests around quota?

1. the file doesn't exceed the maximum file size allowed
2. the teamspace has sufficient quota to upload this file

This applies to both containers and drawings

carmenfan: Can we also spllit these case specific tests to new arrays with variables and do `...variableName` for better readability

carmenfan (changes requested):

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/drawings.js
```diff
@@ -0,0 +1,63 @@
...
+
+	try {
+		req.body = await schema.validate(req.body);
+		await next();
+	} catch (err) {
+		respond(req, res, createResponseCode(templates.invalidArguments, err?.message));
+	}
+};
+
+Drawings.validateNewRevisionData = validateMany([singleFileUpload('file', fileFilter(ACCEPTED_DRAWING_EXT), fileUploads.drawingSizeLimit, true), checkQuotaIsSufficient, validateRevisionUpload]);
```
carmenfan, line 61: Can we also have a middleware to use the `fileExtensionFromBuffer` function from `utils/helper/typecheck` to verify the buffer is indeed what the extension say it is?

We already do this for images. From what I can tell the library can understand dwgs

carmenfan (changes requested):

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/drawings.js
ChristosTsiotsias: I have tested 3 dwg models from

image

and the ext I get is undefined. It works every time with pdfs

backend/tests/v5/resources/models/tooBig.dwg
carmenfan, line 1: do we really need these guys or could we fabricate them on the fly before the tests?

If they're a few KB it's not big deal, but e.g. this one it's 19MB -> adding MBs of files into the respository is not ideal, it will slow down cloning (i.e slow down tests and deployments)

carmenfan (changes requested):

backend/src/v5/middleware/dataConverter/multer.js
```diff
@@ -70,6 +70,21 @@ const ensureFileIsImage = async (req, res, next) => {
 	await next();
 };
 
+// Currently only supports pdf
+MulterHelper.ensureFileIsValid = async (req, res, next) => {
```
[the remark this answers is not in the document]
carmenfan: If we're doing a generalised multer like this then we should just chain it with `singleFileUpload` inthe multer middleware here so it's always applied

carmenfan (approved):

PR #5035, opened 2024-06-27
ISSUE #5033 refactor multiple models middleware

### LINKED ISSUE #5033: Address APT-314
This is a dev task to address a bug identified in APT-314.

For details see https://github.com/3drepo/3D-Repo-Product-Team/issues/523

carmenfan:
This fixes #5033

#### Description
Refactored multiple models middleware to facilitate patch requests on multiple models

#### Test cases
case mentioned in the APT ticket should now be resolved.

### DIFF
--- a/backend/src/v4/middlewares/checkPermissions.js
+++ b/backend/src/v4/middlewares/checkPermissions.js
@@ -111,7 +111,7 @@ function checkPermissions(permsRequest) {
 }
 
 function checkMultiplePermissions(permsRequest) {
-	return function (req, res, next) {
+	return async function(req, res, next) {
 		const models = [];
 
 		// POST request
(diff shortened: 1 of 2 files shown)


backend/src/v4/middlewares/checkPermissions.js
```diff
@@ -124,22 +124,29 @@ function checkMultiplePermissions(permsRequest) {
...
+			models.push(...req.body.flatMap((entry)=> entry?.model || []));
+		}
+		try {
+			await validateUserSession(req, res);
+
 			const username = req.session.user.username;
 			const account = req.params.account;
 
-			const promises = [];
+			if(!models.length) {
```
ChristosTsiotsias, line 137: could this check be before the try ? or is there a reason its here ?

backend/src/v4/middlewares/checkPermissions.js
```diff
@@ -124,22 +124,29 @@ function checkMultiplePermissions(permsRequest) {
...
+			const permRes = await Promise.all(models.map((model) => {
+				return checkPermissionsHelper(username, account, null, model, permsRequest, getPermissionsAdapter);
+			}));
 
-			return Promise.all(promises);
-		}).then(validatePermissions.bind(null, next))
-			.catch(err => {
-				next(err);
-			});
+			await validatePermissions(next, permRes);
```
ChristosTsiotsias, line 145: not related to this pr but the other place validatePermissions is called (line 106) is passing next as second param... not sure where its called and if its affecting anything just wanted to point it out 😆

backend/src/v4/middlewares/checkPermissions.js
ChristosTsiotsias: oh I guess its to catch the error and return next(err)....

ChristosTsiotsias (approved):

backend/src/v4/middlewares/checkPermissions.js
carmenfan: yes 😆

backend/src/v4/middlewares/checkPermissions.js
carmenfan: yea L106 is a old javascript syntax , you'll notice that it's calling the [`bind` function](https://www.w3schools.com/js/js_function_bind.asp) on `validatePermissions` instead of calling validation permissions directly there. Long story short the concept of `this` in super confusing in javascript and before ES6 this was the way to get a function called in the context of something else.

So they are actually doing the same thing for the most part and not a bug on either side.

PR #5054, opened 2024-07-08
2D/3D POC demo

carmenfan:
making a PR to spin up a deployment of 2D/3D for demo purposes


carmenfan:
destroying in favour of #5111

PR #5065, opened 2024-07-15
Issue 5055 - Federation getModelStats endpoint now returns ticket count instead of issues/risks count

### LINKED ISSUE #5055: Display number of tickets on the dashboard instead of issues/risks
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/530

With the change mentioned in https://github.com/3drepo/3D-Repo-Product-Team/issues/530, not every teamspaces will have issues/risks available. We would also want all customers to move onto Tickets thus it would make a lot more sense to show the tickets count instead of issues/risks on the dashboard

image

Upon click, it should direct them to the tabular view page with the federation pre-selected.

This will require the[ stats endpoint to change](https://staging.dev.3drepo.io/docs/#/Models/getModelStats) and return the number of open tickets instead.

### Goals
- [ ] As a user I would like to see the number of open tickets on a federation in the dashboard.

ChristosTsiotsias:
This fixes #5055

Federation getModelStats endpoint should now return the number of open tickets amonth the other information it usually returns instead of the number of issues and risks.

If a ticket is based on a template that has custom statuses set then the status (open/closed) of the ticket should be based according to the statuses of this template.

#### Test cases

Create a federation and some custom tickets in it
Set some of them to closed and leave some open
Call getModelStats endpoint for this federation
The result JSON should have a tickets property that is a number and returns the total open tickets in this federation instead of a tickets prop that has an issues and risks prop which are the number of issues and risks in the federation


backend/src/v5/processors/teamspaces/projects/models/commons/tickets.js
```diff
@@ -341,4 +344,33 @@ Tickets.getTicketList = (teamspace, project, model,
...
+	let openTicketsCount = 0;
+	const templateToStatuses = { };
+
+	for (let i = 0; i < tickets.length; i++) {
+		const ticket = tickets[i];
+		const templateId = UUIDToString(ticket.type);
+
+		if (!templateToStatuses[templateId]) {
+			// eslint-disable-next-line no-await-in-loop
+			const { config } = await getTemplateById(teamspace, ticket.type, { config: 1 });
```
carmenfan, line 361: I think instead of this, we may as well get all the templates available within the teamspace in one go.

It's not going to be a lot and we're only getting hte configs out

This way we don't have to interrupt the linter

carmenfan (changes requested):

Amantini1997 (changes requested):

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.js
```diff
@@ -23,6 +23,8 @@ const {
 	modulePropertyLabels,
```
Amantini1997, line 23: Let's say I have 2 users in the same project:
- The one on the left is looking at a given federation item from the dashboard;
- The one on the right is looking at the same federation in the viewer;
image

The user on the right closes/re-opens a ticket
Should the user on the left see the tickets count updating in realtime?

sanmont3drepo: I think this could be a separate issue.

ChristosTsiotsias: No from what I see only FEDERATION_UPDATE_TICKET event is thrown when a ticket is updated :/

sanmont3drepo: moved this into https://github.com/3drepo/3D-Repo-Product-Team/issues/557

carmenfan (approved):

Amantini1997 (approved):

sanmont3drepo (approved):

carmenfan (approved):

PR #5093, opened 2024-08-07
Issue 5043 - Milestone 4: Drop 2D Pins

### LINKED ISSUE #5043: [2D MVP] Milestone 4: Drop 2D Pins
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/380
We are providing a way for the users to drop pins in the 2d area. 

- When dropping a pin in  the 2d area, the plane that its being dropped to is the lower plane.

### Goals
- [ ] As a collaborator+, I want to be able to drop a pin in the 2D drawing.
- [ ] As a collaborator+, I want to be able to edit a pin in the 2D drawing.

### Tasks
- [ ] Add pin drop from 2d area.
- [ ] Add the ability to edit the pins like with 3d

### Related Resources
[MockUp](https://www.figma.com/design/uTWnaYZ8IyoTn2i2xKtdBo/2D---3D?node-id=11-11332&t=j1A2qZU5bKmgEQil-0)

Amantini1997:
This fixes #5043 

#### Description
DrawingViewerService uses an event that emits the pointClick (when not calibrating) just like ViewerService.
The coords component then uses that event.

#### Test cases
Try to add/move a pin while not calibrating, but with existing calibration.
P.S. the 2d viewer always allows to set an arrow. This is wrong as that feature should only exist while calibrating, so keep that into account as when the calibrating toolbar button is active, you can't drop pins


frontend/src/v5/ui/routes/viewer/tickets/ticketsList/ticketsList.styles.ts
```diff
@@ -88,3 +88,7 @@ export const ActionMenu = styled(ActionMenuBase).attrs({
 export const MenuItem = styled(MenuItemBase)`
 	padding: 5px 12px;
 `;
+
+export const Actions = styled.div` // Can be removed when #5009 merged in
```
The-Daniel, line 92: This can now be deleted

sanmont3drepo (changes requested):

sanmont3drepo (approved):

PR #5097, opened 2024-08-09
Milestone 3: Step 2 - 2D calibration points

### LINKED ISSUE #4910: [2D MVP] Milestone 3: Step 2 - 2D calibration points
### Description
This issue is to address step 2 in calibration wizard.  This defines the 2D vector in the viewer,
based on #4909, it should show a split screen.  The sidebar in the 3d view is the same as the step before.
The toolbar  consists of  **Home**, **Perspective/Ortographic view**, **Turntable/Helicopter view**,  **sectioning**, **attribute data**. 

Depends on #4941 and #4942.

<details>
   <summary> 2d viewer </summary>
 
 In the toolbar there is now a calibration button
image

When clicked it gets activated and two buttons appear in the sub toolbar
image

  - **Calibration vector** : when activated this button sets the viewer in a Vector mode. there also a couple of buttons that appears in the _subtoolbar_ to manage the  start and end points of the vector. 
  They are disabled while there is no vector defined.  When the user defines the start point the "clear start" button gets enabled and when the end point is the defined the "clear end" buttons is enabled. 
  Clicking on each of this buttons will delete its corresponding point from the vector
  
  Once its activated the user will define a vector in the 2d viewer. If the vector has been defined then the "Continue" button is enabled in the steps bar.
</details>

<details>
  <summary> Calibration format  </summary>

```js
  {
    // this will most likely live in the drawing history
    calibration: {
        horizontal: {
            model: [[0,0,0], [1,1,1]],
            drawing: [[0,0], [1,1]]
        },
        verticalRange: [ 0, 10],
        // considering different container can be in different units we'd need to make sure we know how to scale it when viewing on a different model.
       units: "m"
     }
}
```

</details>

### Goals
- [ ] As a calibration user I want to define a 2d vector to calibrate with 3d

### Tasks
- [x] If this is a re-calibration show the vector with the previous calibration data.
- [ ] Make the viewer react to step 3 of the calibration wizard
- [x] Add a 2d vector component using the ui layer: copy the look and feel from unity viewer
- [ ] Add support for snapping.
- [x] Add the buttons in the toolbar following the behaviour from the description of the issue
- [x] Add help box on top (look at design)
- [x] Make sure vector start and end coordinates do not overlap

### Related Resources
[MockUp](https://www.figma.com/file/uTWnaYZ8IyoTn2i2xKtdBo/2D---3D?type=design&node-id=7-34655&mode=design&t=lugkcNhj94mlP2AS-0)

Amantini1997:
This fixes #4910

#### Description
2d calibration mode is only available when calibrating step 2.
The 2d vector data is correctly stored and shown again when re-calibrating

#### Test cases
- Opening a drawing should not show the calibration
- 2d vector can only be set when is calibrating step 2

### DIFF
--- a/frontend/src/v5/ui/routes/dashboard/projects/calibration/calibrationHeader/calibrationHeader.component.tsx
+++ b/frontend/src/v5/ui/routes/dashboard/projects/calibration/calibrationHeader/calibrationHeader.component.tsx
@@ -26,6 +26,7 @@ import { ContainersHooksSelectors, FederationsHooksSelectors } from '@/v5/servic
 import { EMPTY_VECTOR } from '@/v5/ui/routes/dashboard/projects/calibration/calibration.constants';
 import { useContext } from 'react';
 import { CalibrationContext } from '../calibrationContext';
+import { ViewerParams } from '@/v5/ui/routes/routes.constants';
 
 const STEPS = [
 	formatMessage({ defaultMessage: '3D Alignment', id: 'calibration.step.3dCalibration' }),
@@ -35,7 +36,7 @@ const STEPS = [
 
 export const CalibrationHeader = () => {
 	const history = useHistory();
-	const { teamspace, project, containerOrFederation } = useParams();
+	const { teamspace, project, containerOrFederation } = useParams<ViewerParams>();
 	const { step, setStep, isStepValid, vector3D, drawingId, origin } = useContext(CalibrationContext);
 	const selectedModel = FederationsHooksSelectors.selectFederationById(containerOrFederation)
 		|| ContainersHooksSelectors.selectContainerById(containerOrFederation);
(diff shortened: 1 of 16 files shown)


frontend/src/v5/ui/routes/dashboard/projects/calibration/calibrationHeader/calibrationHeader.component.tsx
```diff
@@ -35,14 +35,15 @@ const STEPS = [
...
-	const { step, setStep, isStepValid, vector3D, drawingId, origin } = useContext(CalibrationContext);
+	const { teamspace, project, containerOrFederation } = useParams<ViewerParams>();
+	const { step, setStep, isStepValid, vector3D, vector2D, drawingId, origin } = useContext(CalibrationContext);
 	const selectedModel = FederationsHooksSelectors.selectFederationById(containerOrFederation)
 		|| ContainersHooksSelectors.selectContainerById(containerOrFederation);
 	const isLastStep = step === 2;
 
 	const getIsStepValid = () => {
 		if (step === 0) return !!(vector3D[0] && vector3D[1]);
+		if (step === 1) return !!(vector2D[0] && vector2D[1]);
```
The-Daniel, line 46: Vector2D does not seem to be reset when the user clicks to change the arrow, unlike in the 3D calibration. This means that if the user clicks to start a new arrow the continue button is still valid which will use the old values for the 2D vector.
Gif showing how for 3D calibration the continue buttons gets disabled until the arrow is complete. There is also a secondary bug here where the user can go back to step 1 whilst in the middle of calibrating the 2D vector so it is still enabled but the toolbar option is hidden.
arrows
In fact just going backwards after setting the 2D vector keeps the 2D calibration tool vector so you can change it whilst in the 3D step. Is this okay?

The-Daniel (changes requested):

The-Daniel (approved):

sanmont3drepo (approved):

PR #5104, opened 2024-08-13
Issue 5064 - Calibration endpoints

### LINKED ISSUE #5064: [2D MVP] Milestone 3 - endpoint to store and fetch calibration
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/379

We need an endpoint for the front end to talk to in order to store the calibration values against a drawing.

Calibration needs to be stored against a revision, and a drawing can be calibrated many times, but each calibration we will store as a record so it is clear who calibrated/recalibrated at what time. When the user requests a calibration against a revision, they will be provided with the latest calibration values.

##### Add calibration
endpoint: `POST /teamspaces/{teamspace}/projects/{project}/drawings/{drawing}/revisions/{revision}/calibrations`
permissions: collaborator+
description: Add a calibration against the drawing
payload:
````js
{
    horizontal: {
        model: [[0,0,0], [1,1,1]], //2 3d points to make a vector
        drawing: [[0,0], [1,1]] //2 2d points to make a vector
    },
    verticalRange: [ 0, 10],
    units: "m"
}
````

NOTES:
Calibrations should be stored in `drawings.revisions.calibrations` and a record should look something like this:
````js
{
    _id: LUUID(xxx),
   project: LUUID(xxx),
   drawing: LUUID(xxx),
   rev_id: LUUID(xxx),
   createdAt: ISODATE(),
   createdBy: "username",
   horizontal: {
      model: [[0,0,0], [1,1,1]], //2 3d points to make a vector
      drawing: [[0,0], [1,1]] //2 2d points to make a vector
  },
  verticalRange: [ 0, 10],
  units: "m" 
}
````

##### Stats endpoint
Update to the drawing stats endpoint created in https://github.com/3drepo/3drepo.io/issues/4985 , to add the calibration state of a drawing
````js
{
   calibration: "calibrated"; // Can be calibrated, uncalibrated or unconfirmed
  
 //... other drawing stats as before
}
````
NOTES:
- Calibration status is defined as follows:
  - `Uncalibrated`: there is no calibration associated with the drawing at all
  - `Unconfirmed`: There are calibration associated with the drawing, but not to the latest revision of the drawing
  - `Calibrated`: There is a calibration assoicated with the latest revision of the drawing

##### Get a calibration
endpoint: `GET /teamspaces/{teamspace}/projects/{project}/drawings/{drawing}/revisions/{revision}/calibrations`
permissions: viewer+
description: Get a calibration against the drawing
payload:
````js
{
    horizontal: {
        model: [[0,0,0], [1,1,1]], //2 3d points to make a vector
        drawing: [[0,0], [1,1]] //2 2d points to make a vector
    },
    verticalRange: [ 0, 10],
    units: "m"
}
````

NOTES:
This should pull the latest calibration against the latest revision of the drawing. If there is no calibration against hte latest revision, pull the next available (i.e. latest calibration of the previous revision

##### Confirm a calibration
endpoint: `POST /teamspaces/{teamspace}/projects/{project}/drawings/{drawing}/revisions/{revision}/calibrations?usePrevious=true`
permissions: collaborator+
description: Confirm a calibration against the drawing revision by copying a previous revisions' calibration
payload: none

NOTES:
- This is only responsive if the revision of the drawing is in unconfirmed state (i.e. it has no recorded calibration but there are previous calibration revisions.
- On this request, we take a copy of the latest calibration from the previous revision and copy it in as our own.
- This in affect is the same as posting a calibration, without sending the data (i.e. we should record the time and user who did the confirmation)
- If the revision is not in `unconfirmed` state, it should return with an error.

### Goals
- [ ] As a collaborator+, I should be able to post a calibration against a drawing revision
- [ ] As a viewer+ I should be able to fetch the calibration information of a drawing
- [ ] As a viewer+ I should be able to see the calibration status of the latest revision of the drawing
- [ ] As a collaborator+, I should be able to confirm a calibration of a drawing

### Tasks
- TBC

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #5064

#### Description

Get Calibration and Add Calibration endpoints added
Drawing Stats endpoint amended to return the calibration status of the drawing
Unit and E2E tests added
Swagger documentation added

#### Test cases
Get Calibration endpoint should return the last available calibration for the specified revision
Error should be returned if there are no calibrations to return

Add Calibration endpoint should create a new calibration for the specified revision provided the payload is valid
If usePrevious is set to true in the query then the new calibration should get its data from the last available calibration. If there is no calibration to copy an error should be returned.

Drawing stats should now return the calibration status of the drawing based on whether there is a calibration for the last revision of the drawing or not

### CODE REVIEW

File: backend/src/v5/processors/teamspaces/projects/models/drawings.js [hunk 1]
```diff
@@ -40,16 +42,19 @@ Drawings.getDrawingList = async (teamspace, project, user) => {
...
 	let latestRev;
+	let calibration;
+
 	const [settings, revCount] = await Promise.all([
-		getDrawingById(teamspace, drawing, { number: 1, status: 1, type: 1, desc: 1, calibration: 1 }),
+		getDrawingById(teamspace, drawing, { number: 1, status: 1, type: 1, desc: 1 }),
 		getRevisionCount(teamspace, drawing, modelTypes.DRAWING),
 	]);
 
 	try {
```
carmenfan, line 52: put this function in a helper function and have that function called as part of the Promise.all above?
I don't think there is a dependency with the 2 functions in that array right?

backend/src/v5/processors/teamspaces/projects/models/drawings.js
```diff
@@ -40,16 +42,19 @@ Drawings.getDrawingList = async (teamspace, project, user) => {
...
 		latestRev = await getLatestRevision(teamspace, drawing, modelTypes.DRAWING,
-			{ statusCode: 1, revCode: 1, timestamp: 1 });
+			{ _id: 1, statusCode: 1, revCode: 1, timestamp: 1 });
+		calibration = await getDrawingCalibrationStatus(teamspace, drawing, latestRev?._id);
 	} catch {
-		// do nothing. A drawing can have 0 revision.
+		calibration = calibrationStatuses.UNCALIBRATED;
```
carmenfan, line 57: I  think we should leave it as undefined here. No revisions, no calibration

backend/src/v5/processors/teamspaces/projects/models/calibrations.js
carmenfan, line 1: rename this to `drawings.calibrations`, or have drawings as a folder here and have `index.js` and `calibrations.js` inside

backend/src/v5/routes/teamspaces/projects/models/drawings/calibrations.js
```diff
@@ -0,0 +1,261 @@
...
+const Calibrations = require('../../../../../processors/teamspaces/projects/models/calibrations');
+const { Router } = require('express');
+const { getUserFromSession } = require('../../../../../utils/sessions');
+const { modelTypes } = require('../../../../../models/modelSettings.constants');
+const { respond } = require('../../../../../utils/responder');
+const { revisionExists } = require('../../../../../middleware/dataConverter/inputs/teamspaces/projects/models/commons/revisions');
+const { templates } = require('../../../../../utils/responseCodes');
+const { validateNewCalibration } = require('../../../../../middleware/dataConverter/inputs/teamspaces/projects/models/calibrations');
+
+const getLatestCalibration = async (req, res) => {
```
carmenfan, line 28: this is just getCalibration

backend/src/v5/utils/responseCodes.js
```diff
@@ -66,6 +66,8 @@ ResponseCodes.templates = {
 
 	// Drawing related error
 	drawingNotFound: { message: 'Drawing not found.', status: 404 },
+	calibrationNotFound: { message: 'Calibration not found.', status: 404 },
+	revisionNotUnconfirmed: { message: 'Revision is not is an unconfirmed state.', status: 400 },
```
carmenfan, line 70: actually maybe we should just use calibrationNotFound instead of having a different error.

The action is youo're trying to confirm a calibration, and there is none so calibration not found makes sense.

backend/src/v5/models/calibrations.js
```diff
@@ -0,0 +1,55 @@
...
+Calibrations.getDrawingCalibrationStatus = async (teamspace, drawing, latestRevId) => {
+	const calibrations = await db.find(teamspace, CALIBRATIONS_COL, { drawing }, { rev_id: 1 });
+
+	if (!calibrations.length) {
+		return calibrationStatuses.UNCALIBRATED;
+	} if (calibrations.some(({ rev_id }) => UUIDToString(rev_id) === UUIDToString(latestRevId))) {
+		return calibrationStatuses.CALIBRATED;
+	}
+
+	return calibrationStatuses.UNCONFIRMED;
```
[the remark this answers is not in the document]
carmenfan: Can we make sure we have a test case for this?

carmenfan (changes requested):

backend/src/v5/models/calibrations.js
```diff
@@ -0,0 +1,55 @@
...
+const Calibrations = {};
+
+Calibrations.addCalibration = async (teamspace, project, drawing, revision, calibration) => {
+	const formattedData = {
+		_id: generateUUID(),
+		project,
+		drawing,
+		rev_id: revision,
+		createdAt: new Date(),
+		...calibration,
```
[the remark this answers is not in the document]
ChristosTsiotsias: the createdBy prop is added in the route file, sould I move it ?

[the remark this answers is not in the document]
carmenfan: maybe it's better to pass it down and add it here, so it's explicit

This is a bit confusing to read right now as the user is not part of the calibration information

carmenfan:
@ChristosTsiotsias  we'll also have to edit the `deleteDrawing` logic to delete the associated calibrations

carmenfan (changes requested):

backend/src/v5/models/calibrations.js
```diff
@@ -0,0 +1,64 @@
...
+			formattedProjection[`latestCalibration.${key}`] = 1;
+		}
+	}
+
+	return db.aggregate(
+		teamspace, CALIBRATIONS_COL, [
+			{ $match: { rev_id: { $in: revIds } } },
+			{ $sort: { createdAt: -1 } },
+			{ $group: { _id: '$rev_id', latestCalibration: { $first: '$$ROOT' } } },
+			{ $project: { _id: 1, 'latestCalibration._id': 0 } },
```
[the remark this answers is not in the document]
ChristosTsiotsias: I know it looks kinda weitrd but I did it because if I do something like this to exclude id (cause we dont want to fetch it)

image

Im getting this error

image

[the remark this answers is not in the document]
carmenfan: as the error message suggest, you're trying to do inclusion and exclusion in a single projection, it has to be either or...?

you shouldn't need latestCalibration._id:0 in that, it should be gone by default because you didn't include it?

carmenfan (changes requested):
- Date should be converted to timestamp before returning:
image

- Confirming a calibration should not copy the `createdAt` (or author even, I didn't check that one!) on the new calibration. it should record who confirmed it at what time (please make sure we have a e2e test for this one!) 
image

- I missed this one (sorry!) but can we also have a calibration status on the revision list endpoint? Every revision entry should also have it's calibration status shown here (the UI needs it to list)
image

- Can we add the calibration status on new revision update chat event?
  - This should also be added on the model settings update event as the general calibration status would've changed for the drawing.
- Can we add the calibration status on update revision chat event?

backend/src/v5/services/eventsListener/components/modelEvents.js
```diff
@@ -164,8 +168,17 @@ const revisionUpdated = async ({ teamspace, project, model, data, sender, modelT
 		[modelTypes.DRAWING]: chatEvents.DRAWING_REVISION_UPDATE,
 	};
 
-	await createModelMessage(modelEvents[modelType], { ...data, _id: UUIDToString(data._id) },
-		teamspace, project, model, sender);
+	let calibration;
+	if (modelType === modelTypes.DRAWING) {
```
carmenfan, line 172: I dont' think we need clibration here. the only situation this is trigged right now is if something is set to void i think?

calibration wouldn't have changed and shouldn't be in the dataset.

carmenfan (changes requested):

carmenfan (approved):

PR #5105, opened 2024-08-13
ISSUE #5009 - Make V5 card headers match the style of V4 card headers

### LINKED ISSUE #5009: Fix tickets card header style
### Description
In the viewer, the title of the tickets card uses styles different from the other viewer cards. Namely, the font-size is smaller and there is less gap between the icon and the text

### Suggested solution
Fix the tickets card so the future viewer cards can re-use that component

### Goal
- [ ] Viewer cards should use the same layout

### Tasks
- [ ] fix the style in the re-usable tickets card component

The-Daniel:
This fixes #5009

#### Description
Slightly adjust the V5 card header so that it matches the existing V4 headers

#### Test cases
Check the card headers in all 3 stages of the tickets card
The spacing and font style should match that of the v4 cards (e.g. groups, sequences, compare)


frontend/src/v5/ui/components/viewer/cards/card.styles.ts
```diff
@@ -44,13 +44,15 @@ export const CardContainer = styled.div`
 
 export const CardHeader = styled.div`
 	${({ theme }) => theme.typography.h3}
+	font-size: 17px;
 	background: ${({ theme }) => theme.palette.primary.contrast};
 	display: flex;
```
Amantini1997, line 49: you can delete this as you change the display mode later

Amantini1997 (changes requested):

Amantini1997 (approved):

sanmont3drepo (dismissed):

Amantini1997 (approved):

PR #5131, opened 2024-08-29
ISSUE #5129 - [2D MVP] Milestone 1: Collaborators should be able to see drawing category

### LINKED ISSUE #5129: [2D MVP] Milestone 1:  Collaborators should be able to see drawing category
### Description
Product issue:  https://github.com/3drepo/3D-Repo-Product-Team/issues/377.

- [ ] Collaborator cannot see the drawing category:

The-Daniel:
This fixes #5129 
#### Description
This bug was caused because by default MUI Selects search the children for the key that matches the value provided and use the label provided with that option. A collaborator does not have access to the drawing categories list, so the Categories Select had no children hence no matching child was found and no value was rendered.

I have fixed this by tweaking how our selects render the value. A select will still search the children for the matching key as before but it now has a fallback where it will render the value if no matching child is found. This will also fix a problem I found where the options of a select change (e.g. using the API to alter a custom ticket) and selects appearing empty when they in fact have a value that just no longer exists in the list.

I also prevented the API call for fetching the drawing categories unless you are an admin. This keeps the console a little more clean of errors.

#### Test cases
- View the drawings settings whilst you are a collaborator. The category select should show the correct value - not an empty space. The select should still be disabled
- Other selects should still render as before

### DIFF
--- a/frontend/src/v5/ui/controls/inputs/select/select.component.tsx
+++ b/frontend/src/v5/ui/controls/inputs/select/select.component.tsx
@@ -23,7 +23,7 @@ import {
 } from '@mui/material';
 import { FormInputProps } from '@controls/inputs/inputController.component';
 
-export type SelectProps = MuiSelectProps & FormInputProps;
+export type SelectProps = Omit<MuiSelectProps, 'children'> & FormInputProps & { children: any[] };
 
 export const Select = ({
 	required = false,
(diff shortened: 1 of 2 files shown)


carmenfan (approved):
working now , thank you 😄

Amantini1997 (approved):

sanmont3drepo (approved):

PR #5144, opened 2024-09-06
Issue #5127 - Milestone 1 realtime events fixes

### LINKED ISSUE #5127: [2D MVP] Milestone 1: Realtime upload fixes
### Description
Product issue:  https://github.com/3drepo/3D-Repo-Product-Team/issues/377.

### Bugs
3 things here:
- [ ] Screen that isn't uploading is getting the new container/processing event, but it never processed the finished event (this may be a race condition on the messages, if it is please ignore)
- [ ] Screen that did the upload is processing all the events, but when it is done it's showing `ok` instead of the name of th e revision
- [ ] The rendering of the processing statuses should be different? (Should be inline with container)

### Related Resources

MockUp : `link to mockup`

sanmont3drepo:
This fixes #5127 

#### Description
- Fixed double entry for revision when you are the one uploading it
- Fixed calibration status being up to date
- Fixed latest revisions being up to date
- Refactored a bit to remove redundant actions in redux


frontend/test/test.helpers.ts
```diff
@@ -35,41 +40,54 @@ export const spyOnAxiosApiCallWithFile = (api, method) => {
...
+			resolvePromise = null;
 		}
-	), middlewares);
+		
+		return st;
+	}		
+	, middlewares);
 	
-	const waitForActions = (func, waitActions) =>  { 
+	const waitForActions:WaitForActions = (dispatchingfunction, waitActions, debug?: boolean) =>  {
```
The-Daniel, line 84: Is the idea that if we want to debug a test that uses `waitForActions` we append `true` to it?

The-Daniel (changes requested):

frontend/test/test.helpers.ts
sanmont3drepo: Is just to make it do a console log. After doing that I added a way to run it through vscode so maybe this is no longer necessary.

frontend/src/v5/helpers/store.helpers.ts
```diff
@@ -0,0 +1,26 @@
...
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { AnyAction, Store } from 'redux';
+
+let store: Store<unknown, AnyAction> = undefined;
+
+export const setStore = (newStore) => store = newStore;
+
+export const dispatch = (action: AnyAction) => store.dispatch(action);
```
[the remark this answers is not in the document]
sanmont3drepo: yup, that's the idea. I didnt want to change all the references to this file ATM I would create too much noise on the PR. I just wanted to put this out there so in the near  future we can replace all other references to this one.

frontend/src/v5/helpers/store.helpers.ts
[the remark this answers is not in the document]
sanmont3drepo: In nature they seem to tackle very different problems. one seems to be more like helpers for the state within the store ( `/v5/store/store.helpers.ts`) and the other more about helpers for using the store. I think renaming one of them would be more appropriate. 
Any suggestions?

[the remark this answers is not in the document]
sanmont3drepo: renamed to `redux.helpers.ts`

The-Daniel (approved):

frontend/src/v4/modules/store.ts
```diff
@@ -22,6 +22,7 @@ import { applyMiddleware, compose, createStore } from 'redux';
 import createSagaMiddleware from 'redux-saga';
```
Amantini1997, line 22: I am looking at the same drawings list from 2 different users (A and B). User A changes the latest active revision. For as long as that drawing is not expanded for User B (i.e. for as long and they haven't expanded the revisions list), the revisions updates via WS are not considered 
 
https://github.com/user-attachments/assets/68840cc7-3d4e-4269-a78d-a8a6fadcd040

The same behaviour is observable for containers

frontend/src/v5/store/drawings/revisions/drawingRevisions.selectors.ts
```diff
@@ -18,26 +18,42 @@
...
 
-export const selectRevisionsPending = createSelector(
+export const selectActiveRevisions = createSelector(
 	selectRevisions,
-	selectDrawingById,
-	(revisions, drawing) => revisions.length !== drawing.revisionsCount,
+	(revisions) => revisions.filter((val) => !val.void),
+); 
+
+export const selectLastRevision = createSelector(
```
Amantini1997, line 45: Can this be renamed to `selectLatestRevision<Name>` or something similar as it doesn't actually return the revision itself, but its name/code?

Amantini1997 (changes requested):

Amantini1997 (approved):

PR #5146, opened 2024-09-06
Issue 5091 - Image List property type

### LINKED ISSUE #5091: Introduce a new property to custom tickets: Image List
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/382

We want the ability to store multiple images (`imageList`) in a single property(`image`). On an existing property the user should be able to add/remove images.

It should conform to the same mechanics as all properties (i.e. readOnly, immutable etc)

These images should contribute to the quota calculation, and respects the maximum image file size that is configured (basically same middleware checks as single image.

Like other complex properties, `default` does not need to be supported

#### Schema

This will work very similarly to images in comments, image list property will take in an array of images in base64 format:
````js
{
  //... other properties
  imageListProp: ["<base 64 image 1>", "<base 64 image 2>", "<base 64 image 3>"]
}
````

If it is an edit, it will accept a combination of recognised reference ids and base64 images
````js
{
  //... other properties
  imageListProp: ["<base 64 image 1>", "<UUID of existing image>", "<base 64 image 3>"]
}
````

#### NOTES:
- Images should be removed if it is no longer referenced
- There should be a entry in the tickets log to show if someone has added/removed images to this field
- If the array is empty, unset the property.

### Goals
- [x] As a user I want to add multiple images into a single property
- [x] As a user I want to add additional images to an existing `imageList` property
- [x] As a user I want to remove existing images to an existing `imageList` property
- [x] As a user I want to be able to clear all images from the property.

### Tasks
- [x] Modify the schema to accept multiple images and set null to empty array
- [x] Modify the proc / models to store multiple images
- [x] Remove multiple images if prop is null
- [x] Add the log for adding / removing images
- [x] Edit quota code to take into consideration all images
- [x] Edit unit and e2e tests

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #5091

#### Description
Image list property is added in custom tickets
New ticket resources are added via the property
Resources are removed if they are not referenced (update ticket)
The property contributes to quota calculation
Unit and e2e tests updated

#### Test cases
The user should be able to create a template with an image list property
The user should be able to create a ticket with an image list property
New resources should be added if a valid image buffer is provided via the property
Resources should be retained if their ref is provided
Resources should be removed if their ref is not provided
Retrieving a ticket should show the image list refs deserialized

### CODE REVIEW

File: backend/src/v5/schemas/tickets/validators.js [hunk 1]
```diff
@@ -104,6 +104,9 @@ Validators.propTypesToValidator = (propType, isUpdate, required) => {
 		return imposeNullableRule(Yup.array().of(types.strings.title));
 	case propTypes.IMAGE:
 		return types.embeddedImage(isNullable);
+	case propTypes.IMAGE_LIST:
+		return imposeNullableRule(Yup.array().of(isUpdate ? types.embeddedImageOrRef() : types.embeddedImage())
```
carmenfan, line 108: can't we do .min(1) instead of having a weird transform rule?

backend/src/v5/utils/helper/yup.js
```diff
@@ -117,7 +118,8 @@ YupHelper.types.dateInThePast = YupHelper.types.date.test(('date-in-the-past', '
 }));
 
 const imageValidityTests = (yupType, isNullable) => yupType.test('image-validity-test', 'Image is not valid', async (value, { createError, originalValue }) => {
-	if (isUUIDString(originalValue)) {
+	const isImageRef = isUUIDString(originalValue) || (isUUID(originalValue) && !isBuffer(originalValue));
```
carmenfan, line 121: why do you need the extra bits when we didn't need it before? 🤔 

I think the check should be done in the middleware in which case it should also be in string form?

and surely if it's a UUIDString it'll never be a buffer value?

carmenfan (changes requested):

backend/src/v5/schemas/tickets/validators.js
ChristosTsiotsias: I did this because I saw this

image

If I set min(1) it will not allow an empty array as an input isnt it ?

backend/src/v5/utils/helper/yup.js
```diff
@@ -147,7 +149,10 @@ YupHelper.types.embeddedImage = (isNullable) => imageValidityTests(
 YupHelper.types.embeddedImageOrRef = () => imageValidityTests(
 	Yup.mixed()
 		.transform((currValue, orgVal) => {
-			if (orgVal) return isUUIDString(orgVal) ? stringToUUID(orgVal) : Buffer.from(orgVal, 'base64');
+			if (orgVal && typeof orgVal === 'string') {
```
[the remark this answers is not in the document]
ChristosTsiotsias: the reason is because we are running the validation twice here 


so the first time is converting a UUID string to UUID but the second time is trying to convert a UUID to buffer. The only other place we are using embeddedImageOrRef is comments but we are only validating them once so we dont have the same problem

backend/src/v5/utils/helper/yup.js
ChristosTsiotsias: its the same reason as below :/ the first time this is run its actually a UUID string so it returns true. The second time though its a UUID and it proceeds to extract the extention which is undefined and it returns error. 

With this fix, if I pass a ref it will return true the first time cause of the first OR clause and the second time because of the second clause while a base 64 string is not caught any of the times and goes to check the extention

backend/src/v5/schemas/tickets/validators.js
carmenfan: yea we shouldn't I don't think

backend/src/v5/utils/helper/yup.js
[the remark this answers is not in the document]
carmenfan: ok fair enough. in which case can we use isString from typechecker?

backend/src/v5/utils/helper/yup.js
carmenfan: ok, why do you need to check !isBuffer? can a buffer be a buffer and UUID at the same time? surely that's just a UUID if that's the case?

backend/src/v5/schemas/tickets/index.js
```diff
@@ -272,7 +272,7 @@ Tickets.processReadOnlyValues = (oldTicket, newTicket, user) => {
 	}
 };
 
-const uuidString = Yup.string().transform((val, orgVal) => UUIDToString(orgVal));
+const uuidString = Yup.string().transform((val, orgVal) => UUIDToString(orgVal)).nullable();
```
carmenfan, line 275: is this really correct? You're imposing nullable on everything else that was using this to begin with.

And why would we antipate this to be nullable? surely we don't want to return `[null]` as the  image list?

backend/src/v5/utils/helper/yup.js
```diff
@@ -19,7 +19,7 @@ const { UUIDToString, stringToUUID } = require('./uuids');
 const { fileExtensionFromBuffer, isUUID, isUUIDString } = require('./typeCheck');
 const Yup = require('yup');
 const { fileUploads } = require('../config');
-const { isBuffer } = require('lodash');
+const { isString } = require('lodash');
```
carmenfan, line 22: you should be getting isString from typecheck

carmenfan (changes requested):

backend/src/v5/schemas/tickets/index.js
ChristosTsiotsias: the reason is that when a ticket is updated and an image prop is set to null there is an error thrown here

image

ChristosTsiotsias: I changed it now and added nullable to the specific props

carmenfan (approved):

PR #5147, opened 2024-09-06
Issue 5108 - Processing PDF

### LINKED ISSUE #5108: [2D MVP] Processing PDF
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/378
We need to process the PDF that is coming in and turn them into Raster Image or PDF.

If the PDF does not contain an image, it should return with a processing failed message.

There are plenty of PDF libraries in NodeJS that may be able to do this. for e.g. 
https://www.npmjs.com/package/pdf-extractor

 A bit of time should be spent on figuring out which library to use.

Refer to https://github.com/3drepo/3drepo.io/issues/5090 for how it should integrate into the pipeline, this should sit as an independent service within nodejs to process the html, and it should be called by the modelProcessor. Success/Failures should be communicated in a similar manner as the DWG which is being processed by the bouncer.

NOTE: only first page of the PDF should be considered.

### Goals
- [ ] As a collaborator I would like to upload a PDF of a drawing and see it in the 2D Viewer.

### Tasks
TBC

carmenfan:
This fixes #5108

#### Description
This requires https://github.com/3drepo/local/issues/141 of the local repository

- Processes PDF into SVG and produces a thumbnail
- Support encoding on SVG (i.e. compressed file will be streamed out with the right headers)
- unified the naming of `mime` and `mimeType`

#### Test cases
- PDF should now process and proces an SVG and thumbnail
- DWG should work like before


ChristosTsiotsias (approved):

ChristosTsiotsias:
looks good but there is one failing test

ChristosTsiotsias (changes requested):
looks good but there is one failing test

PR #5161, opened 2024-09-13
Issue #4941 - Snapping for calibration

### LINKED ISSUE #4941: [2D MVP] Milestone 3: Snapping in 2d viewer
### Description
For setting the 2d vector for calibration we want to provide the user with snapping for convenience. 

### Goals
- [x] As a user I want to be able to snap to lines in the 2d viewer
- [x] Snapping should occur as the cursor moves
- [x] Cursor should snap to Corners, Intersections and Curves/Lines
- [x] There is a method that can be called externally that will return the snapped point
- [x] It is possible to enable or disable snapping

### Tasks
- [x] Add skeleton infrastructure for testing snapping behaviour
- [x] Test polyfil to process paths into line segments
- [x] Make functions to translate between container rect and svg coordinate system for svgImage
- [x] Make cursor for debugging
- [x] Find out which [SVG elements](https://developer.mozilla.org/en-US/docs/Web/SVG) need to be supported
- [x] Add support for `polyline`
- [x] Add support for `path`
- [x] Choose a suitable acceleration structure for Line intersection tests
- [x] Get some more SVGs from ODA for testing
- [x] Extract node-points from primitives for closest point tests and add to collector
- [x] Add logic to compute intersections and add to collector
- ~Add acceleration structure for points~
- [x] Integrate point tests into snapping logic and add type to results structure
- [x] Add check to make sure svg element has width and height properly set (otherwise the mapping breaks)
- [x] Make cursor update on zoom
-  ~PolyCollector should not add duplicate points~
- [x] Add Bezier Curve closest point test
- [x] Add Curve & Line intersection test
- [x] Add Curve & Curve intersection test
- [x] Add curve self-intersection test
- [x] Add rect support
- [x] Add circle support
- [x] Add ellipse support
- [x] Add line support
- [x] Add polyline support
- [x] Add polygon support
- [x] Add path support
- [x] Optimise the intersection tests (maybe)

sanmont3drepo:
This fixes #4941 

#### Description
- Added snapping for calibration


frontend/src/v5/ui/components/viewer/drawingViewer/viewerLayer2D/viewerLayer2D.component.tsx
```diff
@@ -46,60 +69,66 @@ export const ViewerLayer2D = ({ viewBox, active, value, onChange }: ViewerLayer2
...
-		if (!active && !offsetEnd) {
-			resetArrow();
+		if (snapping) {
+			const radius = 10 / viewBox.scale;
+			const res = snap(mousePos, snapHandler, radius);
+			mousePos = res.mousePos;
+
+			if (snapType !== res.snapType) {
+				setSnapType(res.snapType);
+			}
```
Amantini1997, line 110: The code above is the same as the previous useEffect
image
(I moved `code A` up to better show the differences)

Amantini1997 (changes requested):

Amantini1997 (approved):

PR #5169, opened 2024-09-20
Issue 5149 User Permission - Warning modal when changing permissions

### LINKED ISSUE #5149: User Permissions - Warning modal when making a change
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/537
Be warned when any user permissions are changed.

### Goals
- [ ] As a teamspace/project admin, show me a modal that requires confirmation before changing a single or multiple users permissions with information regarding the change.
- [ ] As a teamspace/project admin, allow me to suppress this modal on each session by adjusting the toggle option in the user permissions UI.

### Tasks
- [ ] Add a boolean switch in users permissions page (design is just a guide for this)
- [ ] Use local storage to save the switch permission settings ( like with visual settings)
- [ ] Dispatch a warning message like in the design whenever the user is changing something)  

### Related Resources
[MockUp](https://www.figma.com/design/eI1KkbqCaxI7AeMeXnPE1V/Permissions?node-id=1-38)

Amantini1997:
This fixes #5149

#### Description
Changing a user permissions prompts a modal if the flag is set to false. 
It is possible to toggle the flag at any point from the pages where you can change permissions

#### Test cases
- Try to change the permissions from:
   - Users tab in teamspace page;
   - Project permission tabs in project page (both 1 or a batch);
   - Users permissions tabs in project page; (both 1 or a batch);

- Make sure setting the modal flag to false will not trigger the modal
- Make sure, after logging out and logging back in, the flag is set to false


frontend/src/v4/routes/components/customTable/components/cellSelect/cellSelect.component.tsx
```diff
@@ -91,8 +93,13 @@ export class CellSelect extends PureComponent<IProps, IState> {
 		const selectedValue = event.target.value;
 
 		if (this.state.selectedValue !== selectedValue) {
-			this.setState({selectedValue});
-			this.props.onChange(event, selectedValue);
+			updatePermissionsOrTriggerModal({
```
sanmont3drepo, line 96: I dont think this is the best place to add the permissions confimation. cellSelect seems to be used widely in other places. foe example when changin the risks properties:
cellselectproblemst

sanmont3drepo (changes requested):

frontend/test/mocks/react-children-utilities.js
```diff
@@ -0,0 +1,18 @@
+/**
```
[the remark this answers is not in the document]
Amantini1997: Unfortunately, this is due to Jest (indirectly) importing a library called `react-children-utilities`. The latter probably uses CommonJS, which relies on a different export system, and Jest does not recognise it and triggers the error `SyntaxError: Cannot use import statement outside a module`

frontend/src/v4/routes/components/permissionsTable/permissionsTable.component.tsx
```diff
@@ -110,7 +112,7 @@ interface IProps {
 	context?: string;
 	onSelectionChange?: (selectedUsers) => void;
 	onFilterChange?: (filteredUsers) => void;
-	onPermissionsChange?: (permissions) => void;
+	onPermissionsChange?: (permissions, permissionsType, permissionsCount?) => void;
```
sanmont3drepo, line 115: dont change base components that are being used in other parts of the system (like in invitationdialogs)

sanmont3drepo (changes requested):

frontend/src/v4/routes/components/customTable/components/cellSelect/cellSelect.component.tsx
```diff
@@ -91,7 +91,6 @@ export class CellSelect extends PureComponent<IProps, IState> {
 		const selectedValue = event.target.value;
 
 		if (this.state.selectedValue !== selectedValue) {
```
sanmont3drepo, line 93: im worried about the deleted line here. why was this deleted?

Amantini1997: Imagine this scenario:
image
The user has the value set to `Teamspace admin` and attempts to change it to `User`.
The confirmation modal is prompted, and if they cancel the action from the modal, the value should still be `Teamspace admin`, but if this line is not deleted, the shown value is `User` instead.

So I rely on the `componentDidUpdate` to update the state instead

sanmont3drepo: I think we can get rid of the whole `selectedValue` state because that seems intended for uncontrolled inputs and we are not using those types of inputs in in v4

sanmont3drepo (approved):

sanmont3drepo (dismissed):

carmenfan (changes requested):
as discussed, let's remove the confirmation modal logic from the users page
image

sanmont3drepo (approved):

frontend/src/v4/modules/userManagement/userManagement.sagas.ts
```diff
@@ -282,6 +286,13 @@ export function* fetchProject({ project }) {
 
 export function* updateProjectPermissions({ permissions }) {
 	try {
+		const shouldUpdate = yield updatePermissionsOrTriggerModal({
```
sanmont3drepo, line 289: it seems that updating the permissions fails
failed

sanmont3drepo (changes requested):

sanmont3drepo (approved):

carmenfan (approved):

PR #5178, opened 2024-09-25
ISSUE #5177 fixes drawing import issues on local 

### LINKED ISSUE #5177: Drawing queue importParams.json is passing down an absolute path
We're passing the file path inside ImportParams.json and at the moment it's passing down an absolute path of where the file is in the web server, which is getting lost in translation if the bouncer worker is being ran in a different box that mounted the share space in a different drive (or linux vs windows)

It should be utillisng the placeholder token so it gets replaced with a path that is relevant to the consumer of the event

carmenfan:
This fixes #5177

#### Description
Use placeholder in importParams to identify where the file is

#### Test cases
cross platform import should now work

### DIFF
--- a/backend/src/v5/services/modelProcessing.js
+++ b/backend/src/v5/services/modelProcessing.js
@@ -79,7 +79,7 @@ const queueDrawingUpload = async (teamspace, project, model, revId, data, fileBu
 			database: teamspace,
 			project: model,
 			revId,
-			file,
+			file: `${SHARED_SPACE_TAG}/${Path.join(revId, `${revId}${data.format}`)}`,
 		};
 
 		await mkdir(pathToRevFolder);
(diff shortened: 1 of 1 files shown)


backend/src/v5/services/modelProcessing.js
```diff
@@ -98,6 +98,7 @@ const queueDrawingUpload = async (teamspace, project, model, revId, data, fileBu
 
 		publish(events.QUEUED_TASK_UPDATE, { teamspace, model, corId: revId, status: processStatuses.QUEUED });
 	} catch (err) {
+		console.log(err);
```
Charence, line 101: @carmenfan do we want to keep this console log?

carmenfan: @Charence  what console log? 🙈  (removed)

Charence (approved):

PR #5228, opened 2024-10-28
Issue 5198 - Markdown - textbox pushes canvas out of the view

### LINKED ISSUE #5198: Markdown - Text box pushes canvas out of view
### Description

There is a strange visual glitch when a text box in an image markup editor overflows the side of the canvas

### Steps to replicate
- Open an image markup editor (this can be by either editing an image in safetibase/issues, or with the new markup editor for tickets)
- Add a textbox towards the right side of the canvas
- Keep typing

### Current Behaviour
- The image is pushed to the left out of the canvas

### Expected Behaviour
- The image remains fixed (only for custom tickets, disregard changes for safetibase and issues)

- The text is forced multiline when reaching the right side
text

- In the bottom it gets cut off 
image

image

Amantini1997:
This fixes #5198

#### Description
The code had me refactoring it a bit in order to fix this behaviour as the logic from v4 was very convoluted.

#### Test cases
Create a text (plain or with a callout) and try writing some text that should break according to the box view. The text should never push the canvas out of the box and, on overflow, the text should disappear without following the caret (i.e., if you are typing and the caret is out of view, the textbox should always display the first line, ignoring the current caret position)


sanmont3drepo (approved):

frontend/src/v4/routes/components/screenshotDialog/components/typingHandler/typingCalloutHandler.component.tsx
The-Daniel, line 1: Textboxes aren't being saved (all other markup objects are fine)

https://github.com/user-attachments/assets/d7525b93-c098-4da9-8074-94b629f9a109

The-Daniel (changes requested):

The-Daniel (approved):

sanmont3drepo (approved):

PR #5229, opened 2024-10-29
Issue 5226 - File size validation does not inform the user if the file is too big

### LINKED ISSUE #5226: File size validation does not inform the user if the file is too big
### Description
In the upload files modal, if a file exceeds the max size, we do not inform the user, who will eventually not be able to upload the files (as the button is disabled) but without knowing the reason

- This is happening in: both 

### Stakeholders
TBD

### Steps to replicate
- Open the upload files modal for drawing/container
- Upload a file larger that 1GB

### Current Behaviour
The file shows in the modal without any error

### Expected Behaviour
I guess an alert modal should pop up informing the user about the problem and displaying the file in the modal

image

### Linked Dev Tickets
https://github.com/3drepo/3D-Repo-Product-Team/issues/609

Amantini1997:
This fixes #5226 

#### Description
The file validation is triggered upon loading the item and a text shows up next the file size if too big. Also, the field are disabled

#### Test cases
Since the changes should upload to both containers and drawings, try and upload a too large file in both modals


frontend/src/v5/ui/components/shared/uploadFiles/uploadList/uploadListItem/uploadListItemTitle/uploadListItemTitle.component.tsx
The-Daniel, line 1: Error is not applied when uploading directly to an existing container. Drawings work as expected.
image

image

The-Daniel (changes requested):

sanmont3drepo (approved):

PR #5263, opened 2024-11-20
ISSUE #5230 - IP Address match in cookie authentication

### LINKED ISSUE #5230: Tighten cookie authentication
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/528

We want to tighten our check on cookie authentication to ensure the cookie came from the same source/IP.

When we authenticate via cookie, we should check the user agent and IP still matches the one that was registered with when the cookie was established, otherwise, destroy the session and notify the user that they've been logged out (in a similar manner as when the user logs in from a different browser/incognito)

Ensure whatever change we make here still works in the plugins.

### Goals
- [ ] As a user I want to ensure that if my session cookie is compromised and used elsewhere, the session is invalidated

### Tasks
TBC

ChristosTsiotsias:
This fixes #5230 

#### Description
Cookie authentication now requires IP Address match to be successful. The IP address is saved in the session and compared to each requests IP address. If they differ the session is destroyed and a notLoggedIn error is returned.

Plugin should work as before without any changes.

#### Test cases
Log in to IO from a browser
Use a vpn and connect to another server
Make another request in IO
An error should be returned and you should be logged out


backend/src/v4/services/api.js
```diff
@@ -98,7 +98,7 @@ APIService.createAppAsync = async (config, v5Init = true) => {
 	const app = express();
 	app.use(cookieParser());
 
-	if (config && !config.using_ssl && config.public_protocol === "https") {
+	if (config && ((!config.using_ssl && config.public_protocol === "https") || config.trust_proxy)) {
```
carmenfan, line 101: why was this added? 😮

backend/src/v5/middleware/auth.js
```diff
@@ -24,7 +25,25 @@ const { validateMany } = require('./common');
...
+const isIpAddressValid = async (req, res, next) => {
+	const ipMatch = req.session.ipAddress === (req.ips[0] || req.ip);
+
+	if (req.session.user.isAPIKey || ipMatch) {
+		await next();
+		return;
+	}
+
+	try {
+		req.session.destroy(() => {
```
carmenfan, line 37: Can we get a log here ?

backend/src/v5/middleware/auth.js
```diff
@@ -24,7 +25,25 @@ const { validateMany } = require('./common');
...
+			res.clearCookie('connect.sid', { domain: config.cookie_domain, path: '/' });
```
carmenfan, line 38: `connect.sid` is a constant in `utils/sessions.constant.js`, so we should reference that

carmenfan: probably should clear the CSRF token too

backend/src/v5/middleware/auth.js
```diff
@@ -24,7 +25,25 @@ const { validateMany } = require('./common');
 
 const AuthMiddleware = {};
 
-AuthMiddleware.validSession = async (req, res, next) => {
+const isIpAddressValid = async (req, res, next) => {
```
carmenfan, line 28: function name is a bit misleading, should be something like `ipAddressMatched`?

carmenfan: we also should be matching user agent (i.e. if someone logged in via chrome but now trying to use the same cookie on firefox that should not work)

backend/src/v5/middleware/auth.js
```diff
@@ -15,6 +15,7 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+const config = require('../utils/config');
```
carmenfan, line 18: instead of getting the whole config, just get the cookie_domain within it

backend/tests/v5/e2e/routes/users.test.js
```diff
@@ -28,7 +28,7 @@ const { updateProfile } = require('../../../../src/v5/models/users');
 
 const { templates } = require(`${src}/utils/responseCodes`);
 
-const { loginPolicy } = require(`${src}/utils/config`);
+const config = require(`${src}/utils/config`);
```
carmenfan, line 31: just add cookie domain in the destructured assignment

backend/tests/v5/unit/middleware/auth.test.js
carmenfan, line 1: also need a test in e2e for the chat message. Existing users using the same cookie should get a notification that they've been logged out.

carmenfan (changes requested):

backend/tests/v5/helper/services.js
```diff
@@ -774,7 +774,7 @@ ServiceHelper.generateView = (account, model, hasThumbnail = true) => ({
 	...(hasThumbnail ? { thumbnail: ServiceHelper.generateRandomBuffer() } : {}),
 });
 
-ServiceHelper.app = async () => (await createServer()).listen(8080);
+ServiceHelper.app = async () => (await createServer(config)).listen(8080);
```
carmenfan, line 777: why did you need this change?

backend/src/v5/middleware/auth.js
ChristosTsiotsias: we are currently not storing the browser in the session, just whether it is a web session or not

image

should I change it so we store the browser upon login as well ?

backend/src/v4/services/api.js
ChristosTsiotsias: because I have to have trust proxy set to true to be able to test it in the e2e tests, otherwise x-forwarded-for header is ignored and the ip is always the same

image

backend/tests/v5/e2e/routes/users.test.js
ChristosTsiotsias: Im getting the trust_proxy not the cookie domain, and if I use deconstruct I cannot set it 🤔 (check line 804) this is to be able to test it cause otherwise all IPs received from the tests look the same

backend/src/v4/services/api.js
carmenfan: i don't think you need a e2e test for this, just make sure the unit test works it's fine.

To allow trust proxy when it's a http protocol is a bit weird.

backend/tests/v5/e2e/routes/users.test.js
carmenfan: or, test it with user agent change?

carmenfan (approved):

PR #5275, opened 2024-11-29
Issue 5273 - fixed custom sequences not working

### LINKED ISSUE #5273: Custom sequences are not fetching groups properly
2 issues with custom sequences here:

#### Steps to replicate
- [create a custom sequence](https://3drepo.github.io/3drepo.io/#api-Sequences-createSequence) with 2 frames in a model, using the viewpoint syntax, something like the following (but with teamspace/model/sharedIds info adjusted):
````json
{
    "name": "Test",
    "frames": [
        {
            "dateTime": 1704096000000,
            "viewpoint": {
                "hidden_group": {
                    "name": "test",
                    "objects": [
                        {
                            "account": "carmen",
                            "model": "de886fa8-c253-4c78-920c-968b543fe026",
                            "shared_ids": [
                                "95dd5c73-18d9-4bd2-b98d-c8f57e392990",
                                "490c907c-e34c-40ba-b1ac-dfbc00b9bb10",
                                "7162e705-e080-43f8-9990-da02cfe3f089",
                                "fff6cb1e-2a7b-47c9-8d9c-11ba6203e205"
                            ]
                        }
                    ],
                    "account": "carmen",
                    "model": "de886fa8-c253-4c78-920c-968b543fe026"
                }
            }
        },
        {
            "dateTime": 1704098000000,
            "viewpoint": {
                "shown_group": {
                    "name": "test",
                    "objects": [
                        {
                            "account": "carmen",
                            "model": "de886fa8-c253-4c78-920c-968b543fe026",
                            "shared_ids": [
                                "95dd5c73-18d9-4bd2-b98d-c8f57e392990",
                                "490c907c-e34c-40ba-b1ac-dfbc00b9bb10",
                                "7162e705-e080-43f8-9990-da02cfe3f089",
                                "fff6cb1e-2a7b-47c9-8d9c-11ba6203e205"
                            ]
                        }
                    ],
                    "account": "carmen",
                    "model": "de886fa8-c253-4c78-920c-968b543fe026"
                }
            }
        }
    ]
}
`````
- Inspect the sequence in the UI

##### Initial state doesn't fetch the group
When you click on the sequence, it opens the sequence and  suppose to display the first state, the sequence data clearly shows there's a reference to a group, but the group is not being fetched and thus the objects are not being hidden:

( Interestingly it also seems to be fetching the sequence twice... )

##### Group fetches are called with null teamspace and model
If you skip to the next frame in the sequence, you can see the a group fetch is attempted, but returns a 404. Inspecting the URL will see that `null` is given for both the teamspace and the model, instead the current teamspace/model.

#### Expected Behaviour
The groups should be fetched correctly for all frames and the sequence should perform as expected

sanmont3drepo:
This fixes #5273 

#### Description
- Moved fetch groups for viewpoint into a helper
- Groups are now being fetched for a custom sequence frame in sequence sagas
- Updated sequence redux `updateFrameWithViewpoint` to also use datetime 

#### Test cases
- Create a custom sequence for a model and play it

### DIFF
--- a/frontend/src/v4/modules/sequences/sequences.redux.ts
+++ b/frontend/src/v4/modules/sequences/sequences.redux.ts
@@ -35,7 +35,7 @@ export const { Types: SequencesTypes, Creators: SequencesActions } = createActio
 	setOpenOnTodaySuccess: ['openOnToday'],
 	fetchFrame: ['date'],
 	prefetchFrames: [],
-	updateFrameWithViewpoint: ['sequenceId', 'stateId', 'viewpoint'],
+	updateFrameWithViewpoint: ['sequenceId', 'stateId', 'dateTime', 'viewpoint'],
 	setStepInterval: ['stepInterval'],
 	setStepScale: ['stepScale'],
 	fetchActivitiesDefinitions: ['sequenceId'],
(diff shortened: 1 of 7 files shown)


Amantini1997 (approved):

carmenfan (changes requested):
As discussed with @Amantini1997, the sequence below is fetching ifc_guids instead of shared_ids for the groups:

https://issue-5273.dev.3drepo.io/v5/viewer/DevTeamWorkspace/9e311110-1bf4-11ee-b179-49de4ef0b5f5/bddf8590-0998-4df6-a2a1-2cc85d2dd5a6

Because it is a container sequence, so it is fetching the group within the container scope, however, it is passing in the revision ID of the federation: 
image

In short: there's 2 possibilities, the sequence may be on the federation level (in which case, sending the revision ID is correct), or container level (in which case, we need the revision of the container, which is most likely just /master/head)

carmenfan (changes requested):
federation groups is now calling with master/head - but it was working to begin with.. the issue is on container groups (where the sequence came from a container itself).

@sanmont3drepo let's have a call and I'll show you.

PR #5295, opened 2024-12-10
ISSUE #5294 remove nanoid dependency

### LINKED ISSUE #5294: remove nanoid dependency
### Description
In response to https://github.com/3drepo/3drepo.io/pull/5291

`nanoid` seems to be only used in v4 queue, to instantiate a `uid` that doesn't seem to be used anymore. So instead of updating the dependency, we should just remove it.

### Goals
- Address https://github.com/3drepo/3drepo.io/pull/5291
- remove unnecessary dependency

### Tasks
- remoove reference to nanoid

carmenfan:
This fixes #5294
#### Description
- Remove unnecessary reference to nanoid
- Add resolution criteria (mocha is used in v4 tests and that references nanoid)

#### Test cases
- should function as before

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -61,7 +61,6 @@
     "moment-timezone": "0.5.45",
     "mongodb": "3.7.3",
     "multer": "1.4.5-lts.1",
-    "nanoid": "3.3.1",
     "nodemailer": "6.9.14",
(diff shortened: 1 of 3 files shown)


carmenfan:
merged into #5296

PR #5303, opened 2024-12-12
Issue 5268 - Ticket filters: Number

### LINKED ISSUE #5268: Ticket filters: number filter form
### Description
Parent issue: https://github.com/3drepo/3drepo.io/issues/5267
We need to be able to provide users with a way of filtering tickets by number typed properties. 

<details>
<summary>The criterias available for number properties are "Equals", "Does not Equals", "Greater than", "Greater than or equal to", "Less than", "Less than or equal to", "In range", "Not in range", "Exists" or "Does not exists" </summary>
image
</details>

- If the criteria chosen is "Exists" or "Does not exists" the form becomes valid and the filter can be added to the list.

<details>
<summary> If the criteria selected is "Equals", "Does not Equals", "Greater than", "Greater than or equal to", "Less than" or "Less than or equal to" then the form shows a single textfield with a (+) button to add more values. 
</summary>
image
</details>

- If the criteria selected is "In Range"  or "Not in range" then the form shows two textfields with a (+) button to add more values. 

### Goals
- [ ] As a user I want to be able to filter tickets by number typed fields.

### Tasks
- [ ] Create filter form specific to numbers (operators and input type)
- [ ] Add input validation

### Related Resources
[MockUp ](https://www.figma.com/design/nboKoEopFrXIrOAYTJQHru/Filters?node-id=823-44241&node-type=section&t=Abr4f9cZ2vBp6Zzq-0)

Amantini1997:
This fixes #5268 

#### Description
Added support for number filters:
- allowed operations;
- range input type;
- validation

#### Test cases
Switch from no values (e.g. 'exists'), to 'in range', to 'greater than' (in different orders) and make sure validation is always working

#### Worth noting
Switching from a single-value to a range input was breaking the code.
Consider the following:
1. In the filter form, you have a "greater than" operator selected, and only one field;
2. you switch to "in range";
3. `FilterFormValues` catches the change and replaces the single-value inputs with the range ones (<- this is what crashes the app)
4. a `useEffect` in the same file also catches the operator change and resets correctly the fields to account for range values

The range input expects an array and tries to access that array at indices 0 and 1.
However, the re-render happens before the useEffect is invoked, so the data passed to the range input is the same passed to the single value input. Internally, react-hook-form does not handle well this change in data-type and, as it fails access those values, it crashes.
To solve the problem, I prevent the re-render until the useEffect reset the field values.


sanmont3drepo (changes requested):

frontend/src/v5/ui/components/viewer/cards/cardFilters/filterForm/filterFormValues/rangeInput/rangeInput.component.tsx
```diff
@@ -0,0 +1,41 @@
...
+		if (dirty?.[0] || dirty?.[1]) {
+			trigger(name);
+		}
+	}, [error?.[0]?.message, error?.[1]?.message]);
+
+	return (
+		<RangeContainer>
+			<Input name={`${name}.0`} formError={!!error} />
+			<FormattedMessage id="rangeInputs.to" defaultMessage="to" />
+			<Input name={`${name}.1`} formError={!!error} />
```
sanmont3drepo, line 38: Would only show the error if the two fields are dirty

sanmont3drepo (changes requested):

sanmont3drepo (approved):

PR #5307, opened 2024-12-13
ISSUE #5305 - Broken Popover in Storybook

### LINKED ISSUE #5305: JobPopoverCircle & PopoverCircle throws errors in storybook
### Description
JobPopoverCircle & PopoverCircle are not working in storybook

- This is happening in: both

### Steps to replicate
- run `yarn build-storybook`
- navigate to 'https://3drepo.lan/storybook/?path=/docs/info-popovercircle--docs' or  https://3drepo.lan/storybook/?path=/docs/info-jobpopovercircle--docs

### Current Behaviour
-  the page shows and error

### Expected Behaviour
- the page should show the popover component 

image

Amantini1997:
This fixes #5305 

#### Description
The library `contrast-color` was trying to access the window object using `this`, but in storybook, that was not accessible.
The library, for the usage that we do, was not more than just a function, so the former got replaced with a custom function

#### Test cases
Test popover library components as well as components that rely on the old contrast-color library


frontend/.storybook/preview-head.html
```diff
@@ -0,0 +1,3 @@
+<script>
+	window.global = window;
+</script>
```
sanmont3drepo, line 3: is this necessary?

sanmont3drepo (changes requested):

frontend/src/v5/helpers/colors.helper.ts
```diff
@@ -28,7 +27,23 @@ export type RgbGroupColor = GroupColor & { color?: RgbArray };
 export type HexGroupColor = GroupColor & { color?: string };
 
 // Misc
-export const isLight = (hex: string, freshold?: number) => contrastColor({ bgColor: hex, threshold: freshold ?? 127 }) !== '#FFFFFF';
+export const contrastColor = ({ hex = '#FFFFFF', threshold }) => {
```
sanmont3drepo, line 30: Isn't this the same as in `frontend/src/v5/helpers/contrast-color.ts`

PR #5328, opened 2024-12-31
Issue 5287- Endpoints to get the md5 hash of the original model file 

### LINKED ISSUE #5287: Endpoints to get the md5 hash of the original model file
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/547

To add endpoints for federation/containers to provide the md5 hash of the original files

#### Additional endpoints
route: `HTTP GET /teamspaces/{teamspace}/projects/{project}/containers/{container}/revisions/{revision}/files/original/info`
description: get the details of the original file uploaded to that revision of the container
permissions: viewer+
response: md5 hash of the original file 
````js
{
    container: "<container ID>",
    code: "<revision code>",
    uploadedAt: 223984739284723, // timestamp of when it was uploaded
    hash: "<md5 hash of the original file uploaded to this revision>",
    filename: "abc.rvt",
    size: 329487234 // size in Bytes
}
````
route: `HTTP GET /teamspaces/{teamspace}/projects/{project}/federations/{federation}/revisions/{revision}/files/original/info`
description: get the details of the original file uploaded to all the containers within the federation
permissions: viewer+
response: md5 hash of the original file 
````js
[{
    container: "<container ID>",
    code: "<revision code>",
    uploadedAt: 223984739284723, // timestamp of when it was uploaded
    hash: "<md5 hash of the original file uploaded to this revision>",
    filename: "abc.rvt",
    size: 329487234 // size in Bytes
},
{
    container: "<container ID>",
    code: "<revision code>",
    uploadedAt: 223984739284723, // timestamp of when it was uploaded
    hash: "<md5 hash of the original file uploaded to this revision>",
    filename: "def.dwg",
    size: 329487234 // size in Bytes
}]
````
NOTE:
- for each of the container, we need to make sure the user has at least viewer permissions to the container. if not, we should omit the information about that container.

### Goals
- [ ] As a user, I want to be able to get detailed information about the original file of a model, so I can verify if the same file is being used in applications outside of 3D Repo.
- [ ] As a admin, I want to make sure information about the files are not leaked to users who do not have at least viewer permission to the container.

### Tasks
TBC by the developer

### PR DESCRIPTION
**Pull request by Bogdan-Crn**
This fixes #5287 

Description

- Created endpoint for containers to allow retrieval of the MD5 data for the container
- Created endpoint for federations to allow retrieval of the MD5 data for all the containers in the federation that the user has access to
- Created a new function in FileRefs to update a specific .ref file, this is to allow the caching of MD5 hashes for future reference

Test cases

- Unit tests for containers getting the MD5 data, first time the database should not have the MD5Hash property and the second time the system should have set it in the db
- Unit tests for federation getting the MD5 data, similar to above 
- Unit test for the FileRefs new function
- E2E tests for containers and federations testing scenarios where the user has different levels of rights

### CODE REVIEW

File: backend/tests/v5/unit/processors/teamspaces/projects/models/federations.test.js [hunk 1]
```diff
@@ -359,6 +367,97 @@ const testGetTicketGroupById = () => {
...
+		const mockConatiners = [{ _id: '1', name: 'test1', permissions: [{ user: 'user1' }] }, { _id: '2', name: 'test2', permissions: [] }, { _id: '3', name: 'test3' }];
+		ModelSettings.getContainers.mockImplementation((teamspace, containers) => {
+			if (containers.length > 1) {
+				return mockConatiners;
+			}
+
+			return mockConatiners.filter((container) => container._id === containers[0]);
+		});
+
+		// it should get an empty array if user doesn't have rights to the container
```
carmenfan, line 383: redundant comments should be removed 😆 that's just the test description

carmenfan (changes requested):

Bogdan-Crn:
Moved most of the modelList logic to the filesManager.
Updated the federation check to see if the user has access to the container.
Updated what happens if the federation is empty.
Updated the tests to contain generateRandomString() where necessary.

backend/src/v5/processors/teamspaces/projects/models/commons/modelList.js
```diff
@@ -65,68 +62,35 @@ ModelList.getModelList = async (teamspace, project, user, modelSettings) => {
...
+			{ rFile: 1, timestamp: 1, fileSize: 1, tag: 1 });
 	}
 
-	// check if anything is in there
-	if (!rev.rFile?.length) return returnValue;
+	if (!rev.rFile?.length) return {};
 
-	const code = UUIDParse.unparse(rev._id.buffer);
+	const code = rev.tag;
 	const uploadedAt = new Date(rev.timestamp).getTime();
```
carmenfan, line 82: Leave this as a Date object for now and do the serialisation in the middleware

backend/src/v5/processors/teamspaces/projects/models/commons/modelList.js
```diff
@@ -65,68 +62,35 @@ ModelList.getModelList = async (teamspace, project, user, modelSettings) => {
...
-			{ rFile: 1, timestamp: 1, fileSize: 1 });
+			{ rFile: 1, timestamp: 1, fileSize: 1, tag: 1 });
 	}
 
-	// check if anything is in there
-	if (!rev.rFile?.length) return returnValue;
+	if (!rev.rFile?.length) return {};
 
-	const code = UUIDParse.unparse(rev._id.buffer);
+	const code = rev.tag;
```
carmenfan, line 81: Sorry I probably wasn't being clear last time 😅 

Can we keep this as tag on the return value?

backend/src/v5/services/filesManager.js
```diff
@@ -214,4 +216,25 @@ FilesManager.storeFileStream = async (teamspace, collection, id, dataStream, met
 	await insertRef(teamspace, collection, { ...meta, ...refInfo, _id: id, mimeType });
 };
 
+FilesManager.getMD5FileHash = async (teamspace, container, filename) => {
+	const refEntry = await getRefEntry(teamspace, `${container}.history.ref`, filename);
+
+	if (Object.keys(refEntry).includes('MD5Hash')) {
```
carmenfan, line 222: I think you can probably abbreviate this bit with ES6 style syntax?

Also can we call it `md5Hash` just for consistency in naming convention?

```suggestion
const {size, md5Hash: hash }= await getRefEntry(teamspace, `${container}.history.ref`, filename);

if(hash) {
  return {size, hash};
}
```

backend/src/v5/services/filesManager.js
```diff
@@ -214,4 +216,25 @@ FilesManager.storeFileStream = async (teamspace, collection, id, dataStream, met
...
+		return {
+			hash: refEntry.MD5Hash,
+			size: refEntry.size,
+		};
+	}
+
+	const file = await FilesManager.getFile(teamspace, `${container}.history`, filename);
```
carmenfan, line 229: getting filesManager to append history is a bit too coupled (whilst we're only using it in history atm, we can in theory get md5 for anything).

I would rename container to collection and get the processor to pass in `${container}.history` instead

carmenfan (changes requested):
Looking good, just a few minor comments 😆

backend/src/v5/processors/teamspaces/projects/models/commons/modelList.js
Bogdan-Crn: Ok, so instead of code the key will be named tag, what about the value that it holds? Should that stay as the tag of the revision?

carmenfan: yea same data just call it tag on the response is good 👍

carmenfan (approved):

PR #5329, opened 2025-01-02
Issue 5244 - Custom Tickets Query Filters 

### LINKED ISSUE #5244: Add the ability to add query ticket list API
### Description

We want to be able to do the query server side to better serve large amount of tickets with custom property filtering.

#### Endpoint changes

Endpoint: https://www.3drepo.io/docs/#/Tickets/GetTicketList
additional query string: `query`, `limit`, `skip`

##### Query
see https://github.com/3drepo/3drepo.io/wiki/Custom-Ticket-Query-Filters

##### Limit
Limit should be supported to restrict the number of results returned by the endpoint.

e.g. `limit=10` denotes we should only return 10 results

If undefined, we will return all tickets as appropriate

##### Skip
Skip should be supported to allow the user to skip the first n entries for pageination purposes

e.g. `skip=100` denotes we should skip the first 100 results returned.

by default, skip is 0.

### Goals
- [x] As 3D Repo admin, I want to ensure modules and properties cannot start with `$` to ensure queries are not ambiguous.
- [ ] As a 3D Repo user, I want to be able to filter for tickets that matches a specific condition
- [ ] As a 3D Repo API user, I want to be able to pageinate my results to better handle the size of data.

### Tasks
- [x] Add limit and skip to the existing params
- [x] Create a query schema to validate query param
- [x] Dissallow module names / properties to start with dollar sign
- [ ] Amend the processor to convert all these to a mongo query
- [ ] Add extra unit tests
- [ ] Add extra e2e tests
- [ ] Amend documentation

ChristosTsiotsias:
This fixes #5244

The user should now be able to use query filters to fetch the appropriate tickets.
Various filter operators are available to be used as property filters (eq, neq, ss, nss etc)
The template code and ticket code ('template code':'ticket number') can be used as a filter
Module properties can also be used in filters
Properties names cannot start with $ from now on
Unit and E2E tests have been added
Swagger documenation added

#### Test cases
Query filters should be working as expected
Calling the endpoint without a query should also be working
Trying to create a module with a property that starts with $ should throw an error


backend/src/v5/routes/teamspaces/projects/models/common/tickets.js
```diff
@@ -599,6 +600,12 @@ const establishRoutes = (isFed) => {
 	 *         required: false
 	 *         schema:
 	 *           type: string
+	 *       - name: query
+	 *         description: Query string that defies tickets to be included in the response
```
carmenfan, line 604: we're going to need to give our users a bit more help than this. bear in mind any integrators will need to have access to info on how to construct this string.

So I think either we lay out the schema in swagger, or link them to a README in github (bit like the chat event README)

carmenfan: I think we're missing limit and skip?

backend/src/v5/schemas/tickets/templates.js
```diff
@@ -55,7 +55,7 @@ const pinColSchema = Yup.lazy((val) => {
 	});
 });
 
-const blackListedChrsRegex = /^[^.,[\]]*$/;
+const blackListedChrsRegex = /^(?!\$)[^.,[\]]*$/;
```
carmenfan, line 58: i think we need to block double quotes as well? otherwise that'll also cause problems..

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/tickets.test.js
```diff
@@ -664,11 +671,11 @@ const testGetTicketList = () => {
 			const baseRouteParams = { key: users.tsAdmin.apiKey, modelType, projectId: project.id, model };
 
 			const checkTicketList = (ascending = true) => (tickets) => {
-				// check the list is sorted by created at, ascending order
+				// check the list is sorted by updated at, ascending order
```
carmenfan, line 674: why did you need to change this?

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/tickets.test.js
```diff
@@ -664,11 +671,11 @@ const testGetTicketList = () => {
...
 				let lastTicketTime;
 
 				for (const entry of tickets) {
-					const createdAt = entry.properties[basePropertyLabels.CREATED_AT];
+					const createdAt = entry.properties[basePropertyLabels.UPDATED_AT];
```
carmenfan, line 678: if you have to change this then change the variable name too lol

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/tickets.test.js
```diff
@@ -687,6 +694,8 @@ const testGetTicketList = () => {
 				['the model returning only tickets updated since now', { ...baseRouteParams, options: { updatedSince: Date.now() + 1000000 } }, true, []],
 				['the model returning tickets sorted by updated at in ascending order', { ...baseRouteParams, options: { sortBy: basePropertyLabels.UPDATED_AT, sortDesc: false }, checkTicketList: checkTicketList() }, true, model.tickets],
 				['the model returning tickets sorted by updated at in descending order', { ...baseRouteParams, options: { sortBy: basePropertyLabels.UPDATED_AT, sortDesc: true }, checkTicketList: checkTicketList(false) }, true, model.tickets],
+				['the model has tickets with query filter imposed', { ...baseRouteParams, options: { query: `'${textProp.name}::${queryOperators.EQUALS}::${model.tickets[5].properties[textProp.name]}'` } }, true, [model.tickets[5]]],
+				['the model has tickets with template query filter imposed', { ...baseRouteParams, options: { query: `'$${defaultQueryProps.TEMPLATE}::${queryOperators.EQUALS}::${templatesToUse[1].code}'` } }, true, model.tickets.filter((t) => t.type === templatesToUse[1]._id)],
```
carmenfan, line 698: add test cases for skips and limits

carmenfan (changes requested):

backend/src/v5/schemas/tickets/templates.js
carmenfan: also `:`

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/tickets.test.js
ChristosTsiotsias: the test is actually passing UPDATED_AT

carmenfan: by default the tickets should be sorted by createdAt in descending order unless it's specified by the query string?

carmenfan (changes requested):
eq isn't working on numbers (lte and gte are working though...)

image

Also not working on dates:
image

Boolean equals also not working:
image

$ticketCode with ss isn't working but it doesn't throw an error (if it's not a supported operator? if it is then we need to make sure it works)
image


In general could we make a set of test sto ensure all property types with their supported operator is working as expected? (may have to be e2e to ensure we're actually running the mongo queries)

carmenfan (changes requested):

carmenfan (approved):

PR #5344, opened 2025-01-14
Issue 5327 - Fixed warnings when drag and dropping file not showing

### LINKED ISSUE #5327: Container/drawing filesize limit error does not trigger when drag 'n' dropping a file
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/637

When dragging and dropping a file which exceeds the upload limit there is no error triggered. However adding a file using the browse button has no such problems. 

Also note that for sizes under 1KiB the size unit is missing. In the screenshot below the filesize should read `663B` as it is a 663 Byte file. It is probably unlikely a user will upload such a small file, however it is an easy fix

I believe this comes from this PR https://github.com/3drepo/3drepo.io/pull/5229. 
NOTE: this was found before https://github.com/3drepo/3drepo.io/pull/5277 so the  package upgrade is not responsible

### Steps to replicate
- In your local set the uploadSizeLimit in your config to something very low (e.g. 10)
- Open the upload modal
- Click browse and upload a file. This should trigger the error as expected
- Drag and drop the same file

### Current Behaviour

- No error is triggered when dragging and dropping a file that is too large

### Expected Behaviour

- An error is triggered when adding a file that is too large regardless of method

image

sanmont3drepo:
This fixes #5327 

#### Description
- Changed how files are being validated when uploading a revision or a drawing

#### Test cases
- Drag and dropping a small file in drawings or revisions uploads will not trigger a "max file size" error 
- Uploading a small file  by clicking browse in drawings or revisions uploads will not trigger a "max file size" error 
- Drag and dropping a large file in drawings or revisions uploads will trigger a "max file size" error 
- Uploading a large file  by clicking browse in drawings or revisions uploads will trigger a "max file size" error 
- Uploading a multipage pdf to drawings will show the "multipage pdf" warning

### DIFF
--- a/frontend/src/v5/helpers/intl.helper.ts
+++ b/frontend/src/v5/helpers/intl.helper.ts
@@ -22,30 +22,8 @@ type ByteSizeType = {
 	unit: string
 };
 
-const InformationUnits = {
-	B: formatMessage({ id: 'units.information.B', defaultMessage: 'B' }),
-	kB: formatMessage({ id: 'units.information.kB', defaultMessage: 'kB' }),
-	MB: formatMessage({ id: 'units.information.MB', defaultMessage: 'MB' }),
-	GB: formatMessage({ id: 'units.information.GB', defaultMessage: 'GB' }),
-	TB: formatMessage({ id: 'units.information.TB', defaultMessage: 'TB' }),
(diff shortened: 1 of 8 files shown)


Amantini1997:
I still get no size units on very small files 
image

Amantini1997 (approved):

The-Daniel (approved):

PR #5370, opened 2025-01-28
Issue 5270 - Ticket Filters: Integration with backend

### LINKED ISSUE #5270: Ticket filters: integration with backend.
### Description
We are going to use the filtering from the backend in order to not fetch all the tickets data. Based on https://github.com/3drepo/3drepo.io/issues/5244 we should  

### Goals
- [ ] As a user I want to be able to filter tickets

### Tasks
- [ ] Add fetch endpoints in API files 
- [ ] Modify sagas to use the endpoints
- [ ] Modify selectors to show the filtered tickets
- [ ] Wire up ticket card to use these 

### Related Resources
[Endpoint specification](https://github.com/3drepo/3drepo.io/issues/5244)

Amantini1997:
This fixes #5270

#### Description
Backend is wired in.
2 out of 3 special properties (ticketTitle and ticketId) have been renamed to match what the backend expects.
Also the operators changed a bit for the same reason.

Sending a backend request for the filtered tickets returns the whole tickets. Instead of storing them, I made it such that we store those ids and use them to list the actual tickets.
The filtered tickets are fetched everytime the tickets or the filters change (add/remove/edit) while in the ticketsListView, as well as whenever the user reaches such a view

#### Test cases
Try out the different properties and focus on performance and lack of unnecessary backend requests.

**BEWARE**: template is a "select type", which does not exist yet in this PR and allows any operators. However, the backend doesn't allow that and returns an error message if the operator is not supported. This is expected


sanmont3drepo (changes requested):

frontend/src/v5/services/api/tickets.ts
```diff
@@ -59,14 +59,24 @@ export const fetchFederationTemplate = async (
 	return data;
 };
 
+type TicketsQueryParams = {
+	extraProperties?: string[],
```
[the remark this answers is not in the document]
Amantini1997: `extraProperties` was already used before, I didn't change it, and my guess is that it originated from what Swagger calls it 
image

I'm not opposed to change it if you think it'll make it clearer

frontend/src/v5/store/tickets/tickets.selectors.ts
```diff
@@ -79,8 +80,8 @@ export const selectTickets = createSelector(
 	selectTicketsRaw,
 	selectTicketsGroups,
 	(state, modelId) => modelId,
-	(state) => state,
-	(ticketsList, groups, modelId, storeState): ITicket[] => {
+	(ticketsList, groups, modelId): ITicket[] => {
+		const storeState = getState();
```
The-Daniel, line 84: Why has this been changed to use `getState()`?

frontend/src/v5/store/tickets/tickets.sagas.ts
```diff
@@ -41,12 +41,12 @@ import { selectTicketByIdRaw, selectTicketsGroups } from './tickets.selectors';
...
 import { getSanitizedSmartGroup } from './ticketsGroups.helpers';
 
-export function* fetchTickets({ teamspace, projectId, modelId, isFederation, filters }: FetchTicketsAction) {
+export function* fetchTickets({ teamspace, projectId, modelId, isFederation, propertiesToInclude }: FetchTicketsAction) {
 	try {
 		const fetchModelTickets = isFederation
 			? API.Tickets.fetchFederationTickets
 			: API.Tickets.fetchContainerTickets;
-		const tickets = yield fetchModelTickets(teamspace, projectId, modelId, filters);
+		const tickets = yield fetchModelTickets(teamspace, projectId, modelId, { propertiesToInclude: propertiesToInclude });
```
The-Daniel, line 49: This object can be simplified

The-Daniel (changes requested):

sanmont3drepo (approved):
The templates filters are not working but I think that should be fixed in https://github.com/3drepo/3drepo.io/pull/5391

frontend/src/v5/store/tickets/tickets.selectors.ts
Amantini1997: Because of the way selectors cache data. In order for them to avoid recomputing the output, they check the current parameters against the previous ones. When you pass the `storeState` as we did before, any changes, even the ones not affecting tickets, was causing this to re-compute the data, which impacted negatively the performance

The-Daniel (approved):

sanmont3drepo (approved):

Issue #5356, opened 2025-01-21
Add external IDP support


carmenfan:
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/638

We want to replace internal mongodb to store user information with an external IDP. By the end of this issue, users should be stored in the external IDP, and any user related information should be fetched through IDP's API.

The only exception being teamspace association, teamspace invites and API keys.

### Goals
- [ ] As a user I want to be able to log in
- [ ] As a user I want to be able to sign up
- [ ] As a 3DR Admin I want to manage users with an external IDP
- [ ] As a 3DR Admin I want to ensure existing users will be able to continue to use the service
- [ ] As a 3DR Admin I want to ensure any features related to users (teamspace assignment, issues/risks/tickets, user settings) still work as intended

### Tasks
- [ ] Estbablish configuration settings
- [ ] Decommission and replace:
  - [ ] Sign up
    - [ ] Users will have to have an entry in mongo for api/teamspace association
  - [ ] Log in
    - [ ] session management changes incorporating validity check on jwt
  - [ ] Email verification, passwords
- [ ] Update association
  - [ ] Profile page
  - [ ] Migraiton scripts
  - [ ] teamspace related data (members list, ticket assignment)
- [ ] Data migration

PR #5377, opened 2025-01-31
Issue 5339 - Multiple jobs support

### LINKED ISSUE #5339: v5 endpoint to Jobs and multiple jobs support
## Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/610

We want to implement v5 endpoints for jobs with the following changes:
1. a user can have more than one job
2. a user can have 0 job.

##### Add new job
Endpoint: POST /teamspaces/:teamspace/jobs
permissions: teamspace admin
payload: 
````js
{
  name: "job name",
  colour: "#000000", //RGB hex, optional
  users: ["user1"], //optional
}
````

response
````js
{
  _id: "UUID string"
}
````

##### Edit an existing job
Endpoint: PATCH /teamspaces/:teamspace/jobs/:jobID
permissions: teamspace admin
payload: 
````js
{
  name: "new name",
  users: ["user1", "user2"]
}
````

##### Remove an existing job
Endpoint: Delete /teamspaces/:teamspace/jobs/:jobID
permissions: teamspace admin

##### Teamspace members
URL: https://staging.dev.3drepo.io/docs/#/Teamspaces/getTeamspaceMembers
Changes: Should now return an array of "jobs" instead of "job"

### Additional work
- As specified above, we want to add an ID to job instead of using the name of the job as unique identification. We will have to do a migration on existing data, and also data that references jobs (i.e. tickets/issues/risks). We will also need to make adjustment on any associated endpoints.
- Care needs to be taken with how jobs are used. We need to check everywhere that references/uses jobs logic and ensure they still work as intended. This includes endpoints within v4
- We should also remove v4 job endpoints to avoid clashes (with approach described in https://github.com/3drepo/3drepo.io/issues/5317)

## Goals
- [ ] As ts admin I want to be able to do everything I need to do on the old interface
 - [ ] As ts admin I want the data within the old data is preserved and presented in this new interface
- [ ] As a ts admin I want to be able to edit the name fo the job if necessary
- [ ] As a ts admin I want to be able to assign a user to multiple jobs if they are involved in more than one type of work
- [ ] As a system integrator I want to be able to support scenarios such as a user with no associated jobs if this is how it's setup on the other system.

## Tasks
- [x] Create endpoint for adding job
- [x] Create endpoint for editing job
- [x] Create endpoint for removing job 
- [x] Add documentation for endpoints
- [x] Add unit tests for endpoints
- [x] Add e2e tests for endpoints
- [x] Change getMembers endpoint to return multiple jobs
- [x] Write script for changing default jobs
- [x] Check logic that uses jobs and make sure it works
- [x] Remove old v4 endpoints

- Move to frontegg:
  - [x] roles to be saved on frontegg
  - [x] roles updated on frontegg
  - [x] roles removed in frontegg
  - [x] users roles to be obtained from frontegg
    - [x] GetRoles
    - [x] getRolesByUsers
    - [x] GetRolesToUsers
    - [x] getRolesByAccess
  - [x] migration script to copy roles to frontegg
  - [x] create teamspace need to create the first role
  - [x] getTeamspaceMember info - doing some extra logic for the ID
  - [x] ensure v4 is wired up correctly
    - [x]  v4 roles file had additional job related work - wire it to v5.
    - [x] Ideally the entire v4 should point to v5 processor at this point (models/users, models/roles etc)]
      - [x] special attention to:
        - [x] models/users - has a lot of teamspace member checks that needs moving
  - [x] check every file that is referencing models/roles and make sure they call the right function
  - [x] getAccessibleTeamspaceList needs to be removed from models and adjust the callees to get info from frontegg]
  - [x] Adjust service helper for creating jobs
  - [x] check references for `'roles.db'`
  - [x] make sure the new roles middleware checks for user provision flag
    - [ ] Make sure e2e tests also checks for this

- Clean up
  - [x] Ensure deprecated endpoints are repsonded accordingly (v5 jobs)
  - [x] v4 roles (teamspace_member) thing, should be deprecated? check #5480 
  - [ ] v5 roles (teamspace_member thing) 
  - [ ] v5 models/modelSettings `getUsersWithNoAccess` should be isTeamspaceMembers that take in more than one user name
  - [ ] projectSettings create index should be done inside project settings...
  - [ ] check v5 notifications for add/remove users and roles
  - [ ] frontend beware: invitations needs to pass in roleIds
  - [ ] Check removing user from teamspace would auto remove him/her from the role.
  - [ ] Can we cache the isTeamspaceMember check?
    - [ ] check memberExists middleware - if it's  only ever used for the user or 3rd party

- To test
  - [ ] Invitations roles are recorded correctly (actually we don't need to? we're just sending it to frontegg?)
  - [ ] Create teamspace - adding owner to the right job
  - [ ] test invites
  - [ ] test migration script - teamspace member related changes might've messed it up
  - [ ] find user to add to teamspace prompt still works

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #5339 

#### Description

Jobs are renamed to roles and have UUID ids and name
Migration script for updating issues, risks, tickets, templates and ticket logs added
Users can now be asssigned to multiple roles
Endpoints for creating / updating / deleting a role added
Respective v4 endpoints removed and an appropriate response is returned
Swagger documentation added
Unit and e2e tests added

#### Test cases

Migration script should run and update all necessary collections
Users should be able to create/update/delete a role using v5 endpoints
Old v4 endpoints should return appropriate error and point to the new endpoints

### CODE REVIEW

File: backend/src/v5/middleware/dataConverter/inputs/teamspaces/roles.js [hunk 1]
```diff
@@ -0,0 +1,95 @@
...
+					const usersWithNoAccess = getArrayDifference(teamspaceUsers, values);
+					if (usersWithNoAccess.length) {
+						return context.createError({ message: `User(s) ${usersWithNoAccess} have no access to the teamspace` });
+					}
+				}
+
+				return true;
+			}),
+
+		color: Yup.string().matches(/^#[0-9A-Fa-f]{6}$/, 'color is not a valid RGB hex format'),
```
carmenfan, line 62: Add a new type in yupHelper for RGB, we're bound to need it again

backend/src/v5/middleware/dataConverter/inputs/teamspaces/roles.js
```diff
@@ -0,0 +1,95 @@
...
+				return true;
+			}),
+		users: Yup.array().of(types.strings.title)
+			.test('users-uniqueness-check', 'users must be unique', (values) => {
+				if (!values?.length) return true;
+				return values.length === uniqueElements(values).length;
+			})
+			.test('check-users-teamspace-access', async (values, context) => {
+				if (values?.length) {
+					const teamspaceUsers = await getAllUsersInTeamspace(req.params.teamspace);
```
carmenfan, line 52: create a function in models/teamspace to do the check given the list of users.

backend/src/v5/models/roles.js
```diff
@@ -15,11 +15,30 @@
...
+
+const COL_NAME = 'roles';
 const db = require('../handler/db');
+const { generateUUID } = require('../utils/helper/uuids');
+const { templates } = require('../utils/responseCodes');
 
 const Roles = {};
 
+const findOne = async (teamspace, query, projection) => {
+	const template = await db.findOne(teamspace, COL_NAME, query, projection);
```
carmenfan, line 28: template is probably not the right name lol

backend/src/v5/models/roles.js
```diff
@@ -28,4 +47,41 @@ Roles.grantTeamspaceRoleToUser = (teamspace, username) => db.grantRole(teamspace
 
 Roles.revokeTeamspaceRoleFromUser = (teamspace, username) => db.revokeRole(teamspace, TEAM_MEMBER, username);
 
+// --- JOBS RENAMED TO ROLES --- TO BE KEPT
+
+Roles.getRolesToUsers = (teamspace) => findMany(teamspace, {}, { _id: 1, users: 1 });
+
+Roles.getRoles = (teamspace, projection) => findMany(teamspace, {}, projection);
+
+Roles.getRoleById = (teamspace, roleId, projection) => findOne(teamspace, { _id: roleId }, projection);
```
carmenfan, line 56: Unless linter hates it, we can just do this? 
```suggestion
Roles.getRoleById = (teamspace, _id, projection) => findOne(teamspace, { _id }, projection);
```

backend/src/v5/models/roles.js
```diff
@@ -28,4 +47,41 @@ Roles.grantTeamspaceRoleToUser = (teamspace, username) => db.grantRole(teamspace
...
+Roles.getRoleByName = (teamspace, roleName, projection) => findOne(teamspace, { name: roleName }, projection);
```
carmenfan, line 57: Same here
```suggestion
Roles.getRoleByName = (teamspace, name, projection) => findOne(teamspace, { name }, projection);
```

backend/src/v5/routes/routesManager.js
```diff
@@ -45,7 +45,7 @@ RoutesManager.init = (app) => {
 
 	app.use('/v5/teamspaces/', TeamspaceRoutes);
 	app.use('/v5/teamspaces/:teamspace/settings', TeamspaceSettingsRoutes);
-	app.use('/v5/teamspaces/:teamspace/jobs', TeamspaceJobRoutes);
+	app.use('/v5/teamspaces/:teamspace/roles', TeamspaceRoleRoutes);
```
carmenfan, line 48: Can we add a deprecated redireect response for the old jobs endpoint?

backend/src/v5/routes/teamspaces/projects/models/common/general.js
```diff
@@ -986,11 +987,11 @@ const establishRoutes = (modelType) => {
 
 	/**
 	 * @openapi
-	 * /teamspaces/{teamspace}/projects/{project}/{type}/{model}/jobs:
+	 * /teamspaces/{teamspace}/projects/{project}/{type}/{model}/roles:
```
carmenfan, line 990: again, add deprecate redirect

backend/src/v5/routes/teamspaces/roles.js
```diff
@@ -0,0 +1,265 @@
...
+		respond(req, res, templates.ok);
+	} catch (err) {
+		// istanbul ignore next
+		respond(req, res, err);
+	}
+};
+
+const deleteRole = async (req, res) => {
+	const { teamspace, role } = req.params;
+	const updatedRole = req.body;
```
carmenfan, line 68: I don't think we expect a body on deletes. It's definitely not in the HTTP convention 😆

backend/src/v5/routes/teamspaces/roles.js
```diff
@@ -0,0 +1,265 @@
...
+};
+
+const establishRoutes = () => {
+	const router = Router({ mergeParams: true });
+	/**
+	* @openapi
+	* /teamspaces/{teamspace}/roles:
+	*   get:
+	*     description: Get the list of roles within this teamspace
+	*     tags: [Teamspaces]
```
carmenfan, line 86: this applies to every route here

Can we use the roles tag?

backend/src/v5/schemas/tickets/index.js
```diff
@@ -54,13 +54,13 @@ const generatePropertiesValidator = async (teamspace, project, model, templateId
...
-				case presetEnumValues.JOBS_AND_USERS:
+				case presetEnumValues.ROLES_AND_USERS:
 					{
 						const excludeViewers = prop.name === basePropertyLabels.ASSIGNEES;
 						const users = await getUsersWithPermissions(teamspace, project, model, excludeViewers);
-						const accessibleJobs = await getJobsByUsers(teamspace, users);
+						const accessibleRoles = await getRolesByUsers(teamspace, users);
 
-						values = [...accessibleJobs, ...users];
+						values = [...accessibleRoles.map(UUIDToString), ...users];
```
carmenfan, line 63: are the role IDs converted to UUID at any point? we want make sure we store UUIDs on the tickets. I don't thnk I see a conversion

There's also need to be a convertion on the middleware on the response in GET requests for tickets

backend/src/v5/utils/responseCodes.js
```diff
@@ -94,6 +97,9 @@ ResponseCodes.templates = {
 	// Invalid Arguments
 	invalidArguments: { message: 'The arguments provided are not valid.', status: 400 },
 
+	// Resource not available
+	resourceNotAvailable: { message: 'Resource not available.', status: 410 },
```
carmenfan, line 101: Can we rename this to `endpointDecomissioned`
```suggestion
	endpointDecomissioned: { message: 'Endpoint no longer available.', status: 410 },
```

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/general.test.js
```diff
@@ -65,10 +67,16 @@ const generateBasicData = () => {
 		calibration: ServiceHelper.generateCalibration(),
 	};
 
-	data.jobs = [
-		{ _id: ServiceHelper.generateRandomString(), users: [viewer.user] },
-		{ _id: ServiceHelper.generateRandomString(), users: [collaborator.user] },
-		{ _id: ServiceHelper.generateRandomString(), users: Object.values(data.users).map(({ user }) => user) },
+	data.roles = [
```
carmenfan, line 70: since we're doing this in multiple places, can we just have a generateRole that takes in a user array and use that everywhere?

backend/tests/v5/e2e/routes/teamspaces/index.test.js
```diff
@@ -252,12 +254,12 @@ const testGetTeamspaceMembers = () => {
 			expect(res.body.code).toEqual(templates.teamspaceNotFound.code);
 		});
 
-		test('should return list of users with their jobs with valid access rights', async () => {
+		test('should return list of users with their roles with valid access rights', async () => {
```
carmenfan, line 257: Can we add test case for:
1. user with no roles (we'd expect an empty array)
2. user with more than one role (we'd expect array length > 1)

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/tickets.test.js
carmenfan, line 1: Can we add a test here to make sure we're getting a converted role ID to string?

Since I don't see a middleware change I've pretty sure we're spitting out whatever we're storing in the database, so we're either storing a string UUID or we're outputting jibberish

backend/tests/v5/e2e/routes/teamspaces/roles.test.js
```diff
@@ -0,0 +1,214 @@
...
+			['teamspace does not exist', users.tsAdmin.apiKey, generateRandomString(), false, templates.teamspaceNotFound],
+			['user is not a member of the teamspace', users.nobody.apiKey, undefined, false, templates.teamspaceNotFound],
+			['user is a ts admin', users.tsAdmin.apiKey, undefined, true],
+			['user is a member', users.normalUser.apiKey, undefined, true],
+		])('', (desc, key, ts, success, expectedRes) => {
+			test(`should ${success ? 'succeed' : `fail with ${expectedRes.code}`} if ${desc}`, async () => {
+				const expectedStatus = success ? templates.ok.status : expectedRes.status;
+				const res = await agent.get(route(key, ts)).expect(expectedStatus);
+				if (success) {
+					expect(res.body.roles).toEqual(expect.arrayContaining(roles));
```
carmenfan, line 84: There's an [out of order equal array check function](https://github.com/3drepo/3drepo.io/blob/master/backend/tests/v5/helper/services.js#L350) in service helper. would be more appropriate to use that as arrayContaining would only check for the subset

backend/tests/v5/e2e/routes/teamspaces/roles.test.js
```diff
@@ -0,0 +1,214 @@
...
+	const getRolesRoute = (key, ts = teamspace) => `/v5/teamspaces/${ts}/roles${key ? `?key=${key}` : ''}`;
+	const addRoleRoute = (key, ts = teamspace) => `/v5/teamspaces/${ts}/roles${key ? `?key=${key}` : ''}`;
+
+	describe.each([
+		['user does not have a valid session', undefined, undefined, roleData, false, templates.notLoggedIn],
+		['teamspace does not exist', users.tsAdmin.apiKey, generateRandomString(), roleData, false, templates.teamspaceNotFound],
+		['user is not a member of the teamspace', users.nobody.apiKey, undefined, roleData, false, templates.teamspaceNotFound],
+		['user is not a teamspace admin', users.normalUser.apiKey, undefined, roleData, false, templates.notAuthorized],
+		['user is a ts admin and data is valid', users.tsAdmin.apiKey, undefined, roleData, true],
+		['data is invalid (missing name)', users.tsAdmin.apiKey, undefined, { ...roleData, name: undefined }, false, templates.invalidArguments],
```
carmenfan, line 116: Can we add name clash as a test case

carmenfan: Can we also add a test that post 2 roles with the same name pretty much in parallel and make sure one of them fails (to test the unqiue index thing)

backend/tests/v5/e2e/routes/teamspaces/roles.test.js
```diff
@@ -0,0 +1,214 @@
...
+
+	describe.each([
+		['user does not have a valid session', undefined, undefined, roles[0]._id, roleData, false, templates.notLoggedIn],
+		['teamspace does not exist', users.tsAdmin.apiKey, generateRandomString(), roles[0]._id, roleData, false, templates.teamspaceNotFound],
+		['user is not a member of the teamspace', users.nobody.apiKey, undefined, roles[0]._id, roleData, false, templates.teamspaceNotFound],
+		['user is not a teamspace admin', users.normalUser.apiKey, undefined, roles[0]._id, roleData, false, templates.notAuthorized],
+		['user is a ts admin and data is valid', users.tsAdmin.apiKey, undefined, roles[0]._id, roleData, true],
+		['role does not exist', users.tsAdmin.apiKey, undefined, generateRandomString(), roleData, false, templates.roleNotFound],
+		['data is invalid (empty body)', users.tsAdmin.apiKey, undefined, roles[0]._id, { }, false, templates.invalidArguments],
+		['data is invalid (users with non ts members)', users.tsAdmin.apiKey, undefined, roles[0]._id, { ...roleData, users: [users.nobody.user] }, false, templates.invalidArguments],
```
carmenfan, line 155: Can we test:

- name already exists
- name is unchanged but submitted again (so detecting a name clash but it's the same role as the one we're trying to update)

backend/tests/v5/unit/models/roles.test.js
```diff
@@ -75,9 +83,230 @@ const testRevokeTeamspaceRoleFromUser = () => {
...
+	describe('Get Roles to users', () => {
+		test('should get list of roles within the teamspace with the users', async () => {
+			const expectedResult = [
+				{ _id: 'roleA', users: ['a', 'b', 'c'] },
+			];
+			const fn = jest.spyOn(db, 'find').mockImplementation(() => expectedResult);
+			const teamspace = 'ts';
+			const res = await Roles.getRolesToUsers(teamspace);
+			expect(res).toEqual(expectedResult);
+			expect(fn.mock.calls.length).toBe(1);
```
carmenfan, line 98: why aren't we using `expect().toHaveBeenCalledTimes(1)`
and `expect().toHaveBeenCalledWith()` etc?

If it's because it's just copy and pasted, let's just change it whilst we're editing

backend/tests/v5/unit/models/roles.test.js
```diff
@@ -75,9 +83,230 @@ const testRevokeTeamspaceRoleFromUser = () => {
 	});
 };
 
+// --- JOBS RENAMED TO ROLES --- TO BE KEPT
+
+const testGetRolesToUsers = () => {
+	describe('Get Roles to users', () => {
+		test('should get list of roles within the teamspace with the users', async () => {
+			const expectedResult = [
+				{ _id: 'roleA', users: ['a', 'b', 'c'] },
```
carmenfan, line 92: use generateRandomString()

backend/tests/v5/unit/models/roles.test.js
carmenfan: same with the functions below,

backend/tests/v5/unit/models/roles.test.js
carmenfan: same with the functions below,

backend/src/v5/routes/teamspaces/teamspaces.js
```diff
@@ -191,10 +191,13 @@ const establishRoutes = () => {
...
 	 *                         example: 3D Repo Ltd
-	 *                       job:
-	 *                         type: string
-	 *                         description: Job within the teamspace
-	 *                         example: Project Manager
+	 *                       roles:
+	 *                         type: array
+	 *                         items:
+	 *                           type: string
+	 *                           format: uuid
```
carmenfan, line 198: Since I don't see a change here or an addition of middleware here, I'm guessing  are we converting the UUIDs to string here?

Can we sanity check all the routes are converting UUIDs to string, and the database is storing UUIDs?

backend/src/v5/models/roles.js
```diff
@@ -28,4 +47,41 @@ Roles.grantTeamspaceRoleToUser = (teamspace, username) => db.grantRole(teamspace
...
+Roles.removeUserFromRoles = async (teamspace, userToRemove) => {
+	await updateMany(teamspace, { users: userToRemove }, { $pull: { users: userToRemove } });
+};
+
+Roles.getRolesByUsers = async (teamspace, users) => {
+	const roles = await findMany(teamspace, { users: { $in: users } }, { _id: 1 });
+	return roles.map((j) => j._id);
+};
+
+Roles.createRole = async (teamspace, role) => {
```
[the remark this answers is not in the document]
ChristosTsiotsias: I guess I should add it in the migration script as well as initTeamspace ?

backend/src/v5/middleware/dataConverter/inputs/teamspaces/roles.js
```diff
@@ -0,0 +1,95 @@
...
+	.then(() => true).catch(() => false);
+
+const validateRole = async (req, res, next) => {
+	const isUpdate = !!req.roleData;
+
+	let schema = Yup.object().shape({
+		name: (isUpdate ? types.strings.title : types.strings.title.required())
+			.test('check-name-is-unique', 'Role with the same name already exists', async (value) => {
+				if (value !== req.roleData?.name) {
+					const nameTaken = await nameExists(req.params.teamspace, value);
```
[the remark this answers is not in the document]
ChristosTsiotsias: It doesn't because value will be undefined so there isn't a role with undefined name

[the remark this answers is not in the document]
carmenfan: knowing how the frontend sends data, it is very possible for the network request to come in with the existing name of the role even if they have no intention to update it.

I don't think we can make assumption on that, it's technically valid for someone trying to update a role with the same name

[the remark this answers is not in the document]
ChristosTsiotsias: I dont get it 🤔 if the value is the same then  if (value !== req.roleData?.name) will return false and the test will return true straight away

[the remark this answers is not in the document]
carmenfan: Ah true you do have that if statement..


But I think you're better off making an exclusion in your query so there's no chance of async errors. it's always better to make those checks atomic, and you won't need that if statement.

But thinking about it if we make the unqiue index change on the role name else where these are all not very important anyway 😆

backend/src/v5/models/roles.js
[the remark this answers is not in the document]
carmenfan: > I guess I should add it in the migration script as well as initTeamspace ?

yes please

carmenfan (changes requested):
As discussed, we need to consider #5356 for this.

We want to instead migrate to frontegg to store our roles in their groups (so for e.g. if a new role is created, we call this : https://developers.frontegg.com/api/identity/user-groups/groupscontrollerv1_creategroup)

So edit and update should be going through frontegg (sorry 😢  🙈 )

backend/tests/v5/e2e/routes/users.test.js
```diff
@@ -130,6 +130,11 @@ const testEndpointRoutes = () => {
 				.send({ user: testUser.user, password: testUser.password })
 				.expect(templates.pageNotFound.status);
 		});
+
```
carmenfan, line 133: Not the right place for this test

PR #5389, opened 2025-02-04
ISSUE #5374 - Tickets filters: initial filters

### LINKED ISSUE #5374: Ticket filters: initial filters
### Description
Parent issue: https://github.com/3drepo/3drepo.io/issues/5188

In the ticket list we need to show relevant tickets. That means showing the tickets that are being filtered by query string and when we dont have a query string
- When there's a query param `ticketSearch` that is an array of ticket codes. We should set a filter by ticket codes using the param.

<details>
<summary> When there's no query param the initial filtering should filter by not completed statuses.</summary>

This means filtering status values in `Status` (basic property) and in `safetibase:'Treatment Status'`

### Completed in `Status` basic property 
- For the basic property `Status` get the status config from the template config (`template.config.status`) 
- From the status config get the names with types 'done' and 'void'
- Set a filter to `Status` _not equal_ the previously obtained names.

### Completed in safetibase `Treatment Status` property
- `safetibase:'Treatment Status'` set the filter to _not_equal_  `['Agreed (Fully)', 'Void']`
</details>

### Goals
- [ ] As a user I want to be see relevant tickets initially

### Tasks
- [ ] Rename `HandleTicketsCardSearchParams` to `HandleInitialFilters` or `InitialFiltersSetter` or something more relevant like that
- [ ] Update `HandleTicketsCardSearchParams` by changing how filters are set.  This is an array with ticket codes so set the ticket code filter with this equals this array
- [ ] Remove 'ticketTemplates' and 'ticketCompleted' from the query handling
- [ ] Modify`HandleTicketsCardSearchParams` to set the completed filter(as specified earlier) if there is no 'ticketSearch' param

Amantini1997:
This fixes #5374 

#### Description
Set default ticket filters:
- if the URL contains valid ticket codes, use them
- set to hide any status that is either "closed" or "void". If safetibase exists in any template, filter according the same logic for treatment status

#### Test cases
1. Append ticket codes in the URL like `?ticketSearch=ALL:1,ALL:2`
- Only those ticket codes should be used for filters

2. Leave the URL as is
- Only status "close" and "void" should be used in filters as "not equals"

3. Have a project with safetibase templates
- The filters should be for both status and safetibase

4. Have a template with custom statuses
- The filters should include the equivalent of "close" and "void" for custom status


frontend/src/v5/store/tickets/card/ticketsCard.redux.ts
```diff
@@ -93,6 +94,13 @@ export const setPinToDrop = (state: ITicketsCardState, { pinToDrop }: SetPinToDr
 };
 
 const getFilterKey = ({ module, property, type }: CardFilter): TicketFilterKey => `${module}.${property}.${type}`;
+export const setFilters = (state: ITicketsCardState, { filters }: SetFiltersAction) => {
+	filters.forEach((filter) => {
+		const path = getFilterKey(filter);
+		state.filters[path] = filter.filter;
+	});
+};
+
```
sanmont3drepo, line 103: I dont like having a redux action for something that you could already do with other actions (upsert filter). you could just do  `filters.foreach(TicketsCardActionsDispatchers.upsertFilter)` in defaultticketfiltersSetter  instead of adding a specific action for setting all the filters (unless we come up with a different use for TicketsCardActionsDispatchers.setFilters).

sanmont3drepo (changes requested):

The-Daniel (approved):

sanmont3drepo (approved):

Issue #5403, opened 2025-02-13
onPRClose action can only destroy deployments following issue-xxx format


Charence:
### Description
`onPRClose.yml` expects all issues to follow the `isssue-xxx` format where `xxx` is the issue number: https://github.com/3drepo/3drepo.io/blob/master/.github/workflows/onPRClose.yml

This will need to change if we have branches that don't follow the format, e.g. qa, issue-123b, etc..

### Suggested solution
- Use branch name instead of recreating from issue number to destroy deployment
- Otherwise, accept manual intervention required to remove those branches

### Goal
- Destroy `isssue-xxx` branches as before
- Destroy other branches that don't following that naming format, e.g. qa, issue-123b, etc.

carmenfan:
> Use branch name instead of recreating from issue number to destroy deployment

The PR should have data about the branch we're merging from so we should 100% do that instead of guessing

Charence:
https://github.com/3drepo/3drepo.io/actions/runs/13313446883/job/37181644451

Charence:
Helm chart names should not have uppercase/underscores: https://helm.sh/docs/chart_best_practices/conventions/

PR #5410, opened 2025-02-14
Testing testbranch closure

Charence:
Testing #5403


Charence:
`testbranch` deployment not deleted as desired because name matches `testbranch`

image

Charence:
`testbranch` is no longer protected (now set to `staging`) so this deployment should be destroyed on PR close.

Charence:
Removed as expected:
image
image

PR #5413, opened 2025-02-14
ISSUE 5403 - Better handle destroy for issue-xxx and text only branch names

### LINKED ISSUE #5403: onPRClose action can only destroy deployments following issue-xxx format

Charence:
This fixes #5403 

#### Description
- `issue-xxx` destroyed as before
- `qa`, `testbranch`, etc. will now also be destroyed
- Do not destroy `staging`

#### Test cases
- All scenarios tested via PR close (see #5403 and referenced test PRs)


.github/workflows/onPRClose.yml
```diff
@@ -50,13 +50,16 @@ jobs:
   destroy-deployed-branch:
     name: Call Azure Destroy Pipeline
     runs-on: ubuntu-latest
-    needs: get-details
     steps:
+    - name: Set new branch name value
```
carmenfan, line 54: This naming is misleading since this is an destruction action, we're not really creating/setting a branch name.

We're trying to figure out what the deployment charname is so maybe change it for a better description?

.github/workflows/onPRClose.yml
```diff
@@ -50,13 +50,16 @@ jobs:
...
+      run: |
+        echo 'chartName='$( echo ${{ github.head_ref || github.ref_name }} | sed "s/_/-/" | awk '{print tolower($0)}' ) >> "$GITHUB_OUTPUT"
+      id: getChartName
     - name: Azure Pipelines Action
-      if: needs.get-details.outputs.issue-number
+      if: steps.getChartName.outputs.chartName != 'staging'
```
carmenfan, line 59: Could we also check master? just for the odd chance that are merge master to staging

carmenfan (changes requested):

Charence:
Tested

carmenfan (approved):

PR #5417, opened 2025-02-19
Issue 5387 - provide empty values on negative searches

### LINKED ISSUE #5387: Ticket filter: Need to also provide empty values on negative searches
### Description

We need to include tickets that doesn't have a value on the negative searches.

For e.g. if I want all the status that isn't closed, I want to see any tickets that doesn't have a status.

### Goals
- [ ] As a user, when I'm filtering tickets any tickets that does not equal to something, I expect tickets with no value to appear
- [ ] As a user, when I'm filtering tickets any tickets that does not contain, I expect tickets with no value to appear
- [ ] As a user, when I'm filtering tickets any tickets that is not within the range of something, I expect tickets with no value to appear

### Tasks

Bogdan-Crn:
This fixes #5387 

#### Description
Updated the query logic to ensure that on negative searches empty values are retrieved.
Fixed the tests to reflect the new logic.


carmenfan (changes requested):
as mentioned by @ChristosTsiotsias  it seems like the behaviour is already working as intended.

Can we remove the code changes and just leave the e2e test and merge that in?

Sorry for making you go around in circles!

PR #5419, opened 2025-02-20
ISSUE #5418 kill the server if we run into Topology is closed

### LINKED ISSUE #5418: Session manager doesn't recover from Topology is closed error
### Description
`mongo-connect` within the session management doesn't seem to be able to deal with recovery when it runs into a `Topology is closed` error.

### Steps to reproduce
- Run this on the browser: `https://{domain}/api/A%20Test/24770c45-8082-4d51-bcdd-b9bfad3eabcd/revision/master/head/meta/all.json?key={key}`
- Then run this: `https://{domain}/api/v5/login`

### Current Behaviour
- The first one errors (as expected), and the second one also errors, looking at the server log it is failing with `MongoError: Topology is closed`
- Subsequent requests relying on a cookie session also fails.

### Expected Behaviour
- It should not happen, subsequent requests should run fine

carmenfan:
This fixes #5418

#### Description
- The session manager is stuck in a weird state that is struggling to connect to mongo through the provided connection.
- This situation doesn't trigger mongo-connect's autoReconnect mechanism, and there is no way to manually replace the connection (see https://github.com/jdesboeufs/connect-mongo/issues/485)
- So instead, I added in a kill switch to just kill off the application (following the unhandled promise rejection route), which in a deployment would kill the pod and generate a new one.

#### Test cases
- in the mentioned situation, the 2nd request should now kill off the server.
  - An email should be generated to explain the situation

### DIFF
--- a/backend/src/v4/response_codes.js
+++ b/backend/src/v4/response_codes.js
@@ -416,6 +416,10 @@
 	 */
 	responseCodes.respond = function (place, req, res, next, resCode, extraInfo, format, cache, customHeaders) {
 
+		// Topology is closed mongo error is typically coming from the session management and the library
+		// doesn't let us recover from it (so far). So kill this pod and let it respawn.
+		const killServer = resCode?.name === "MongoError" && resCode?.message?.includes && resCode.message.includes("Topology is closed") ? resCode : false;
+
 		resCode = utils.mongoErrorToResCode(resCode);
 
 		if (!resCode || valid_values.indexOf(resCode.value) === -1) {
@@ -489,6 +493,10 @@
 
 			// log bandwidth and http status code
 			systemLogger.logInfo(genResponseLogging(resCode, meta, req), undefined, logLabels.network);
+		}
+
+		if (killServer) {
+			return Promise.reject(killServer);
 		}
 	};
 


backend/src/v4/response_codes.js
```diff
@@ -416,6 +416,10 @@
 	 */
 	responseCodes.respond = function (place, req, res, next, resCode, extraInfo, format, cache, customHeaders) {
 
+		// Topology is closed mongo error is typically coming from the session management and the library
+		// doesn't let us recover from it (so far). So kill this pod and let it respawn.
+		const killServer = resCode?.name === "MongoError" && resCode?.message?.includes && resCode.message.includes("Topology is closed") ? resCode : false;
```
ChristosTsiotsias, line 421: can we not remove the middle part ? ( && resCode?.message?.includes &&)

backend/src/v4/response_codes.js
```diff
@@ -490,6 +494,10 @@
 			// log bandwidth and http status code
 			systemLogger.logInfo(genResponseLogging(resCode, meta, req), undefined, logLabels.network);
 		}
+
+		if (killServer) {
```
ChristosTsiotsias, line 498: I'd make killServer a boolean eg

const killServer = resCode?.name === "MongoError" && resCode?.message?.includes && resCode.message.includes("Topology is closed");

and then do return Promise.reject(resCode);

just for clarity but its just preference

ChristosTsiotsias (changes requested):

backend/src/v4/response_codes.js
carmenfan: @ChristosTsiotsias resCode gets reassigned right after I assigned `killServer` so the data would've been lost.

So to do what you want I'd have to assign 2 variables instead of one

ChristosTsiotsias (approved):

PR #5429, opened 2025-02-25
ISSUE #5423 - display fullname instead of username

### LINKED ISSUE #5423: Replace usernames with user's first name and last name in v4 UI
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/638

With using an external IDP, we are decommissioning the concept of usernames in favour of an internal user ID (in the form of a UUID). 

So new users will no longer have a context relevant username, which will not make any sense to the user. We want to replace any instances where we're using username as the user representation, with the first name and last name of the user (just like v5).

Pages to update:
- [x] Teamspace -> users tab
  - [x] Remove username in the listing
Image
  - [x] Remove username on the promp on add user
  Image
    - [x] When selected use display first name and second name instead of username (just leave it as email if it makes life easier)
  - [x] Issues & Risks
    - [x] owner, comment author, system log should all show full name instead of the username
  - [ ] v4 group
    - [ ] owner should be displayed in full name
- [x] top banner
  - [x] Replace username with email here: 
Image
  - [x] remove username from User profile

### Goals
- As a user, I do not want to use username to identify the user that is being referred to

### Tasks
TBC

Amantini1997:
This fixes #5423

#### Description
Display first name and last name (or "Unknown user" if the former are not available) instead of username in v4 components

#### Test cases
See issue description

#### Notes
- We are not addressing the mentions and [replies ](https://github.com/3drepo/3drepo.io/pull/5429#discussion_r2012317998) further than this (now its using the lastname and hover doesnt work well)  because it would need  a more complex solution that its out of the scope of this PR.


carmenfan (dismissed):
LGTM!

frontend/src/v4/helpers/comments.ts
The-Daniel, line 1: image
Username is still displayed in the user popover

The-Daniel, line 1: Not sure what should happen here but tagging a user in issues/risks comments it as a username
image
image

frontend/src/v4/routes/components/newUserForm/newUserForm.component.tsx
The-Daniel, line 1: v5 signin refers to username
image
As does the user sign up. 
image

The-Daniel, line 1: Perhaps this is outside the scope of this issue but the suggestion for adding a new user to a teamspace only occurs when you type in the username exactly
image
(i.e. "Santiag" will not trigger the suggestion)

The-Daniel (changes requested):

The-Daniel (approved):

frontend/src/v4/routes/components/newUserForm/newUserForm.component.tsx
carmenfan: yea, also intentional, full email or full username is required or we risk basically exposing the whole user list

frontend/src/v4/helpers/comments.ts
Amantini1997: I suspect it may take longer that a couple hours as it's v4 logic. But I will give it a go. If I see the light at the end of the tunnel soon enough, I will push the changes

carmenfan: just leave the email bit, all good.

carmenfan (changes requested):
We're asking user to type their username here when we remove someone from the teamspace:
image


So this becomes a long uuid on the new users 😆 .... can we just convert this to a good old simple confirmation? i.e. just ask `are you sure` and they say `yay` or `nay`

frontend/src/v4/helpers/comments.ts
```diff
@@ -236,32 +238,32 @@ const convertActionCommentToText = (comment: IComment) => {
...
-				text = comment.action.propertyText + ' updated by ' + comment.owner;
+				text = comment.action.propertyText + ' updated by ' + author;
 				break;
 		}
 	}
 
 	if (0 === text.length) {
 		if (!comment.action.from || !comment.action.to) {
-			text = comment.action.propertyText + ' updated by ' + comment.owner;
+			text = comment.action.propertyText + ' updated by ' + author;
```
sanmont3drepo, line 261: -  When you reply a comment the mention uses the username:

reply

sanmont3drepo (changes requested):

frontend/src/v4/helpers/comments.ts
sanmont3drepo: @The-Daniel  two things:
 -  [ ] Full name is preferred instead of just last name
 -  [ ] Hover on the name is not showing the user card like with the username
 
image

sanmont3drepo (approved):

PR #5456, opened 2025-03-11
Issue 5442 - IDP: Frontend changes on authentication

### LINKED ISSUE #5442: IDP: Frontend changes on authentication
### Description
Parent issue: https://github.com/3drepo/3drepo.io/issues/5356
Requires: https://github.com/3drepo/3drepo.io/issues/5412

With #5412, the way we authenticate users has changed, and the following flows are now removed:

- all AAD routes (link/unlink, sign up, authenticate etc)
- forgot password
- reset password
- signup
- verify
- login (username and password)

So any frontend logic around these flows can also be removed.

Instead, authentication now exclusively happen via an OAuth flow with frontegg. A page has been added in #5356. However we need to add an additional authentication flow when the user tries to access any data on a specific teamspace.

The session now requires to be authenticated against a particular teamspace in order for the data to be released. When the user first login, they will be authenticated against a teamspace already (typically the last teamspace they were authenticated against), this information is available via [`GET login` request](https://github.com/3drepo/3drepo.io/blob/ISSUE_5412/backend/src/v5/routes/users.js#L31)

To authenticate the user against another teamspace, the OAuth workflow will need to be triggered again with a different endpoint to fetch the authentication link: https://github.com/3drepo/3drepo.io/blob/1da40bc88cabadd421b65530dac312c538178f0e/backend/src/v5/routes/authentication.js#L61 Note that this route is only available if the user has a valid session with the server and is a member of the teamspace. Upon success, the `authenticatedTeamspace` field via `GET login` request will be updated.

Update profile also needs adjusting:
- Email is now considered a unique identifier thus should no longer be editable.
- SSO related fields (i.e. link and unlink) can all be removed.
- CountryCode is no longer a required field

> Don't worry about password update, that will be tackled in https://github.com/3drepo/3drepo.io/issues/5438

### Specification
- [Auth flow for different teamspaces](https://github.com/3drepo/3D-Repo-Product-Team/issues/649#issuecomment-2596167915)

### Goals
- [ ] As a Teamspace Admin, I want to ensure any additional security restriction (SSO, MFA etc) are respected before the user is granted access to the data.
- [ ] As a user, I want to access teamspace data via the frontend UI as before, any additional flow show be easy to follow.
- [ ] As a user I want to be able to update my profile information

### Tasks
TBC

Amantini1997:
This fixes #5442

#### Description
- Removed all the old SSO code and paths
- Remove all the login/signup/password recovery code and paths
- Email is now read-only and countryCode is not required
- Authentication via frontegg is fully supported
- Changing teamspace re-triggers authentication for that teamspace

**Notes:**
Changing teamspace is not super smooth, this is already known

#### Test cases
Login with users and navigate the app. The only real differences should be the way you login, and the update everytime the teamspace changes


The-Daniel (changes requested):

frontend/src/v5/services/api/sso.ts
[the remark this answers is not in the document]
Amantini1997: Potentially, but I talked to Carmen about this and she agreed it's better to leave them for now

frontend/test/auth/auth.sagas.spec.ts
```diff
@@ -15,15 +15,14 @@
...
 import { mockServer } from '../../internals/testing/mockServer';
 import { createTestStore } from '../test.helpers'
-import { selectIsAuthenticated, selectLoginError } from '@/v5/store/auth/auth.selectors';
+import { selectAuthenticatedTeamspace, selectIsAuthenticated } from '@/v5/store/auth/auth.selectors';
 import { DialogsTypes } from '@/v5/store/dialogs/dialogs.redux';
 
 describe('Auth: sagas', () => {
-	const username = 'Jason';
-	const password = 'Friday13';
+	const authenticatedTeamspace = 'WE_HATE_PINEAPPLE_ON_PIZZA_TEAMSPACE';
```
The-Daniel, line 25: \>:(

The-Daniel (approved):

carmenfan (changes requested):
It's a bit trippy at the moment if you click on a teamspace that needs re-authenticating, it's taking me to the teamspace page, then it does a refresh to be in the same page.

If i'm feeling this out correctly we're doing this:
click on teamspace -> go to teamspace dashboard -> check if we need reauth -> redirect for re-auth (with redirectURI set to the dashboard)

Could we  do this instead?
click on teamspace -> check if we need reauth -> redirect for re-auth (with redirectURI set to the dashboard)

so when we click on the teamspace, we check if we need reauth on the teamspace listing page, redirect to the auth url if we need reauth, or go to the dashboard otherwise.

frontend/assets/icons/thirdParty/microsoft.svg.tsx
The-Daniel, line 1: If you go back in browser history when teamspace is authenticating you get stuck on the teamspace page with the auth modal open
image

The-Daniel, line 1: After you change teamspace and it authenticates if you try to go back in the browser history the page will reload and you're still on the same page. I think when it authenticates it's adding a new entry to the history when it shouldn't

The-Daniel, line 1: If you try to access a teamspace that you do not have access to (or one that does not exist) then you get stuck with the authentication modal
image
Also I think you might need to remove the padding from this modal title so that it looks vertically centred

The-Daniel (changes requested):

carmenfan (approved):
Happy with the general flow. I'll leave the details with @The-Daniel  and @sanmont3drepo

frontend/src/v5/ui/components/shared/modalsDispatcher/templates/infoModal/authenticatingModal/authenticatingModal.component.tsx
```diff
@@ -14,18 +14,18 @@
...
-	const location = useLocation();
-	const initialLocation = useRef(location);
+	const currentLocation = useLocation();
+	const initialLocation = useRef(currentLocation);
 
 	useEffect(() => {
-		if (location !== initialLocation.current) onClickClose();
-	}, [location]);
+		if (currentLocation.pathname !== initialLocation.current.pathname) onClickClose();
+	}, [currentLocation]);
```
The-Daniel, line 28: Nice. This doesn't have to be done in this PR but I think this should be moved to the base modal component so it does this for _all_ modals.

I made an issue for this a while ago https://github.com/3drepo/3D-Repo-Product-Team/issues/441

The-Daniel (approved):
Seems good to me

sanmont3drepo (approved):

PR #5495, opened 2025-04-10
Issue 5480 - Get teamspace member through IDP endpoint

### LINKED ISSUE #5480: Get teamspace member through IDP endpoint
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/720

For details, see Product issue.

In summary:
- We want to rewrite get teamspace members logic to get [this information from frontegg](https://github.com/3drepo/3drepo.io/blob/ISSUE_5356/backend/src/v5/services/sso/frontegg/components/accounts.js#L69)
  -  This will return userIds from frontegg, we would want to turn all the userIds into usernames
  - There may be users that are not recognised by 3D Repo from this request, those users should then be processed and a user entry needs to be created (similar to when [someone we don't know logged in](https://github.com/3drepo/3drepo.io/blob/ISSUE_5356/backend/src/v5/middleware/sso/frontegg.js#L75))
- We will need to make sure v4 logic also do the same thing (easiest thing is probably to rewrite them to v5 functions)
- We want to clean up any old logic that is reading/writing teamspace access to `system.users`
- We will need to make sure utility scripts related to teamspace memberships are also rewired

### Goals
- [ ] As a 3D Repo Admin, I want to control teamspace access through frontegg panel.
- [ ] As a 3D Repo Admin, I want to make sure there is only one place storing teamspace membership so there is no risk of data going out of sync.

### Tasks
TBC by the developer

### Related Resources
Frontegg API: https://developers.frontegg.com/api/overview

Bogdan-Crn:
This fixes #5480

#### Description
Updated v5 to accommodate using frontegg and return the same output.
Update the v4 logic for getTeamspaceMembers.
Updated the utility scripts to reflect the changes in getting the teamspace members.
Updated some of the tests helpers

#### Test cases
Update v5 tests to reflect the changes
Update necessary v4 tests to reflect the changes


backend/src/scripts/utility/teamspaces/removeTeamspaceData.js
carmenfan, line 1: actually can we just remove this script? It's never intended to be used anymore. (superseded by removeTeamspace)

backend/src/v4/models/user.js
```diff
@@ -590,12 +593,11 @@ User.hasReachedLicenceLimitCheck = hasReachedLicenceLimit;
 User.getMembers = async function (teamspace) {
 	const promises = [];
 
-	const getTeamspaceMembers = User.findUsersInTeamspace(teamspace, {
-		user: 1,
-		customData: 1
-	});
-	const getJobInfo = usersWithJob(teamspace);
+	const { refId: tenantId } = await getTeamspaceSetting(teamspace, { refId: 1 });
```
carmenfan, line 596: This seems overly complicated to just get the list of users in teamspaces....

Can't we just use `getTeamspaceMembersInfo` from v5 processor?

In general we want to avoid v4 doing all these from scratch. it should just interface with a single functino from v5 that v5 itself uses (so there's a single point of change if the logic needs to be altered)

backend/src/v4/models/user.js
```diff
@@ -630,6 +632,12 @@ User.findUsersInTeamspace =  async function (teamspace, fields) {
 	return await db.find("admin", COLL_NAME, query, fields);
 };
 
+User.findUsersInTeamspaceById =  async function (userId) {
```
carmenfan, line 635: I guess we don't need this if we change the function above?

backend/src/v5/models/users.js
```diff
@@ -18,7 +18,9 @@
 const { USERS_COL, USERS_DB_NAME } = require('./users.constants');
 const { createResponseCode, templates } = require('../utils/responseCodes');
 const { generateHashString, sanitiseRegex } = require('../utils/helper/strings');
+const { createNewUserRecord } = require('../processors/users');
```
carmenfan, line 21: this is cyclic dependency, files in model folder should never include something from the processor folder.

The idea is that if you need a bigger context, this logic should be happening in the processor folder.

backend/src/v5/models/users.js
```diff
@@ -165,4 +167,30 @@ User.ensureIndicesExist = async () => {
 	}
 };
 
+User.getMemberInfoFromId = async (userId) => {
```
carmenfan, line 170: again if we need this function this should be done in the processor.

If this needs to stay here for whatever reason then the naming is wrong, `getMember` is a teamspace context, we're looking at user records in dependent of teamspaces here.

backend/src/v5/processors/teamspaces/index.js
```diff
@@ -43,9 +43,11 @@ const { removeUserFromAllProjects } = require('../../models/projectSettings');
 const Teamspaces = {};
 
 const removeAllUsersFromTS = async (teamspace) => {
-	const members = await getMembersInfo(teamspace);
+	const { refId: tenantId } = await getTeamspaceSetting(teamspace, { refId: 1 });
```
carmenfan, line 46: I think what you really want here is a common function that getAllMembersInTeamspace

Which will do the frontegg stuff and convert all the userIds to usernames.

Then things like this function, `getTeamspaceMembersInfo` will just call that common function.

There's a lot of duplication everywhere just to get the members list.

backend/src/v5/models/users.js
carmenfan: Having read the other file I think what you really want here is just a function that given an array of userIds it returns you the array of corresponding usernames?

backend/tests/v4/setup.js
```diff
@@ -16,5 +16,7 @@
  */
 const mockRequire = require('mock-require');
 const FronteggMock = require('../v5/helper/fronteggMock');
+const FrontEggAccountsMock = require('../v5/helper/fronteggMock/components/accounts')
 
 mockRequire('../../src/v5/services/sso/frontegg', FronteggMock);
+mockRequire('../../src/v5/services/sso/frontegg/components/accounts', FrontEggAccountsMock);
```
carmenfan, line 22: bit weird that you have to do this when L21 should've done the whole thing 🤔

backend/tests/v5/e2e/routes/teamspaces/index.test.js
```diff
@@ -131,9 +130,17 @@ const testGetTeamspaceMembers = () => {
 
 				return data;
 			});
-			expectedData.push({ job: DEFAULT_OWNER_JOB, user: teamspace });
 
-			ServiceHelper.outOfOrderArrayEqual(res.body.members, expectedData);
+			const resWithoutAdmin = res.body.members.filter(
```
carmenfan, line 134: not entirely sure why you had to do this, `outOfOrderArrayEqual` should've done the trick? why do you have to split the array and check it separately?

backend/tests/v5/helper/fronteggMock/components/accounts.js
```diff
@@ -39,7 +39,12 @@ Accounts.createAccount = (name) => {
 	return Promise.resolve(accountId);
 };
 
-Accounts.getAllUsersInAccount = (accountId) => Promise.resolve(usersInAccount[accountId] ?? []);
+Accounts.getAllUsersInAccount = (accountId) => {
+	if (usersInAccount[accountId]) {
+		return Promise.resolve(usersInAccount[accountId].map((id) => ({ id, email: id })));
```
carmenfan, line 44: this will be a ticking timebomb if you hack it this way lol

addUserToAccount takes in an email, so you should really be returning the email that was passed in originaly (maybe you need another dictionary of userIdToEmail, or change the usersInAccount object)

backend/src/v4/models/user.js
Bogdan-Crn: Initially I didn't use it because of what getTeamspaceMembersInfo was returning, will change it now

backend/tests/v4/setup.js
Bogdan-Crn: I know,  but when I run the tests without that line it seems to be using the real accounts

backend/src/v5/processors/teamspaces/index.js
```diff
@@ -100,12 +101,31 @@ Teamspaces.getTeamspaceListByUser = async (user) => {
...
 
-Teamspaces.getTeamspaceMembersInfo = async (teamspace) => {
+Teamspaces.getAllMembersInTeamspace = async (teamspace) => {
 	const { refId: tenantId } = await getTeamspaceSetting(teamspace, { refId: 1 });
 	const frontEggUsers = await getAllUsersInAccount(tenantId);
-	const membersList = await Promise.all(frontEggUsers.map((user) => getMemberInfoFromId(user.id)));
-	const jobsList = await getJobsToUsers(teamspace);
+	const membersList = await Promise.all(frontEggUsers.map(async (user) => {
+		try {
+			return getUserInfoFromId(user.id);
```
carmenfan, line 109: As discussed previously, I think it's worth passing all the userIds over and do a single query... in the event where there's over 1000 users in the teamspace, we're doing 1000 db query instead of 1.

carmenfan (changes requested):
Looking good, just a couple more things 🙈

backend/src/v5/processors/teamspaces/index.js
carmenfan: Actually ignore my previous comment, 2 things:

1. We need to do the query with `email`, it's very possible that the userId get out of sync (Frontegg could potentially generate a new ID due to users being orphaned, in which they remove the record, then get added back In again). So we actually want to find users by their email, and update the userId in the database record if it's mismatched. (again if possible let's do a `findMany` instead of doing multiple database queries)
1. For any one we couldn't find, we still want to create the user entry, but I don't think it's necessary to call `getUserById` below, I think you have everything you need from the response of `getAllUsersInAccount`, you just need to tweak the function in `services/sso/frontegg/components/accounts`. this is the data you get from the endpoint  
image

backend/src/v5/processors/teamspaces/index.js
```diff
@@ -100,12 +101,31 @@ Teamspaces.getTeamspaceListByUser = async (user) => {
 	return Promise.all(tsList.map(async (ts) => ({ name: ts, isAdmin: await isTeamspaceAdmin(ts, user) })));
 };
 
-Teamspaces.getTeamspaceMembersInfo = async (teamspace) => {
+Teamspaces.getAllMembersInTeamspace = async (teamspace) => {
```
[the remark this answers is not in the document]
carmenfan: Once you've fixed this bug, have a look at the list.  What I wanted to test is the behaviour of an invitee (i.e. that invited email who is not currently a 3drepo user). 

I think frontegg will return it as if it's an existing user of the teamspace (they are differentiated by the status) so we will end up creating the user record on the database and listing it as if it is a user (i.e. I think  we will see 3 users on the list, instead of 2 if you do the same thing in staging. We need to do a bit more logic around matching email address to invitees and ignoring them inside this function

It might be too confusing to understand, so might want to leave this till I come back!

backend/src/v5/processors/teamspaces/index.js
```diff
@@ -104,23 +104,49 @@ Teamspaces.getTeamspaceListByUser = async (user) => {
...
-
-			return getUserInfoFromId(userId);
-		}
-	}));
-
-	return membersList;
+	const membersData = await getUserInfoFromEmailArray(frontEggUsers.map((user) => user.email));
+
+	// check id discrepancies
+	membersData.forEach(async ({ user, email, userId }) => {
```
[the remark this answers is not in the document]
Bogdan-Crn: I was sure I did it as a map but clearly I missed it.

backend/src/v5/processors/teamspaces/index.js
```diff
@@ -104,23 +104,49 @@ Teamspaces.getTeamspaceListByUser = async (user) => {
...
+		const processedInvites = teamspaceInvites.map((invite) => invite._id);
+
+		const unprocessedUsers = frontEggUsers.filter(
+			({ email }) => !foundUsersEmails.includes(email) && !processedInvites.includes(email),
+		);
+
+		const newUsers = await Promise.all(unprocessedUsers.map(async ({ id, email }) => {
+			logger.logDebug(`User not found: ${id}, creating user based on info from IDP...`);
+
+			const userData = await getUserById(id);
```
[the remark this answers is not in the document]
carmenfan: @Bogdan-Crn look deeper 😆 

We're filtering it ourselves https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/services/sso/frontegg/components/accounts.js#L116

You'll see that the api from frontegg is returning you a lot more (https://developers.frontegg.com/api/identity/user-management/userscontrollerv3_getusers) 
image

backend/src/v5/processors/teamspaces/index.js
```diff
@@ -104,23 +104,49 @@ Teamspaces.getTeamspaceListByUser = async (user) => {
 Teamspaces.getAllMembersInTeamspace = async (teamspace) => {
 	const { refId: tenantId } = await getTeamspaceSetting(teamspace, { refId: 1 });
 	const frontEggUsers = await getAllUsersInAccount(tenantId);
```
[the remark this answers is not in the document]
Bogdan-Crn: renamed the constant from ```frontEggUsers``` to ```accountUsers```

backend/src/v5/processors/teamspaces/index.js
```diff
@@ -104,23 +104,49 @@ Teamspaces.getTeamspaceListByUser = async (user) => {
...
+		const { id } = frontEggUsers.filter((frontEggUserData) => frontEggUserData.email === email)[0];
+		// update the mongo userId to match
+		if (userId !== id) await updateUserId(user, id);
+	});
+
+	// check for unprocessed users
+	if (frontEggUsers.length !== membersData.length) {
+		const teamspaceInvites = await getTeamspaceInvites(teamspace);
+
+		const foundUsersEmails = membersData.map((userData) => userData.email);
```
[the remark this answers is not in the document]
Bogdan-Crn: Re did the logic so that we have minimal looping through the data.

backend/src/v5/processors/teamspaces/index.js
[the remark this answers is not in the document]
Bogdan-Crn: updated ```getAllUsersInAccount``` to return the information needed to create an account in mongo

carmenfan:
@Didi-Kwan one for you to test.

The UX has not changed, it's just the underlying logic has been updated and we wanted to make sure we didn't break anything!

Didi-Kwan (approved):
Sanity Test is complete. No bug detected. Here is the [test summary report](https://asitesol-my.sharepoint.com/:x:/g/personal/ylokkwan_asite_com/EQfK1-OHjbJGkGxQMjKbNPABwupoiZxwJotopdi8DkuJqQ).

PR #5501, opened 2025-04-14
Issue 5493 - several changes to Frontegg behaviour

### LINKED ISSUE #5493: Reset IDP session if the teamspace to reauthenticate against has special settings
### Description
When we re-authenticate a user against a particular teamspace, we want to determine if we should destroy the frontegg session to force the users to login again.

This may be necessary if the teamspace has special configurations like enterprise SSO.

### Goals
- [ ] As a IT admin of a teamspace with SSO configured, I want to ensure my users log in via the configured SSO so that I can impose any security restrictions I want prior to them accessing the data.

carmenfan:
This fixes #5493

#### Description
- authenticate endpoint now accepts an optional email address
  - this is used to determine the active tenant ID to authenticate against
  - If the user does not exist, impose a random tenant ID 
  - if the email address it not provided, do not specify the tenant ID
  - email address is added as a login hint (i.e. it will be prefilled in frontegg's login form
- Frontegg session is removed on authenticate against a teamspace route if:
  - the teamspace has SSO configuration, and the SSO configuration has claimed the domain of the user
- Set MFA policy by default
  - Create teamspace now sets MFA policy by default (enabled)

#### Test cases
- If authenticate url is called with email in the query string:
  - a tenant ID will be specified in the redirect url
  - the email provided will be specified in the redirect url
- If the authenticate url is called without an email
  - it will behave the same as before
- If a user tries to authenticate against a teamspace that has an SSO configuration
  - If the user's email is associated with the configuration, it will remove the frontegg session, prompting the user to login again (which should redirect to the SSO Identity provider)
- When createTeamspace script is called
  - the newly created teamspace will automatically be MFA enabled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,11 +21,11 @@ matrix:
         - docker
 
       before_install:
+        - sudo apt-get -y install gnupg curl expect
         - curl -s https://packagecloud.io/install/repositories/cloudamqp/rabbitmq/script.deb.sh?any=true | sudo bash
-        - sudo apt-get install rabbitmq-server=4.0.8-1
+        - sudo apt-get -y install rabbitmq-server=4.0.8-1
         - sudo systemctl start rabbitmq-server
         - sudo apt remove mongodb && sudo apt purge mongodb && sudo apt autoremove && sudo rm -rf /var/lib/mongodb
-        - sudo apt-get -y install gnupg curl expect
         - curl -fsSL https://www.mongodb.org/static/pgp/server-8.0.asc | sudo gpg -o /usr/share/keyrings/mongodb-server-8.0.gpg --dearmor
         - echo "deb [ arch=amd64,arm64 signed-by=/usr/share/keyrings/mongodb-server-8.0.gpg ] https://repo.mongodb.org/apt/ubuntu noble/mongodb-org/8.0 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-8.0.list
         - sudo apt-get update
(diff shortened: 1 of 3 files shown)


backend/src/v5/services/sso/frontegg/components/accounts.js
```diff
@@ -179,6 +202,25 @@ Accounts.addUserToAccount = async (accountId, email, name, emailData) => {
...
+			[HEADER_TENANT_ID]: accountId,
+		};
+
+		const { data } = await get(`${config.vendorDomain}/team/resources/sso/v1/configurations`, headers);
+		return data.flatMap(({ enabled, domains }) => {
+			if (!enabled) return [];
+			return domains.map(({ domain }) => domain);
+		});
+	} catch (err) {
+		logger.logError(`Failed to get claimed domains within an account: ${JSON.stringify(err?.response?.data)} `);
```
ChristosTsiotsias, line 219: just in case it fails to get the domains cant we just return [ ] ? cause in this case the app will crash

ChristosTsiotsias (changes requested):

backend/src/v5/services/sso/frontegg/components/accounts.js
carmenfan: how did you trigger an app crash? 

This should never happen.

None the less I'll add a catch case on removeFronteggSessionIfNeeded and return a 500.

ChristosTsiotsias (approved):

PR #5521, opened 2025-04-30
ISSUE #5519 extend the CSRF cookie upon validation

### LINKED ISSUE #5519: CRSF cookie needs to refresh along side the session cookie
### Description

With 5.16 we've reduced the cookie session to 30mins, and this is extended everytime a api request has been made and therefore the session will be available for as long as the user is active.

However the CSRF cookie is only established at the point of login, and this cookie expires when the 30mins is up.

This results in the user being logged out if they ever refresh the page after the initial 30mins period, as the application can no longer find the original CSRF token.

### Steps to reproduce
- log into 3D Repo
- Keep the session active over 30mins  (e.g. click on something to trigger an api request 10mins in, so the session expires 40mins from the intial login)
- after the initial 30mins, refresh the page

### Current behaviour
- the user gets logged out after the refresh

### Expected behaviour
- the user session should still be active.

carmenfan:
This fixes #5519
#### Description
The CSRF cookie will now be extended everytime a validation occurs (i.e. ever authenticated endpoint). This should make the expiration time matches the session cookie.

Also fixed a bug where the isLoggedIn middleware is actually bypassing the ip/referer check.

#### Acceptance Criteria
- if the user refreshes their page after the initial 30mins, they should still be logged in

### DIFF
--- a/backend/tests/v5/e2e/routes/teamspaces/projects/index.test.js
+++ b/backend/tests/v5/e2e/routes/teamspaces/projects/index.test.js
@@ -169,7 +169,10 @@ const testCreateProject = () => {
 
 		test('should fail if multiple projects are being sent at similar times with the same name', async () => {
 			const payload = { name: ServiceHelper.generateRandomString() };
-			const [res1, res2, res3] = await Promise.all(times(3, () => agent.post(route()).send(payload)));
+			const prom1 = agent.post(route()).send(payload);
+			const [res1, res2, res3] = await Promise.all([
+				prom1,
+				...times(2, () => agent.post(route()).send(payload))]);
 
 			expect(res1.statusCode).toBe(templates.ok.status);
 			expect(res2.statusCode).toBe(templates.invalidArguments.status);
--- a/backend/tests/v5/unit/middleware/permissions/index.test.js
+++ b/backend/tests/v5/unit/middleware/permissions/index.test.js
@@ -23,8 +23,8 @@ jest.mock('../../../../../src/v5/utils/permissions');
 const Permissions = require(`${src}/utils/permissions`);
 const { templates } = require(`${src}/utils/responseCodes`);
 
-jest.mock('../../../../../src/v5/utils/sessions');
-const Sessions = require(`${src}/utils/sessions`);
+jest.mock('../../../../../src/v5/middleware/auth');
+const Sessions = require(`${src}/middleware/auth`);
 const PermMiddlewares = require(`${src}/middleware/permissions`);
 const { determineTestGroup, generateRandomString } = require('../../../helper/services');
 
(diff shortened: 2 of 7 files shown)


Didi-Kwan (approved):
Testing is complete. It's good to go now :) Here is the [test summary report](https://asitesol-my.sharepoint.com/:x:/g/personal/ylokkwan_asite_com/ESrAyk12RrhNlJgOVUIiZkQBr7rPSDuWswoRW0RqkdddrA?e=HkM0gX).

backend/tests/v5/e2e/routes/teamspaces/projects/index.test.js
```diff
@@ -169,7 +169,10 @@ const testCreateProject = () => {
 
 		test('should fail if multiple projects are being sent at similar times with the same name', async () => {
 			const payload = { name: ServiceHelper.generateRandomString() };
-			const [res1, res2, res3] = await Promise.all(times(3, () => agent.post(route()).send(payload)));
+			const prom1 = agent.post(route()).send(payload);
```
Bogdan-Crn, line 172: why did we have to create a new promise instead of using the ```...times()```

backend/tests/v5/unit/middleware/permissions/index.test.js
```diff
@@ -23,8 +23,8 @@ jest.mock('../../../../../src/v5/utils/permissions');
 const Permissions = require(`${src}/utils/permissions`);
 const { templates } = require(`${src}/utils/responseCodes`);
 
-jest.mock('../../../../../src/v5/utils/sessions');
-const Sessions = require(`${src}/utils/sessions`);
+jest.mock('../../../../../src/v5/middleware/auth');
+const Sessions = require(`${src}/middleware/auth`);
```
Bogdan-Crn, line 27: If we are looking at Auth now for session then we might want to change it from Session to Auth

Bogdan-Crn (changes requested):

backend/tests/v5/e2e/routes/teamspaces/projects/index.test.js
carmenfan: because github is sometimes failing the request on prom1 because prom2 or prom3 got processed first.

By making prom1 be a separate command gives a bigger latency for it to suceed.

The alternative would be to change the expected results to not always look for prom1  - i can do that if you prefer? @Bogdan-Crn

backend/tests/v5/unit/middleware/permissions/index.test.js
carmenfan: @Bogdan-Crn It was changed because it was a badly written test to begin with. 

It was always calling the auth middleware (as you can see, I didn't make any changes to the `permissions/index.js` in the src folder. So it was relying on the auth middleware to be implemented a way and call `utils/sessions` instead of mocking what it's calling. so it wasn't really a real unit test.

Bogdan-Crn: No, I mean from being ```const Sessions``` to ```const Auth``` to reflect that we are looking at Auth for that not Sessions

Bogdan-Crn (approved):

PR #5555, opened 2025-06-02
Issue 5526 - Adjustments to createTeamspace utility script

### LINKED ISSUE #5526: Adjustments to createTeamspace utility script
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/791

At the moment, we have done the bare minimum to keep the script running. We'd like to make some additional changes to the script to make it easier for the systems team to manage support requests:

#### Support user in email form
At the moment, the script takes in a username to be the teamspace's first admin. As we are moving away from usiing usernames, it is often difficult to find out what is the username of a user. It would be a lot more convenient if we can put in the user's email and/or username.

Also, in the situation where this is a fresh deployment, we may not have a user. We need to be able to put in an email that is not associated with a user, and it will automatically create the user in the system.

#### Impose an existing account ID instead of creating a new account
We want the option to be able to input an existing Frontegg Account ID instead of always creating a new one. So that if the account is created by a different application (or deployment) we can just reuse the information that is already existing in frontegg.

This means we will be adding an additional (optional) parameter to the script, if it is not provided it will create a new account in frontegg, otherwise it will search for an existing account and pegging this teamspace against that account.

### Goals
- As a 3DR Admin, I want to be able to depict who is the first admin of the teamspace by an email address
  - If the email address is known, it should assign that user as the first admin
  - If the email address it not known, it should create the user and assign that user as the first admin
- As a 3DR Admin, I want to be able to associate an existing FE account to the teamspace by providing the account ID
- As a 3DR Admin, I want to be informed and do not create the teamspace if the provided FE account ID is not found.
- As a 3DR Admin, I want the script to create an account on FE if an account ID is not provided

### Tasks
TBD

Bogdan-Crn:
This fixes #5526

#### Description
**Support user in email form**
This change allows the usage of user or email to be the teamspace's admin and in situation where is a fresh deployment it will create an invite for the email if it does not exist.

**Impose an existing account ID instead of creating a new account**
This change allows the option to input an existing Frontegg Account ID and if the account exists use it to pin the newly created teamspace against it.

#### Acceptance Criteria
- [x] As a 3DR Admin, I want to be able to depict who is the first admin of the teamspace by an email address.
  - [x] If the email address is known, it should assign that user as the first admin.
  - [x] If the email address it not known, it should create the user and assign that user as the first admin.
- [x] As a 3DR Admin, I want to be able to associate an existing FE account to the teamspace by providing the account ID.
- [x] As a 3DR Admin, I want to be informed and do not create the teamspace if the provided FE account ID is not found.
- [x] As a 3DR Admin, I want the script to create an account on FE if an account ID is not provided.


carmenfan (changes requested):
@Bogdan-Crn  there's a conflict!

Bogdan-Crn:
> @Bogdan-Crn there's a conflict!

I'll have a look at it now

backend/tests/v5/scripts/teamspaces/createTeamspace.test.js
```diff
@@ -23,19 +23,45 @@ const {
...
 
 const user = generateUserCredentials();
+const emailUser = generateUserCredentials();
+emailUser.user = 'emailUser';
+emailUser.basicData.email = 'test@email.com';
 const teamspace = generateRandomString();
+const existingAccount = generateRandomString();
+let existingAccountId;
+
+jest.mock('../../../../src/v5/models/users', () => {
```
carmenfan, line 45: Script tests are integration tests, so we don't want to mock a function unless it's absolutely necessary (e.g. it's talking to an external service/ checking for an email etc)

We should create the users using the service helper as necessary.

backend/tests/v5/scripts/teamspaces/createTeamspace.test.js
```diff
@@ -23,19 +23,45 @@ const {
...
+		getUserByUsernameOrEmail: jest.fn((mockUser) => {
+			if (mockUser === 'problemUser') throw new Error('Made up error!');
+			if (mockUser === 'nonExistantUser') throw new Error('User not found.');
+			if (mockUser === 'test@email.com') return { user: 'emailUser' };
+
+			return { user: mockUser };
+		}),
+	};
+});
+UserBillingMock.getSubscriptionLimits.mockResolvedValue({ collaboratorLimit: 'unlimited' });
```
carmenfan, line 58: This will never work. Since a newly created teamspace would not have a quota assigned to it.
image

We need a bypass in the invite function to not check for quota

backend/src/scripts/utility/teamspaces/createTeamspace.js
```diff
@@ -35,10 +45,13 @@ const run = async (teamspace, user) => {
 		throw new Error('Teamspace already exists');
 	}
 
-	await initTeamspace(teamspace, user);
+	await initTeamspace(teamspace, username, accountId);
+	if (!username) {
+		await createInvite(user, teamspace, DEFAULT_OWNER_JOB, undefined, { TEAMSPACE_ADMIN: true });
```
carmenfan, line 50: Recheck the function args, it's not storing the permissions properly. (Red is generated through a standard invite web UI, blue is generated with this script - you can see it's missing the teamspace admin permissions)

image

carmenfan (changes requested):

carmenfan (changes requested):
It seems like the permissions isn't being mapped across from the invitation. Need a more detailed look:

- I created a teamspace with a new user
  - This generated the invite and emails fine
- When I signed up and logged in with the new user, I'm not seeing the teamspace in the dashboard
  - I saw the invite was consumed (i.e. the bson is removed the invitations collection), but the user themselves has not been added to the teamspace (i.e. the permissions is not mapped across to the teamspace settings)

It could be a regression bug from #5480, but need fixing either way 😆

carmenfan (approved):

PR #5578, opened 2025-06-30
Issue 5566 - Tickets table: improve side panel performance

### LINKED ISSUE #5566: Improve performance on tabular view - Side panel
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/806
Opening a ticket in tabular view is rather slow, and it gets worse as the number of tickets increase.

### Goals

- [ ] Opening a ticket in the side panel should be fast
- [ ] The time shall possibly not be affected by the number of tickets displayed in the table

### Tasks
- [ ] Improving the opening time of the side panel

Amantini1997:
This fixes #5566 

#### Description
The side panel was pretty slow to open due to the use in staging of react-router-dom "states". They were forcing refreshes in the whole table everytime the "ticketId" (path segment) and "containerOrFederation" (search params) were changing.

To improve performance, clicking on a new ticketId does not update the url path. Instead, it is added as a search param.
Moreover, I created a new searchParam hook which does not force re-renders, but allows to subscribe to changes.
This new approach still works well with the back/forward navigation in history and allows to share the current URL.

To deal with older versions of the code where a ticket id was stored in the path (as opposed to a search param), a useEffect in the ticketsTableContext deals with it.

#### Acceptance Criteria
- [ ] Side panel should be fast in opening (Take into account the time required by the template to render)
- [ ] It should still be possible to share the current view via URL
- [ ] it should be possible to navigate back and forth in history and to open/close tickets this way

### DIFF
--- a/frontend/src/v5/ui/routes/dashboard/projects/tickets/ticketsTable/ticketsAndResizableTableProvider.component.tsx
+++ b/frontend/src/v5/ui/routes/dashboard/projects/tickets/ticketsTable/ticketsAndResizableTableProvider.component.tsx
@@ -31,7 +31,7 @@ export const TicketsAndResizableTableProvider = ({ children }) => {
 
 	return (
 		<TicketsTableContextComponent>
-			<ResizableTableContextComponent columns={columns} columnGap={1} key={templateId + templatHasBeenFetched}>
+			<ResizableTableContextComponent columns={columns} columnGap={1}>
 				{children}
 			</ResizableTableContextComponent>
 		</TicketsTableContextComponent>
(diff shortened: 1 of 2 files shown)


frontend/src/v5/ui/routes/dashboard/projects/tickets/ticketsTable/ticketsTableContent/ticketsTableGroup/columnsVisibilitySettings/columnsVisibilitySettings.component.tsx
```diff
@@ -117,6 +117,11 @@ export const ColumnsVisibilitySettings = () => {
 		refreshSorting();
 	}, [orderingColumnPropertiesFetched]);
 
+	useEffect(() => {
+		if (!orderingColumnPropertiesFetched) return;
+		refreshSorting();
+	}, [orderingColumnPropertiesFetched]);
```
sanmont3drepo, line 123: this seems like duplicated code

sanmont3drepo (changes requested):

PR #5668, opened 2025-09-15
Issue 5624 - Custom Tickets - Ability to depict default columns on tabular view within the template

### LINKED ISSUE #5624: Custom Tickets - Ability to depict default columns on tabular view within the template
### Description

We want to allow the template creator to depict what are the default columns to be displayed inside tabular view.

This should be done by adding an additional (optional) configuration:

````js
{
  config: {
    tabular: {
      columns: [
        
      ],
    }
    //...other configs
  }
  //...
}
````

#### Notes:
- If this is not defined, it should display the default columns as depicted on or before v5.18.0
- If it is defined, it should display the ticket code, title, and all the columns depicted (in the order specified in the configuration)
- When this is edited, the middleware should check if the columns depicted exists within the template
- If a property that is specified in the columns is being deprecated, it should error and specify that a deprecated property cannot be a default column.

### Goals
- As a TS admin I want to depict the default columns when the template is being viewed instead the tabular view UI so the users can immediately gain useful context that is specific to the template
- As a TS admin I want to see the default tabular view columns if I did not specify a configuration
- As a TS admin I want to be notified if I'm deprecating a property that is currently defined as a default column
- As a TS admin I want to be notified if I am trying to add a default column where the propetyy doesn't exist.

### Tasks
Developer: Please fill in the tasks and your estimates

### PR DESCRIPTION
**Pull request by Bogdan-Crn**
This fixes #5624 

#### Description
We want to allow the template creator to depict what are the default columns to be displayed inside tabular view.

This change allows an additional configuration under config for the issue template that permits the user to define what columns should be showing in tabular view.

Updates have happened to the middleware to check for changes to the templates.

#### Acceptance Criteria

- [x] As a TS admin I want to depict the default columns when the template is being viewed instead the tabular view UI so the users can immediately gain useful context that is specific to the template
- [x] As a TS admin I want to see the default tabular view columns if I did not specify a configuration
- [x] As a TS admin I want to be notified if I'm deprecating a property that is currently defined as a default column
- [x] As a TS admin I want to be notified if I am trying to add a default column where the property doesn't exist.

### CODE REVIEW

File: backend/src/v5/schemas/tickets/templates.js [hunk 1]
```diff
@@ -187,6 +187,13 @@ const configSchema = Yup.object().shape({
 			.test('Custom status', 'values must be unique', (vals) => uniqueElements(vals.map(({ name }) => name)).length === vals.length),
 		default: Yup.mixed().when('values', (values) => (values ? Yup.string().oneOf(values.map(({ name }) => name)).required() : Yup.mixed())),
 	}).default(undefined),
+	tabular: Yup.object({
```
ChristosTsiotsias, line 190: I'd try to enforce the module-property validity test here so that it covers both create/edit a template, something similar to pinMappingTest

backend/tests/v5/unit/schemas/tickets/templates.test.js
```diff
@@ -885,6 +885,7 @@ const testValidate = () => {
 			config: {
```
ChristosTsiotsias, line 885: If the validity test is moved in config schema I'd create another set of tests ni Validate method to ensure the tabular objest is valid

ChristosTsiotsias (changes requested):

backend/tests/v5/unit/schemas/tickets/templates.test.js
```diff
@@ -860,15 +988,48 @@ const testValidate = () => {
...
+			if (i === 0) return { name: generateRandomString(), type: propTypes.TEXT, deprecated: true };
+			return { name: generateRandomString(), type: propTypes.TEXT };
+		}),
+	}];
+	const deprecatedProperties = times(3, (i) => {
+		if (i === 0) return { name: generateRandomString(), type: propTypes.TEXT, deprecated: true };
+		return { name: generateRandomString(), type: propTypes.TEXT };
+	});
+	const tabularColumnsTest = [
+		['tabular column with all required properties filled in', generateBasicSchema(basicModules, basicProperties), true],
```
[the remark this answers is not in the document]
Bogdan-Crn: as part of tabular view the front end could select a base property like the name of the template or code the validation adds those as properties as well

PR #5680, opened 2025-09-22
Issue 5660 - Integrate Apryse to view PDFs

### LINKED ISSUE #5660: Add Apryse as a viewer for PDFs in 2D
For PDFs, instead of converting to SVGs, we want to load them natively in Apryse's (PDFTron) WebViewer.

Tasks 

- [ ] Integrate the Apryse viewer as a React [component](https://docs.apryse.com/web/guides/get-started/react), and toggle it based on the type of document delivered via the /image API.
- [ ] Make the backend check if a drawing is a PDF, and if so deliver the actual asset rather than the converted SVG
- [ ] Disable the drawing conversion job for PDFs
- [ ] Make sure the avatar, navigation etc works the same for the Apryse WebViewer
- [ ] Make sure snapping works for the Apryse viewer
- [ ] Make sure calibration (getting a raster, etc) works for the Apryse WebViewer

sebjf:
This fixes #5660

This PR requires additional checks - labelled as invalid for the moment.

#### Description
This PR integrates the Apryse PDF viewer as an optional element for the underlying viewer in the viewer2D Component, updates the  selection logic to use the actual drawing type delivered from the backend, and then configures the backend to return the uploaded file instead of the converted SVG when asking for PDFs.

The Apryse viewer is integrated as a drawing viewer Component; to support this a number of additional minor changes are made:

1. The Vertical Spatial Boundaries Handler can no longer assume the same drawing src can be loaded into an Image Element, so the image for display in 3D is now retrieved via the `DrawingViewerService`, where the PDF viewer can render the PDF into a blob beforehand. As this is potentially expensive, the handler now only gets this once, on creation.

2. Types concerned with Snapping have been moved out of the Svg Handler and have been made asynchronous, to form a more complete & general purpose interface. The Apryse Component implements this, handling snapping internally.

3. The GetScreenshot method of the drawing viewer now explicitly ignores the camera icon. It is unlikely users would want to see this anyway, but the main reason for it is that `modern-screenshot3` doesn't include it when capturing the Apryse component, so this change ensures screenshots are consistent between viewers. The invocation of `modern-screenshot3` has also been updated to ensure it takes into account the scroll position of DOM elements, which is best practice.


backend/package.json
```diff
@@ -33,6 +33,7 @@
     "@dotenvx/dotenvx": "1.38.3",
     "@elastic/elasticsearch": "7.17.14",
     "@frontegg/client": "5.3.2",
+    "@napi-rs/canvas": "^0.1.80",
```
carmenfan, line 36: lock the versions here
```suggestion
    "@napi-rs/canvas": "0.1.80",
```

backend/package.json
```diff
@@ -69,6 +70,7 @@
     "nodemailer": "6.9.14",
     "openapi-schema-validator": "12.1.3",
     "paypal-rest-sdk": "1.8.1",
+    "pdfjs-dist": "^5.4.149",
```
carmenfan, line 73: this one too

backend/src/v4/routes/config.js
```diff
@@ -140,6 +141,8 @@ function createClientConfig(serverConfig, req) {
 
 	clientConfig.impliedPermission = C.IMPLIED_PERM;
 
+	clientConfig.apryse = config.apryse;
```
carmenfan, line 144: if i'm reading this right, the client config will have the following:

````
  {
      apryseLicense: "xxx" //from L60,
      apyrse: {
         licenseKey: "xxx" //this line
      }
  }
````

Do you need both?

File: backend/src/v5/utils/helper/images.js [hunk 4]
```diff
@@ -15,21 +15,54 @@
...
-	const jpgBuffer = await sharp(buffer, { density })
-		.flatten({ background: '#ffffff' })
-		.toFormat('jpeg')
-		.toBuffer();
+	let imgBuffer;
+	if (mimeType === 'application/pdf') {
+		const pdfJsLib = await loadPdfJsDist();
+		const loadingTask = pdfJsLib.getDocument({
+			data: buffer.buffer,
+			standardFrontDataUrl: '../../../node_modules/pdfjs-dist/standard_fonts',
```
carmenfan, line 45: just to make it a bit robust (if for some reason we move this file it's not going to break without anyone noticing), can we use a relative path to the v5 source directory? 

i.e. use the v5Path from the [interop file](https://github.com/3drepo/3drepo.io/blob/master/backend/src/interop.js)

```suggestion
			standardFrontDataUrl: `${v5Path}/../../node_modules/pdfjs-dist/standard_fonts`,
```

backend/src/v5/utils/helper/images.js
```diff
@@ -15,21 +15,54 @@
...
+
+ImageHelper.createThumbnail = async (buffer, mimeType, width = 600, density = 150) => {
 	if (!buffer) throw new Error('Image not provided');
 
-	const jpgBuffer = await sharp(buffer, { density })
-		.flatten({ background: '#ffffff' })
-		.toFormat('jpeg')
-		.toBuffer();
+	let imgBuffer;
+	if (mimeType === 'application/pdf') {
```
carmenfan, line 41: can we put the mimeTypes to a constant file? maybe in `/utils/helpes/typeCheck.constants` (thought we had one already but it looks like we don't! 🙈

backend/tests/v5/unit/processors/teamspaces/projects/models/drawings/index.test.js
```diff
@@ -605,10 +606,13 @@ const testCreateDrawingThumbnail = () => {
 	describe('Create drawing thumbnail', () => {
 		test('Should store the thumbnail if it has been successfully generated', async () => {
 			const imageRef = generateUUID();
-			const imageBuffer = generateRandomString();
+			const imageBuffer = generateRandomBuffer();
+			const readStream = Readable.from(imageBuffer);
+			const mimeType = 'application/pdf';
```
carmenfan, line 611: get constant from the aforementioned constant file here (applies for the other test cases too!)

carmenfan (changes requested):
Backend side looks good, just a few tiny changes


Need to wait for @sanmont3drepo to review the frontend stuff 😆

backend/src/v4/routes/config.js
sebjf: No that line is out of date 😮 Removed!

carmenfan (approved):
LGTM. I made a commit for some minor housekeeping tweaks (nothing important and worth making you come back to this branch for!)

sanmont3drepo (changes requested):

sebjf:
@sanmont3drepo, I have addressed those comments! I also realised when testing out the changes, that snapping can be quite slow on larger PDFs, so I also tweaked the Apryse snapping to return the same promise if a snapping request is in-flight, and that improves it somewhat.

sanmont3drepo (approved):

carmenfan:
going to run through Didi's test cases if all good we can merge!

sebjf:
Hi @carmenfan, I think you're correct as if I await on `revisionAdded`, the status shows as queued until the thumbnail is generated!

Though there is still the problem of the thumbnail generation blocking the main thread, so the backend can't do anything else.

We can't use another Js library, because they are all ESM nowdays. I'd say we have three options,

1. Create a worker pool for pdfjsdist. The downsides are that it is a lot of boilerplate code, and more significantly, pdfjs will still take 10's of seconds to create the thumbnails, it just won't block anything anymore (the binaries Felix found could so the same thing in a fraction of a second). The advantages are that it is still all handled within the same process and there are no additional dependencies.

2. We use one of the binaries Felix found and call it with `child_process` (or some npm generic wrapper library). The downsides are that it is another dependency we have to deploy, and the conversion is a separate process & none of the binaries Felix found support piping, so we'd have to deal with it writing temporary files, making sure they're cleaned up, handling zombie processes etc etc. The advantages are speed of processing (10x faster, at least), and probably a simpler implementation than a worker pool.

3. We can create our own C++ [addon](https://nodejs.org/api/addons.html) around one of those libraries. Note that some exist for libraries like Poppler, but they still require writing a file to disk. The disadvantages are that it is the most work by far (we'd need to spin up a whole project around it...), but the conversion would run within the same process but with the fast native library, getting the best of both worlds in a seamless integration.

carmenfan (approved):

PR #5689, opened 2025-09-29
Issue #5676 - Filters shortcut on column name

### LINKED ISSUE #5676: Tabular view improvement: Shortcut to filtering on column headers
### Description
We want to allow the users to quickly apply filtering on the column headers instead of having to go through the filter buttona and search the property that they are already staring at... Taking inspiration from excel, we should add a filter icon on all the filterable properties, and upon click, it should bring them to the UI to add/edit filters for that property

### Goals
- [x] As a user I want to be able to quickly filter with fewer clicks by clicking on the name column and using the filtering system from there

### Tasks

- [ ] Task 1
- [ ] Task 2

### Related Resources

Image

sanmont3drepo:
This fixes #5676

#### Description
- Added a button on the filtarable columns to be able to change the filter for that column from there
- Created an icon container style for the buttons on the header of the tickets table
- Simplified the code from tickets headers
- Added sorting by id (this was to avoid the special case of id header to have a different structure)
- Modified the stiling in the headers in order to align everything.
- Refactor of naming of columns to be unified (now EVERYWHERE uses the path to the property in the ticket like 'title', or 'properties.status' , so using get(column, ticket) would get you the value of that column) 

#### Acceptance Criteria
- [x] As a user I want to be able to quickly filter with fewer clicks by clicking on the name column and using the filtering system from there


The-Daniel (dismissed):

The-Daniel (approved):

carmenfan (approved):
all good, no bugs found!

PR #5697, opened 2025-10-15
ISSUE_5670 - Drawing data removed after project deletion

### LINKED ISSUE #5670: Drawing histories are not cleared up properly after a project is removed
When we remove a project, the drawing container entry is removed, but drawing revisions are not.

#### Steps to reproduce
- Create a project with a drawing in it (with one or more revisions)
- take note of the drawing ID
- remove the project
- probe the database (robo mongo, mongoshell etc) and search for the model in drawing history (i.e. run something like `db.getCollection("drawings.history").find({model: "model ID"})`

### Current behaviour
- entries of the drawing exists despite it being removed

### Expected behaviour
all data related to the drawing should be removed

#### NOTE
- I only checked .history, we will have to make sure things like calibrations are also removed.
- We will need a script to find existing orphaned drawing data and remove them to clean up

ChristosTsiotsias:
This fixes #5670

#### Description
Drawing data like drawing history, files and calibrations are now removed after the project is deleted.
MIgration script added to remove all drawing data that do not match an existing project

#### Acceptance Criteria
Create a project
Add a drawing and a drawing revision in it
Delete the project
Inspect the database. The drawing data should be removed

### DIFF
--- a/backend/src/scripts/migrations/5.22/removeOrphanedDrawingRecords.js
+++ b/backend/src/scripts/migrations/5.22/removeOrphanedDrawingRecords.js
@@ -18,36 +18,23 @@
 const { v5Path } = require('../../../interop');
 const { getTeamspaceList } = require('../../utils');
 
-const { deleteDrawingCalibrations } = require(`${v5Path}/models/calibrations`);
-const { modelTypes } = require(`${v5Path}/models/modelSettings.constants`);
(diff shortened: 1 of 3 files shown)


Bogdan-Crn (approved):
Tested it and it looks good

Bogdan-Crn (dismissed):

PR #5702, opened 2025-10-17
Issue 5678 - Shareable url

### LINKED ISSUE #5678: Tabular view improvement: Shareable link for the state of the tabular view
### Description
We want the ability to share a link to the tabular view, showing the state of what exactly the users were seeing (i.e. the formation of the columns, the filters, containers/feds/template selected and the groupings applied.

As discussed with @sanmont3drepo , due to the complexity of the smart filters, we may be in danger of hitting the [max URL length](https://www.baeldung.com/cs/max-url-length#bd-maximum-length-of-a-url-in-different-browsers) imposed by the browsers. To counter that, the following should be considered:

- We don't need a visually understandable URL string to reflect the state of the viewer, consider constructing what you need then compress it into jibberish that we can then decode internally.
- Even with the point above, we can potentailly hit the max length - so we should tell the user in that situation and suggest them to simplify the view

### Goals

- [ ] As a user I want to share a tabular view link and paste into a different browser and get the same results

### Tasks

- [ ] Task 1
- [ ] Task 2

### Related Resources

MockUp : `link to mockup`

sanmont3drepo:
This fixes #5678 

#### Description
- Now the filters, groupby , columns (order and visibility) are in the url so when shared it produces the same view
- Refreshing the page is till shows the same view.

#### Acceptance Criteria
- [x] As a user I want to change the groupby and when I pass the url to someone they can see the same groupby, template , models, tickets
- [x] As a user I want to change the filters and when I pass the url to someone they can see the same filtered tickets
- [x] As a user I want to change the columns and when I pass the url to someone they can see the same columns in the table
- [x] As a user if I change the params tabular view when I navigate back I forth I want to see the same views that I produced before

### DIFF
--- a/backend/VERSION.json
+++ b/backend/VERSION.json
@@ -1,4 +1,4 @@
-{ "VERSION" : "5.18.4",
+{ "VERSION" : "5.19.1",
   "unity" : {
 	"current" : "5.20.0",
 	"supported": []
--- a/backend/package.json
+++ b/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "3drepo.io",
-  "version": "5.18.4",
+  "version": "5.19.1",
   "engines": {
     "node": "22.x.x"
   },
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "3drepo.io-frontend",
-	"version": "5.18.4",
+	"version": "5.19.1",
 	"description": "The frontend for 3drepo.io",
 	"engines": {
 		"node": "22.x.x"


frontend/src/v5/ui/components/viewer/cards/cardFilters/cardFilters.types.ts
```diff
@@ -16,9 +16,26 @@
...
 export type TicketFilterOperator = 'ex' | 'nex' | 'is' | 'nis' | 'eq' | 'neq' | 'ss' | 'nss' | 'rng' | 'nrng' | 'gt' | 'gte' | 'lt' | 'lte';
+
+export enum TicketFilterOperatorEnum {
+	'ex' = 0,
+	'nex' = 1,
+	'is' = 2,
+	'nis' = 3,
+	'eq' = 4,
+	'neq' = 5,
+	'ss' = 6,
```
The-Daniel: I believe you can leave off the `= <number>,` and it will work the same way

frontend/src/v5/ui/routes/dashboard/projects/tickets/ticketsTable/ticketsTableContent/ticketsTableContent.component.tsx
```diff
@@ -30,26 +30,57 @@ import { useEdgeScrolling } from '../edgeScrolling';
...
+		ignoreColumnChange.current = true;
+		
+		if (!colsParam.length) {
+			const allColumns = getAllColumnsNames();
+			const initialColumns = INITIAL_COLUMNS.filter((name) => allColumns.includes(name));
+
+			if (!isEqual(getVisibleSortedColumnsNames(), initialColumns)) {
+				setVisibleSortedColumnsNames(initialColumns);
+				resetWidths();
+				stretchTable(BaseProperties.TITLE);
```
The-Daniel: A minor point perhaps for a future issue, but if the title column is excluded when sharing the link the table will not stretch to fit the screen. Perhaps it should check to see if the title is included in the table and if not fall back on stretching the first or last column

frontend/src/v5/store/users/users.selectors.ts
```diff
@@ -38,7 +38,7 @@ export const selectUser = createSelector(
 	},
 );
 
-const selectCurrentTeamspaceUsers = createSelector(
+export const selectCurrentTeamspaceUsers = createSelector(
```
The-Daniel: This export isn't being used

The-Daniel (approved):

PR #5706, opened 2025-10-27
Issue 5616 - Custom Tickets - Throws an error if the user tries to impose a default property type on a complex type

### LINKED ISSUE #5616: Custom Tickets - Throws an error if the user tries to impose a default property type on a complex type
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/437

As it stands the user could impose a default value for complex types like images and viewpoints.

We don't currently support this thus it may be better to throw an error when a user is trying to create/edit a template to have such feature to manage expectations

NOTE: we should make the logic clean and easy to add more property types for unsupported defaults in the future (i.e. avoid adhoc hard coding!)

### Steps to replicate
create a template with a default viewpoint/image/imagelist

### Current Behaviour
this will pass, but the tickets created on the UI will not contain the default viewpoint.

### Expected Behaviour
this should fail and tell the user that we don't support a default in those properties

Bogdan-Crn:
This fixes #5616 

#### Description
Created a test that does not allow users to create or update a template with a complex type (View, Image, ImageList) that has a default value.

#### Acceptance Criteria
- [x] fails and tell the user that we don't support a default in those properties.

### DIFF
--- a/backend/src/scripts/utility/modelProcessing/resetProcessingFlags.js
+++ b/backend/src/scripts/utility/modelProcessing/resetProcessingFlags.js
@@ -30,15 +30,33 @@ const { v5Path } = require('../../../interop');
 const { logger } = require(`${v5Path}/utils/logger`);
 const { getTeamspaceList } = require('../../utils');
 
+const { deleteIfUndefined } = require(`${v5Path}/utils/helper/objects`);
+const { DRAWINGS_HISTORY_COL } = require(`${v5Path}/models/revisions.constants`);
 const { updateMany } = require(`${v5Path}/handler/db`);
-const { SETTINGS_COL } = require(`${v5Path}/models/modelSettings.constants`);
+const { SETTINGS_COL, processStatuses } = require(`${v5Path}/models/modelSettings.constants`);
 const Path = require('path');
 
 const processTeamspace = async (teamspace, model) => {
-	const query = model ? { _id: model } : {};
+	const drawingStatusQuery = {
+		status: {
+			$exists: true,
+			$nin: [
+				processStatuses.OK,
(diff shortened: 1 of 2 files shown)


backend/src/v5/schemas/tickets/templates.js
```diff
@@ -143,7 +149,7 @@ const propSchema = Yup.object().shape({
 
 		if (type === propTypes.ANY_OF) return res.oneOf(values);
 
-		if (type === propTypes.IMAGE || type === propTypes.IMAGE_LIST) return Yup.mixed().strip();
+		if (complexTypes.has(type)) return Yup.mixed();
```
ChristosTsiotsias: I would put somehting like this here

```
if (complexTypes.includes(type)) {
			return Yup.mixed().test('invalid-default', `Default property cannot be applied to ${type}`, (value) => !value);
		}
```

havent tested it fully but I think this would do the trick and then you won't need the noDefaultsOnComplexTypesTest test case

backend/src/v5/schemas/tickets/templates.js
```diff
@@ -279,6 +285,30 @@ const validTabularPropsTest = (val, context) => {
...
+	const template = TemplateSchema.generateFullSchema(val);
+
+	for (const prop of template.properties) {
+		if (propertyCheck(complexTypes, prop)) {
+			return context.createError({ message: `Property type "${prop.type}" for "${prop.name}" property does not support a default value` });
+		}
+	}
+
+	for (const mod of template.modules) {
+		// console.log(mod);
```
ChristosTsiotsias: remove this and the one in line 302

PR #5736, opened 2025-11-17
ISSUE #5735 teamspace member check clean up

### LINKED ISSUE #5735: Use `getAllUsersInAccount` for `hasAccessToTeamspace`
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/907
Use `getAllUsersInAccount` instead of `getUserStatusInAccount` if we don't have to probe the status. This will help to reduce the amount of traffic to frontegg as `getAllUsersInAccount` is often used and cached.

### Goals
It should function as before, with less calls to frontegg. (See product issue for use case details)

### Tasks
- use `getAllUsersInAccount` instead of `bypassStatusCheck` is set to true

carmenfan:
This fixes #5735

#### Description
- If we're not getting the status, use `getAllUsersInAccount` call instead to get the benefit of accessing a cache that is more likely to be around.
- Remove `hasAccessToTeamspace` from `models/teamspaceSettings` and `utils/permissions`
  - They all redirected to `isTeamspaceMember` in `processors/teamspaces/index`

#### Acceptance Criteria
- Should work as before.
- Should generate less frontegg calls when we're checking if multiple users are teamspace members

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,15 @@ matrix:
 
       before_install:
         - sudo apt-get -y install gnupg curl expect
-        - curl -s https://packagecloud.io/install/repositories/cloudamqp/rabbitmq/script.deb.sh?any=true | sudo bash
-        - sudo apt-get -y --allow-downgrades install rabbitmq-server=4.0.8-1
-        - sudo systemctl start rabbitmq-server
+        # Free port 5672
+        - sudo lsof -i :5672 || true
+        - sudo fuser -k 5672/tcp || true
+        - sudo service rabbitmq-server stop || true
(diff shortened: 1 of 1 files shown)


backend/tests/v5/unit/middleware/permissions/components/teamspaces.test.js
```diff
@@ -74,7 +77,7 @@ const testIsTeamspaceMember = () => {
 
 		test('should respond with error if hasAccessToTeamspace did not resolve', async () => {
```
ChristosTsiotsias: isTeamspaceMember*

ChristosTsiotsias (approved):

PR #5756, opened 2025-12-03
Issue 5745 - Revision tree endpoint

### LINKED ISSUE #5745: Expose a fulltree.json endpoint to internal service through v5.
### Description

Currently we're serving[ fulltree.json through v4](https://3drepo.github.io/3drepo.io/#api-Model-getModelTree). We would like to create an v5 version of the endpoint, **but only served to the internal service.**

This endpoint is also simplier, as there is no need to serve the federation version of this endpoint.

#### Endpoint
route: `GET /teamspaces/{teamspaceName}/projects/{projectId3DRepo}/containers/{containerId}/tree?revId={revision Id}`
payload: Same as v4 container response.

NOTES:
- We want to make sure it is using the v5 middleware
- If revision ID is not provided, fetch the the latest revision
- We want to rewrite the logic into v5 
  - since we're just serving containers, it shoudl be much simplier:
    - We don't need additional logic to patch a federation of the json tree
    - it should just be fetching the file from file service (I think!)
    

### Goals
- As a Asite integrator I want to be able to fetch the full tree from the internal service

### Tasks
- [x] Create the route method (1 hr)
- [ ] Create documentation (1 hr)
- [x] Create processsor method (3 hrs)
- [x] Create model method (4 hrs)
- [ ] Write unit tests (4 hrs)
- [ ] Write e2e tests (4 hrs)
- [ ] User testing (4 hrs)

ChristosTsiotsias:
This fixes #5745

#### Description
New endpoint that allows to fetch a revisions full tree. It should return the fulltree.json file for the provided revision or the latest revision if a revision id is not provided. If there are no revisions to the container an error will be returned. The endpoint is only available on internal service.

#### Test cases
Call the endpoint with a specific revision id.
It should return the full tree for that revision.
Call the endpoiint without a revision.
It should return the full tree for the latest revision.
If there are no revisions an error should be returned.

### DIFF
--- a/backend/src/v5/routes/teamspaces/projects/models/common/assets/index.js
+++ b/backend/src/v5/routes/teamspaces/projects/models/common/assets/index.js
@@ -1,5 +1,5 @@
 /**
- *  Copyright (C) 2022 3D Repo Ltd
+ *  Copyright (C) 2025 3D Repo Ltd
  *
  *  This program is free software: you can redistribute it and/or modify
  *  it under the terms of the GNU Affero General Public License as
--- a/backend/tests/v5/e2e/routes/teamspaces/projects/models/common/assets/index.test.js
+++ b/backend/tests/v5/e2e/routes/teamspaces/projects/models/common/assets/index.test.js
@@ -1,5 +1,5 @@
 /**
- *  Copyright (C) 2021 3D Repo Ltd
+ *  Copyright (C) 2025 3D Repo Ltd
  *
  *  This program is free software: you can redistribute it and/or modify
  *  it under the terms of the GNU Affero General Public License as


backend/src/v5/processors/teamspaces/projects/models/containers.js
```diff
@@ -135,4 +136,19 @@ Containers.getSettings = (teamspace, container) => getContainerById(teamspace,
 
 Containers.getRevisionMD5Hash = getModelMD5Hash;
 
+Containers.getTree = async (teamspace, container, revision) => {
```
carmenfan: Can we move this to `..../models/common/assets/jsonAssets.js`?

Just being wary of what's coming ahead we'll be better off putting this to a separate file.

backend/src/v5/routes/routesManager.js
```diff
@@ -40,7 +40,12 @@ const { modelTypes } = require('../models/modelSettings.constants');
 
 RoutesManager.init = (app) => {
 	const internal = app.get(BYPASS_AUTH);
+
 	app.use('/v5/teamspaces/:teamspace/projects', ProjectRoutes(internal));
+	app.use('/v5/teamspaces/:teamspace/projects/:project/containers', CreateModelGeneralRoutes(modelTypes.CONTAINER, internal));
+	app.use('/v5/teamspaces/:teamspace/projects/:project/federations', CreateModelGeneralRoutes(modelTypes.FEDERATION, internal));
```
carmenfan: Why did you have to move federation/drawing?

backend/src/v5/routes/teamspaces/projects/models/common/general.js
```diff
@@ -248,837 +261,920 @@ const establishRoutes = (modelType) => {
...
-	 *                   format: uuid
-	 *                   description: Model ID
-	 *                   example: ef0855b6-4cc7-4be1-b2d6-c032dce7806a
-	 */
-	router.post('/', isAdminToProject, validateAddModelData(modelType), addModel(modelType));
+	if (isInternal) {
+		if (modelType === modelTypes.CONTAINER) {
+			/**
+			* @openapi
+			* /teamspaces/{teamspace}/projects/{project}/container/{container}/tree:
```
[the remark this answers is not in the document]
carmenfan: yup typo here

backend/tests/v5/unit/processors/teamspaces/projects/models/containers.test.js
```diff
@@ -553,6 +554,59 @@ const testGetMD5Hash = () => {
 	});
 };
 
+const testGetTree = () => {
```
carmenfan: move the file as per the src code

backend/src/v5/routes/routesManager.js
ChristosTsiotsias: I did this because I wanted to achieve 100% coverge in tests because of this if statement

image

I could alternatively istanbul ignore this line

carmenfan: @ChristosTsiotsias  yea let's istanbul ignore this 

In theory we don't need this if statement since it'll never be false the way we use it. but it's good to be there for clarity.

carmenfan (approved):

Issue #5764, opened 2025-12-04
Create v5 version of endpoints used in unity (that requires federation endpoints)


carmenfan:
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/900

The following endpoints to be copied over to v5. They should also be allowed through the internal configuration of api services:

| Desc| v4 | v5 |
|--------|--------|--------|
| Get Model Properties | `GET /<teamspace>/<modelID>/revision/<revision>/modelProperties.json`  | `GET .../{modelType}/<modelId>/assets/properties` |
| Get repoAssets| `GET /<teamspace>/<modelID>/revision/<revision>/repoAssets.json`| `GET .../{modelType}/<modelId>/assets/repobundles` | 
| Get Assets legacy | `GET /<teamspace>/<modelID>/revision/<revision>/assetsMeta` | `GET .../{modelType}/<modelId>/assets/bundles/unity/meta` |
| Get superMeshes.json | `GET /<teamspace>/<containerID>/revision/<revisionID>/supermeshes.json.mpc` | `GET .../{modelType}/<modelId>/assets/bundles/meta` |

### Goals
- Move over to v5 in order ot use v5 permissions middleware, with better maintainability and inclusion of robust tests

### Tasks
- [x] Create a common middleware that gets accessible containers from federation/user session.
- [x] Middle ware for processing revision ID 
  - [x] use it every where
- [x] Create v5 endpoint for:
  - [x] model properties
  - [x] repo assets
  - [x] assetsMeta
    - [x] Setup route
    - [x] Get nodes from stash graph
    - [x] middleware to convert IDs to string
  - [x] supemeshes.json
- [x] Refactor the writeStreamRespond to take size/filename in the options objects 
- [x] Check swagger is working
- [ ] Add e2e tests for no revision in sub model

PR #5771, opened 2025-12-15
Issue 5752 - Get object metadata endpoint 

### LINKED ISSUE #5752: Create a v5 version of `meta/all.json` endpoint available for the internal service
### Description

Currently we're serving [meta/all.json through v4](https://3drepo.github.io/3drepo.io/#api-Meta-getAllMetadata). We would like to create an v5 version of the endpoint, but only served to the internal service.

This endpoint is also simplier, as there is no need to serve the federation version of this endpoint.

#### Endpoint
route: `GET /teamspaces/{teamspaceName}/projects/{projectId3DRepo}/containers/{containerId}/metadata?revId={revision Id}`
payload: Same as v4 container response.

NOTES:
- We want to make sure it is using the v5 middleware
- If revision ID is not provided, fetch the the latest revision
- We want to rewrite the logic into v5 
  - since we're just serving containers, it should be simplier.

### Goals
- As a Asite integrator I want to be able to fetch the metadata from the internal service

### Tasks
Developer: Please fill in the tasks and your estimated time

ChristosTsiotsias:
This fixes #5752

#### Description
Implementation of meta/all.json endpoint in V5. The endpoint is internal-service only
The user should now be able to get all the objects of a revision together with their metadata.
The data returned should be casted (UUIDS to string and metadata in key-value form)
The user can provide a revision Id or if not the latest revision will be returned.

#### Test cases
- [ ] Call meta/all.json endpoint with a revision id
- [ ] You should get a list of objects of this revision with their metadata casted
- [ ] If you call it with no revision Id you should get the data of the last revision
- [ ] If there are no revisions on the model an error should be thrown
- [ ] If the service is not internal you should not be able to call the endpoint

### DIFF
--- a/backend/src/v5/middleware/permissions/components/models.js
+++ b/backend/src/v5/middleware/permissions/components/models.js
@@ -41,7 +41,6 @@ const permissionsCheckTemplate = (type, callback) => async (req, res, next) => {
 	const { session, params } = req;
 	const user = getUserFromSession(session);
 	const { teamspace, project, model } = params;
-
 	try {
 		const modelInProject = await checkModelExists(teamspace, project, model, type);
 		if (!modelInProject) {
--- a/backend/src/v5/routes/teamspaces/index.js
+++ b/backend/src/v5/routes/teamspaces/index.js
@@ -147,6 +147,64 @@ const createTeamspace = async (req, res) => {
 const establishRoutes = (isInternal) => {
 	const router = Router({ mergeParams: true });
 
+	/**
+	* @openapi
+	* /teamspaces/{teamspace}/quota:
+	*   get:
+	*     description: Gets quota information about a teamspace
+	*     tags: [v:external, v:internal, Teamspaces]
+	*     parameters:
+	*       - name: teamspace
+	*         description: name of teamspace
+	*         in: path
+	*         required: true
+	*         schema:
+	*           type: string
+	*     operationId: getQuotaInfo
+	*     responses:
+	*       401:
+	*         $ref: "#/components/responses/notLoggedIn"
+	*       200:
+	*         description: Gets the quota information of the user
+	*         content:
+	*           application/json:
+	*             schema:
+	*               type: object
+	*               properties:
+	*                 freeTier:
+	*                   type: boolean
+	*                   description: Whether or not the user has a paid subscription
+	*                   example: true
+	*                 expiryDate:
+	*                   type: number
+	*                   description: The closest expiry date of a users active plan (in epoch)
+	*                   example: 1233445
+	*                 data:
+	*                   type: object
+	*                   properties:
+	*                     used:
+	*                       type: number
+	*                       description: The number of bytes the user is currently using
+	*                       example: 1000000
+	*                     available:
+	*                       type: number
+	*                       description: The number of bytes the user can use
+	*                       example: 1000000
+	*                 seats:
+	*                   type: object
+	*                   properties:
+	*                     used:
+	*                       type: number
+	*                       description: The number of collaborators the user is currently using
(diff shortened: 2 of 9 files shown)


backend/src/v5/processors/teamspaces/projects/models/commons/assets/json.js
```diff
@@ -47,4 +50,39 @@ JsonAssets.getTree = async (teamspace, container, revision) => {
...
+			metaObj[key] = value;
+		});
+
+		castedMetadata.metadata = metaObj;
+	}
+
+	return castedMetadata;
+};
+
+JsonAssets.getMetadata = async (teamspace, container, revision) => {
```
carmenfan: I know this is done this way in v4 but this is becasue we're supporting a federation and if we don't stream and the federation have a lot of containers, we risk causing OOM on the pod.

Since we only need to cover containers in this version, and we're not doing any sort of pagination. It's a bit pointless to stream it as we're only doing 1 database query and writing to the stream once.

We may as well do it like the other standard JSON endpoints and just return normally, then the casting can happen in the middleware via yup cast so we can make it consistent

backend/src/v5/routes/teamspaces/projects/models/common/assets/index.js
```diff
@@ -128,6 +141,101 @@ const establishRoutes = (modelType, isInternal) => {
...
+            *                        "IFC GUID": "1oZ0wPs_PE8ANCPg3bIs4j",
+            *                        "AboveGround": "False"
+            *                     },
+            *                     "parents": [
+            *                        "323a9900-ece1-4857-8980-ec96ffc7f681"
+            *                     ]
+            *                  }
+            *               ]
+            */
+			router.get('/meta/all.json', hasReadAccessToContainer, verifyRevQueryParam(modelTypes.CONTAINER), getMetadata);
```
carmenfan: can we change the route as per the issue? 
image

backend/src/v5/routes/teamspaces/projects/models/common/assets/index.js
carmenfan: Changes are done on the wrong file. Metadata is not a model asset (we consider this as the 3D viewer stuff), what we're fetching here are the BIM attributes - i.e. the user context stuff.

It should go here: https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/routes/teamspaces/projects/models/containers/metadata.js

backend/tests/v5/helper/services.js
```diff
@@ -338,11 +338,10 @@ db.addLoginRecords = async (records) => {
 	await DbHandler.insertMany(INTERNAL_DB, 'loginRecords', records);
 };
 
-db.createScene = (teamspace, project, modelId, rev, nodes, meshMap) => Promise.all([
+db.createScene = (teamspace, project, modelId, rev, nodes, meshMap = {}) => Promise.all([
 	db.createRevision(teamspace, project, modelId, rev),
 	DbHandler.insertMany(teamspace, `${modelId}.scene`, nodes),
 	FilesManager.storeFile(teamspace, `${modelId}.stash.json_mpc`, `${UUIDToString(rev._id)}/idToMeshes.json`, JSON.stringify(meshMap)),
```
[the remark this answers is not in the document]
carmenfan: ☝️ 

Also worth noting that I removed `createRevision` call in this function in #5764 So if you're using this function it might not work as expected once the 2 branches merge into each other

The reason being is that I needed a revision regardless of whether am creating a scene, and the test case is creating a scene is re-creating a entry which is throwing dup index.

Considering this function only deal with adding scene data, revision seems to be out of scope and is better to create it outside.

PR #5823, opened 2026-02-10
ISSUE #5799 - Migration script to remove invalid supermeshes

### LINKED ISSUE #5799: DB migration to rename supermesh files
### Description
We've noticed that there is a discrepancy of how we're naming the supermeshes.json file.

3drepobouncer is creating and writing this file into the filesManager with `/revisions/` prepended to the name, but 3drepo.io is fetching it without, thus every new revision that is being viewed the first time, the v4 version of the script was caching a new one under a different name.

As of #5764, we have deprecated the v4 endpoint, the v5 one does not create this file (since we'd expect all models to contain one at this point) - the naming decrepancy is causing models viewing to fail.

For existing models, we are also storing 2 versions of the same file under different name, which is inefficient.

Therefore, we want to write a db migration script to do the following:
1. if there is 2 versions of supermeshes.json, remove the one that is prepended with `/revision/`
2. if there we only have the version prepended with `/revision/`, rename it

### Goals
- As a 3DR maintainer I want to make sure that i'm not using extra storage space than necessary
- As a 3DR user I want to be able to view new models uploaded since the change in #5764

### Tasks
Developer: Please consider the tasks required and provide estimates

ChristosTsiotsias:
This fixes #5799 

#### Description
A Db migration script has been added to remove invalid supemeshes JSON files. Once run this script will store all JSON supermeshes files under name format "{revisionID}/supermeshes.json". It will remove invalid entries if a correct and invalid entry is present, and will store the files under the correct name if a valid entry is missing. 

#### Test Cases
Create a container and add some revisions to it
Manually change the entry in the database from "{revisionID}/supermeshes.json" to "revision/{revisionID}/supermeshes.json"
Run the script
It should remove the invalid entry and store it in the valid name format

### DIFF
--- a/backend/src/scripts/migrations/5.22/removeInvalidSupermeshEntries.js
+++ b/backend/src/scripts/migrations/5.22/removeInvalidSupermeshEntries.js
@@ -43,8 +43,10 @@ const processCollection = async (teamspace, collection) => {
 	}
 
 	await Promise.all([
-		insertManyRefs(teamspace, collection, missingEntries),
-		removeRefsByQuery(teamspace, collection, { _id: { $in: idsToRemove } }),
+		missingEntries.length > 0 ? insertManyRefs(teamspace, collection, missingEntries)
+			: Promise.resolve(),
+		idsToRemove.length > 0 ? removeRefsByQuery(teamspace, collection, { _id: { $in: idsToRemove } })
+			: Promise.resolve(),
 	]);
 };
 


backend/src/scripts/migrations/5.22/removeInvalidSupermeshEntries.js
```diff
@@ -0,0 +1,80 @@
...
+const { getFile, storeFile, removeFilesWithMeta } = require(`${v5Path}/services/filesManager`);
+const { find } = require(`${v5Path}/handler/db`);
+const { logger } = require(`${v5Path}/utils/logger`);
+
+const processCollection = async (teamspace, collection) => {
+	const modelId = collection.split('.')[0];
+	const revisions = await getRevisions(teamspace, undefined, modelId, modelTypes.CONTAINER, false, { _id: 1 });
+
+	/* eslint-disable no-await-in-loop */
+	for (const { _id: revId } of revisions) {
```
carmenfan: This is a bit too overkill..

You don't need to go fetch the revisions and work out all the names.

Just get all the entries from the collection with the ID: `revision/*/supermeshes.json` and `*/supermeshes.json`
if there's a matching `*/supermeshes.json`, delete, if not, rename.

backend/src/scripts/migrations/5.22/removeInvalidSupermeshEntries.js
```diff
@@ -0,0 +1,71 @@
...
+	const goodEntries = await find(teamspace, collection,
+		{ _id: { $regex: '^[0-9a-fA-F-]+\\/supermeshes\\.json$' } }, { _id: 1 });
+
+	const goodEntryIds = createConstantsObject(goodEntries.map(({ _id }) => _id));
+
+	/* eslint-disable no-await-in-loop */
+	for (const { _id } of badEntries) {
+		try {
+			if (!goodEntryIds[_id]) {
+				const file = await getFile(teamspace, collection, _id);
```
carmenfan: this is overkill, just update the name on the .ref entry. we don't need to move the file.

carmenfan (changes requested):

backend/src/scripts/migrations/5.22/removeInvalidSupermeshEntries.js
```diff
@@ -0,0 +1,85 @@
...
+		const refInfo = await getRefEntry(teamspace, collection, id);
+		await insertRef(teamspace, collection, { ...refInfo, _id: correctedId });
+	} catch (err) {
+		logger.logError(`Failed to insert ref entry for ${teamspace}.${collection} with _id: ${correctedId}`);
+	}
+};
+
+const removeInvalidRef = async (teamspace, collection, id) => {
+	try {
+		await removeFilesWithMeta(teamspace, collection, { _id: id });
```
carmenfan: you dont' want to remove the files (since we're reusing the same link)

You just want to remove the refs

There is a function to do this in bulk

backend/src/scripts/migrations/5.22/removeInvalidSupermeshEntries.js
```diff
@@ -0,0 +1,85 @@
...
+const { insertRef, getRefEntry } = require(`${v5Path}/models/fileRefs`);
+const { createConstantsObject } = require(`${v5Path}/utils/helper/objects`);
+const { removeFilesWithMeta } = require(`${v5Path}/services/filesManager`);
+const { find } = require(`${v5Path}/handler/db`);
+const { logger } = require(`${v5Path}/utils/logger`);
+
+const insertMissingRef = async (teamspace, collection, id, correctedId) => {
+	try {
+		const refInfo = await getRefEntry(teamspace, collection, id);
+		await insertRef(teamspace, collection, { ...refInfo, _id: correctedId });
```
carmenfan: use insertManyRefs for bulk insert

carmenfan (changes requested):

PR #5825, opened 2026-02-11
ISSUE_5808 - Users may see multiple teamspaces despite there is only one available

### LINKED ISSUE #5808: Users may see multiple teamspaces despite there is only one available.
If the user is associated with multiple account with the same name inside frontegg, they will see multiple entries of teamspace on the teamspace dashboard despites there is only one entry inside the deployment

Image

### Steps to reproduce
- Create a teamspace,  assigning a user as admin
- Delete teamspace without removing the frontegg entry (i.e. `--removeAssociatedAccount false`)
- Create the teamspace again (i.e. same name, same user as admin)
- Login as that user

### Current behaviour
The user will see 2 entries on the teamspace listing (also the teamspace list API)

### Expected Behaviour
The user should only see the entry that has a matching account ID in the teamspace setting.

Bogdan-Crn:
This fixes #5808

#### Description
These changes check with the the db to make sure that the account id is the same as the ref Id of the teamspace so it avoids duplicate listings.

#### Acceptance Criteria
The user should only see the entry that has a matching account ID in the teamspace setting.

### DIFF
--- a/backend/tests/v5/unit/processors/teamspaces/index.test.js
+++ b/backend/tests/v5/unit/processors/teamspaces/index.test.js
@@ -92,9 +92,9 @@ const testGetTeamspaceListByUser = () => {
 				{ name: generateRandomString(), isAdmin: true },
 				{ name: generateRandomString(), isAdmin: false },
 			];
-			const frontEggData = times(goldenData.length + 1, (index) => (
-				{ name: goldenData[index]?.name || generateRandomString(), refId: generateRandomString() }
-			));
+			const frontEggData = goldenData
+				.map(({ name }) => ({ name, refId: generateRandomString() }))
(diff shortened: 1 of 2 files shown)


backend/tests/v5/unit/processors/teamspaces/index.test.js
```diff
@@ -92,19 +92,26 @@ const testGetTeamspaceListByUser = () => {
 				{ name: generateRandomString(), isAdmin: true },
 				{ name: generateRandomString(), isAdmin: false },
 			];
+			const frontEggData = times(goldenData.length + 1, (index) => (
```
carmenfan: It reads easier if we use es6 syntax

```suggestion
			const frontEggData = [...goldenData, {/* your new entry here */}]
```

PR #5859, opened 2026-03-09
Issue 5785 - Clash test run trigger

### LINKED ISSUE #5785: Endpoint to trigger a clash detection test run
### Description

An endpoint to instigate a manually triggered run of a clash detection test.

This requires a test plan to be present as per https://github.com/3drepo/3drepo.io/issues/5784

### Endpoint
route: `POST /teamspaces/{teamspace}/projects/{project}/clashes/{clashId}/runs`
internal/external: external
payload is not needed.
permissions: project admin

NOTES:
- This only works if the containers specified in the test plan has revisions
- It should then send a task to the clash queue for 3drepobouncer to process
  - clash queue does not exist at the moment
- The process should work similarly to model uploads, a config file will need to be added into the shared folder with info as specified [here](https://github.com/3drepo/3D-Repo-Product-Team/issues/639#issuecomment-3710914159)
  - We will need to run the query to construct the object array base on the latest revisions.
  - Composite ids should be the transformation id (or mesh ID if there is no trans) that is the parent object of the metadata query. the mesh array within the composite object shuold be all meshes where the transformation ID is an ancestor.
  - the file can become large, so we should stream write this as it we go
- We should keep a record on the database about the running instance (like drawing upload, where we add a record in the database with the correlation/revision Id used for the processing (see below for schema) 
  - This record should be stored in a `clashes.runs` collection
- This issue only deals with instigating a test, processing the result will be a separate issue.

### Process
````mermaid
flowchart TD
   User(["User Request"]) --> |Trigger a Clash test|ValidateRequest["Validate Payload"]
   ValidateRequest --> |Translate smart queries to object structure|GenerateConfig["Create Config"]
   GenerateConfig --> RecordInstance["Add test record to DB"] --> |Test run ID is correlation ID| PushToQueue["Add task to clash queue"]
   PushToQueue --> UpdateStatus["Update record status to 'Queued'"]
   PushToQueue --> |Clash config, correlation ID|RabbitMQ[[Rabbit MQ]]
````

NOTE: the message sent to the queue should be the following:
` processClash  ${teamspace} ${project} ${SHARED_SPACE_TAG}/${corId}/clashConfig.json`

### Clash run schema
in `clashes.runs`
````js
{
  _id: "UUID", // this should be the same as the correlation ID passed to the queue
  triggeredBy: "string", //This should be the userId, or "auto:<Trigger event>"
  triggeredAt: "date", // datetime at which this document was added
  completedAt: "date", // ignore for this issue - this should be filled in when the task finishes
  status: "string", // this should be "planned" initially, "queued" if it's been queued and "failed" if for some reason it failed (e.g. didn't insert into the queue)
plan: { /* copy of the test plan document, with revisions of the containers used recorded */}

}
````

### Goals
- As a project admin, I want the ability to trigger a test plan to be executed

### Tasks
- [x] create the route endpoint (1 hr)
- [x] create swagger documentation (1 hrs)
- [x] create middleware to ensure plan has containers of the same federation (3 hrs)
- [ ] create processor method (constuct config obj) (5 hrs)
- [x] create clash processing service (5 hrs)
- [ ] create clash event listener (3hrs)
- [ ] create model method to add run obj in the database (3 hrs)
- [ ] create method to add (2 hrs)
- [ ] user testing (1 day)
- [ ] create unit and e2e tests (1 day)

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #5785 

#### Description

Endpoint for starting a new clash test run.
Clash Service has been added responsible for sending the message to the bouncer.
Route, middleware, processor and model methods added.
Swagger documentation added.
Unit and E2E tests added.

#### Test Cases
Call the endpoint to start a new clash run
The planId should correspond to an existing plan
The plan should point to to containers with active revisions
The run should start and you should get a run Id as a response

### DIFF
--- a/backend/src/v5/processors/teamspaces/projects/models/commons/scenes.js
+++ b/backend/src/v5/processors/teamspaces/projects/models/commons/scenes.js
@@ -83,7 +83,6 @@ Scene.getMeshesWithParentIds = async (teamspace, project, container, revision, p
 	});
 
 	return groupByParent ? results : Array.from(results);
-
 };
 
 Scene.getExternalIdsFromMetadata = (metadata, wantedType) => {
@@ -168,10 +167,10 @@ const fetchMeshBinariesStreams = async (teamspace, container, refObj) => {
 Scene.getTexture = async (teamspace, project, container, textureId) => {
 	const textureNode = await getNodeByQuery(teamspace, project, container,
 		{ _id: textureId, type: nodeTypes.TEXTURE }, {
-		_id: 1,
-		_blobRef: 1,
-		extension: 1,
-	});
+			_id: 1,
+			_blobRef: 1,
+			extension: 1,
+		});
 
 	// eslint-disable-next-line no-underscore-dangle
 	if (!textureNode || !textureNode._blobRef) {

### CODE REVIEW

File: backend/src/v5/processors/teamspaces/projects/clashes.js [hunk 1]
```diff
@@ -25,4 +33,53 @@ Clashes.updatePlan = updatePlan;
...
+			}
+
+			stream.write(JSON.stringify(matchedMeshGroup));
+			first = false;
+		}
+	}
+	stream.write(']}');
+};
+
+Clashes.createRun = async (teamspace, project, plan, userId) => {
```
carmenfan: Can we call this user/username

userId is a term we use for the frontegg ID (whilst for most people they're the same thing, it's not for all) - it can get confusing.

```suggestion
Clashes.createRun = async (teamspace, project, plan, user) => {
```

backend/src/v5/utils/logger.js
```diff
@@ -42,6 +42,7 @@ Logger.labels = {
 	aad: 'AAD',
 	journaling: 'JRN',
 	notifications: 'NOTIF',
+	clashes: 'CLSH',
```
carmenfan: don't think we need this, just let it use MODPRO

backend/tests/v5/unit/services/modelProcessing.test.js
```diff
@@ -29,11 +29,18 @@ const FilesManager = require(`${src}/services/filesManager`);
 
 jest.mock('../../../../src/v5/models/revisions');
 const Revisions = require(`${src}/models/revisions`);
+
+jest.mock('fs', () => {
+	const actualFs = jest.requireActual('fs');
```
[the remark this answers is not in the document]
ChristosTsiotsias: I think its these tests that were the reason I did it 

image
image

PR #5913, opened 2026-03-27
Issue 5791 - Parsing a completed clash run

### LINKED ISSUE #5791: Add logic to process a finished clash test
### Description

With https://github.com/3drepo/3drepo.io/issues/5785, we introduced logic to create an instance of a test run based off a test plan and send the request to 3drepobouncer.

We now need to catch the signal from `callbackq` and process the results.

### The process

````mermaid
flowchart TD
   RabbitMQ[[Rabbit MQ]] --> |message from callbackq| ModPro[Model Processing service]
   ModPro --> |triggers clash completion event| ClashPro[Clash processor: processes results]
   ClashPro -->  |Store Clash result in blob storage| FilesManager
   ClashPro --> |Add file reference to result, completedAt and update status| Database[[Database]]

````
NOTES:
- we expect a callback queue message with the following information:
 ````js
    { 
         //note: this is a default rabbitmq property, bespoke data is stored inside content
        properties: { 
          correlationId  // the id we originally sent the request with, also the clash run ID in db
        },
        content: {
          teamspace: "string",
          project: "string",
          value: "number",
          results: "path to file", // this should be the shared directory folder - similar to where we store models. available if value is 0
        }
    }
````
- The results file can be massive, if we need to read it, it should be read in a streaming manner
  - the schema of the result file can be found here: https://github.com/3drepo/3D-Repo-Product-Team/issues/639#issuecomment-3710914159
  - exceptions are recorded inside the results file, therefore we do need to parse it to figure out the status.
- we want to copy/move the clash results over to the official file storage.
  - if an exception occured - don't store the file, just update the status/message accordingly.

### Goals
- [ ] As a user I want to know if a clash run has completed
- [ ] As a user I want to know that a clash result has been stored in the system once it has been executed.

### Tasks
Developer: please list the tasks and provide estimates

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #5791

#### Description
Functionality for parsing the clash result coming from the bouncer once the clash detection has been completed.
The api is listening to the callback queue and once the message is received the results are read using the path received from the bouncer. The clashes IDs are replaced by external Ids and the categorized into new, active and resolved based on the previous clash run. They are then saved in a file stored in the fileshare. The clash run in the database is then marked as complete and the results path is saved in results property.

Unit, E2E tests have been added.

Additially the property names in the model queue have been changed accordingly

database renamed teamspace
project is no longer model but the actual project Id
container or drawing is received

#### Test cases
Start a clash run from postman and wait for bouncer to conduct the clash detection.
Wait until the clash detection is complete and results are parsed.
The run entry in the database should be marked as complete and there should be a results file populated with the clash results and correctly categorized.

#### Stress Test
image

### CODE REVIEW

File: backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/clashes.js [hunk 1]
```diff
@@ -32,7 +32,8 @@ const validatePlanData = async (req, res, next) => {
 		req.body = await validatePlan(teamspace, project, req.body);
 
 		if (req.planData) {
-			if (isEqual(req.planData, req.body)) {
+			const { _id, ...existingPlanData } = req.planData;
+			if (isEqual(existingPlanData, req.body)) {
```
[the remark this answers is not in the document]
carmenfan: ☝️ 

It also sounds like we're missing some unit test cases here that should be detecting the above.

backend/src/v5/models/clashes.runs.js
```diff
@@ -35,4 +36,21 @@ ClashRuns.createTestRun = async (teamspace, plan, user) => {
 	return _id;
 };
 
+ClashRuns.completeTestRun = async (teamspace, runId, resultId) => {
```
carmenfan: It doesn't appear we process failures or queued status?

backend/src/v5/models/clashes.runs.js
```diff
@@ -35,4 +36,21 @@ ClashRuns.createTestRun = async (teamspace, plan, user) => {
...
+	const run = await db.findOne(teamspace, CLASH_RUNS_COL, query, projection, sort);
+
+	if (!run) {
+		throw templates.testRunNotFound;
+	}
+
+	return run;
+};
+
+ClashRuns.getLastRunFromPlan = (teamspace, planId) => db.findOne(teamspace, CLASH_RUNS_COL, { 'plan._id': planId }, { sort: { createdAt: -1 } });
```
carmenfan: call `runByQuery` so you don't end up duplicate work

carmenfan: I assume you'd also want projection here so you don't pull everything (e.g. copy of the test plan when you're just trying to compare results from previous run)?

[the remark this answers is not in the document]
carmenfan: this is true - what if it's not completed or failed?

carmenfan: Can we make sure we have an index on this collection? 

by plan then by whatever you're usiing for the above query.

backend/src/v5/processors/teamspaces/projects/models/commons/scenes.js
```diff
@@ -101,14 +94,16 @@ Scene.getExternalIdsFromMetadata = (metadata, wantedType) => {
 			if (!idType || idCounted.has(idType)) return;
 
 			idCounted.add(idType);
-			res[idType].push(value);
+			res[idType].push({ key, value });
```
carmenfan: why do we track the key here now? it doesn't look like we ever use it (or am I missing something?)

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -33,56 +38,170 @@ Clashes.updatePlan = updatePlan;
...
-			matched.flatMap(({ parents }) => parents), { groupByParent: true }) : Promise.resolve([]),
-		unwanted.length ? getMeshesWithParentIds(teamspace, project, container, revision,
-			unwanted.flatMap(({ parents }) => parents)) : Promise.resolve([]),
-	]);
+const getMeshDataFromRules = async (teamspace, project, container, revision, rules) => {
+	const { matched, unwanted } = await getMetadataByRules(teamspace, project, container,
+		revision, rules, { _id: 1, parents: 1, metadata: 1 });
+	const matchedTransNodes = matched.flatMap(({ parents }) => parents);
+	const meshes = matched.length
+		? await getNodesByQuery(teamspace, project, container, { type: 'mesh', rev_id: revision, parents: { $in: matchedTransNodes } }, { _id: 1, name: 1, parents: 1 })
```
carmenfan: I don't think this is quite right? If you're fetching immediate child meshes, what if the matched trans node is the root node of the tree?

You need `getMeshesWithParentIds` here, just like `unwantedMeshIds`? The process should be the same.

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -33,56 +38,170 @@ Clashes.updatePlan = updatePlan;
...
+			return acc;
+		}, {});
+
+		const unwantedIdsObj = createConstantsObject(unwantedMeshIds.map((id) => id));
+
+		for (const mesh of meshes) {
+			const idStr = UUIDToString(mesh._id);
+
+			if (!unwantedIdsObj[idStr]) {
+				const parentId = UUIDToString(mesh.name ? mesh._id : mesh.parents[0]);
```
carmenfan: you're mixing ID types here

`mesh._id` is the _unique ID_ of a node, `parent` is the _shared ID_ of its parent node.

the one that is stored in metadata.parents, is shared ID, so you would need the `mesh.shared_id` to have an accurate comparison.

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -33,56 +38,170 @@ Clashes.updatePlan = updatePlan;
...
+				const meshMeta = metadataMapping[parentId];
```
carmenfan: This is not a great plan for 2 reasons:

1. When you don't have rules, you have a scaling issue - getting all the metadata from the scene is a big move - there could be enough to overwhelm the pod if the model is big (we've had many cases of this happening)
2. When you have rules and you have a subset of the metadata, you're missing relationships that are 2 generations apart.

You should focus on getting the mesh list, then from that information, find out the external ID associated with them - work it out backwards (similar approach to groups, you just want to preserve the information about which mesh belongs to which external ID to construct the composite objects), don't try to prefetch/reuse the metadata results you got from the rules query.

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -33,56 +38,170 @@ Clashes.updatePlan = updatePlan;
...
+			result.resolved.push(existingItem);
+		}
+	}
+
+	return result;
+};
+
+Clashes.completeRun = async (teamspace, project, corId, resPath) => {
+	const resStream = createReadStream(resPath, { encoding: 'utf8' });
+	const content = await streamToJSON(resStream);
```
carmenfan: Yea you're definitely not streaming the results here 😆 if you do this, you may as well just do JSON.parse() from a file read....

try reading it in blocks.

like each clash item at a time and process them.

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -33,56 +38,170 @@ Clashes.updatePlan = updatePlan;
...
+
+	let lastRunClashes;
+
+	try {
+		const { plan } = await getTestRunByQuery(teamspace, { _id: corId }, { plan: 1, triggeredAt: 1 });
```
carmenfan: You should be able to do this if you only need the id from the plan object
```suggestion
		const { plan } = await getTestRunByQuery(teamspace, { _id: corId }, { "plan._id": 1, triggeredAt: 1 });
```

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -33,56 +38,170 @@ Clashes.updatePlan = updatePlan;
...
+		const lastCompletedRun = await getTestRunByQuery(teamspace,
```
carmenfan: you have `getLastRunFromPlan` in `clashes.run.js` but you don't use it here?

If you don't need it then let's just remove it 

Image

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -33,56 +38,170 @@ Clashes.updatePlan = updatePlan;
...
+	}
+
+	return result;
+};
+
+Clashes.completeRun = async (teamspace, project, corId, resPath) => {
+	const resStream = createReadStream(resPath, { encoding: 'utf8' });
+	const content = await streamToJSON(resStream);
+
+	let lastRunClashes;
```
carmenfan: Can we put the get info from last run into a helper function so this function is shorter and less "jumpy"

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -33,56 +38,170 @@ Clashes.updatePlan = updatePlan;
...
+
+	try {
+		const { plan } = await getTestRunByQuery(teamspace, { _id: corId }, { plan: 1, triggeredAt: 1 });
+		const lastCompletedRun = await getTestRunByQuery(teamspace,
+			{ 'plan._id': plan._id, completedAt: { $exists: true } },
+			{ result: 1 }, { completedAt: -1 },
+		);
+
+		const { readStream } = await getFileAsStream(teamspace, RUN_HISTORY_COL, lastCompletedRun.result);
+		const { new: newClashes, active, resolved } = await streamToJSON(readStream);
```
carmenfan: again we need to read this better.

No point streaming it just to call `JSON.parse` on the whole object

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -33,56 +38,170 @@ Clashes.updatePlan = updatePlan;
...
+		lastRunClashes = [...newClashes, ...active, ...resolved];
+	} catch (err) {
+		lastRunClashes = [];
```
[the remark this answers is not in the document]
carmenfan: agreed, should distinguish no previous run vs other sort of problems, and we should at least log the error if it's the latter before we carry on.

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -33,56 +38,170 @@ Clashes.updatePlan = updatePlan;
...
+	stream.on('end', () => {
+		resolve(JSON.parse(data));
+	});
+
+	stream.on('error', reject);
+});
+
+const compareRunResults = (existingRunClashes, newRunClashes) => {
+	const existingMap = new Map(existingRunClashes.map((item) => [item.index, item]));
+	const newMap = new Map(newRunClashes.map((item) => [[item.a, item.b].sort().join('-'), item]));
```
carmenfan: Do you need this in a map? All you're doing is preprocess the key and looping through the map

I feel like you could just loop it through as an array and generate the key on the fly?

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -33,56 +38,170 @@ Clashes.updatePlan = updatePlan;
...
+		};
+
+		if (existingMap.has(key)) {
+			result.active.push(clashToAdd);
+		} else {
+			result.new.push(clashToAdd);
+		}
+	}
+
+	for (const [key, existingItem] of existingMap) {
```
carmenfan: We need to stress test this whole thing by generating clash results with a lot of clashes and make sure we're not taking the server down due to OOM.

I'm a bit worried about this function - we're holding 2 copies of clash results in memory at any given time.

ask @sebjf  and/or @Ramadyfm  about the likely upperbound for the number of clash items within a run, let's run a test base on those numbers and make sure we're capable of processing it

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -33,56 +38,170 @@ Clashes.updatePlan = updatePlan;
...
+const compareRunResults = (existingRunClashes, newRunClashes) => {
+	const existingMap = new Map(existingRunClashes.map((item) => [item.index, item]));
+	const newMap = new Map(newRunClashes.map((item) => [[item.a, item.b].sort().join('-'), item]));
+
+	const result = { new: [], active: [], resolved: [] };
+
+	for (const [key, newItem] of newMap) {
+		const clashToAdd = {
+			...newItem,
+			a: { container: newItem.a.split('::')[0], idType: newItem.a.split('::')[1], id: newItem.a.split('::')[2] },
```
carmenfan: You're doing `split` 3 times instead of just storing the results in an interim variable.


```suggestion
            const [ container, idType, id] = newItem.a.split("::");
			a: { container, idType, id},
```

Maybe this should be a helper function so you don't have to duplicate it for processing b

backend/src/v5/processors/teamspaces/projects/clashes.js
carmenfan: We can relieve some pressure off the memory by removing the entries that are already matched.

I believe we can be confident that clashes are unique, i.e. if we've seen the key before, we shouldn't see it again.

backend/src/v5/services/eventsListener/components/modelEvents.js
```diff
@@ -20,13 +20,14 @@ const { addGroupUpdateLog, addImportedLogs, addTicketLog } = require('../../../m
...
 const { getContainerFileName, getLogArchive } = require('../../modelProcessing');
-const { getModelType, isFederation: isFederationCheck, newRevisionProcessed, updateModelStatus } = require('../../../models/modelSettings');
 const { getRevisionByIdOrTag, getRevisionFormat, onProcessingCompleted, updateProcessingStatus } = require('../../../models/revisions');
 const { initialiseAutomatedProperties, onModelNameUpdated, onTemplateUpdated } = require('../../../processors/teamspaces/projects/models/commons/tickets');
+const { isFederation: isFederationCheck, newRevisionProcessed, updateModelStatus } = require('../../../models/modelSettings');
 const { modelTypes, processStatuses } = require('../../../models/modelSettings.constants');
 const { publish, subscribe } = require('../../eventsManager/eventsManager');
 const { DRAWINGS_HISTORY_COL } = require('../../../models/revisions.constants');
 const { EVENTS: chatEvents } = require('../../chat/chat.constants');
+const { completeRun } = require('../../../processors/teamspaces/projects/clashes');
```
carmenfan: Can we rename this for better context?(we have to remember on this file the context is no longer clash)

something like this:

```suggestion
const { completeRun: processClashResults } = require('../../../processors/teamspaces/projects/clashes'
```

carmenfan: actually if we can rename completeRun to something like processResults within the clash processor it'd be nice too.

backend/src/v5/services/modelProcessing.js
```diff
@@ -46,30 +46,50 @@ const {
...
 	try {
-		const { status, database: teamspace, project: model, user, value, message } = JSON.parse(content);
-		if (status) {
-			publish(events.QUEUED_TASK_UPDATE,
-				{ teamspace, model, corId: properties.correlationId, status });
+		const { status, teamspace, project, container, drawing, type, user, value,
+			message, results } = JSON.parse(content);
+		const modelType = container ? modelTypes.CONTAINER : modelTypes.DRAWING;
+
+		if (type === MESSAGE_TYPES.CLASH) {
```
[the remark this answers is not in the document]
carmenfan: ☝️  we need to handle failures (and  have unit/e2e tests)

backend/src/v5/services/modelProcessing.js
```diff
@@ -46,30 +46,50 @@ const {
...
+};
+
 const onCallbackQMsg = ({ content, properties }) => {
 	logger.logInfo(`[Received][${properties.correlationId}] ${content}`);
 	try {
-		const { status, database: teamspace, project: model, user, value, message } = JSON.parse(content);
-		if (status) {
-			publish(events.QUEUED_TASK_UPDATE,
-				{ teamspace, model, corId: properties.correlationId, status });
+		const { status, teamspace, project, container, drawing, type, user, value,
```
carmenfan: I think it's better if we figure out whats' the type of message first, then dereference the other variables as appropriate inside the if statements.

i.e.


```suggestion
  const {type, status, ...msgContent} = JSON.parse(content)
  if(type === MESSAGE_TYPES.CLASH) {
     // specific params we expect from a clash complete message
    const { teamspace, project, results, /* etc */ } = msgContent
  } else if(status) {
    //specific params for status update message
    const { /*...*/ } = msgContent;
    
    
  }
  //...etc
```

backend/src/v5/services/modelProcessing.js
```diff
@@ -46,30 +46,50 @@ const {
...
+			message, results } = JSON.parse(content);
+		const modelType = container ? modelTypes.CONTAINER : modelTypes.DRAWING;
+
+		if (type === MESSAGE_TYPES.CLASH) {
+			const resultsDir = results.replace(SHARED_SPACE_TAG, sharedDir);
+			publish(events.CLASH_RUN_COMPLETED,
+				{ teamspace, project, corId: properties.correlationId, results: resultsDir });
+		} else if (status) {
+			publish(events.QUEUED_TASK_UPDATE, { teamspace,
```
carmenfan, line 110: As discussed with @sebjf  we should also processing statuses from the callbackq for clash tasks

backend/src/v5/services/modelProcessing.js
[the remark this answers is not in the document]
carmenfan: but... considering we don't deal with failures here for importing models, we shouldn't deal with the calsh failures here either

backend/src/v5/utils/responseCodes.js
```diff
@@ -95,6 +95,7 @@ ResponseCodes.templates = {
 
 	// Clashes related error
 	clashPlanNotFound: { message: 'Clash plan not found.', status: 404 },
+	testRunNotFound: { message: 'Clash test run not found.', status: 404 },
```
carmenfan: Since we've left the scope of a clash file here, we should prefix it with clash.

```suggestion
	clashRunNotFound: { message: 'Clash test run not found.', status: 404 },
```

carmenfan (changes requested):

backend/src/v5/models/clashes.runs.js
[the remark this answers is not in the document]
ChristosTsiotsias: this was not used, I forgot to remove it 🤣

ChristosTsiotsias: this method was not used cause I used getTestRunByQuery for it but should I still create the index ? do you reckon I should create it when a teamspace is created and add a migration script for existing teamspaces ?

carmenfan: yes, 100%

my worry is that the queries are going to start becoming really slow as test runs increases.

It could be done anywhere (e.g. on the insert function). we just need ot make sure all the clash run collections are indexed as we wish.

carmenfan (changes requested):
waiting for @ChristosTsiotsias  to fix some bugs, please re-flag when ready!

backend/src/v5/models/clashes.runs.js
```diff
@@ -15,24 +15,62 @@
...
+	await db.updateOne(teamspace, CLASH_RUNS_COL, { project, _id: runId },
+		{ $set: { ...setUpdate, updatedAt: new Date() } });
+};
+
+ClashRuns.updateRunStatus = async (teamspace, project, runId, status, results) => {
+	await updateClashRun(teamspace, project, runId, deleteIfUndefined({
+		status,
+		results,
+	}));
+};
```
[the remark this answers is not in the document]
carmenfan: behaviour has changed. runId is now used as a reference to locate the stored results.

The last completed run can be found by the run with completed status and the latest updatedAt.

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -33,56 +46,241 @@ Clashes.updatePlan = updatePlan;
...
+		}
+	}
+
+	categorizedClashes.resolved = Array.from(knownClashes.values());
+
+	await storeFile(teamspace, RUN_HISTORY_COL, runId, Buffer.from(JSON.stringify(categorizedClashes)));
+	await updateRunStatus(teamspace, project, runId, clashRunStatus.COMPLETED,
+		{ stats: {
+			new: categorizedClashes.new.length,
+			active: categorizedClashes.active.length,
```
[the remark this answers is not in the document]
carmenfan: we'll keep this approach for now and do optimisations at a later stage.

carmenfan (approved):
Approved. Waiting on bouncer deployment before we merge into staging (or it'll break the behaviour)

PR #5928, opened 2026-04-02
ISSUE #5927 fix invitations logic

### LINKED ISSUE #5927: Invitations are broken
### Description
With the mongo driver upgrade, invitations are broken.

v4 invitations are calling a mongo api directly (ensureIndex), which is now decommissioned.

This needs to be replaced by a call to the functions in db.js (which should do nothing if the index is already there)

### Goals
- [ ] As a TS admin, I want to be able to create invitations to invite other users into the teamspace.

carmenfan:
This fixes #5927

#### Description
- change every instance in invitations.js, where it gets the collection and does it's own DB call, to use a function in db.js instead.
- removed `setJob` - doesn't appear to be used

#### Acceptance Criteria
- invitations should work like before

### DIFF
--- a/backend/src/v4/handler/db.js
+++ b/backend/src/v4/handler/db.js
@@ -17,10 +17,10 @@
 
 "use strict";
 
-const {v5Path} = require("../../interop");
-const {_context, disconnect} = require(`${v5Path}/handler/db`);
+const { v5Path } = require("../../interop");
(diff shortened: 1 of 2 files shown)


ChristosTsiotsias (changes requested):
I think its not because of this PR but it seems that the unpack method is never called and even after the creation of a user the invitation stays in the database

ChristosTsiotsias (approved):

PR #5960, opened 2026-04-15
ISSUE_5910 - Config check in queue initialization

### LINKED ISSUE #5910: Add queue config validation
### Description
Whilst reviewing https://github.com/3drepo/3drepo.io/pull/5859#pullrequestreview-4020614658 I've realised that the tests are running and passing despite the test config did not have a clashq defined.

It appears that we are happily sending queue requests to undefined queue names. There should be a configuration validation check at init to ensure all expected values exist

### Goals
- [ ] As a devops, I want to be told if there is an expected value in the compute node config that is undefined
- [ ] As a devops, I want to be told if the provided path to a directory in the compute node is defined, but unreachable.

### Tasks
Developer: Please list out the tasks and your estimates

ChristosTsiotsias:
This fixes #5910

#### Description
A config check is done before initializing the queue to ensure all required values are defined

#### Test cases
Remove a required queue value from the test config
Run the drivers tests
An error should be thrown


backend/src/v5/handler/queue.js
```diff
@@ -178,7 +178,22 @@ Queue.close = async (reset = true) => {
...
+	if (!queueConfig
+			|| !queueConfig.host
+			|| !queueConfig.shared_storage
+			|| !queueConfig.callback_queue
+			|| !queueConfig.model_queue
+			|| !queueConfig.clash_queue
+			|| !queueConfig.event_exchange
+	) {
+		throw Error('Queue is not configured correctly.');
+	}
```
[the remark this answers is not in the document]
carmenfan: Do the check on https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/services/modelProcessing.js through https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/services/initialiser.js

other than `host` the rest of the config has nothing to do with the queue handler - they're all business logic.

Can we also make it more substanticial - define a yup schema and validate against that. (so we can check it's a non 0 length string etc.)

We should also check shared_storage is accessible with the appropriate permissions (check that it has sufficient access rights to the folder)

carmenfan (changes requested):

backend/src/v5/services/modelProcessing.js
```diff
@@ -23,10 +23,13 @@ const { listenToQueue, queueMessage } = require('../handler/queue');
 const { modelTypes, processStatuses } = require('../models/modelSettings.constants');
 const { DRAWINGS_HISTORY_COL } = require('../models/revisions.constants');
 const Path = require('path');
+const Yup = require('yup');
 const { addRevision } = require('../models/revisions');
 const archiver = require('archiver');
+const { constants } = require('fs');
 const { events } = require('./eventsManager/eventsManager.constants');
 const { execFile } = require('child_process');
+const fs = require('fs/promises');
```
carmenfan: just import what you need from fs/promises? there's already an entry in L20

backend/src/v5/services/modelProcessing.js
```diff
@@ -23,10 +23,13 @@ const { listenToQueue, queueMessage } = require('../handler/queue');
 const { modelTypes, processStatuses } = require('../models/modelSettings.constants');
 const { DRAWINGS_HISTORY_COL } = require('../models/revisions.constants');
 const Path = require('path');
+const Yup = require('yup');
 const { addRevision } = require('../models/revisions');
 const archiver = require('archiver');
+const { constants } = require('fs');
```
carmenfan: add it into L21

backend/src/v5/services/modelProcessing.js
```diff
@@ -260,6 +263,27 @@ ModelProcessing.queueClashRun = async (teamspace, project, corId, stream) => {
...
+		model_queue: Yup.string().trim().min(1).required(),
+		clash_queue: Yup.string().trim().min(1).required(),
+		event_exchange: Yup.string().trim().min(1).required(),
+	});
+
+	try {
+		await queueConfigSchema.validate(queueConfig);
+		await fs.stat(queueConfig.shared_storage);
+		// eslint-disable-next-line no-bitwise
+		await fs.access(queueConfig.shared_storage, constants.R_OK | constants.W_OK);
```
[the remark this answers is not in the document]
carmenfan: it's a good point

backend/src/v5/services/modelProcessing.js
```diff
@@ -260,6 +263,27 @@ ModelProcessing.queueClashRun = async (teamspace, project, corId, stream) => {
 	}
 };
 
+ModelProcessing.checkQueueConfig = async () => {
```
carmenfan: Can we we call this `init`?
then we can just build on it if we needed more stuff in the future.

backend/tests/v5/unit/services/modelProcessing.test.js
```diff
@@ -389,17 +395,48 @@ const testQueueClashRun = () => {
...
 
-			// fs.createWriteStream.mockRestore();
 			expect(Queue.queueMessage).not.toHaveBeenCalled();
 		});
 	});
 };
 
+const testCheckQueueConfig = () => {
+	describe('Check Queue Config', () => {
+		describe.each([
```
carmenfan: It'll be slightly cleaner (and less fussy to add more in the future) if you declare an array and pass it into the describe function, something like this:

```suggestion
        const params = ["host", "shared_storage", //etc...];
		describe.each(params.map(p => [p]))
```

carmenfan (changes requested):

carmenfan (approved):

Issue #5959, opened 2026-04-15
Restore Joystick support


rmittal007:
### Description
The Virtual Joystick Control API allows external consumers to inject joystick input into the Unity viewer at runtime. This is achieved by registering a callback function that is polled every frame to retrieve the current joystick values.

### API Description

**1. enableVirtualJoystick:** 
Enables the virtual joystick input system and registers a callback function that will be queried every frame to obtain joystick values.

Callback Requirements:
The callback is called once per frame.
It must return an array of four numbers.

Order of values: 
leftStickX
leftStickY
rightStickX
rightStickY

All values must be between -1.0 and 1.0.

**2. disableVirtualJoystick:** 
Disables virtual joystick input and unregisters any previously registered joystick callback.

PR #5962, opened 2026-04-16
ISSUE #5959 Add Virtual Joystick Support

rmittal007:
This PR exists to make a deployment for the Field team to test


frontend/src/globals/unity-util.ts
```diff
@@ -2740,4 +2747,39 @@ export class UnityUtil {
 	public static createWebRequestHandler(gameObjectName: string) {
 		return this.externalWebRequestHandler && this.externalWebRequestHandler.setUnityInstance(this.unityInstance, gameObjectName);
 	}
-}
+
+	/**
+	 * Enable the virtual joystick feature.
```
sebjf, line 2752: @rmittal007 can you expand on this a little to help out new users/integrators? For example, mention that this is polled, and that the values should be between 0 and 1. Give an example as well: for example, your test case of providing  a static array [0, 0, 1, 0] would suffice, just so whoever wants to call it has a line of working code they can start with!

sebjf (changes requested):
@rmittal007 looks good! Just expand on the comment/summary for the public API a bit to help new users!

sebjf (approved):

PR #5969, opened 2026-04-21
ISSUE #5968 - Fix front egg caching issue

### LINKED ISSUE #5968: Create some debug code for TEAMSPACE_NOT_FOUND issue
### Description
Occasionally, we're finding that staging is stuck on TEAMSPACE_NOT_FOUND when someone is entering a particular teamspace:

Image

- This seems to only happen to a specific teamspace, and it goes away if the pod is restarted.
- API key auth is still reporting teamspace not found
- When it happens, accessing the same teamspace in a different deployment is not affected.

It's unclear what is causing it, thus I want to add some additional debugging so we have more to go by the next time it happens.

carmenfan:
This fixes #5968

#### Description
updateOne is being called with `{upsert: true}` instead of just a `true`. Meaning upsert never happened, thus the cache was never utilised.


carmenfan:
@copilot it appears that the bug was to do with a wrong parameter to signal upsert when updateOne is called (it's suppose to be a boolean operator, not an object with {upsert: true}, see https://github.com/3drepo/3drepo.io/pull/5969/changes/a35449800bcf2ea0ede392a933b7e6edf1c96542). Can you check around the repository to ensure all calls to updateOne or updateMany to db.js have the correct params?

PR #5975, opened 2026-04-23
ISSUE #5953 - Extend the Unity-Util API to allow intercepting and providing responses in offline mode

### LINKED ISSUE #5953: Extend the Unity-Util API to allow intercepting and providing responses in offline mode
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/347

All web requests from the viewer should now go through the ExternalWebRequestHandler. To support fully offline integrations, we need the handler to be able to receive responses from an offline store.

This task is to design & add an API to unity-util, allowing an integration to intercept any requests from the viewer and optionally provide their own responses.

Some general guidelines:

* The API should do everything that Field requires.
* The API should operate using Streams (this is because in the future we may want to process large responses with an event based parser).

Other comments:

* Whether the interception should take place above or below the existing indexeddb cache supported by ExternalWebRequestHandler is an open question. We should evaluate the strengths and advantages of each option.

### Goals

- [ ] The public API, consisting of one or more methods, exists on the UnityUtil class
- [ ] When not called, ExternalWebRequestHandler behaves as normal
- [ ] When used by an integration, all requests can be intercepted and handled.

### Tasks

- [ ] TBC

### Related Resources

MockUp : `link to mockup`

vmavadiya:
This fixes #5953

#### Description
- The interceptor is declared inside ExternalWebRequestHandler and is applied in the createGetRequest method if it is defined or set via UnityUtil’s exposed setOfflineFetchInterceptor method, which is used for fetching data from the offline store.


sebjf (changes requested):
@vmavadiya, the changes look good! I think we should add some unit tests, partly to catch regression issues but also for documentation purposes (i.e. so users can see how a store might be implemented).

I have added a boilerplate file `frontend/test/unity-util/unity-util.spec.ts` and made sure it works. You can either call it from the CLI: 

```
npm test ./test/unity-util/unity-util.spec.ts
```
(from the frontend folder)

Or the following launch config works for me as well:

```
  {
            "type": "node",
            "name": "jest v5:unit specific file",
            "request": "launch",
            "args": [
                "jest",
                "./test/unity-util/unity-util.spec.ts",
            ],
            "cwd": "C:\\3drepo\\3drepo.io\\frontend",
            "console": "integratedTerminal",
            "internalConsoleOptions": "neverOpen",
            "disableOptimisticBPs": true,
            "runtimeExecutable": "yarn",
            "env": {
                "NODE_ENV":"testV5",
                "NODE_CONFIG_DIR":"C:/3drepo/3drepo.io/config",
            },
            "envFile": "C:/3drepo/3drepo.io/config/testV5/.env",
        },
```

Can you add one or two tests to that for your new functionality (e.g. based on the snippet you sent me in Teams). The tests don't have to be very complex, but they should test providing an ArrayBuffer to the Response and getting a ReadableStream and vice-versa.

sebjf (approved):

PR #5979, opened 2026-04-27
Issue 5977 - Add clash report as a property within clash module

### LINKED ISSUE #5977: Custom tickets : Store clash report name in default clash module
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/961
We want to be able to keep track of which clash report the clash came from, and be able to perform grouping/filtering with it.

This can all be done if we add an additional field on the clash default module so we can sync this to/from navisworks.

It should be read only on UI, so if the clash report got renamed, it can be updated appropriately. It would also allow us to update any existing clash which currently does not store this value.

### Goals
- As a user, I want to be able to record/update the clash report name when I sync from navisworks.
- As a user, I want to be able to see which clash report the clash item came from, and perform actions such as "Group by" and "filter" based on it.

### Tasks
- Add clash report as one of the properties within the clash module

carmenfan:
This fixes #5977

#### Description
- add additional property to record clash report within the clash module
  - It should be read only on UI - i.e. disabled on io frontend, but available to update via API

This is added to support https://github.com/3drepo/3drepoPlugin/issues/586

#### Acceptance Criteria
- There should now be a property `Clash Report` within the clash module
  - This will be blank by default
  - This will be read only on the UI
  - The users can edit this through API requests.


ChristosTsiotsias (changes requested):
Is there any unit test we could ament to ensure it will always be added ?

carmenfan:
> Is there any unit test we could ament to ensure it will always be added ?

@ChristosTsiotsias  no we can't enforce that - because we need to ensure we're still supporting the older version of the plugins, which will not be sending the report names. This is why it's only set to `ReadOnlyOnUI` but not required. 

As for unit tests, i don't think it's needed - functions like any standard property. We're just adding a property to a default template, we didn't change any functionality.

ChristosTsiotsias (approved):

PR #5984, opened 2026-04-29
Issue 5794 - Move maps to v5

### LINKED ISSUE #5794: Move maps to v5
### Description

This is part of the effort in migrating all endpoints used by Unity viewer to v5. They will need to be available in both internal and external service.

### endpoints changes

| v4| v5 |
|--------|--------|
| /:teamspace/:model/maps/| /teamspaces/:teamspace/projects/:project/:modelType/:model/maps |
| /:teamspace/:model/maps/hereinfo/ |  /teamspaces/:teamspace/projects/:project/:modelType/:model/maps/here/ | 
| /:model/maps/osm/:zoomLevel/:gridx/:gridy.png / | /teamspaces/:teamspace/projects/:project/:modelType/:model/maps/osm/default/tiles?zoomLevel=i&x=j&y=k |
| /:model/maps/here*/:zoomLevel/:gridx/:gridy.png / | /teamspaces/:teamspace/projects/:project/:modelType/:model/maps/here/*/tiles?zoomLevel=i&x=j&y=k (see notes) |

NOTES:
- here tiles endpoint:
  - here -> here/default/tiles
  - hereaerial -> here/aerial/tiles
  - heretraffic -> here/traffic/tiles
  - etc..
- herebuilding (where it takes a lat long instead) can be ignored 
- Ensure we have a middleware to verify the params

### Code structure
It would good if we can organise the code into something like this
````mermaid
flowchart TD
route[Map route] -->processor[Map processor]
processor --> OSMService[OSM Service]
processor --> HEREService[HERE Service]
OSMService --> httputils[utils/webRequests]
HEREService --> httputils

subgraph Services/maps
    OSMService
    HEREService
end
````

- As usual, route takes in the network requests and utilise middlewares to validate the request
- the request gets sent to the processor
- processor detemines which map type we're after and sends the request to the relevant service
- the services should utilise the webRequest utility to fetch tiles from 3rd party

### Goals
- As 3drepo maintainer I want to ensure all unity viewer endpoints are using v5 endpoints
- As asite collaborator I want to ensure I can access the maps endpoints via the internal service

#### Project Task List
- [x] **HERE Service** (Est: 2d)
  - [x] create the here service (Est: 1d)
  - [x] create the unit tests (Est: 1d)
- [x] **OSM Service** (Est: 2d)
  - [x] create the osm service (Est: 1d)
  - [x] create the unit tests (Est: 1d)
- [x] **Addon Access Middleware** (Est: 1.5d)
  - [x] create the middleware (Est: 1d)
  - [x] create the unit tests (Est: 0.5d)
- [x] **Route Endpoints** (Est: 8 d)
  - [x] create the route endpoints (Est: 1d)
  - [x] create the processor functions (Est: 3d)
    - [x] create the unit tests (Est: 1.5d)
  - [x] create the e2e test (Est: 2d)
- [x] **Manual Testing** (Est: 0.5d)

**Total Estimated Time:** 14 days

### PR DESCRIPTION
**Pull request by Bogdan-Crn**
This fixes #5794

#### Description
As part of the effort to migrate endpoints to v5 this PR moves the maps endpoints used by unity viewer.

This PR has moved almost all maps endpoints to v5 and decommissioned the v4 endpoints.
As part of this move new services for HERE and OST were created as well as new processors and middleware to validate if HERE is enabled on the teamspace.

Unit and e2e tests have been updated to have coverage of the new added endpoints and functions.

#### Acceptance Criteria
As 3drepo maintainer I want to ensure all unity viewer endpoints are using v5 endpoints
As asite collaborator I want to ensure I can access the maps endpoints via the internal service

### CODE REVIEW

File: backend/src/v4/routes/maps.js [hunk 1]
```diff
@@ -16,200 +16,33 @@
...
+router.get("/:model/maps/heretruck/:zoomLevel/:gridx/:gridy.png", routeDecommissioned("GET", "/v5/teamspaces/:teamspace/projects/:project/containers/:container/maps/here/truck/tiles"));
+router.get("/:model/maps/heretruckoverlay/:zoomLevel/:gridx/:gridy.png", routeDecommissioned("GET", "/v5/teamspaces/:teamspace/projects/:project/containers/:container/maps/here/truckoverlay/tiles"));
+router.get("/:model/maps/herelabeloverlay/:zoomLevel/:gridx/:gridy.png", routeDecommissioned("GET", "/v5/teamspaces/:teamspace/projects/:project/containers/:container/maps/here/labeloverlay/tiles"));
+router.get("/:model/maps/heretollzone/:zoomLevel/:gridx/:gridy.png", routeDecommissioned("GET", "/v5/teamspaces/:teamspace/projects/:project/containers/:container/maps/here/tollzone/tiles"));
+router.get("/:model/maps/herepoi/:zoomLevel/:gridx/:gridy.png", routeDecommissioned("GET", "/v5/teamspaces/:teamspace/projects/:project/containers/:container/maps/here/poi/tiles"));
 router.get("/:model/maps/herebuildings/:lat/:long/tile.json", middlewares.isHereEnabled, getHereBuildingsFromLongLat);
```
carmenfan: Decommission this, too. It's never used anyway

backend/src/v5/middleware/dataConverter/queryParams.js
```diff
@@ -0,0 +1,39 @@
...
+
+const QueryParams = {};
+
+const mapsCoordinatesSchema = yup.object({
+	zoomLevel: yup.number().integer().min(0).required(),
+	x: yup.number().integer().min(0).required(),
+	y: yup.number().integer().min(0).required(),
+}).required();
+
+QueryParams.validateMapsCoordinates = async (req, res, next) => {
```
carmenfan: This should live inside a maps-specific middleware.

backend/src/v5/middleware/permissions/components/teamspaces.js
```diff
@@ -67,6 +67,20 @@ TeamspacePerms.isTeamspaceMember = checkTeamspaceMembership(true);
 
 TeamspacePerms.isActiveTeamspaceMember = checkTeamspaceMembership(false);
 
+TeamspacePerms.isAddOnEnabled = (addOn) => async (req, res, next) => {
+	try {
+		const { teamspace } = req.params;
+		const enabled = await isAddOnEnabled(teamspace, addOn);
+		if (!enabled) {
+			respond(req, res, createResponseCode(templates.addOnUnavailable, `${addOn} add-on is not enabled in this teamspace`));
```
carmenfan: Use throw here, so you just respond in the catch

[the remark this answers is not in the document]
carmenfan: We can just be generic here. Don't need a custom message. The context of the endpoint should be clear what we mean.

backend/src/v5/middleware/permissions/index.js
```diff
@@ -70,6 +72,13 @@ Permissions.hasWriteAccessToContainer = validateMany([Permissions.hasAccessToTea
 Permissions.hasAdminAccessToContainer = validateMany([Permissions.hasAccessToTeamspace,
 	getModelIdFromParam(modelTypes.CONTAINER), hasAdminAccessToContainer]);
 
+Permissions.hasAccessToContainersWithCoords = validateMany([
+	Permissions.hasReadAccessToContainer, validateMapsCoordinates]);
+Permissions.hasHereAccessToContainer = validateMany([
+	Permissions.hasReadAccessToContainer, isAddOnEnabled(ADD_ONS.HERE)]);
+Permissions.hasHereAccessToContainerWithCoords = validateMany([
```
carmenfan: This is high coupling. Permissions should only be dealing with permissions. Add the validation middleware within the route file/map specific middleware

backend/src/v5/processors/teamspaces/projects/models/commons/maps.js
```diff
@@ -0,0 +1,110 @@
...
+
+const Maps = {};
+
+Maps.getListOfMaps = async (teamspace) => {
+	const hereEnabled = await isAddOnEnabled(teamspace, ADD_ONS.HERE);
+	const maps = [
+		{ name: 'Open Street Map', layers: [{ name: 'Map Tiles', source: 'OSM' }] },
+
+	];
+	if (hereEnabled && config.here && config.here.apiKey) {
```
carmenfan:
```suggestion
	if (hereEnabled && config?.here.apiKey) {
```

backend/src/v5/processors/teamspaces/projects/models/commons/maps.js
```diff
@@ -0,0 +1,110 @@
...
+		maps.push(...[
+			{ name: 'Here',
```
carmenfan: just push this out into a constant and reference it here

backend/src/v5/processors/teamspaces/projects/models/commons/maps.js
```diff
@@ -0,0 +1,110 @@
...
+					{ name: 'Map Tiles', source: 'HERE_POI' },
+					{ name: 'Traffic Flow', source: 'HERE_TRAFFIC_FLOW' },
+					{ name: 'Truck Restrictions', source: 'HERE_TRUCK_OVERLAY' },
+				] },
+		]);
+	}
+
+	return maps;
+};
+
```
carmenfan, line 36: We can do this better; it is very wordy. 

just have a `getTile(mapProvider, mapName,  /*.. whatever other params you need..*/)` 

`mapType` should reference the `source` name defined in `getListOfMaps` (but move this out into a constant.

So routes calling the processor should just be doing something like `getTile(mapType.OSM, blah, blah)`

And we have a single function that determines whether something should call the OSM or HERE service.

Then it should be the HERE service that inspects `mapName` and calls the right API.

backend/src/v5/processors/teamspaces/projects/models/commons/maps.js
```diff
@@ -0,0 +1,110 @@
...
+const OSMService = require('../../../../../services/maps/osm');
+const config = require('../../../../../utils/config');
+const { isAddOnEnabled } = require('../../../../../models/teamspaceSettings');
+
+const Maps = {};
+
+Maps.getListOfMaps = async (teamspace) => {
+	const hereEnabled = await isAddOnEnabled(teamspace, ADD_ONS.HERE);
+	const maps = [
+		{ name: 'Open Street Map', layers: [{ name: 'Map Tiles', source: 'OSM' }] },
```
carmenfan: This is a bit high coupling, the map service shouldn't need to know the type of maps each service supports.

For this and also the HERE list, get the map service to provide you a list and concat them instead.

backend/src/v5/routes/teamspaces/projects/models/common/maps.js
```diff
@@ -0,0 +1,1096 @@
...
+		const { teamspace } = req.params;
+		const maps = await Maps.getListOfMaps(teamspace);
+		respond(req, res, templates.ok, maps);
+	} catch (err) {
+		// istanbul ignore next
+		respond(req, res, err);
+	}
+};
+
+const getHereBaseInfo = async (req, res) => {
```
carmenfan: Looking around... I don't think this is ever called.

Can you verify? If it isn't used, then we should axe this function altogether.

backend/src/v5/routes/teamspaces/projects/models/common/maps.js
```diff
@@ -0,0 +1,1096 @@
...
+		const info = await Maps.getHereBaseInfo();
+
+		respond(req, res, templates.ok, info);
+	} catch (err) {
+		// istanbul ignore next
+		respond(req, res, err);
+	}
+};
+
+const getOSMTile = async (req, res) => {
```
carmenfan: This is a crazy amount of replication to do the same thing.

With the suggested change to the processor, you can utilise the same function for all routes.

something like this:


```suggestion
const getTile = (mapProvider, mapType) => async (req, res) => {
  try { 
    const { zoomLevel, x, y } = req.query;
	const tileData = await Maps.getTile(mapProvider, mapType, zoomLevel, x, y);
  /*...etc */
}
```

And you call this in the route declaration with something like this:

````js
  router.get('/osm/tiles', hasAccessToContainersWithCoords, getTile(mapProvider.OSM, osmMapType.STANDARD));
````

File: backend/src/v5/routes/teamspaces/projects/models/common/maps.js [hunk 11]
```diff
@@ -0,0 +1,1096 @@
...
+	 *           image/png:
+	 *             schema:
+	 *               type: string
+	 *               format: binary
+	 */
+	router.get('/osm/tiles', hasAccessToContainersWithCoords, getOSMTile);
+
+	/**
+	 * @openapi
+	 * /teamspaces/{teamspace}/projects/{project}/containers/{container}/maps/here/default/tiles:
```
carmenfan: Instead of defining all the endpoints like this, we should be able just to put the map type into a variable?

i.e. 
```suggestion
	 * /teamspaces/{teamspace}/projects/{project}/containers/{container}/maps/here/{mapType}/tiles:
```

We could even do it to include osm (we can slightly change the endpoint to fit):

```js
	 * /teamspaces/{teamspace}/projects/{project}/containers/{container}/maps/{mapProvider}/{mapType}/tiles:
```

We'll need middleware to validate the variables (and probably convert them to the expected constant values if necessary). This will save a lot of duplication...

backend/src/v5/services/maps/here.constants.js
```diff
@@ -0,0 +1,53 @@
...
+	TRUCK_RESTRICTIONS: 'explore.day',
+	AERIAL: 'satellite.day',
+	TRAFFIC: 'logistics.day',
+	TERRAIN: 'topo.day',
+	HYBRID: 'explore.satellite.day',
+	GREY: 'lite.day',
+};
+
+HereConstants.RESOURCES = {};
+resources.forEach((resource) => {
```
carmenfan: We have something that does this already. If you need the reference in a constant case, you could add a flag to that function so it does it for you.

https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/utils/helper/objects.js#L45

carmenfan: I also don't think you need to declare the resources as a separate variable unless you reference them in other places (which it appears that you don't)

So do something like this:

```suggestion
HereConstants.RESOURCES =createConstantsObject([
	'base',
	'flow',
	'label',
	'blank',
], true);
```

backend/src/v5/services/maps/here.constants.js
```diff
@@ -0,0 +1,53 @@
...
+	'flow',
+	'label',
+	'blank',
+];
+
+HereConstants.BASE_URL = 'maps.hereapi.com';
+
+HereConstants.HERE_TRAFFIC_DOMAIN = 'traffic.maps.hereapi.com';
+
+HereConstants.FEATURES = {
```
carmenfan: These are deviating a bit from our coding style in the project.

The variable that stores an enum-like constant object is usually declared in camelCase. It should be `feature` here. 

(also applies to the other object declarations in this file)

backend/src/v5/services/maps/here.js
```diff
@@ -0,0 +1,205 @@
...
+		domain,
+		'v3',
+		resource,
+		'mc',
+		zoomLevel,
+		x,
+		y,
+		'png8',
+	].join('/');
+	const uriQuery = [
```
carmenfan: It'll be cleaner if you generate an object instead and use the querystring functionality https://nodejs.org/api/querystring.html

backend/src/v5/services/maps/here.js
```diff
@@ -0,0 +1,205 @@
...
+const { BASE_URL, FEATURES, HERE_TRAFFIC_DOMAIN, RESOURCES, STYLES } = require('./here.constants');
+const { get, getArrayBuffer } = require('../../utils/webRequests');
+const config = require('../../utils/config');
+const { logger } = require('../../utils/logger');
+const { templates } = require('../../utils/responseCodes');
+
+const HereService = {};
+
+const generateTileURI = (domain, resource, zoomLevel, x, y, { features, style } = {}) => {
+	const uriPath = [
```
carmenfan: I think it'd be more readable if you just do it in a string instead of utilising `join`. A lot of these things are set in stone anyway (e..g domain never changes I think?)

so 

something like this:


```suggestion
	const uriPath = `https://${BASE_URL}/v3/${resource}...`
```

So it's clear the structure of the endpoint and much easier to decode when we need to do changes.

backend/src/v5/services/maps/here.js
```diff
@@ -0,0 +1,205 @@
...
+		const uri = `https://${BASE_URL}/v3/info?apiKey=${config.here.apiKey}`;
+		const { data } = await get(uri);
+		return data;
+	} catch (error) {
+		logger.logError(`Failed to get HERE base info: ${error?.response?.data} `);
+		throw templates.mapsRequestFailed;
+	}
+};
+
+HereService.getTile = async (zoomLevel, x, y) => {
```
carmenfan: similar here...

Instead of duplicating, have a single helper function that everything calls.

This will change naturally after applying the change request I noted in the maps processor. So not going to provide an example here!

backend/src/v5/services/maps/osm.js
```diff
@@ -0,0 +1,45 @@
...
+
+const config = require('../../utils/config');
+const { logger } = require('../../utils/logger');
+const { templates } = require('../../utils/responseCodes');
+
+const OSMService = {};
+
+OSMService.getTile = async (zoomLevel, x, y) => {
+	try {
+		let domain = 'a.tile.openstreetmap.org';
```
carmenfan: don't use openstreetmap as the default.

If it's not configured, let's make it not available.

OSM is going to get really mad at us otherwise...

backend/src/v5/utils/responseCodes.js
```diff
@@ -48,6 +48,7 @@ ResponseCodes.templates = {
 	ssoRestricted: { message: 'This teamspace only accepts Single Signed On users. Please link your account with an authority.', status: 401 },
 	domainRestricted: { message: 'Your email does not belong in a domain that is accepted by this teamspace. Please contact your teamspace administrator.', status: 401 },
 	moduleUnavailable: { message: 'This module is not available in this teamspace.', status: 400 },
+	addOnUnavailable: { message: 'This add-on is not available in this teamspace.', status: 400 },
```
carmenfan:
```suggestion
	addOnUnavailable: { message: 'This feature is not available in this teamspace.', status: 400 },
```

backend/src/v5/utils/responseCodes.js
```diff
@@ -105,6 +106,9 @@ ResponseCodes.templates = {
 	// Queue related
 	queueConnectionError: { message: 'There was a problem connecting to the queue. Please contact support.', status: 500 },
 	queueInsertionFailed: { message: 'Failed to insert an item into the queue. Please contact support.', status: 500 },
+
+	// MAPs related:
+	mapsRequestFailed: { message: 'Request failed to retrieve map data.', status: 400 },
```
carmenfan:
```suggestion
	mapsRequestFailed: { message: 'Failed to retrieve map data, please try again later.', status: 500 },
```

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/maps.test.js
```diff
@@ -0,0 +1,320 @@
...
+
+const genRoute = ({
+	teamspace,
+	projectId,
+	containerId,
+	path = '',
+	key,
+	query,
+}) => {
+	const searchParams = new URLSearchParams();
```
carmenfan: It's interesting that this is a completely different approach to constructing the query string from your main code implementation 😆 

You can use the helper function in the service helper; it removes undefined options. https://github.com/3drepo/3drepo.io/blob/master/backend/tests/v5/helper/services.js#L412

So something like this can replace all that logic you added here:


```suggestion
const qs = {key, ...query};
	`/v5/teamspaces/${teamspace}/projects/${projectId}/containers/${containerId}/maps${path}${createQueryString (qs)'}`
```

backend/src/v5/routes/routesManager.js
```diff
@@ -57,6 +58,7 @@ RoutesManager.init = (app) => {
 	app.use('/v5/teamspaces/:teamspace/projects/:project/containers/:model/assets', CreateAssetsRoutes(modelTypes.CONTAINER, internal));
 	app.use('/v5/teamspaces/:teamspace/projects/:project/containers/:model/assets/bundles', CreateBundleRoutes(modelTypes.CONTAINER, internal));
 	app.use('/v5/teamspaces/:teamspace/projects/:project/containers/:container/metadata', CreateMetadataRoutes(internal));
+	app.use('/v5/teamspaces/:teamspace/projects/:project/containers/:container/maps', CreateMapsRoutes());
```
carmenfan: This needs to work for both federations and containers This will mean we also need e2e tests for federations.

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/maps.test.js
```diff
@@ -0,0 +1,320 @@
...
+
+		const route = ({
+			key = users.tsAdmin.apiKey,
+			projectId = project.id,
+			containerId = container._id,
+		} = {}) => genRoute({ teamspace, projectId, containerId, key });
+
+		describe.each([
+			['the user does not have a valid session', route({ key: null }), false, templates.notLoggedIn],
+			['the user is not a member of the teamspace', route({ key: users.nobody.apiKey }), false, templates.teamspaceNotFound],
```
carmenfan: For this and all the HERE tests, can we make sure we have a test for if the teamspace is not configured with HERE enabled?

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/maps.test.js
```diff
@@ -0,0 +1,320 @@
...
+};
+
+describe(ServiceHelper.determineTestGroup(__filename), () => {
+	beforeAll(async () => {
+		server = await ServiceHelper.app();
+		agent = await SuperTest(server);
+	});
+
+	afterAll(() => ServiceHelper.closeApp(server));
+
```
carmenfan, line 242: we also need internal server tests.

backend/tests/v5/e2e/routes/teamspaces/projects/models/common/maps.test.js
```diff
@@ -0,0 +1,320 @@
...
+			projectId = project.id,
+			containerId = container._id,
+			query = {
+				zoomLevel: 10,
+				x: 2,
+				y: 3,
+			},
+		} = {}) => genRoute({ teamspace: ts, projectId, containerId, path, key, query });
+
+		describe('Endpoint protection and validation', () => {
```
carmenfan: Can we structure it the same way as other tests?

with the test cases in an array etc.

backend/tests/v5/unit/middleware/dataConverter/queryParams.test.js
```diff
@@ -0,0 +1,153 @@
...
+const QueryParams = require(`${src}/middleware/dataConverter/queryParams`);
+const { templates } = require(`${src}/utils/responseCodes`);
+const { determineTestGroup } = require('../../../helper/services');
+
+jest.mock('../../../../../src/v5/utils/responder');
+const Responder = require(`${src}/utils/responder`);
+
+const testValidateMapsCoordinates = () => {
+	describe('validateMapsCoordinates', () => {
+		test('should call next() and cast query values to numbers if query is valid', async () => {
```
carmenfan: can we make the test cases array here to reduce duplication? 

also it would be good if we can add more test cases (e.g. drop every query param separately and ensure they all break, additional params are ignored)

backend/tests/v5/unit/middleware/permissions/components/teamspaces.test.js
```diff
@@ -282,10 +282,51 @@ const testNotUserProvisioned = () => {
...
+			expect(TeamspaceSettings.isAddOnEnabled).toHaveBeenCalledWith(teamspace, addOn);
+		});
+
+		test('should respond with add-on module unavailable if the add-on is not enabled', async () => {
+			TeamspaceSettings.isAddOnEnabled.mockResolvedValueOnce(false);
+			await TSMiddlewares.isAddOnEnabled(addOn)({ params: { teamspace } }, {}, mockCB);
+
+			expect(mockCB).not.toHaveBeenCalled();
+			expect(Responder.respond).toHaveBeenCalledTimes(1);
+			expect(Responder.respond.mock.results[0].value.code).toEqual(templates.addOnUnavailable.code);
```
carmenfan: You should be able to do this with `toHaveBeenCalledWith`?

esp  now we're making the error message generic, but even previously, you should be able to do it with [`objectContaining`](https://jestjs.io/docs/expect#expectobjectcontainingobject)

backend/tests/v5/unit/middleware/permissions/components/teamspaces.test.js
```diff
@@ -282,10 +282,51 @@ const testNotUserProvisioned = () => {
...
+		});
+
+		test('should respond with error if the isAddOnEnabled throws error', async () => {
+			const error = new Error();
+			TeamspaceSettings.isAddOnEnabled.mockRejectedValueOnce(error);
+			await TSMiddlewares.isAddOnEnabled(addOn)({ params: { teamspace } }, {}, mockCB);
+
+			expect(mockCB).not.toHaveBeenCalled();
+			expect(Responder.respond).toHaveBeenCalledTimes(1);
+			expect(Responder.respond.mock.results[0].value.code).toEqual(error.code);
```
carmenfan: similar here

backend/tests/v5/unit/models/teamspaceSettings.test.js
```diff
@@ -725,6 +725,33 @@ const testGetTeamspaceInvites = () => {
 	});
 };
 
+const testIsAddOnEnabled = () => {
+	describe('Is addOn enabled', () => {
+		const teamspace = generateRandomString();
+
+		test('should return true if addOn is enabled', async () => {
+			const fn = jest.spyOn(db, 'findOne').mockResolvedValue({ addOns: { [ADD_ONS.VR]: true } });
```
carmenfan:
```suggestion
			const fn = jest.spyOn(db, 'findOne').mockResolvedValueOnce({ addOns: { [ADD_ONS.VR]: true } });
```

backend/tests/v5/unit/models/teamspaceSettings.test.js
```diff
@@ -725,6 +725,33 @@ const testGetTeamspaceInvites = () => {
...
+			const addOnName = ADD_ONS.VR;
+
+			await expect(Teamspace.isAddOnEnabled(teamspace, addOnName)).resolves.toEqual(true);
+
+			expect(fn).toHaveBeenCalledTimes(1);
+			expect(fn).toHaveBeenCalledWith(teamspace, TEAMSPACE_SETTINGS_COL, { _id: teamspace },
+				addOnsProjection, undefined);
+		});
+		test('should return false if addOn is not found', async () => {
+			const fn = jest.spyOn(db, 'findOne').mockResolvedValue({ addOns: { } });
```
carmenfan:
```suggestion
			const fn = jest.spyOn(db, 'findOne').mockResolvedValueOnce({ addOns: { } });
```

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/maps.test.js
```diff
@@ -0,0 +1,284 @@
...
+const OSMService = require(`${src}/services/maps/osm`);
+
+const testGetListOfMaps = () => {
+	describe('Get list of maps', () => {
+		const teamspace = generateRandomString();
+		test('should return only Open Street Map if HERE add-on is not enabled', async () => {
+			config.here = { };
+			const maps = await Maps.getListOfMaps(teamspace);
+			expect(maps).toEqual([
+				{ name: 'Open Street Map', layers: [{ name: 'Map Tiles', source: 'OSM' }] },
```
carmenfan: declare constants in the main code base and reference them here

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/maps.test.js
```diff
@@ -0,0 +1,284 @@
...
+			]);
+		});
+
+		test('should return all maps if HERE add-on is enabled and config is set', async () => {
+			config.here = { apiKey: 'testApiKey' };
+			TeamspaceSettings.isAddOnEnabled.mockResolvedValueOnce(true);
+			const maps = await Maps.getListOfMaps(teamspace);
+			expect(maps).toEqual([
```
carmenfan: similar here

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/maps.test.js
```diff
@@ -0,0 +1,284 @@
...
+			expect(tileData).toEqual('defaultTileData');
+		});
+	});
+};
+
+const testGetHereAerialTile = () => {
+	describe('Get HERE aerial tile', () => {
+		test('should return HERE aerial tile data', async () => {
+			HereService.getAerialTile.mockResolvedValueOnce('aerialTileData');
+			const tileData = await Maps.getHereAerialTile(10, 5, 5);
```
carmenfan: similar to the rest of the file, can you use a random number generator (e.g. https://github.com/3drepo/3drepo.io/blob/master/backend/tests/v5/helper/services.js#L445)

And reference the same numbers in your results.

We want to avoid hard-coding as much as possible.

(applies to the other tests here)

backend/tests/v5/unit/processors/teamspaces/projects/models/commons/maps.test.js
```diff
@@ -0,0 +1,284 @@
...
+			expect(HereService.getTile).toHaveBeenCalledWith(10, 5, 5);
+			expect(tileData).toEqual('defaultTileData');
+		});
+	});
+};
+
+const testGetHereAerialTile = () => {
+	describe('Get HERE aerial tile', () => {
+		test('should return HERE aerial tile data', async () => {
+			HereService.getAerialTile.mockResolvedValueOnce('aerialTileData');
```
carmenfan: use constants from main source code. (similar to the rest of the tests here)

carmenfan (changes requested):

backend/src/v5/routes/teamspaces/projects/models/common/maps.js
Bogdan-Crn: the getHereBaseInfo is called by the `/here` endpoint

Bogdan-Crn: I've updated the function to be just getBaseInfo and to return based on mapProvider

backend/src/v5/services/maps/here.js
Bogdan-Crn: yeah, looking at it there will always be a url and resource and it will fail the middleware validation if the zoomLevel, x and/or y are not present

backend/src/v5/services/maps/osm.js
Bogdan-Crn: Ok, I only kept it because on v4 that was the default if here was not available. Should I do a OSM check as well similar to `isAddonEnabled`?

backend/src/v5/utils/responseCodes.js
Bogdan-Crn: why do we want a 500 instead of a 400 error?

backend/src/v5/routes/routesManager.js
Bogdan-Crn: Will update to have federations as well, for some reason I thought this one is only applicable to containers on v5 🥲

backend/src/v5/utils/responseCodes.js
carmenfan: Because it's an internal error that's unexpected, rather than a user-related one.

We already validated the request, so for situations where they don't have the subscription, or they didn't provide the right parameters, we'd have returned the relevant errors.

If we're returning this, it meanw the user did everything right, but we still couldn't pull the tile for them in which case something unexpected happened, not a user fault.

This is based on my understanding of where this is used, unless I've mistaken?

Bogdan-Crn: That is triggered when we're trying to pull the tile from either OSM or HERE but we get back an error from them (i.e. unauthorised) so from our side either the api key is not correct or the service we're trying to reach is not available ( or in the last case the params provided to the service are not right).
That's why I thought a 400 would be fine because it's a bad request made towards OSM or HERE

carmenfan: No Imo this is 500.

This issue is when communicating with OSM/Here. With a bad key, OSM/HERE would return a 4xx error (because we're their client), so we should return a 5xx error to our own clients, since the fault lies with us (we have to update our config on the server, it has nothing to do with the clients). Our end users didn't make a bad request, provided their teamspace has the add-on enabled, and the parameters are correct.

The client isn't concerned about how we fetch the tiles, and issues like the API key not working aren't their fault.

This is similar to a MongoDB query, and it returned an error. We validated the request before processing it, so this is our issue, and we should be returning a 500.

Bogdan-Crn: Fair enough, that does make sense, thank you.

backend/src/v5/services/maps/osm.js
carmenfan: nah there's no add on for OSM, we provide that to all teamspace,

Just simply check if we have configuration settings for OSM, if there is do the magic, if not return an error

sanmont3drepo (dismissed):

sebjf:
@carmenfan @Bogdan-Crn this branch has been updated with the latest version of the viewer, which uses the v5 maps apis.

carmenfan (dismissed):

carmenfan (approved):

PR #5991, opened 2026-05-01
Issue 5978 - Add groups option to loadModel

### LINKED ISSUE #5978: Support loading models by floor/partition
### Description
Product issue: 

Bouncer is being given the option to cut up models by floor (and in the future, other arbitrary partitions and groupings). We should expose this logic to unity-util.

This functionality does not at yet cover the tree.

### Goals

- [ ] Unity-util has an option to load only a subset of RepoBundles based on groupings
- [ ] When loading only a few such bundles, all other viewer features (such as groups & overrides that may have been placed on unloaded parts of the model) should continue to work.

### Tasks

- [ ] TBC

### Related Resources

MockUp : `link to mockup`

sebjf:
This fixes #5978

#### Description
This PR adds a parameter to the model initialisation structure passed to unity to allow the viewer to only load assets with specific tags.

This PR does not update the viewer because the necessary viewer changes are already in staging/5.22.2.

It is not necessary to use this to load a subset of a Container - the backend can also pass a munged bundles response to get the same results.


frontend/src/globals/unity-util.ts
```diff
@@ -1794,6 +1803,7 @@ export class UnityUtil {
 		isFederation: boolean = false,
 		initView = null,
 		clearCanvas = true,
+		assetGroups: string[] | undefined = undefined,
```
carmenfan: I guess you had to type define it as `undefined` because you're using `undefined`?

Could we use null instead?


```suggestion
		assetGroups: string[] = null,
```

frontend/src/globals/unity-util.ts
```diff
@@ -1813,6 +1823,11 @@ export class UnityUtil {
 		if (initView) {
 			params.initView = initView;
 		}
+
+		if (assetGroups && assetGroups.length) {
```
carmenfan: This is super minor but since i'm bouncing it back... 🙈 
```suggestion
		if (assetGroups?.length) {
```

carmenfan (changes requested):

carmenfan (approved):

PR #6017, opened 2026-05-18
Issue 6008 - update clash plan schema to take in ticket configuration

### LINKED ISSUE #6008: Add tickets object to clash plan schema
### Description

We want to extend the clash plan schema to include information about tickets, should users wish to generate custom tickets as output.

The schema is specified within the parent issue, https://github.com/3drepo/3drepo.io/issues/5796

### Goals
- As a clash manager I want to manage the clashes in 3DR custom tickets
  - As a clash manager I want to specify which federation the custom tickets should be managed within
  - As a clash manager I want to specify which user should be automatically assigned the ticket when it is created
  - As a clash manager I want to specify which user should the tickets be created/updated under

### Tasks
- [x] Update clash plan schema with the additional details
  - [x] update validation and tests
- [x] check why validation seems to assume required fields regardless of whether it is an update or new plan.
- [x] move clash plan schema to middleware
- [x] is PlanId stored as string? No
- [x] update e2e tests
  - [x] Ensure update tests matches patch  behaviour
  - [x] ensure to test updating parts of the ticket object
  - [x] test with null to remove
- [x] On get plans (both for routes and when we trigger a run) - filter out unnecessary data

#### cleanup/fixing previous tasks
- [x] plans should be stored at project level (it's not referencing with project...)
- [x] Check update - does it accept nulls?
- [x] generateBasicData clean up on clashes e2e test

carmenfan:
This fixes #6008

This PR depends on #5913 

#### Description
- Updated clash plan schema
  - Refactored clash plan schema into clash middleware (it's not used anywhere else and it is tightly knitted with middleware behaviour
  - Added ticket config to the schema 
- Changed update plan behaviour
  - We defined it as a `PATCH` endpoint but it was running on `PUT` behaviour. It should now allow the user to only specify the data it wishes to update instead of the whole config.
  - Setting something null will also remove the field.
- Adding project ID context into plans
  - Plans belong to projects; we should not allow the user to trigger/alter a plan within the same teamspace if they have privileges within a different project.

#### Acceptance Criteria
- You can now create a plan with ticket specifications
  - creator specified must have commenter+ privilege within the specified federation
  - Specified federation must reside in the same project.
  - You can specify the list data to pre-populate when the ticket is created
- You can now update a plan by only specifying parts that you wish to update
  - You can specify `null` to remove optional fields (e.g. tickets)
    - You should not be able to remove required fields


backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/clashes.js
```diff
@@ -15,30 +15,213 @@
...
+		if (!id) {
+			return true;
+		}
+		const fn = isFederation ? getFederationById : getContainerById;
+
+		const modelExists = await fn(teamspace, id, { _id: 1 }).catch(() => false);
+		return modelExists ? modelsExistInProject(teamspace, project, [id]) : false;
+	};
+
+	const imposeCondition = (schema, required, canRemove) => {
```
ChristosTsiotsias, line 54: arent these 2 mutually exclusive ? can we not just have one param ?

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/clashes.js
```diff
@@ -15,30 +15,213 @@
...
+
+const generatePlanSchema = (teamspace, project, user, isUpdate) => {
+	const modelExistsTest = (isFederation) => async (id) => {
+		if (!id) {
+			return true;
+		}
+		const fn = isFederation ? getFederationById : getContainerById;
+
+		const modelExists = await fn(teamspace, id, { _id: 1 }).catch(() => false);
+		return modelExists ? modelsExistInProject(teamspace, project, [id]) : false;
```
ChristosTsiotsias, line 51: could we not just do modelsExistInProject?

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/clashes.js
```diff
@@ -15,30 +15,213 @@
...
+				teamspace, project, req.body.tickets, req.planData?.tickets);
+		}
 
 		if (req.planData) {
-			const existingPlanData = pick(req.planData, fieldsToCompare);
-			const incomingPlanData = pick(req.body, fieldsToCompare);
+			Object.keys(req.body).forEach((key) => {
+				const bodyVal = req.body[key];
+				if (isObject(bodyVal)) {
+					const combinedObj = { ...req.planData[key], ...bodyVal };
```
ChristosTsiotsias, line 210: why dont we do 

`if (isEqual(bodyVal, req.planData[key])) {`

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/clashes.js
```diff
@@ -15,30 +15,213 @@
...
+					if (isEqual({ data: combinedObj }, { data: req.planData[key] })) {
+						delete req.body[key];
+					}
+				} else if (isArray(bodyVal)) {
+					if (getCommonElements(req.planData[key], bodyVal).length === req.planData[key].length) {
```
ChristosTsiotsias: or `if(uniqueElements(req.planData[key], bodyVal).length)`

backend/tests/v5/e2e/routes/teamspaces/projects/clashes.test.js
```diff
@@ -33,8 +33,11 @@ const { getFileAsStream } = require(`${src}/services/filesManager`);
 const { getPlanById } = require(`${src}/models/clashes.plans`);
 const { stringToUUID } = require(`${src}/utils/helper/uuids`);
 const { templates } = require(`${src}/utils/responseCodes`);
+const { statuses: defaultStatuses } = require(`${src}/schemas/tickets/templates.constants`);
 const fs = require('fs');
 const path = require('path');
+const { UUIDToString } = require('../../../../../../src/v5/utils/helper/uuids');
```
ChristosTsiotsias: `${src}/

backend/src/v5/models/clashes.plans.js
```diff
@@ -32,23 +33,45 @@ const getPlanByQuery = async (teamspace, query, projection) => {
...
 
-ClashPlans.updatePlan = async (teamspace, planId, data, user) => {
-	await db.updateOne(teamspace, CLASH_PLANS_COL, { _id: planId },
-		{ $set: { ...data, updatedAt: new Date(), updatedBy: user } });
+ClashPlans.updatePlan = async (teamspace, project, planId, data, user) => {
+	const toSet = { updatedAt: new Date(), updatedBy: user };
+	const toUnset = {};
+	const collectUpdates = (searchObj, prefix = '') => {
+		Object.keys(searchObj).forEach((key) => {
+			if (isObject(searchObj[key]) && !isUUID(searchObj[key])) {
```
ChristosTsiotsias: wouldnt it be better to do Object.entries ?

ChristosTsiotsias (changes requested):

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/clashes.js
carmenfan: They're not necessarily mutually exclusive. The 2 fields have different purposes: 

`Required` is to ensure that the value exists when you create a clash run, and `canRemove` imposes nullable when you update a clash. Whilst required parameters can never be removed, an optional field could be un-nullable (like an immutable field)

I'd prefer to keep them separate so it's considered separately.

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/clashes.js
carmenfan: potentially, but this eliminates any corrupted data going through (and we 100% have orphaned models swimming about)

carmenfan: But also `modelExistsInProject` doesn't guarantee it's a federation/container so we needed the previous check to make sure we're not plugging in the wrong model type.

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/clashes.js
carmenfan: because it wouldn't be reliable on an update. Suppose that i have an existing ticket object with details, and I only want to change the template I'm using, then my `bodyVal.ticket` would only have `template` whereas `req.planData.ticket` has the whole shebang.

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/clashes.js
carmenfan: unique calls lodash's `uniq` function, which only takes in a single array.... do you mean something else here?

If we use `getArrayDifference`, it only returns the elements in 2nd array thatares not in the first, so it doesn't give you the whole picture.... 🤔

backend/src/v5/models/clashes.plans.js
carmenfan: probably neither here nor there, but sure!

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/clashes.js
ChristosTsiotsias: ok makes sense, you then could maybe just do 

`isEqual(combinedObj, req.planData[key])`

ChristosTsiotsias (approved):

PR #6027, opened 2026-05-20
Issue 6024 - `hiddenOnUI` support on custom ticket properties

### LINKED ISSUE #6024: Add `hiddenOnUI` ticket property config
### Description
  Product issue: 3drepo/3D-Repo-Product-Team#639
  Parent issue: #5796

  Introduce a new ticket property config option called `hiddenOnUI`.

  This should allow a property to be stored in ticket data, validated as normal, returned where required by the API, and used by backend logic, while making it clear to the frontend that the property should not be displayed in the
  ticket UI.

  This is needed for internal traceability fields such as cloud clash `Clash Plan ID` and `Clash Run ID`, which are useful to the system but not meaningful to users reviewing clash tickets (used in #6018)

  ### Goals
  - [ ] As a developer maintaining ticket templates, I want a reusable `hiddenOnUI` property config so backend/internal fields can be stored without being rendered in the UI.
  - [ ] As a user reviewing tickets, I do not want internal system fields displayed in the UI when they are not meaningful during normal review.
  - [ ] As a developer using ticket data, I want hidden UI fields to remain stored and available to backend/API logic.

  ### Tasks

<details><summary>Tasks suggested by genAI (for reference only)</summary>
<p>

  - [ ] Add `hiddenOnUI` to supported ticket property config options in `src/v5/schemas/tickets/templates.constants.js`.
  - [ ] Update ticket template validation so `hiddenOnUI` is accepted as a boolean property config.
  - [ ] Update ticket template serialisation/API output as needed so frontend can identify hidden UI fields.
  - [ ] Update frontend ticket rendering logic so properties with `hiddenOnUI` are not displayed in the ticket UI.
  - [ ] Ensure `hiddenOnUI` does not prevent ticket data validation, storage, import, export, or backend access.
  - [ ] Add/update backend unit tests for template constants and template validation.
  - [ ] Add/update frontend tests for hiding `hiddenOnUI` properties.

</p>
</details>

carmenfan:
This fixes #6024

#### Description
- backend:
  - Added `hiddenOnUI` configuration in the template schema
  - Refactored some tests to be more consistent to how we deal with test data creation
- frontend:
  - Filter out all `hiddenOnUI` properties when we process templates

#### Acceptance Criteria
- You can create a default property/ module property with `hiddenOnUI` set to true.
  - If it's set to true, it will not be visible on the 3DRepo UI.
    - If it is the case that this property is a required value, the frontend will not be able to validate this, but the backend will return with a relevant error.


sanmont3drepo (approved):
LGTM from the frontend side! 👍

ChristosTsiotsias (approved):

PR #6062, opened 2026-06-10
ISSUE #6061 adjust scripts to check clash runs

### LINKED ISSUE #6061: Update zombie processing and resetProcessingFlag scripts to check for clash runs.
• ### Description

  The existing zombie processing utilities detect and reset zombified model and drawing uploads, but clash runs can also get stuck in processing states. Clash runs should be included in the same zombie detection/reset flow.

  The zombie processing email template also needs correcting so detected zombie entries are grouped and formatted clearly by type.

  ### Goals
  - [ ] As a system admin, I can detect zombified clash runs alongside zombified models and drawings.
  - [ ] As a system admin, I can reset zombified clash runs so they are cancelled cleanly instead of remaining stuck.
  - [ ] As a user reviewing processing alerts, I can see zombie models, drawings, and clash runs grouped separately in the notification email.
  - [ ] As a user reviewing processing alerts, I only see sections in the email for zombie entry types that actually exist.

  ### Tasks
  - [ ] Update `detectZombieProcessing` to detect clash runs in non-terminal statuses.
  - [ ] Return/process zombie entries grouped by type: models, drawings, and clash runs.
  - [ ] Update `resetProcessingFlags` to reset zombie clash runs to `aborted` with error reason `cancelled manually`.
  - [ ] Update clash result processing to no-op when the clash run is already `aborted`.
  - [ ] Update the zombie processing email template data shape to accept grouped `zombieEntries`.
  - [ ] Update the zombie processing HTML email template to render one table per zombie entry type.
  - [ ] Do not render table markup for empty zombie entry groups.
  - [ ] Add/update script tests for detecting and resetting zombie clash runs.
  - [ ] Add/update clash processor tests for ignored aborted runs.
  - [ ] Add/update mailer template tests for grouped zombie tables and omitted empty sections.

  ### Related Resources
  Split out from the clash run abort/superseded work in #6048.

carmenfan:
This fixes #6061 

 #### Description
  - Added clash runs to zombie processing detection.
  - Updated `resetProcessingFlags` so it can reset models, drawings, or clash runs via `--type models|drawings|clashes`.
  - Replaced the script’s model-specific argument with a generic `--id` argument for targeting a specific item.
  - Reset clash runs are marked as `aborted` with error reason `Cancelled manually`.
  - Updated clash result processing to ignore runs that have already been aborted.
    - This used to be impossible until this script has been updated.
  - Fixed the zombie processing email template to render grouped tables for models, drawings, and clash runs.
  - Empty zombie groups are not rendered in the email.

  #### Acceptance Criteria
  - [x] As a system admin, I can detect zombified clash runs alongside zombified models and drawings.
  - [x] As a system admin, I can reset processing flags for only models, only drawings, or only clash runs.
  - [x] As a system admin, I can target a specific item by id when resetting processing flags.
  - [x] As a system admin, I can reset a zombified clash run so it is marked as aborted with reason `Cancelled manually`.
  - [x] As the model processing system, I do not process clash results for runs that have already been aborted.
  - [x] As a user receiving zombie processing alerts, I can see zombie models, drawings, and clash runs grouped separately.
  - [x] As a user receiving zombie processing alerts, I only see email tables for zombie entry types that have results.


backend/src/scripts/utility/modelProcessing/detectZombieProcessing.js
```diff
@@ -62,15 +78,17 @@ const run = async (teamspace, limit, notify) => {
 	}
 
 	const teamspaces = teamspace ? [teamspace] : await getTeamspaceList();
-	const results = (await Promise.all(teamspaces.map((ts) => processTeamspace(ts)))).flat();
+	const results = { models: [], drawings: [], clashRuns: [] };
+	await Promise.all(teamspaces.map((ts) => processTeamspace(ts, results)));
+	const totalResults = Object.values(results).reduce((sum, entries) => sum + entries.length, 0);
```
ChristosTsiotsias: I think Object.values(results).flat().length would be simpler

backend/src/scripts/utility/modelProcessing/detectZombieProcessing.js
```diff
@@ -29,29 +28,46 @@ const { getTeamspaceList } = require('../../utils');
...
 
-const processTeamspace = async (teamspace) => {
+const processTeamspace = async (teamspace, results) => {
 	const expiredTimestamp = new Date(new Date() - TIME_LIMIT);
 	const zombieQuery = {
 		status: { $exists: true, $not: { $regex: `(${processStatuses.OK})|(${processStatuses.FAILED})` } },
 		timestamp: { $lt: expiredTimestamp },
 	};
+	const zombieClashRunQuery = {
+		status: { $exists: true, $nin: [clashRunStatus.COMPLETED, clashRunStatus.FAILED, clashRunStatus.ABORTED] },
```
ChristosTsiotsias, line 46: why not check if its planned or processing ?

backend/src/scripts/utility/modelProcessing/resetProcessingFlags.js
```diff
@@ -46,46 +58,80 @@ const processTeamspace = async (teamspace, model) => {
...
+			],
+		},
+	});
 
 	const modelAction = { $unset: { status: 1 } };
 	const drawingAction = { $set: { status: processStatuses.FAILED } };
+	const clashRunAction = {
+		$set: {
+			status: clashRunStatus.ABORTED,
+			results: { error: { reason: MANUAL_CANCELLATION_REASON } },
```
ChristosTsiotsias, line 79: when we manually abort there is no error property we set result: { reason} } ... is this intentional ?

ChristosTsiotsias (changes requested):

backend/src/scripts/utility/modelProcessing/detectZombieProcessing.js
carmenfan: more prone to error, for e.g. you already forgotten about queued 😆 .

the intention is to find anything that is not in a terminated state, so the logical reasoning makes more sense to go for $nin... If in the future we add more interim states they'll be automatically covered.

backend/src/scripts/utility/modelProcessing/detectZombieProcessing.js
carmenfan: It might be more readable, but it'll take up more memory and ops - you're copying the contents of the array of array into a flat list. 

For readability, i'll change it to `const totalResults = results.models.length + results.drawings.length + results.clashRuns.length;`

backend/src/scripts/utility/modelProcessing/resetProcessingFlags.js
carmenfan: nope, it's an oversight. well spotted!

ChristosTsiotsias (approved):

PR #6063, opened 2026-06-11
Issue 6001 - Email notifications on unexpected errors from event driven situations

### LINKED ISSUE #6001: Email notifications on unexpected errors from event driven situations
### Description
There are quite a few [event-driven listeners](https://github.com/3drepo/3drepo.io/tree/master/backend/src/v5/services/eventsListener/components) that process data. If it fails, we would like to know.

We should inspect all the event listeners in the above folder, for any situation where we have a try/catch, we should detemine whether we should send a system email notification so we know about it.

for e.g.[ `queueTasksCompleted`](https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/services/eventsListener/components/modelEvents.js#L62),  we would like to be notified if it failed for reasons other than the model was removed. 

The notification should contain the event which failed, the message details and the error reason (call stack as well if it's available).

### Goal
- [ ] As a 3DR maintainer I would like to know of unexpected errors within the event listener component of the application

### Tasks
 - [x] determine suppression rules for notifications  ~ 0.5 d
 - [ ] create a error notification utility  ~ 1 d
 - [ ] create unit tests for error notification utility  ~ 1 d
 - [ ] update model events  ~ 0.5 d
 - [ ] update unit tests for model events  ~ 0.5 d
 - [ ] update auth events  ~ 0.5 d
 - [ ] update unit tests for auth events  ~ 0.5 d
 - [ ] update user events  ~ 0.5 d
 - [ ] update unit tests for user events  ~ 0.5 d

Bogdan-Crn:
This fixes #6001

#### Description

- Add error notifications for event-driven listeners in ```components```.
- Review existing ```try/catch``` blocks in the listeners and determine which failures are unexpected and should trigger a system email.
- Include the event name, message details, and the error reason in the notification payload, with stack trace information when available.
- Do not notify for expected or handled cases, such as ```queueTasksCompleted``` when the model has already been removed.

#### Acceptance Criteria

- [ ]  Unexpected failures in event listeners trigger a system email notification.
- [ ]  The notification includes the failed event, message details, and error reason.
- [ ]  Stack trace information is included when available.
- [ ]  Expected/known cases that are intentionally handled do not generate notifications.
- [ ]  Existing listener behavior is preserved for successful and expected failure paths.


backend/src/v5/services/eventsListener/components/clashEvents.js
carmenfan, line 1: There's new events here since you did the changes, can you make sure they're also covered

backend/src/v5/services/eventsListener/components/modelEvents.js
```diff
@@ -89,11 +115,24 @@ const revisionAdded = async ({ teamspace, project, model, revId, modelType, cali
...
 		if (modelTypes.DRAWING === modelType) {
-			await createDrawingThumbnail(teamspace, project, model, revId).catch((err) => {
-				// It is not critical error if we failed to create a thumbnail.
-				// So catch the error and proceed
+			try {
+				await createDrawingThumbnail(teamspace, project, model, revId);
+			} catch (err) {
 				logger.logError(`Failed to create thumbnail for drawing ${teamspace}.${model}.${revId}: ${err?.message}`);
-			});
+				await listenerErrorNotification.notifyListenerFailure({
```
carmenfan: leave this one - we don't care enough

backend/src/v5/services/eventsListener/components/modelEvents.js
```diff
@@ -115,59 +154,110 @@ const revisionAdded = async ({ teamspace, project, model, revId, modelType, cali
...
-			if (err.stack) {
-				logger.logError(err.stack);
+					zipPath ? [{ filename: 'logs.zip', path: zipPath }] : undefined,
+				);
+			} catch (err) {
+				logger.logError('Failed to send email for model import failures');
+				if (err.stack) {
+					logger.logError(err.stack);
+				}
+				await listenerErrorNotification.notifyListenerFailure({
```
carmenfan: you can leave this one. this only failed if we sent an email import failure email

If that failed this one probably' won't have much luck sending either...

backend/src/v5/services/eventsListener/components/modelEvents.js
```diff
@@ -115,59 +154,110 @@ const revisionAdded = async ({ teamspace, project, model, revId, modelType, cali
...
+					eventName: events.MODEL_IMPORT_FINISHED,
+					listenerName: 'modelProcessingCompleted.importFailureEmail',
+					component: 'modelEvents',
+					payload: {
```
carmenfan: This applies to all of the calls, but if we have to reconstruct the payload data, you may as well keep it intact in the function arguments and do a deconstruction after

i.e. 
````js
const modelProcessingCompleted = async (payload) => {
 const { teamspace, project, model, revId, user, modelType, data } = payload
}
````

Then you can just use payload here

File: backend/src/v5/services/eventsListener/listenerErrorNotification.js [hunk 4]
```diff
@@ -0,0 +1,161 @@
...
+
+const shouldSuppressListenerError = (error) => {
+	const { status, message } = normaliseError(error);
+	const safeMessage = `${message ?? ''}`;
+	const hasMissingContext = ENTITY_PATTERN.test(safeMessage) && MISSING_PATTERN.test(safeMessage);
+
+	return status === 404 || hasMissingContext;
+};
+
+ListenerFailureNotifier.notifyListenerFailure = async ({
```
carmenfan: This seems an absolute overkill.
If the payload should be redacted, then simply don't send it over. The whole point of this is to alert us to look further, so we just need sufficient information to investigate - we don't need everything if they are deemed sensitive.

I'm not sure why you need so many functions to work around the data... This is simpler than you might think.

Just create a new template in the mailer service that accepts the exact parameters you've been passing everywhere, and call `sendSystemEmail` directly. This file probably doesn't even need to exist.

Any parameter validation should be done via Yup within your template specification.

carmenfan (changes requested):

carmenfan (approved):

PR #6073, opened 2026-06-17
ISSUE #6072 - Move IO build and deploy to GitHub Actions

### LINKED ISSUE #6072: Look into deploying 3drepo via github actions
move from azure pipelines to github so everything is in one place (bouncer and io)

nkuberaf:
This fixes #6072

#### Description
Moves the 3drepo.io build and deployment path from Azure Pipelines into GitHub Actions, and removes the Azure pipelines entirely (clean cutover  we can revert or fix-forward if anything misbehaves once it's on `master`).

Long-lived environments keep Azure parity: pushes to `master` and `staging` build and push the Docker image, then deploy the `io` Helm chart with `image.tag` set to the triggering commit SHA. PR environments are driven through the existing ChatOps workflow:
- `/deploy` builds and deploys the PR branch instance.
- `/destroy` removes the PR branch Helm release.

The release name is derived from the branch name and normalised to a DNS-1035-friendly Helm/Kubernetes string (commented inline).

#### Key changes (addressing review feedback)
- **Per-branch deploy config** (`HELM_CHART_VERSION`, `CUSTOM_HELM_OVERRIDE`) moved out of the workflow `env` into `.github/deploy/overrides.env`, which the deploy job sources from the **checked-out branch**. Because `repository_dispatch` always runs the workflow definition from the default branch, this is what lets per-branch overrides actually take effect again. Protected with `merge=ours`.
- **Destroy split out** into its own `.github/workflows/destroy.yml`; `buildAndDeploy.yml` is now deploy-only.
- **Docker build context moved** from `.azure/Docker` to `.github/docker`, and the rest of `.azure/` deleted. `onPRClose.yml` now dispatches the new destroy workflow instead of the Azure pipeline.
- **`peter-evans/slash-command-dispatch` bumped** `v3` -> `v5`.
- **`devops-sanity-check`** updated to read `CUSTOM_HELM_OVERRIDE` from `overrides.env` so the "no APP_ settings merged" guard keeps working after the move.
- Branch-name sanitisation commented; `kubectl` bumped to `v1.34.2`; fixed a malformed Helm deploy step; added a CD usage README.

#### Acceptance Criteria
- Pushes to `master` and `staging` build and push `3drepo.io` images tagged with the release name and commit SHA.
- Pushes to `master` and `staging` deploy the `io` Helm chart using `image.tag` set to the triggering commit SHA.
- Commenting `/deploy` on a PR builds and deploys that PR branch instance, honouring `.github/deploy/overrides.env` from that branch.
- Commenting `/destroy` on a PR (or closing the PR) removes that PR branch Helm release and refuses protected `master`/`staging` releases.
- The `devops-sanity-check` validates `CUSTOM_HELM_OVERRIDE` in `.github/deploy/overrides.env` so `APP_` overrides are not merged accidentally.
- The Azure pipelines (`.azure/`) are removed.

> Note: the ChatOps `/deploy` and `/destroy` paths run the workflow from the default branch (inherent to `repository_dispatch`), so they can only be fully validated once this is merged to `master`.


.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,276 @@
...
+on:
+  push:
+    branches:
+      - master
+      - staging
+  pull_request:
+    branches:
+      - master
+      - staging
+  workflow_dispatch:
```
carmenfan: Why do we want this? (this feels very deja vu)

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,276 @@
+name: Build and Deploy 3drepo.io
+
+on:
+  push:
+    branches:
+      - master
+      - staging
+  pull_request:
```
carmenfan: From what I'm reading in the documentation, a pull request trigger would happen on pretty much any aactions around the PR https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request

This would be very wasteful and we'd be triggering a build when someone simply opened or close a PR... We want the azp style of action where someone has to do a command to trigger a deployment.

There's also no reason to restrict it to master/staging. If I'm doing sub feature and merging it into a feature branch, I should be able to do a deployment if I want to.

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,276 @@
...
+concurrency:
+  group: build-deploy-${{ github.event_name == 'pull_request' && github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' && github.ref != 'refs/heads/staging' }}
+
+env:
+  DOCKER_BUILDKIT: 1
+  IMAGE_REPOSITORY: 3drepo.io
+  DOCKERFILE_PATH: .azure/Docker/Dockerfile
+  HELM_CHART_VERSION: '5.23.0'
+  CUSTOM_HELM_OVERRIDE: config.ADDITIONAL_CONFIG=hello
```
carmenfan: you will also have to make adjustments to this part of the testing script https://github.com/3drepo/3drepo.io/blob/master/.github/workflows/automaticTesting.yml#L209

To make sure we don't merge in custom overrides

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,276 @@
...
+jobs:
+  prepare:
+    name: Prepare build metadata
+    runs-on: ubuntu-24.04
+    outputs:
+      branch_name: ${{ steps.metadata.outputs.branch_name }}
+      checkout_ref: ${{ steps.metadata.outputs.checkout_ref }}
+      commit_sha: ${{ steps.metadata.outputs.commit_sha }}
+      branch_tag: ${{ steps.metadata.outputs.branch_tag }}
+      should_push: ${{ steps.metadata.outputs.should_push }}
```
carmenfan: In what situation woudl this script be triggered and you don't want it to push or deploy?

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,276 @@
...
+            commit_sha="$GIT_SHA"
+            branch_tag="$PR_HEAD_REF"
+          else
+            source_branch="$REF_NAME"
+            checkout_ref="$GIT_SHA"
+            commit_sha="$GIT_SHA"
+            branch_tag="$REF_NAME"
+          fi
+
+          branch_name="$(
```
carmenfan: I am guessing this is sanitising the branch name to be a k8s friendly string, it would be good to make a comment here as it's not obvious.

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,276 @@
...
+            printf '%s' "$source_branch" |
+            tr '[:upper:]' '[:lower:]' |
+            sed -E 's/[^a-z0-9-]+/-/g; s/^-+//; s/-+$//; s/-+/-/g' |
+            cut -c1-53 |
+            sed -E 's/-+$//'
+          )"
+          if [[ -z "$branch_name" ]]; then
```
carmenfan: do we ever expect it to be undefined? (if codex vibe coded this it does have a habit of being too safe and add branches that never gets entered.)

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,276 @@
...
+            branch_name="branch"
+          fi
+
+          should_push=false
```
carmenfan: i'm not sure if any of these booleans should be kept (in my mind, this should be like the azp script and only gets triggered when we want to deploy.)

If it's kept then I think we need some comments to explain what situation do we expect the script to do what, because as it stands it's very difficult to understand.

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,276 @@
...
+          set -euo pipefail
+          docker push "$REGISTRY/${IMAGE_REPOSITORY}:$BRANCH_TAG"
+          docker push "$REGISTRY/${IMAGE_REPOSITORY}:$COMMIT_SHA"
+
+      - name: Log out from Azure Container Registry
+        if: always() && needs.prepare.outputs.should_push == 'true'
+        run: docker logout "${{ secrets.REGISTRY_LOGIN_SERVER }}"
+
+  validate-helm:
+    name: Validate Helm deployment
```
carmenfan: why do we need this step? by the looks of your script it's only done if it's a PR triggered run?

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,276 @@
...
+
+      - name: Set Kubernetes context
+        uses: azure/k8s-set-context@v4
+        with:
+          method: kubeconfig
+          kubeconfig: ${{ secrets.KUBE_CONFIG_STAGING }}
+
+      - name: Add Helm repository
+        run: |
+          helm repo add 3drepo https://3drepo.github.io/DevOps/
```
carmenfan: Why do we need the helm reg. from the devops github? it appears you're doing verything through the acr one?

carmenfan (changes requested):

.github/workflows/buildAndDeploy.yml
nkuberaf: We also have this on azure side i need to see why we actually have it

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,257 @@
...
+concurrency:
+  # Cancel superseded runs for the same operation/ref so repeated /deploy
+  # comments or quick successive staging pushes do not race each other.
+  group: deploy-${{ github.event_name == 'repository_dispatch' && github.event.action || github.ref }}-${{ github.event.client_payload.pull_request.number || github.event.client_payload.github.payload.issue.number || github.ref_name }}
+  cancel-in-progress: true
+
+env:
+  DOCKER_BUILDKIT: 1
+  IMAGE_REPOSITORY: 3drepo.io
+  DOCKERFILE_PATH: .azure/Docker/Dockerfile
```
carmenfan: can we also move the docker file outside of .azure? 

infact do we still need anything in the azure folder or should it be deleted? I'd imagine if it triggers a build too it's going to conflict with what we're doing here.

.github/workflows/chatOperations.yml
```diff
@@ -10,7 +10,10 @@ jobs:
         uses: peter-evans/slash-command-dispatch@v3
```
carmenfan: Can we look into using the latest version fo the script? (keif committed this in ~3 years ago)

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,257 @@
...
+on:
+  # Keep Azure parity: merges to master/staging are pushes and should build
+  # and deploy those long-lived releases automatically.
+  push:
+    branches:
+      - master
+      - staging
+  # ChatOps dispatch comes from .github/workflows/chatOperations.yml when
+  # someone comments /deploy or /destroy on a PR.
+  repository_dispatch:
```
carmenfan, line 8: This is once again going to be problematic when we make updates to the file

From my understanding this would trigger the default branch version of this script.

Meaning if we update the helm version for staging, or add custom queue and add to the custom config here, it won't pick it up.

So more thinking needs to be done here.

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,257 @@
...
+              exit 1
+            fi
+
+            source_branch="$PR_HEAD_REF"
+            checkout_repository="$PR_HEAD_REPO"
+            checkout_ref="$PR_HEAD_SHA"
+            commit_sha="$PR_HEAD_SHA"
+          fi
+
+          release_name="$(
```
carmenfan: not addressed:

Image

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,257 @@
...
+            -f "$RUNNER_TEMP/values.yaml" \
+            "${helm_set_args[@]}"
+
+      - name: Log out from registries
+        if: always()
+        run: |
+          docker logout "${{ secrets.REGISTRY_LOGIN_SERVER }}" || true
+          helm registry logout "${{ secrets.REGISTRY_LOGIN_SERVER }}" || true
+
+  destroy:
```
carmenfan: Considering the script is called `buildAndDeploy`, having a destroy script here doesn't make a lot of sense.

carmenfan (changes requested):

.github/workflows/buildAndDeploy.yml
nkuberaf: .azure removed, docker folder moved over .github/docker which is then referenced in deploy workflow

.github/workflows/buildAndDeploy.yml
nkuberaf: comment added

.github/workflows/buildAndDeploy.yml
nkuberaf: Split the destroy job out into its own .github/workflows/destroy.yml  (triggered by the destroy-command dispatch). buildAndDeploy.yml is now deploy-only,  with its trigger and concurrency narrowed accordingly.

.github/workflows/buildAndDeploy.yml
nkuberaf: Per-branch config is done via a sourced overrides.env which is in the repo, gets checkd out with the target branch and is sourced at deploy time. This will give us Azure's per-branch override behaviour.

.github/workflows/destroy.yml
```diff
@@ -0,0 +1,87 @@
+name: Destroy 3drepo.io instance
+
+on:
+  # Dispatched by .github/workflows/chatOperations.yml when someone comments
+  # /destroy on a PR, and by .github/workflows/onPRClose.yml when a PR is
+  # closed so the ephemeral branch deployment is torn down.
+  repository_dispatch:
+    types:
+      - destroy-command
+
```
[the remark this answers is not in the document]
nkuberaf: peter-evans/slash-command-dispatch defaults event-type-suffix to -command, so /deploy and /destroy dispatch as deploy-command / destroy-command will work

.github/workflows/automaticTesting.yml
```diff
@@ -210,13 +210,27 @@ jobs:
...
-        uses: mikefarah/yq@master
-        with:
-          cmd: if ( $(yq eval '.variables.customHelmOverride | ( contains("APP_") )' .azure/build-and-deploy.yaml) === "true" ); then echo $(yq eval '.variables.customHelmOverride' .azure/build-and-deploy.yaml); exit 1;  fi
+        # Per-branch helm overrides now live in .github/deploy/overrides.env
+        # (sourced by buildAndDeploy.yml). Fail the merge if CUSTOM_HELM_OVERRIDE
+        # still carries a branch-specific APP_ setting that shouldn't reach
+        # master/staging.
+        run: |
+          set -euo pipefail
+          # Treat a missing file or absent/commented key as "no overrides":
```
carmenfan: in what situation do you expect this file to be missing?

Considering it also records the helm version we need - surely it'll always be here?

I think it's better to not see absence as no overrides, if someone accidentally deleted this file, it would causes the deployment to fail thus we want to flag that up.

carmenfan: Can you take a look through all the scripts added in this PR and cut out any excessive comments? It's a lot of text that is irrelevant, and we're just bloating the file.

e.g., we don't really need an explanation of why we're doing a checkout or where the files used to be. Comments should help understand what it is trying to do, not explain the historical baggage.

I don't think this section needs any comments. The name of the action already clearly identifies what we're checking here.

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,254 @@
...
+
+env:
+  DOCKER_BUILDKIT: 1
+  IMAGE_REPOSITORY: 3drepo.io
+  DOCKERFILE_PATH: docker/3drepo.io/Dockerfile
+  # Per-branch overridable values (chart version, custom Helm overrides) are NOT
+  # set here. They live in .github/deploy/overrides.env and are sourced from the
+  # deployed branch's checkout below, because repository_dispatch always runs
+  # this workflow definition from the default branch.
+  DEPLOY_OVERRIDES_FILE: .github/deploy/overrides.env
```
carmenfan: probably better to rename this to config.env or something
This file depicts the deployment configurations. We don't really want to define helm chart value here as a fallback

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,254 @@
...
+      # Sourcing is safe here: forked PRs are refused in metadata and only
+      # write-access users can /deploy, so this only runs trusted branches.
+      - name: Load deploy overrides
+        run: |
+          set -euo pipefail
+
+          HELM_CHART_VERSION=5.23.0
+          CUSTOM_HELM_OVERRIDE=
+
+          if [[ -f "$DEPLOY_OVERRIDES_FILE" ]]; then
```
carmenfan: we should assume it is always present, if it's not present we should fail the pipeline

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,254 @@
...
+      # Read per-branch deploy values from the checked-out branch (see
+      # DEPLOY_OVERRIDES_FILE comment above). Defaults are applied first so a
+      # branch without the file, or with only some keys, still deploys.
+      # Sourcing is safe here: forked PRs are refused in metadata and only
+      # write-access users can /deploy, so this only runs trusted branches.
+      - name: Load deploy overrides
+        run: |
+          set -euo pipefail
+
+          HELM_CHART_VERSION=5.23.0
```
carmenfan: I don't think we want to have a fall back helm chart version. we want to fail if we can't find it as we woulnd't expect a future build to work with an existing helm chart (at least not reliably)

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,254 @@
...
+          CUSTOM_HELM_OVERRIDE=
```
carmenfan: is there any point in declaring a variable with nothing?

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,254 @@
...
+
+          if [[ -f "$DEPLOY_OVERRIDES_FILE" ]]; then
+            set -a
+            # shellcheck disable=SC1090
```
carmenfan: can you go through these files and remove comments?

Assuem whoever is reading this is capable of reading code - you should only need comments where it's not obvious (or it's out of ordinary) The verbosity of comments is too high, it actually makes this harder to read.

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,254 @@
...
+            source "$DEPLOY_OVERRIDES_FILE"
+            set +a
+          else
+            echo "No $DEPLOY_OVERRIDES_FILE on this branch; using defaults."
```
carmenfan: as mentioned above, fail here if we can't find the file.

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,254 @@
...
+        uses: azure/k8s-set-context@v4
+        with:
+          method: kubeconfig
+          kubeconfig: ${{ secrets.KUBE_CONFIG_STAGING }}
+
+      # Reuse the same default values source that the Azure pipeline used,
+      # then override image.tag, branchName, and any temporary custom config.
+      - name: Get default Helm values file
+        env:
+          TESTS_USER: ${{ secrets.TESTS_USER }}
```
carmenfan: you should be able to use secrets directly in the commands - the env vars are not required

`       curl -fsS -u "${{ secrets.TESTS_USER }}:${{ secrets.TESTS_PASSWORD }}" "${{ secrets.HELM_DEFAULTS_URL }}" \
      --output "$RUNNER_TEMP/values.yaml"`

.github/workflows/buildAndDeploy.yml
```diff
@@ -0,0 +1,254 @@
...
+      # immutable deploys. Helm uses the commit SHA tag below.
+      - name: Build Docker image
+        env:
+          REGISTRY: ${{ secrets.REGISTRY_LOGIN_SERVER }}
+          IMAGE_BRANCH_TAG: ${{ needs.metadata.outputs.image_branch_tag }}
+          COMMIT_SHA: ${{ needs.metadata.outputs.commit_sha }}
+        run: |
+          set -euo pipefail
+          docker build \
+            -t "$REGISTRY/${IMAGE_REPOSITORY}:$IMAGE_BRANCH_TAG" \
```
carmenfan: similar to the job above, you could just reference REGISTRY_LOGIN_SERVER directly instead of putting it into an env var?

.github/workflows/destroy.yml
```diff
@@ -0,0 +1,87 @@
...
+jobs:
+  destroy:
+    name: Destroy PR instance
+    if: github.event.client_payload.pull_request != null || github.event.client_payload.branch != null
+    runs-on: ubuntu-24.04
+    steps:
+      # Derive the release name the same way buildAndDeploy.yml does so /destroy
+      # targets the release that /deploy created. The branch comes from the
+      # ChatOps PR payload, or from the `branch` field sent by onPRClose.yml.
+      - name: Resolve release name
```
carmenfan: it would be good to put common jobs into a sub workflow that is referenced by both script so we avoid duplication

.github/workflows/onPRClose.yml
```diff
@@ -52,18 +52,16 @@ jobs:
...
-      id: getChartName
-    - name: Azure Pipelines Action
-      if: ${{ (steps.getChartName.outputs.chartName != 'staging') && (steps.getChartName.outputs.chartName != 'master') }}
-      uses: Azure/pipelines@v1.2
-      with:
-        azure-devops-project-url: https://dev.azure.com/3drepo/3drepo.io
-        azure-pipeline-name: 'destroy'
-        azure-devops-token: ${{ secrets.AZURE_DEVOPS_TOKEN }}
-        azure-pipeline-variables:  '{"branchName": "${{ steps.getChartName.outputs.chartName }}"}'
+      # Hand off to .github/workflows/destroy.yml (same path as a /destroy
```
carmenfan: instead of calling chat operations, wouldn't it make sense to just call the destroy workflow directly? seems a bit convoluted.

docker/3drepo.io/src/getCustomLogins.py
carmenfan, line 1: is this still relevant? I don't think we support custom login pages anymore, can you check if it's still referenced, and evaluate?

carmenfan (changes requested):

docker/3drepo.io/src/getCustomLogins.py
nkuberaf: Couldnt find any reference to it. Removed

.github/workflows/buildAndDeploy.yml
nkuberaf: renamed

.github/workflows/destroy.yml
nkuberaf: new script added in derive-release-name and is referenced in the deployment yaml

.github/workflows/onPRClose.yml
nkuberaf: yeah probably cleaner ive referenced that directly now.

.github/workflows/buildAndDeploy.yml
nkuberaf: directly references the server now

.github/workflows/buildAndDeploy.yml
nkuberaf: Yeah removed

.github/workflows/buildAndDeploy.yml
nkuberaf: CUSTOM_HELM_OVERRIDE= declaration is gone (it defaults to empty at read time only if the key is genuinely absent)

carmenfan (changes requested):
please fix conflict and reflag for review!

.github/actions/derive-release-name/action.yml
```diff
@@ -0,0 +1,34 @@
+name: Derive release name
```
carmenfan: check if we can use a sub workflow or is there differences

.github/actions/derive-release-name/action.yml
carmenfan, line 1: `derive-release-name` should be camel case

carmenfan (changes requested):

.github/actions/derive-release-name/action.yml
nkuberaf: The reusable workflow runs the release-name step as its own separate job (own runner, passed in/out via job inputs/outputs)  the  action ran it inline as a step within the existing job: same logic, but the reusable-workflow version matches our _-prefixed convention at the cost of an extra job per runso ive changed it to this.

.github/actions/derive-release-name/action.yml
nkuberaf: changed to a workflow file

carmenfan (approved):

PR #6104, opened 2026-06-30
ISSUE 6023 - Clash Plan Name placeholder

### LINKED ISSUE #6023: Add clash plan name automated placeholder
### Description
  Product issue: 3drepo/3D-Repo-Product-Team#639
  Parent issue: #5796

  Extend the existing automated property placeholder/value logic introduced in https://github.com/3drepo/3D-Repo-Product-Team/issues/882

  The system already supports automated values such as `{model_name}`, `{template_code}`, and `{ticket_number}` through `supportedPatterns` and the replacement flow in:

  `src/v5/processors/teamspaces/projects/models/commons/tickets.js`

  Add a new supported placeholder for clash plan name, for example:

  `{clash_plan_name:id}`

  This placeholder should allow a read-only ticket property to be populated with the clash plan name of the clash plan ID provided, at runtime, when cloud clash tickets are fetched.

  ### Goals
  - [ ] As a user reviewing generated clash tickets, I want to see the clash plan name so that the ticket is understandable without exposing internal IDs.
  - [ ] As a user reviewing generated clash tickets, I want to see the clash plan name updated within the existing clash ticket if the plan name is updated after the tickets have been created.
  - [ ] As a developer creating cloud clash tickets, I want to provide clash plan context at runtime so generated ticket fields can be populated automatically.

  ### Tasks
  - [x] Add a clash plan name placeholder to `supportedPatterns`, e.g. `clash_plan_name`. (0.5 hour)
  - [x] Modify the logic so that ticket properties with this placeholder are initialized on ticket creation (1.5 days)
  - [x] Modify the logic so that ticket properties are updated upon clash plan update (1.5 days)
  - [x] Add/update unit tests for supported placeholder validation. (1 day)
  - [x] Add/update unit tests for automated property replacement with clash plan name. (1 day)
  - [x] Add/update integration/e2e coverage for a generated clash ticket using the clash plan name placeholder. (1 day)

ChristosTsiotsias:
This fixes #6023

#### Description

A new placeholder is introduced in ticket templates: {clash_plan_name}. This will be replaced by the clash name based on which the ticket is created upon ticket creation. 

Also, when a clash plan name is updated, this should update all tickets related to this clash plan by replacing all references of the placeholder with the new plan name.

To achieve this Clash Plan Name property in Cloud Clash module has a value of {clash_plan_name} instead of being assigned the clash name during ticket creation

#### Test cases
Create some tickets based on a clash run results.
Update the clash plan's name
The tickets' Clash Plan Name property in Cloud Clash module should be updated with the new name,


backend/src/v5/models/tickets.js
```diff
@@ -182,6 +185,9 @@ Tickets.getTicketsByQuery = (teamspace, project, model, query, projection) => Db
 Tickets.getTicketsByTemplateId = (teamspace, templateId, projection) => DbHandler.find(teamspace,
 	TICKETS_COL, { teamspace, type: templateId }, projection);
 
+Tickets.getTicketsByClashPlanId = (teamspace, project, templateId, clashPlanId, projection) => DbHandler.find(teamspace,
```
carmenfan, line 188: this it too highly coupled.

This is a ticket document representation. It should not need to understand clash plans.

I feel like this is a processor level process and we should be using getTicketsByQuery.

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.js
```diff
@@ -439,8 +440,11 @@ const findPropertiesToUpdate = (template, placeholdersToFind) => {
 const updatePropertiesWithAutomatedValue = async (teamspace, tickets, template, propertiesToUpdate) => {
 	const patternToVal = {};
 	const updatedTickets = tickets.map((ticket) => cloneDeep(ticket));
+	const { CLOUD_CLASH } = presetModules;
+	const { [CLOUD_CLASH]: cloudClashProps } = modulePropertyLabels;
 
 	const modelIdToName = {};
+	const clashPlanIdToName = {};
```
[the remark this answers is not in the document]
carmenfan: This is true, you're confusing the 2 approaches that we've discussed

If you're going for the automated field approach (i.e. not a placeholder, but more like a level of risk where we automatically calculate the field, then this is not the place to update it, as this is placeholder-related logic.


So either do a placeholder and update it at run time, or keep it separate (you're targetting a specific field here, i don't feel like your intention have that much in common with the rest of the logic here), I would refactor the logic and just do it in onClashPlanNameUpdated  (or a separate helper function)

backend/src/v5/schemas/tickets/templates.constants.js
```diff
@@ -110,6 +110,13 @@ TemplateConstants.statusTypes = createConstantMapping(['open', 'active', 'review
 
 TemplateConstants.riskLevelsToNum = (value) => riskLevelsArr.indexOf(value);
 
+TemplateConstants.supportedPatterns = createConstantMapping([
+	'model_name',
+	'template_code',
+	'ticket_number',
+	'clash_plan_name',
```
carmenfan, line 117: again, with your approach you're not using placeholder, so don't add it.

backend/src/v5/schemas/tickets/templates.constants.js
```diff
@@ -161,7 +168,7 @@ TemplateConstants.presetModulesProperties = {
 		createPropertyEntry('Clash Plan ID', propTypes.TEXT, { [propOptions.REQUIRED]: true, [propOptions.IMMUTABLE]: true, [propOptions.HIDDEN_ON_UI]: true }),
 		createPropertyEntry('Clash Run ID', propTypes.TEXT, { [propOptions.REQUIRED]: true, [propOptions.IMMUTABLE]: true, [propOptions.HIDDEN_ON_UI]: true }),
 		createPropertyEntry('Clash ID', propTypes.LONG_TEXT, { [propOptions.REQUIRED]: true, [propOptions.IMMUTABLE]: true, [propOptions.HIDDEN_ON_UI]: true }),
-		createPropertyEntry('Clash Plan Name', propTypes.TEXT, { [propOptions.REQUIRED]: true, [propOptions.IMMUTABLE]: true, [propOptions.READ_ONLY_ON_UI]: true }),
+		createPropertyEntry('Clash Plan Name', propTypes.TEXT, { [propOptions.REQUIRED]: true, [propOptions.IMMUTABLE]: true, [propOptions.READ_ONLY_ON_UI]: true, value: `{${TemplateConstants.supportedPatterns.CLASH_PLAN_NAME}}` }),
```
carmenfan, line 171: I think this is a readOnly with no default.

carmenfan (changes requested):

backend/src/v5/models/tickets.js
```diff
@@ -122,7 +119,7 @@ Tickets.updateTickets = async (teamspace, project, model, oldTickets, data, auth
 			});
 			return {
 				updateOne: {
-					filter: { _id: oldTicket._id, teamspace, project, model },
+					filter: { _id: oldTicket._id, teamspace, project, ...(model ? { model } : {}) },
```
carmenfan: `deleteIfUndefined()` will sort you out here?

But also, why would you run into this situation?

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.clashes.js
```diff
@@ -387,13 +387,17 @@ const prepareContext = async (teamspace, project, federation, plan, runId, templ
...
+const processNewClashTickets = async (teamspace, project, federation, template, ticketsToCreate, creator, planName) => {
 	if (ticketsToCreate.length === 0) return;
 
 	// run it through a validation pass to ensure we have all the default values set, and also
 	// we're not creating tickets with invalid data (e.g. if there's a unique/read only property we're setting incorrectly)
 	const validatedTicketsToCreate = (await validateTickets(teamspace, project, federation, template,
-		ticketsToCreate, { author: creator })).map(({ newTicket }) => newTicket);
+		ticketsToCreate, { author: creator })).map(({ newTicket }) => {
+		const ticket = cloneDeep(newTicket);
+		ticket.modules[CLOUD_CLASH][CLASH_PLAN_NAME] = planName;
```
carmenfan: I'm a bit confused why you need to do that when we already add the plan name in processClashes?  The data should already be in the ticket?

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.js
```diff
@@ -517,20 +508,15 @@ const updatePropertiesWithPattern = async (teamspace, project, model, template,
...
-
-	await Promise.all(templates.map(async (template) => {
-		const fullTemplate = generateFullSchema(template);
-		const propertiesToUpdate = findPropertiesToUpdate(fullTemplate, [supportedPatterns.CLASH_PLAN_NAME]);
-
-		if (propertiesToUpdate.length) {
-			const tickets = await getTicketsByClashPlanId(teamspace, project, template._id, planId,
-				{ _id: 1, number: 1, project: 1, model: 1, modules: 1 });
+Tickets.onClashPlanNameUpdated = async (teamspace, project, planId, planName) => {
+	const tickets = await getTicketsByQuery(teamspace, project, undefined,
```
carmenfan: I don't think you really need a read here right?

You can just do a `updateMany` where planId matches. There isn't a lot of reason why you need to do a find to pull it out?

Also I don't think we want to go through the usual update route, have a special method for this which just takes in the query and the update, this way you don't have to add all the fudging around the model field `model`, and we also dont' want to trigger a ticket update event or record an entry for the ticket logging as technically the ticket wasn't updated by anyone or reason.

carmenfan (changes requested):

backend/src/v5/models/tickets.js
carmenfan: ignore this comment - My later comment would imply that this change should be reverted

carmenfan:
@copilot resolve the merge conflicts in this pull request

carmenfan (approved):

PR #6107, opened 2026-07-06
Issue 6105

### LINKED ISSUE #6105: Expose SetSnapMode to UnityUtil
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/1002

The UnityUtil function should look like:

`UnityUtil.setSnapMode(profile: "Off" | "Navigation" | "Drawing");`

The existing enable/disable snapping methods should be left in place.

### Goals

- [x] User can set snapping mode by profile name

### Related Resources

MockUp : `link to mockup`

vmavadiya:
This fixes #6105 

#### Description
-Added setSnapMode method inside UnityUtil class

#### Acceptance Criteria
- User can set Snap mode "Off", "Navigation" or "Drawing"


frontend/src/globals/unity-util.ts
```diff
@@ -1289,6 +1289,15 @@ export class UnityUtil {
 		UnityUtil.toUnity('DisableSnapping', undefined, undefined);
 	}
 
+	/**
+	 * Set the snap mode.
+	 * @param mode - The snap mode to be set like "Off", "Navigation" or "Drawing"
+	 * @category Configurations
+	 */
+	public static setSnapMode(mode) {
+		UnityUtil.toUnity('SetSnapMode', undefined, mode);
```
sebjf: The state should be `UnityUtil.LoadingState.VIEWER_READY` instead of undefined - even though the current if..else block in the handler falls through, this is an implementation detail. Typically, all setup messages like this one should be after VIEWER_READY as that is when the viewer is definitely ready to receive messages! 

Also, we should have have mode declared as a string in TypeScript as we are trying to make unityutil fully typed!

sebjf (changes requested):
Just one change!

frontend/src/globals/unity-util.ts
vmavadiya: Hi @sebjf 

Could we used **enum** or **literal union** too declare mode?

sebjf (approved):

PR #6141, opened 2026-07-15
ISSUE #6137 - Always show an issue's/risk's pin when in the details view

### LINKED ISSUE #6137: Inconsistency when editing an issue with start/end sequence date set, with its pins and shapes.
### Description
There's an inconsistency when editing an issue with start/end sequence date set, with its pins and shapes.
When editing the issue if the sequence is selected , the shapes appear all the time while the pin only when in the window of the start/end date

### Steps to replicate
- Create an issue in a container that has a sequence.
- Select a valid start/end date in the 4d tab in the issue
- Create shapes 
- Drop a pin
- Drag the slider outside the designated start/end date while still editing the issue

### Current Behaviour
- Shapes are shown all the time while the pin dissapears outside the issue's 4d time window

### Expected Behaviour
- Shapes and pin should be shown all the time or shapes and pin should dissapear at the same time outside the start/end date window

( 
comment by @sanmont3drepo : @carmenfan   I think that it should appear all the time otherwise you could edit pins that you cant see:

https://github.com/user-attachments/assets/0729ade9-ea52-42e3-9441-7d24a34d3e1b

)

When editing the issue:

https://github.com/user-attachments/assets/1b8d7cb4-5f8f-4fa9-926c-e5966b3f4bec

When the issue view is the list:

https://github.com/user-attachments/assets/012ceb26-4970-429a-a20d-ea1b87a0a253

Produce issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/1018

The-Daniel:
This fixes #6137 

#### Description
When the user is within details view they may be editing the issue/risk. Therefore it is required that the pin is visible so that they can properly edit it. This means the pin will be visible even if the sequence is outside the range given in the 4D tab, and even if the "show pins" setting in list view is turned off


frontend/src/v4/modules/risks/risks.selectors.ts
```diff
@@ -172,7 +172,7 @@ export const selectPins = createSelector(
 
 		let pinsToShow = [];
 
-		if (showPins) {
+		if (showPins && !showDetails) {
```
sanmont3drepo, line 175: When Im creating an issue a pin appears even though I've never dropped one

https://github.com/user-attachments/assets/d2d78efe-7f36-4f9b-ac9f-9ca4888fe987

sanmont3drepo (changes requested):

sanmont3drepo (approved):

sanmont3drepo (approved):

PR #6143, opened 2026-07-15
ISSUE #6142 added utility script for broken revisions

### LINKED ISSUE #6142: Utility script to fix revisions with overlapping imports
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/549

If a scene collection contains nodes from multiple imports that all have the same rev_id, then that revision is in an indeterminate state and cannot be loaded or re-regenerated.

A number of revisions have ended up in this state, and we need a script to fix them before the migration can continue.

### Goals

- [x] Scene collections for revisions that have had at least once successful import can be recovered.
- [x] Other revisions are unaffected.

### Related Resources

MockUp : `link to mockup`

sebjf:
This fixes #6142

#### Description
A utility script to fix broken revisions caused by repeated imports to the same revision id.

The script operates by iterating over all completed revisions. It identifies the root node for that revision and collects all referenced nodes by shared id. If the collection contains any nodes with that revision id that are not in the hierarchy, they are deleted.

If multiple successful imports have taken place to the same revision id, then a revision may have multiple root nodes. In this case, the 'correct' root node is distinguished by the tree, as only one tree can exist at any given time.

The script therefore handles both cases where the multiple imports have taken place, and also streamed imports that were aborted before, or after, a successful import.

The script requires at least one successful import to have run however. It cannot clean up abandoned imports, and neither will it clean up nodes that are not associated with a revision in the history collection. These nodes are not read by bouncer and so do not cause issues for bundle generation.

The script will not run for revisions flagged as incomplete.

#### Acceptance Criteria
- [x] The script does not affect collections that do not have orphaned nodes
- [x] The script cleans up revisions that have had imports aborted halfway through
- [x] The script cleans up revisions that have had multiple successful imports


backend/src/scripts/utility/scene/repairFailedImports.js
```diff
@@ -0,0 +1,235 @@
...
+ * the traditional tree but also material and metadata nodes in one go.
+ *
+ * This script will only clean up the scene collection. It will not clean up
+ * zombie stash files, if bouncer failed during bundle generation. It will also
+ * not delete any ref nodes or their binaries on disk - a separate script must
+ * be used for that.
+ */
+
+const { v5Path } = require('../../../interop');
+const { uuidToString } = require('../../../v4/utils');
```
carmenfan: you already import UUIDToString from v5 at L48, so I think you can use that instead?

backend/src/scripts/utility/scene/repairFailedImports.js
```diff
@@ -0,0 +1,235 @@
...
+const { stringToUUID } = require('../../../v5/utils/helper/uuids');
```
carmenfan: this should go into L48 (import both functions in one line)

backend/src/scripts/utility/scene/repairFailedImports.js
```diff
@@ -0,0 +1,235 @@
...
+	await deleteMany(teamspace, `${collection}.scene`, { rev_id: revision, shared_id: { $in: idsToDelete } });
+};
+
+const processRevision = async (teamspace, collection, revision) => {
+	const rootSharedId = await getRootNodeSharedId(teamspace, collection, revision);
+	await cleanupOrphanedNodesForRevision(teamspace, collection, revision, rootSharedId);
+};
+
+const processContainer = async (teamspace, container) => {
+	const history = await find(
```
carmenfan: It would be better to call [`getRevisionsByQuery`](https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/models/revisions.js#L93) so if there's any changes it will automatically be accomodated by the main source code function and it won't break the script

backend/src/scripts/utility/scene/repairFailedImports.js
```diff
@@ -0,0 +1,235 @@
...
+		if (children) {
+			queue.push(...children);
+		}
+	}
+
+	return { allNodes, referencedIds };
+};
+
+const getRootNodeSharedId = async (teamspace, collection, revision) => {
+	const rootNodes = await find(
```
carmenfan: similar here, would be better to call [`getNodesByQuery`](https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/models/scenes.js#L32)

backend/src/scripts/utility/scene/repairFailedImports.js
```diff
@@ -0,0 +1,235 @@
...
+		.filter((node) => !referencedIds.has(uuidToString(node.shared_id)))
+		.map((node) => node.shared_id);
+
+	if (idsToDelete.length === 0) {
+		return;
+	}
+
+	logger.logInfo(`\t\tRemoving ${idsToDelete.length} orphaned nodes from ${collection}/${UUIDToString(revision)}`);
+
+	await deleteMany(teamspace, `${collection}.scene`, { rev_id: revision, shared_id: { $in: idsToDelete } });
```
carmenfan: How painful would it be to also delete the relevant blob file from scene.ref?

They're essentially orphans, and ideally we'd want to clean up the storage as it could be quite chunky)

backend/src/scripts/utility/scene/repairFailedImports.js
```diff
@@ -0,0 +1,235 @@
...
+const { getTeamspaceList, getCollectionsEndsWith } = require('../../utils');
+
+const Path = require('path');
+
+const { deleteMany, find } = require(`${v5Path}/handler/db`);
+const { getFile } = require(`${v5Path}/services/filesManager`);
+const { UUIDToString } = require(`${v5Path}/utils/helper/uuids`);
+
+const getReferencedIdsFromHierarchy = async (teamspace, collection, revision, rootSharedId) => {
+	const allNodes = await find(
```
carmenfan: `getNodesByQuery` here too

carmenfan (changes requested):
@sebjf, some feedback, though most is for long-term maintenance considerations.

If this is more likely to be a one-off that we won't merge into the main branch, it's good to go, and you can ignore the feedback (logically,y this makes sense to me)

But if we want it to be a permanent member of the utils... we're also going to need some tests added or it won't pass the github actions check for code coverage (sorry 🙈, but Copilot should be fairly good at doing them! )

Image

backend/src/scripts/utility/scene/repairFailedImports.js
```diff
@@ -105,83 +109,94 @@ const getRootNodeSharedId = async (teamspace, collection, revision) => {
...
+		revision,
+		idsToDelete,
+		{ _blobRef: 1 },
+	);
+	// eslint-disable-next-line no-underscore-dangle
+	const blobRefNames = [...new Set(nodesToDelete.filter((n) => n._blobRef).map((n) => n._blobRef.buffer.name))];
 
-	await deleteMany(teamspace, `${collection}.scene`, { rev_id: revision, shared_id: { $in: idsToDelete } });
+	await deleteMany(teamspace, `${container}.scene`, { rev_id: revision, shared_id: { $in: idsToDelete } });
+	await deleteMany(teamspace, `${container}.scene.ref`, { _id: { $in: blobRefNames } });
```
carmenfan: We do have another utility script that would crawl the FS and find orphaned files, so this is technically fine, but it's probably better to call [removeFilesWithMeta](https://github.com/3drepo/3drepo.io/blob/a4008f49d662ffb44ee3142aa65a5d58be6980e7/backend/src/v5/services/filesManager.js#L118) to remove the blob files as well. The params should be identical to the ones you provided here

backend/src/scripts/utility/scene/repairFailedImports.js
```diff
@@ -191,7 +206,7 @@ const processTeamspace = async (teamspace, specificContainer) => {
 		}
 		logger.logInfo(`\t- Checking container ${container}`);
 		// eslint-disable-next-line no-await-in-loop
-		await processContainer(teamspace, container);
+		await processContainer(teamspace, projectLookup[container], container);
```
carmenfan: The script might trip over on the off chance that we have an orphan .scene collection that isn't in a project (they're once again corrupted data from the ancient 3drepo years, I've seen them around before, and since it doesn't bother anyone, I haven't really cleaned them up 🙈 )

It might be safer to add an if statement here and proceed only if projectLookup[container] is not defined, but I'm not sure what the implications are for your main migration script...

backend/tests/v5/scripts/scene/repairFailedImports.test.js
```diff
@@ -0,0 +1,582 @@
...
+ * Sets up test data with multiple containers and scenarios by composing different import states
+ * Creates the same scenarios in both teamspaces to test multi-teamspace behavior
+ * @returns {Object} Test data object with revision objects
+ */
+const setupData = async () => {
+	const teamspace1 = `teamspace_${generateRandomString()}`;
+	const teamspace2 = `teamspace_${generateRandomString()}`;
+
+	const createScenarios = async (teamspace) => {
+		// Scenario 1: Clean container (successful import only)
```
carmenfan, line 341: We have to do data setup a lot for e2e tests, so there are a few helper functions you can use in [helper/services](https://github.com/3drepo/3drepo.io/blob/a4008f49d662ffb44ee3142aa65a5d58be6980e7/backend/tests/v5/helper/services.js#L4) - It might be useful to utilise instead of hand-crafting the data, so if there's any change in anything, it will automatically be patched in the future. Should hopefully also make your life slightly easier, as you don't have to hand-craft the relationships or make up names. 

The data generator function will supply you with UUID strings where appropriate and auto convert them to UUID to commit. 

Some of them might not be useful since your scene construction is in more details than what we deal with typically, but I think you can use some of them to reduce the line count here

|  Data type |  Data generator fn | Write to DB function  | Notes |
|---|---|---|---|
|  Teamspace  | None: we just use `generateRandomString`  |  db.createTeamspace ||
|  Project  |  generateRandomProject | db.createProject  | This takes an array of model ids to setup the relationship|
|  Model |  generateRandomModel | db.createModel  ||
| Revision | generateRevisionEntry | db.createRevision | all the params are optional (only need to use them if you need to tweak something) |
| scene nodes | generateBasicNode, generateMeshNode,  generateTextureNode | db.createScene | generateBasicNode fabricates the basic info for the node and you can add specify parent(s) and any additional data. createScene takes in all the nodes and commits them, also managing the blob ref stuff. |
| JSON files  |  none | db.addJSONFile | |

backend/tests/v5/scripts/scene/repairFailedImports.test.js
```diff
@@ -0,0 +1,582 @@
...
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const { determineTestGroup } = require('../../helper/utils');
+const { db: { reset: resetDB }, generateRandomString, generateRandomBuffer, generateRandomNumber } = require('../../helper/services');
+const { utilScripts, src } = require('../../helper/path');
+
+const stashFiles = new Map(); // Map of "${teamspace}/${collection}/${path}" -> content
+jest.doMock(`${src}/utils/logger`);
+jest.doMock(`${src}/services/filesManager`, () => ({
```
carmenfan: Utility script tests are considered e2e tests, so there is a file storage setup for the tests. You don't actually need to mock this; instead, you can commit the file as normal (it will be stored and retrieved as you expect.

But this is fine too, just an FYI.

backend/tests/v5/scripts/scene/repairFailedImports.test.js
```diff
@@ -0,0 +1,582 @@
...
+			`void_${generateRandomString()}`,
+			[
+				(ts, c, rev) => addSuccessfulImport(ts, c, rev),
+				(ts, c, rev) => addFailedImport(ts, c, rev, 2),
+			],
+			{ void: true },
+		);
+
+		// Create projects for each container
+		await createProject(teamspace, 'Clean Project', [cleanRevision]);
```
carmenfan: If they are not dependent on each other then it'll be quicker to do Promise.all:

```` js
await Promise.all[
  createProject(teamspace, 'Clean Project', [cleanRevision]),
  createProject(teamspace, 'Corrupted Project', [corruptedRevision])  
]
````

Same with the revisions creations above

File: backend/tests/v5/scripts/scene/repairFailedImports.test.js [hunk 12]
```diff
@@ -0,0 +1,582 @@
...
+			mixedCorruptedRevision,
+			incompleteRevision,
+			multiRootRevision,
+			noRootRevision,
+			corruptTreeRevision,
+			voidRevision,
+		};
+	};
+
+	const teamspace1Scenarios = await createScenarios(teamspace1);
```
carmenfan: same here
````js
const [teamspace1Scenarios, teamspace2Scenarios]  = await Promise.all([
  createScenarios(teamspace1),
  createScenarios(teamspace2),
])
````

**Review by carmenfan — CHANGES_REQUESTED** [review]

File: backend/src/scripts/utility/scene/repairFailedImports.js (hunk 8 above)
Comment (sebjf) [reply backend/src/scripts/utility/scene/repairFailedImports.js]: > not sure what the implications are for your main migration script...

I will add a similar check to the the 782 script so it will ignore them too, if it doesn't already!

File: backend/tests/v5/scripts/scene/repairFailedImports.test.js [hunk 13]
```diff
@@ -0,0 +1,582 @@
...
+		};
+		const shuffled = [...sceneNodes].sort(() => Math.random() - 0.5);
+		const nodes = shuffled.slice(0, Math.min(10, shuffled.length));
+		for (const node of nodes) {
+			// eslint-disable-next-line no-underscore-dangle
+			node._blobRef = blobRef;
+		}
+	}
+	return refNodes;
+};
```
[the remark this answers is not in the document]
sebjf: Resolved by using the inbuilt methods

PR #6146, opened 2026-07-16
ISSUE #6134 - Improved usability of dattimepicker

### LINKED ISSUE #6134: Improve UX of the date component
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/657
The date component currently forces the users to specify a date and time, and you can only operate via the calendar UI, which can be annoying if for e.g. you're years off or doesn't actually care about the actual hour/time

For fixing this we are changing the UI showing date and time at first glance to be not force setting the time 

### Goals
- [x] As a user I want to be able to quickly put in the date/time if I am years away from the current date
- [x] As a user I don't want to specify the time if I don't require a specific time/date

### Tasks

- [ ] Task 1
- [ ] Task 2

### Related Resources
Figma design
https://www.figma.com/design/80RzVQHY7yzsRs1e2ArnsQ/Date-Component?node-id=0-1&t=WJwKFjiTFz2J6tQ2-1

Date time flow - mouse
https://www.figma.com/proto/80RzVQHY7yzsRs1e2ArnsQ/Date-Component?node-id=0-1&p=f&viewport=1068%2C281%2C0.05&t=WJwKFjiTFz2J6tQ2-0&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=2%3A20280&show-proto-sidebar=1

Date Time flow - keyboard
https://www.figma.com/proto/80RzVQHY7yzsRs1e2ArnsQ/Date-Component?node-id=0-1&p=f&viewport=1068%2C281%2C0.05&t=WJwKFjiTFz2J6tQ2-0&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=2%3A32038&show-proto-sidebar=1

sanmont3drepo:
This fixes #6134 

#### Description
- Changed wording of some error codes
- Removed unused error codes
- Replaced some error codes that are irrelevant to user (e.g. if bouncer failed to launch, we shouldn't tell the user about it, it should just tell them the process failed)
- Add bouncer error code to email notification for better referencing.

#### Acceptance Criteria
- [x] As a user I want to be able to quickly put in the date/time if I am years away from the current date
- [x] As a user I don't want to specify the time if I don't require a specific time/date

### DIFF
--- a/backend/yarn.lock
+++ b/backend/yarn.lock
@@ -5969,11 +5969,6 @@ node-releases@^2.0.36:
   resolved "https://registry.yarnpkg.com/node-releases/-/node-releases-2.0.46.tgz#d188a129a83f5e03a101aacb58f260f2ee8faaa1"
   integrity sha512-GYVXHE2KnrzAfsAjl4uP++evGFCrAU1jta4ubEjIG7YWt/64Gqv66a30yKwWczVjA6j3bM4nBwH7Pk1JmDHaxQ==
 
-node-releases@^2.0.27:
-  version "2.0.27"
-  resolved "https://registry.yarnpkg.com/node-releases/-/node-releases-2.0.27.tgz#eedca519205cf20f650f61d56b070db111231e4e"
-  integrity sha512-nmh3lCkYZ3grZvqcCH+fjmQ7X+H0OeZgP40OierEaAptX4XofMh5kwNbWh7lBduUzCcV/8kZ+NDLCwm2iorIlA==
-
 nodemailer@9.0.1:
   version "9.0.1"
   resolved "https://registry.yarnpkg.com/nodemailer/-/nodemailer-9.0.1.tgz#871c04d8423fc0c790289f4a8f6f71f1b3563e8c"
--- a/frontend/src/v5/ui/controls/inputs/datePicker/dateTimePicker.component.tsx
+++ b/frontend/src/v5/ui/controls/inputs/datePicker/dateTimePicker.component.tsx
@@ -179,10 +179,8 @@ export const DateTimePicker = ({
 										enableAccessibleFieldDOMStructure={false}
 										format={getLocaleDateFormat()}
 										value={dateValue}
-										onChange={(newValue: Dayjs | null, context: any) => {
-											if (newValue?.isValid() && !context.validationError) {
-												setDateValue(newValue);
-											}
+										onChange={(newValue: Dayjs | null) => {
+											setDateValue(newValue);
 										}}
 										onFocus={() => setView(DatePickerView.calendar)}
 										slotProps={{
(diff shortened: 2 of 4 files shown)


frontend/src/v5/ui/controls/inputs/datePicker/dateTimePicker.component.tsx
The-Daniel, line 1: I seem to remember that the advantage of this datepicker is that you don't have to set the time. Should that mean that the time should start in an unset state? Currently it opens at 23:59, but it is possible to change it to what appears like an unset state of `hh:mm` and the form looks valid. Upon saving the time is set to 23:59.

Image
The time can be set to an errored state by manually deleting either just the hours or just the minutes. There is no error message and the save button is not disabled

Image
Deleting both removes the error. But saving sets the time to 23:59

frontend/src/v5/ui/controls/inputs/datePicker/dateTimePicker.styles.ts
The-Daniel, line 1: It is possible for the dateTimePicker to extend out of the viewport. The user is therefore unable to save or clear the date

Image

frontend/src/v5/ui/controls/inputs/datePicker/dateTimePicker.component.tsx
The-Daniel, line 1: Sometimes after changing the date I am not able to reopen the datepicker. I've only had it happen when changing between set and unset states, so I wonder if it's because the anchor element is changing. Sometimes the datepicker will show up in the very top left of the screen which suggests to me it can't find the anchor element it was provided
https://github.com/user-attachments/assets/8cd74be8-d808-4a24-9630-5cff12038ccc

backend/yarn.lock
The-Daniel, line 1: How do we determine whether to use this new dateTimePicker or the old one?

For a ticket the due date will use the new design, but custom date properties will use the old one
Image

frontend/src/v5/ui/controls/inputs/datePicker/dateTimePicker.component.tsx
The-Daniel: Tested this on staging and it is also happening there, so feel free to spin off into a separate issue

The-Daniel (changes requested):

frontend/src/v5/ui/controls/inputs/datePicker/dateTimePicker.component.tsx
sanmont3drepo: it means that is not compulstory to set the time because you can just pick a date, doesnt mean that there is no time there

backend/yarn.lock
sanmont3drepo: we always use the new one It doesnt make sense to support two versions that provides the same functionality

frontend/src/v5/ui/controls/inputs/datePicker/dateTimePicker.component.tsx
sanmont3drepo: I think I've fixed this, including the reposition of the datepicker.

The-Daniel (approved):

carmenfan (approved):

PR #6164, opened 2026-07-22
ISSUE #6161 - Added safeguard when the group doesnt have shared_ids

### LINKED ISSUE #6161: Old views can mess up selection
### Description
Old views without shared_ids will cause the viewer to not be able to select meshes anymore. This is because in the functions that use hide nodes it expect shared_ids to be an array.  

### Tasks
- [ ] avoid using shared_ids if its not there
- [ ] console log an error about the objects

sanmont3drepo:
This fixes #6161 

#### Description
- If a fetched group doesnt have shared_ids now it assigns an empty array
- If a fetched group doesnt have shared_ids now it outputs an error in the console for debugging

### DIFF
--- a/frontend/src/v4/services/api/groups.ts
+++ b/frontend/src/v4/services/api/groups.ts
@@ -18,6 +18,18 @@
 import { splitValuesIfNecessary } from '../../helpers/requests';
 import { API as api } from './default';
 
+
+const sanitizeGroup = ({objects}) => {
+	objects.forEach(obj => {
+		if (!(obj.shared_ids?.forEach)) {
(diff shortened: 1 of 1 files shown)


carmenfan (changes requested):
We ened to fix the group list API too:

https://issue-6161.dev.3drepo.io/v5/viewer/DevTeamWorkspace/7b6c7bd0-054a-11ee-ac5b-59e952832776/f2d6c937-9998-48f1-9eae-e2de37fa68d6

Image

carmenfan (approved):

Issue #6145, opened 2026-07-16
Custom tickets: New property type `tags`


carmenfan:
### Description

Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/999

Add support for a new custom ticket property type: `tags`. This is similar to `manyOf` — the value is an array of strings — but unlike `manyOf`, no predefined list of allowed values is defined on the template. Each tag string is bounded by the same max length as the `text` property type (120 chars), and the array must contain no duplicates.

Additionally, expose a new endpoint so the UI can retrieve the distinct tag values recorded across all tickets for a given template property, to power a `manyOf`-style filter/selection UI.

#### Endpoint
Route:  `GET /teamspaces/{teamspace}/projects/{project}/{modelType}/{model}/tickets/templates/{template}/properties/{module::propertyName}/values`
description: Get the distinct list of values that are currently used within this template, within the container/federation.
Permissions: viewer+
Response: { values: ["a", "b", "c"]}

---

### Goals

- [ ] As a template author, I can define a property of type `tags` on a custom ticket template without specifying a `values` list, and the template saves and validates correctly.
- [ ] As a template author, I can set a `default` on a `tags` property (an array of strings, each ≤ 120 chars); duplicate entries in the default are silently deduplicated.
- [ ] As a template author, marking a `tags` property as `immutable` is rejected with a validation error, since tags are not immutable by design.
- [ ] As a template author, I can mark a `tags` property as `readOnly`, and its value can only be set via import or system-side logic.
- [ ] As a user creating or updating a ticket, I can set a `tags` property to an array of strings each ≤ 120 characters; any duplicate entries are silently removed before saving.
- [ ] As a user creating or updating a ticket, if a tag string exceeds 120 characters, the request is rejected with a clear validation error.
- [ ] As a user updating a ticket, I can send `null` to clear a non-required `tags` field.
- [ ] As a developer importing tickets in bulk, `tags` properties are validated and processed consistently with other array property types.
- [ ] As a UI developer, I can receive the distinct list of tag strings stored across all tickets belonging to that template, within the container/federation.
- [ ] As a viewer with at least read access to the model, the above endpoint is accessible; unauthenticated or unauthorised requests are rejected.
- [ ] As a consumer of the API, if I call the distinct-values endpoint on a property that is not of type `tags`, the request is rejected with an appropriate error.

---

### Tasks
Developer: please determine the tasks and your estimates.

<details><summary>AI generated tasks for reference</summary>
[...]

PR #6179, opened 2026-07-30
Issue 6148 - Automatically fill in metadata in cloud clash tags upon creation

### LINKED ISSUE #6148: Ability to configure adding metadata to cloud clash tags property
### Description

Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/999
Follows on from #6145 (tags property type support).

Two related changes:

1. Add a `tags` property to the `cloudClash` preset module — editable by users after creation, set by the system at ticket creation time.
2. Extend `valuesAtCreation` in the clash plan schema to support metadata placeholder strings for `tags` properties. A placeholder like `{$meta-IFC::IFCType}` is resolved at ticket-creation time by fetching the corresponding metadata field for each of the two clashing objects; the resolved values are merged into the tags array alongside any static string entries.

**Placeholder resolution rules:**
- Placeholders follow the pattern `{$meta-<field>}`, matching the metadata field reference syntax used in selection rules. If the metadata is within a category, it is prefixed with `<group>::<field>`.
- At creation time, the system fetches the metadata value for object A and object B using their IDs from the clash result.
- If both objects share the same resolved value, only one entry is added (deduplication applies).
- If the metadata field is absent on an object, that object contributes nothing for that placeholder.
- Static strings in the same array are included as-is; the final tags value is the full deduplicated result.
- Placeholders are **not** re-evaluated when clashes are re-processed (existing tickets are not updated).
- Validation of `valuesAtCreation` should accept placeholder strings in the `value` array when the target property is of type `tags`; the values are validated further at processing time.

---

### Goals

- [ ] As a template author, I can add a `tags` property to a ticket template via the `cloudClash` preset module, and it behaves as a standard editable `tags` property.
- [ ] As a clash plan author, I can include a `valuesAtCreation` entry targeting a `tags` property, with a `value` array that mixes static strings and metadata placeholders (e.g. `["apples", "{$meta-IFC::IFCType}"]`).
- [ ] As the system processing new clash tickets, each `{$meta-<group>::<field>}` placeholder in a `tags` value is resolved by looking up that metadata field on object A and object B from the clash result; distinct resolved values are each added as a tag entry.
- [ ] As the system processing new clash tickets, if a metadata field is absent on a clashing object, that object contributes no entry for that placeholder (silent skip).
- [ ] As the system processing new clash tickets, the final `tags` value is the deduplicated union of static strings and all resolved metadata values.
- [ ] As the system re-processing active clashes (existing tickets), the `tags` value is **not** updated — it is set once at creation and left unchanged.
- [ ] As a plan author, if I provide a `valuesAtCreation` entry with placeholder strings targeting a non-`tags` property type, the plan is rejected with a validation error.

---

### Tasks

Developer: please determine the tasks and your estimates.

<details><summary>AI-generated tasks for reference:</summary>

<p>

- [ ] Add `createPropertyEntry('Tags', propTypes.TAGS, ...)` to `[presetModules.CLOUD_CLASH]` in `schemas/tickets/templates.constants.js`
- [ ] Define a placeholder pattern constant (e.g. `META_PLACEHOLDER_REGEX = /^\{\$meta-(.+)::(.+)\}$/`) in an appropriate constants file
- [ ] Update `generatePlanSchema` in `middleware/dataConverter/inputs/teamspaces/projects/clashes.js`:
  - In `ticketSchema.valuesAtCreation`, when the target property type is `tags`, allow array items to be either plain strings or placeholder strings matching the pattern; reject placeholder syntax in non-`tags` entries
- [ ] Implement a helper `resolveTagPlaceholders(teamspace, project, clash, placeholders)` in `processors/teamspaces/projects/models/commons/tickets.clashes.js` (or a shared util):
  - For each placeholder, extract `<group>` and `<field>`
  - Fetch metadata for object A and object B by their IDs (from `clash.a` / `clash.b`)
  - Collect resolved values, deduplicate, skip missing
- [ ] In `processClashes` (`processors/.../tickets.clashes.js`), when building a new ticket:
  - For each `valuesAtCreation` entry targeting a `tags` property, expand any placeholders via the helper before writing to the ticket
  - Static strings pass through unchanged; final array is deduplicated
- [ ] Unit tests: preset module includes `Tags` property; placeholder resolution (both values same → one entry; different → two entries; missing → skip; mixed static + placeholder); non-`tags` property with placeholder rejected at plan validation
- [ ] Unit tests for new `resolveTagPlaceholders` helper
- [ ] E2e / integration test: end-to-end clash processing with a plan that has a `tags` `valuesAtCreation` entry containing placeholders; verify resulting ticket tags

</p> </details>

carmenfan:
This fixes #6148

This depends on #6145 
#### Description
- Added `Tags` as a preset property on the `cloudClash` module.
- Added support for `cloudClash.Tags` values at clash ticket creation.
- Added `$meta` placeholder resolution for `cloudClash.Tags` only.
- Resolved placeholder values from object A and object B metadata during new clash ticket creation.
- Supports internal 3D Repo IDs via scene `shared_id` lookup and external IDs via metadata ID keys.
- Deduplicates final tag values from static entries and resolved metadata.
- Leaves existing clash tickets unchanged when clashes are re-processed.
- Added unit and e2e coverage for tag preset, metadata placeholder resolution, deduplication, missing metadata, and update behavior.

#### Acceptance Criteria
- [x] As a template author, I can add a `tags` property to a ticket template via the `cloudClash` preset module, and it behaves as a standard editable `tags` property.
- [x] As a clash plan author, I can include a `valuesAtCreation` entry targeting `cloudClash.Tags`, with a `value` array that mixes static strings and metadata placeholders.
- [x] As the system processing new clash tickets, each `{$meta-...}` placeholder in `cloudClash.Tags` is resolved by looking up that metadata field on object A and object B from the clash result.
- [x] As the system processing new clash tickets, if a metadata field is absent on a clashing object, that object contributes no entry for that placeholder.
- [x] As the system processing new clash tickets, the final `tags` value is the deduplicated union of static strings and all resolved metadata values.
- [x] As the system re-processing active clashes, the `tags` value is not updated — it is set once at creation and left unchanged.


backend/src/v5/processors/teamspaces/projects/models/commons/tickets.clashes.js
```diff
@@ -245,6 +251,95 @@ const generateBaseNewTicket = async (teamspace, project, federation, template, d
...
+	tagInstruction.value.forEach((value) => {
+		const field = value.match(META_PLACEHOLDER_REGEX)?.[1];
+		if (field) {
+			fields.push(field);
+		} else {
+			staticValues.push(value);
+		}
+	});
+
+	return fields.length ? { staticValues, fields } : undefined;
```
ChristosTsiotsias: Im a bit confused .... do we only want to populate metadata values ? what if there are only static values ?

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.clashes.js
```diff
@@ -245,6 +251,95 @@ const generateBaseNewTicket = async (teamspace, project, federation, template, d
...
 };
 
+const getTagValueInstruction = (valuesAtCreation = []) => {
+	const tagInstruction = valuesAtCreation.find(({ module, property }) => module === CLOUD_CLASH
+		&& property === TAGS);
+
+	if (!tagInstruction) return undefined;
+
+	const staticValues = [];
+	const fields = [];
```
ChristosTsiotsias: I'd rename this to metadataValues

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.clashes.js
```diff
@@ -245,6 +251,95 @@ const generateBaseNewTicket = async (teamspace, project, federation, template, d
...
+					key: { $in: idTypesToKeys[idType] }, value: { $in: ids } } } });
+			});
+
+			if (!queriesOrCondition.length) return [];
+
+			return getMetadataByQuery(teamspace, container,
+				{ rev_id: revision, $or: queriesOrCondition }, { metadata: 1 });
+		},
+	));
+	return metadataEntries.flat().flatMap(({ metadata }) => {
```
ChristosTsiotsias, line 319: do we need both flat and flatMap ?

backend/tests/v5/e2e/services/modelProcessing/clashes.test.js
```diff
@@ -163,21 +166,37 @@ const testParseClashResults = () => {
...
+	const [tagObjectAId, tagObjectBId] = times(2, () => ServiceHelper.generateUUIDString());
 	const planWithPreviousRun = ServiceHelper.generateClashPlan(modelA._id, modelA._id);
 	const planForBouncerError = ServiceHelper.generateClashPlan(modelA._id, modelB._id);
 	const planWithoutPreviousRun = ServiceHelper.generateClashPlan(
 		modelB._id,
 		modelB._id,
 		{ federation: ticketFederation, template: ticketTemplate, creator: user.user },
 	);
-	planWithoutPreviousRun.tickets.template = stringToUUID(planWithoutPreviousRun.tickets.template);
+	const tagPlan = ServiceHelper.generateClashPlan(modelA._id, modelB._id,
```
ChristosTsiotsias, line 183: maybe we could add some static valuesAtCreation to another plan to ensure they work on their own too

ChristosTsiotsias (changes requested):

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.clashes.js
carmenfan: they're already assigned via the construction of `baseNewTicket`. `populateTagPlaceholders` only overwrites the field in the ticket if there are metadata placeholders.

There is a unit test for this and is passing:
image

carmenfan: but since it's confusing, i'll add a comment

backend/tests/v5/e2e/services/modelProcessing/clashes.test.js
carmenfan: theres already a unit test for that. e2e is more just a quick integration test instead of a thorough test for all corner cases so i'd rather we don't add more lines to a already very complicated test case (unless you really think it's necessary!)

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.clashes.js
carmenfan: they're not values though. they're the metadata keys we're looking for. I will rename this to metadataFields

backend/src/v5/processors/teamspaces/projects/models/commons/tickets.clashes.js
carmenfan: they're serving different purposes.

`metadataEntries` is an array of array due to this construction:
image

so the data looks like this: `[[<metadata entries from container 1>], [<metadata entries from container 2]]`

so the first `flat()`is to flatten this into a single array of metadata enttries.

Without this, the `({metadata})` on the line you referenced will not pattern match and we would need another loop inside to loop through the inner array.

ChristosTsiotsias (approved):

PR #6188, opened 2026-08-04
Issue 6159 - Support for tags properties in tickets

### LINKED ISSUE #6159: Tags property support in custom tickets
### Description
Product issue:  https://github.com/3drepo/3D-Repo-Product-Team/issues/999
For a flexible workflow for grouping tickets we want to implement a tag system. This will allow tickets to be grouped  arbitrarily following any user criteria.

### Goals
- As the user I want to be able to group tickets arbitrarily  by adding tags to the ticket
- As the user I want to be able to filter  tickets by tags

<details>

### Endpoints
- `tag` type for templates; a new type for properties called `tag` will be available for templates it will accept an array of strings as payload and return an array when fetching the value

- And endpoitn with al tags. This endpoint is similar to manyOf — the value is an array of strings — but unlike manyOf, no predefined list of allowed values is defined on the template. Each tag string is bounded by the same max length as the text property type (120 chars), and the array wont contain duplicates.

`/teamspaces/{teamspace}/projects/{project}/{modelType}/{model}/tickets/templates/{template}/properties/{module::propertyName}/values`
</details>

### Tasks

- [x] Add tags interface 
- [x] Add support for grouping by tags
- [x] Add support for filtering by tags

### Related Resources

[tagProperty.html](https://github.com/user-attachments/files/30306678/tagProperty.html)

sanmont3drepo:
This fixes #6159

#### Description
- Added support for a new `tags` property type in custom ticket templates: a chip-based free-text input for adding/removing tags on a ticket.
- Added support for grouping and filtering tickets by tags in the tabular view, with tag options fetched from the selected federations/containers.
- Fixed various styling and validation inconsistencies in the tags input (spacing, required-field validation, optional-field handling) so it behaves consistently with other ticket properties.
- Fixed tags rendering, column sizing, and filter option aggregation across models in the tabular view.

#### Acceptance Criteria
- [x] As the user I want to be able to group tickets arbitrarily  by adding tags to the ticket
- [x] As the user I want to be able to filter  tickets by tags


backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/commons/tickets.js
The-Daniel, line 1: When editing an existing ticket it does not let you clear the tags field. This is on non-required tag fields
Image

frontend/src/v5/ui/routes/viewer/tickets/ticketsForm/properties/tagProperty.component.tsx
The-Daniel, line 1: I'm not sure what the use case for an immutable tags property is, so feel free to ignore, but an immutable tags property will become disabled on ticket creation even when the field is empty. oneOf/manyOf remain interactable after ticket creation until a value is set
Image

The-Daniel (changes requested):

frontend/src/v5/store/tickets/tickets.validators.ts
```diff
@@ -77,7 +78,7 @@ const propertyValidator = ({ required, name, type }: PropertyDefinition) => {
 			},
 			{ name }),
 		);
-		if (type === 'manyOf') {
+		if (type === 'manyOf' || type === 'tags') {
```
The-Daniel: I feel like the wording for this error is a bit strange for tags. Perhaps split into two separate errors with different messages or change this one to something that can be shared between the two like "Must have at least one value"
Image

The-Daniel (changes requested):

The-Daniel (dismissed):

carmenfan (dismissed):

PR #6197, opened 2026-08-11
Issue #6124 - Custom ticket pin icons: refactor to backend endpoints

### LINKED ISSUE #6124: Custom ticket pin icons: refactor to backend endpoints
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/404

Currently the set of valid pin icons (`DEFAULT`, `RISK`, `ISSUE`, `MARKER`) is hardcoded
in the backend schema. This task replaces that with a server-managed registry backed by
SVG assets stored in the repository, exposed via three new API endpoints. The ticket
template schema is updated to derive its allowed icon values from the registry.

The design must be extensible: adding a new built-in icon should require no code change
(only new asset files), and the architecture should leave a clear hook for future
customer-supplied custom icons served through the same endpoints.

##### Get list of icons available
Route: `GET /teamspaces/{teamspace}/settings/tickets/pinIcons`
Description: Return all known icon names
Permissions: member of teamspace
response: 
````js
{icons: ["ISSUES", "RISKS", .... //etc]}
````

##### Get standard SVG of pinIcon
Route:  `GET /teamspaces/{teamspace}/settings/tickets/pinIcons/{pinIcon}/normal.svg`
Description: Return the SVG of the normal state of the pin icon
Permissions: member of teamspace
mimeType: `image/svg+xml`
Repsonse: SVG of the icon

##### Get selected SVG of pinIcon
Route:  `GET /teamspaces/{teamspace}/settings/tickets/pinIcons/{pinIcon}/selected.svg`
Description: Return the SVG of the normal state of the pin icon
Permissions: member of teamspace
mimeType: `image/svg+xml`
Repsonse: SVG of the icon

### Goals

- [ ] As a user with access to a teamspace, I can get a listing showing all available pin icon
      names.
- [ ] As a user with access to a teamspace, I can fetch the SVG for the pin icon in its default state.
- [ ] As a user with access to a teamspace, I can fetch the SVG for the pin icon in its selected state
- [ ] Requesting an SVG for an unknown `{pinIcon}` or unknown variant returns `404`.
- [ ] As a teamspace admin, I can use any icon name returned by the list endpoint as
      the `icon` value (static or conditional) on a `coords` property — schema
      validation derives its allowed values from the registry, not a hardcoded array.
- [ ] Existing templates using the current hardcoded icon names (`DEFAULT`, `RISK`,
      `ISSUE`, `MARKER`) continue to validate correctly after migration — those names
      must remain in the built-in registry.

### Tasks

<details><summary>AI generated task list</summary>
<p>
- [ ] Add SVG assets for the four existing icon names (`DEFAULT`, `RISK`, `ISSUE`,
      `MARKER`) in both variants under a new directory, e.g.
      `src/v5/assets/pinIcons/{iconName}/normal.svg` and
      `src/v5/assets/pinIcons/{iconName}/selected.svg`.
- [ ] Create a pin icon registry module (e.g. `src/v5/utils/pinIcons.js`) that:
      - Derives the list of available icons by scanning the assets directory for
        subdirectories containing both `normal.svg` and `selected.svg` — adding a new
        icon requires only dropping asset files, no code change.
      - Exposes a helper to resolve the filesystem path for a given icon + variant.
      - Provides a hook point for merging customer-supplied custom icons in future
        without changing the route or schema layer.
- [ ] Add processor functions (`getPinIconList`, `getPinIconSVG`) in
      `src/v5/processors/teamspaces/settings.js`.
- [ ] Add three routes to `src/v5/routes/teamspaces/settings.js` under
      `hasAccessToTeamspace` middleware, with OpenAPI (`@openapi`) JSDoc:
      - `GET /tickets/pinIcons`
      - `GET /tickets/pinIcons/:pinIcon/normal.svg`
      - `GET /tickets/pinIcons/:pinIcon/selected.svg`
- [ ] Update `pinIconSchema` in `src/v5/schemas/tickets/templates.js` to validate
      against the registry list rather than a hardcoded `oneOf` array.
- [ ] Add/extend unit and e2e tests in `tests/v5/` covering:
      - List endpoint returns all registered icon names.
      - SVG endpoints return correct `Content-Type` and file content.
      - SVG endpoint for unknown icon name / variant returns `404`.
      - Template validation accepts registry icons and rejects unknown names.

</p>
</details>

### PR DESCRIPTION
**Pull request by Bogdan-Crn**
This fixes #6124 

#### Description

- Replaced hardcoded ticket pin icon enum validation with a registry-driven approach backed by SVG assets under backend/resources/pinIcons.
- Added a new pin icon service that:
  - Discovers icon names from file assets using the pattern <ICON_NAME>.<variant>.svg.
  - Supports normal and selected variants.
  - Validates completeness of variant pairs per icon.
  - Returns PIN_ICON_NOT_FOUND for unknown icon/variant or invalid asset layout.
- Added new teamspace settings endpoints (member access):
  - `GET /teamspaces/{teamspace}/settings/tickets/pinIcons`
  - `GET /teamspaces/{teamspace}/settings/tickets/pinIcons/{pinIcon}/{variant}`
- Wired new endpoints through settings route and processor layers and returned SVG responses with mime type image/svg+xml.
- Added a new response code template for pin icon not found (404): `PIN_ICON_NOT_FOUND`.
- Updated ticket template schema to derive allowed icon values dynamically from the pin icon registry rather than `templates.constants` hardcoded values.
- Removed hardcoded pin icon constants from `templates.constants`.
- Added built-in SVG assets for existing icons to preserve compatibility:
  - DEFAULT, RISK, ISSUE, MARKER (normal + selected variants)
- Added test coverage:
  - E2E tests for pin icon list and SVG retrieval endpoints, including auth/permission/error scenarios.
  - Unit tests for pin icon service discovery, validation, and file loading behavior.
  - Unit schema tests confirming direct and conditional icon mapping continue to validate for built-in icons.

#### Acceptance Criteria

- [ ]  As a user with access to a teamspace, I can get a listing showing all available pin icon names.
- [ ]  As a user with access to a teamspace, I can fetch the SVG for the pin icon in its default state.
- [ ]  As a user with access to a teamspace, I can fetch the SVG for the pin icon in its selected state.
- [ ]  Requesting an SVG for an unknown {pinIcon} or unknown variant returns 404.
- [ ]  As a teamspace admin, I can use any icon name returned by the list endpoint as the icon value (static or conditional) on a coords property; schema validation derives allowed values from the registry, not a hardcoded array.
- [ ]  Existing templates using current hardcoded icon names (DEFAULT, RISK, ISSUE, MARKER) continue to validate correctly after migration.

### CODE REVIEW

File: backend/resources/tickets/pinIcons/DEFAULT.normal.svg
Comment (carmenfan) [inline backend/resources/tickets/pinIcons/DEFAULT.normal.svg:1]: can we put these in `backend/resources/tickets/pinIcons`?

So it's very clear these are for tickets not anything else

File: backend/src/v5/routes/teamspaces/settings.js [hunk 1]
```diff
@@ -281,6 +302,76 @@ const establishRoutes = () => {
...
+	*         description: returns the requested pin icon SVG
+	*         content:
+	*           image/svg+xml:
+	*             schema:
+	*               type: string
+	*               format: binary
+	*       404:
+	*         $ref: "#/components/responses/pinIconNotFound"
+	*/
+	router.get('/tickets/pinIcons/:pinIcon/:variant', hasAccessToTeamspace, getPinIcon);
```
carmenfan: There should be a middleware here to validate the `pinIcon` and `variant`

backend/src/v5/processors/teamspaces/settings.js
```diff
@@ -16,6 +16,7 @@
  */
 
 const { addTemplate, getAllTemplates, updateTemplate } = require('../../models/tickets.templates');
+const { getIcon, getIconNames } = require('../../services/pinIcons');
```
carmenfan: I don't think I agree that pinIcons is a service. Service is more like a independent service that is facilitating something (like SSO, queue, eventManager, maps etc)

Pins is simply a subfeature within custom tickets.

I think the better approach here is to refactor tickets related settings to it's own processor (settings.tickets.js in here) and manage it there.

backend/src/v5/schemas/tickets/templates.constants.js
```diff
@@ -247,6 +247,4 @@ TemplateConstants.supportedPatterns = createConstantMapping([
 
 TemplateConstants.idTypeLabels = createConstantMapping(idTypeLabels);
 
-TemplateConstants.pinIcons = createConstantMapping(['DEFAULT', 'RISK', 'ISSUE', 'MARKER']);
```
carmenfan, line 252: I think this should stay here, you could rename it as defaultPinIcons for anticipation of future change.


In the future it'd be up to the processor to just do a union logic to return the full list.

backend/src/v5/schemas/tickets/templates.js
```diff
@@ -41,9 +40,10 @@ const TemplateSchema = {};
 
 const defaultFalse = stripWhen(Yup.boolean().default(false), (v) => !v);
 const nameSchema = types.strings.title.min(1);
+const pinIconNames = require('../../services/pinIcons').getIconNames();
```
carmenfan: I see what you're trying to do here, but again, it's a bit overkill in the wrong direction.

In the future, it would need to be provided with a teamspace as a parameter to figure out the icon list anyway, so you will still have to run the function at runtime. I wouldn't worry about it for now.Keepp it as it is before

backend/src/v5/services/pinIcons.js
```diff
@@ -0,0 +1,64 @@
...
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+const Path = require('path');
+const fs = require('fs');
+
+const { templates } = require('../utils/responseCodes');
+
+const ICONS_DIR = Path.resolve(__dirname, '../../../resources/pinIcons');
```
carmenfan: add resource into the [`interop`](https://github.com/3drepo/3drepo.io/blob/master/backend/src/interop.js)  and work your path base on that instead

backend/src/v5/services/pinIcons.js
```diff
@@ -0,0 +1,64 @@
...
+const variants = ['normal', 'selected'];
+const filenameRegex = /^(?<name>.+)\.(?<variant>[^.]+)\.svg$/;
+
+const getIconDetails = () => {
```
carmenfan: We want to keep the solution extensible but we don't want to overengineer. This is leaning towards being overengineered.

- Just assume a single directory. if we end up with multiple directories, whoever that's adding the additional directory will have to edit this function anyway. So let's just keep it simple.

- Just run this function once to establish your icon library. It's unlikely we'll store user defined icons this way (more likely to be in the database/fileshare). The list is not going to change with current implementation so just call once and keep the object in memory instead of doing fs operations everytime `getIconNames` and `getIcon` is called.

backend/src/v5/services/pinIcons.js
```diff
@@ -0,0 +1,64 @@
...
+	const directories = [ICONS_DIR];
+	const iconFiles = {};
+
+	for (const directory of directories) {
+		fs.readdirSync(directory, { withFileTypes: true }).forEach((entry) => {
+			if (!entry.isFile()) throw templates.pinIconNotFound;
```
carmenfan: Not only we need ot make sure it's a file, we need to make sure it's a SVG.

Since we only run this once so it won't be too heavy loaded, it is probably a good idea to use the type checker to ensure it's a SVG to minimise malicious intent. https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/utils/helper/typeCheck.js#L62

carmenfan: also tighten the regex (or add additional lines) to prune out unkonwn variants

backend/src/v5/utils/responseCodes.js
```diff
@@ -75,6 +75,7 @@ ResponseCodes.templates = {
 	templateNotFound: { message: 'Template not found.', status: 404 },
 	ticketNotFound: { message: 'Ticket not found.', status: 404 },
 	commentNotFound: { message: 'Comment not found.', status: 404 },
+	pinIconNotFound: { message: 'Pin icon not found.', status: 404 },
```
carmenfan, line 79: We should only throw this if the Pin Icon is not found despite us accepting it's a recognised pinIcon (e.g. you're trying to get the default-selected icon but we struggled to return the file for some reason

If the user gave us an unknown pin type, or unknown variant, we should really throw Invalid arguments stating the reason in a bespoke message (handled in the middleware)

backend/src/v5/services/pinIcons.js
```diff
@@ -0,0 +1,64 @@
...
+		if (missingVariants.length) throw templates.pinIconNotFound;
+	});
+
+	return Object.keys(iconFiles).sort();
+};
+PinIcons.getIcon = (iconName, variant) => {
+	const iconFiles = getIconDetails();
+	if (!iconFiles[iconName] || !iconFiles[iconName][variant]) throw templates.pinIconNotFound;
+	const iconPath = iconFiles[iconName][variant];
+	return fs.readFileSync(iconPath);
```
carmenfan: Use the promise version of the function call.

So it's up to the user whether to wait for it or do something else whilst waiting

backend/tests/v5/e2e/routes/teamspaces/settings.test.js
```diff
@@ -248,6 +248,65 @@ const testGetRiskCategories = () => {
...
+			['teamspace does not exist', false, templates.teamspaceNotFound, { key: basicData.tsAdmin.apiKey, ts: generateRandomString() }],
+			['user is not a member of the teamspace', false, templates.teamspaceNotFound, { key: basicData.normalUser.apiKey, ts: basicData.noTemplatesTS.name }],
+			['user is a member of teamspace', true, undefined, { key: basicData.normalUser.apiKey }],
+		])('', (desc, success, expectedRes, getTestData) => {
+			test(`should ${success ? 'succeed if' : `fail with ${expectedRes.code}`} if ${desc}`, async () => {
+				const { key, ts } = getTestData;
+				const expectedStatus = success ? templates.ok.status : expectedRes.status;
+				const res = await agent.get(route(key, ts)).expect(expectedStatus);
+				if (success) {
+					expect(res.body.icons).toEqual(['DEFAULT', 'ISSUE', 'MARKER', 'RISK']);
```
carmenfan: expose the constants from the src folder and reference it instead of hard coding it here.

backend/tests/v5/e2e/routes/teamspaces/settings.test.js
```diff
@@ -248,6 +248,65 @@ const testGetRiskCategories = () => {
...
+				}
+			});
+		});
+	});
+};
+
+const testGetPinIcon = () => {
+	describe('Pin icon', () => {
+		const basicData = generateBasicData();
+		const route = (key, pinIcon = 'RISK', variant = 'selected', ts = basicData.teamspace.name) => `/v5/teamspaces/${ts}/settings/tickets/pinIcons/${pinIcon}/${variant}${key ? `?key=${key}` : ''}`;
```
carmenfan: pull these from constants as well

backend/tests/v5/e2e/routes/teamspaces/settings.test.js
```diff
@@ -248,6 +248,65 @@ const testGetRiskCategories = () => {
...
+
+		beforeAll(() => setupTestData(basicData));
+
+		describe.each([
+			['user does not have a valid session', false, templates.notLoggedIn, {}],
+			['teamspace does not exist', false, templates.teamspaceNotFound, { key: basicData.tsAdmin.apiKey, ts: generateRandomString() }],
+			['user is not a member of the teamspace', false, templates.teamspaceNotFound, { key: basicData.normalUser.apiKey, ts: basicData.noTemplatesTS.name }],
+			['icon name is invalid', false, templates.pinIconNotFound, { key: basicData.normalUser.apiKey, pinIcon: generateRandomString() }],
+			['icon variant is invalid', false, templates.pinIconNotFound, { key: basicData.normalUser.apiKey, variant: generateRandomString() }],
+			['user is a member of teamspace and normal icon exists', true, undefined, { key: basicData.normalUser.apiKey, variant: 'normal' }],
```
carmenfan: avoid hard coded `normal` and `selected` here

backend/tests/v5/unit/schemas/tickets/templates.test.js
```diff
@@ -1420,6 +1420,32 @@ const testValidate = () => {
 		expect(output.properties[1].hiddenOnUI).toBeUndefined();
 	});
 
+	test('should accept every built-in pin icon in direct and conditional mappings', () => {
+		['DEFAULT', 'RISK', 'ISSUE', 'MARKER'].forEach((icon) => {
```
carmenfan: avoid hard coding

backend/tests/v5/unit/schemas/tickets/templates.test.js
```diff
@@ -1420,6 +1420,32 @@ const testValidate = () => {
...
+			const directIconTemplate = generateBasicSchema({
+				properties: [{
+					name: generateRandomString(),
+					type: propTypes.COORDS,
+					icon,
+				}],
+			});
+			const conditionalIconTemplate = generateBasicSchema({
```
carmenfan: you can merge the 2 template into one and test it together

backend/tests/v5/unit/schemas/tickets/templates.test.js
```diff
@@ -1420,6 +1420,32 @@ const testValidate = () => {
 		expect(output.properties[1].hiddenOnUI).toBeUndefined();
 	});
 
+	test('should accept every built-in pin icon in direct and conditional mappings', () => {
```
carmenfan: It's probably better to add a test group for this and have a test for each type (so if it breaks it's clear which icon name broke)

i.e.  somethign like this (but probably should rename the group)
```suggestion
	describe('should accept every built-in pin icon in direct and conditional mappings', () => {
	   ['DEFAULT', 'RISK', 'ISSUE', 'MARKER'].forEach((icon) => {
	     test(icon, () => { 

```

backend/tests/v5/unit/services/pinIcons.test.js
```diff
@@ -0,0 +1,83 @@
...
+const { templates } = require(`${src}/utils/responseCodes`);
+
+const PinIcons = require(`${src}/services/pinIcons`);
+
+const toFsEntry = (name, isFile = true) => ({
+	name,
+	isFile: jest.fn().mockReturnValue(isFile),
+});
+
+const fullIconNames = ['RISK.selected.svg', 'DEFAULT.selected.svg', 'ISSUE.selected.svg', 'MARKER.selected.svg', 'RISK.normal.svg', 'DEFAULT.normal.svg', 'ISSUE.normal.svg', 'MARKER.normal.svg'];
```
carmenfan: avoid hard coding

backend/tests/v5/unit/services/pinIcons.test.js
```diff
@@ -0,0 +1,83 @@
...
+
+const testBuiltInProvider = () => {
```
carmenfan: This is a unit test, so the test grouping should match how we've formatted in other files (i.e. one function per test function)

backend/tests/v5/unit/services/pinIcons.test.js
```diff
@@ -0,0 +1,83 @@
...
+	describe('getIconNames', () => {
+		test('should discover complete icon pairs in alphabetical order', () => {
+			expect(PinIcons.getIconNames()).toEqual(['DEFAULT', 'ISSUE', 'MARKER', 'RISK']);
```
carmenfan: too much hard coding!

carmenfan (changes requested):

backend/src/v5/routes/teamspaces/settings.js
Bogdan-Crn: My first time doing that was with a middleware because it made sense but then I found that there was repeating code without a more elegant way of simplifying it. I thought of a utils function but was not sure where to put it and then I realized that I would like to know the path of the icon if in the future we want to allow more custom ones to be updated.

backend/src/v5/processors/teamspaces/settings.js
Bogdan-Crn: Initially I thought about putting some logic in settings but would we want the disk reading feature to sit in processor?

backend/src/v5/schemas/tickets/templates.constants.js
Bogdan-Crn: is the future change (adding more pin icons) going to sit in the same structure /resources/tickets/pinIcons or are we looking at storing it somewhere else?

carmenfan: for the next issue where we're adding more icons, i'd anticipate it to be the same place, just more files in the folder.

carmenfan: in the far far future it's not going to be static files, and it would be customer data dependant on teamspace, so it wont' sit in the repo anyway. and will have to go through the files manager. don't worry about that for this issue.

backend/src/v5/processors/teamspaces/settings.js
carmenfan: This is no different to importing source code, these are static images we're referencing in the code page. so it doesn't matter where it sits. Processor is fine, ticket template schema is also fine.

backend/src/v5/routes/teamspaces/settings.js
carmenfan: You can get the middleware to import code from the processor if required. but if you left the constants in the template schema i don't think you have this issue?


By the time you get to the processor, you don't have to redo these checks; just fetch the file.

backend/src/v5/processors/teamspaces/settings.tickets.js
```diff
@@ -0,0 +1,49 @@
...
+TicketSettings.getTemplateList = (teamspace) => getAllTemplates(
+	teamspace, true, { _id: 1, name: 1, code: 1, deprecated: 1 },
+);
+TicketSettings.getPinIconNames = () => {
+	const iconDetails = getDefaultPinIconsDetails();
+	return Object.keys(iconDetails).sort();
+};
+
+TicketSettings.getPinIcon = (iconName, variant) => {
+	const iconDetails = getDefaultPinIconsDetails();
```
carmenfan: this check feels redundant.

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/commons/tickets.js
```diff
@@ -192,4 +233,16 @@ TicketsMiddleware.validateUpdateMultipleTickets = validateMany([
 
 TicketsMiddleware.templateExists = checkTicketTemplateExists;
 
+TicketsMiddleware.checkPinIconExists = async (req, res, next) => {
+	const { pinIcon, variant } = req.params;
+	const pinIconsDetails = getDefaultPinIconsDetails();
```
carmenfan: this middleware should get the details from tickets settings processor

backend/src/v5/routes/teamspaces/settings.js
```diff
@@ -26,8 +26,10 @@ const Audit = require('../../processors/teamspaces/audits');
 const MimeTypes = require('../../utils/helper/mimeTypes');
 const { Router } = require('express');
 const TeamspaceSettings = require('../../processors/teamspaces/settings');
+const TicketSettings = require('../../processors/teamspaces/settings.tickets');
 const { UUIDToString } = require('../../utils/helper/uuids');
 const { castTicketSchemaOutput } = require('../../middleware/dataConverter/outputs/teamspaces/settings');
+const { checkPinIconExists } = require('../../middleware/dataConverter/inputs/teamspaces/projects/models/commons/tickets');
```
carmenfan: wrong place for the middleware

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/commons/tickets.js
carmenfan, line 1: the change is in the wrong place this should be in `inputs/teamspaces/settings.js` or `inputs/teamspaces/settings.tickets.js`

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/commons/tickets.js
```diff
@@ -232,4 +233,16 @@ TicketsMiddleware.validateUpdateMultipleTickets = validateMany([
 
 TicketsMiddleware.templateExists = checkTicketTemplateExists;
 
+TicketsMiddleware.checkPinIconExists = async (req, res, next) => {
+	const { pinIcon, variant } = req.params;
```
carmenfan: This is a bit over-complicated. I don't think `PinIconDetails` as an object needs to exist and be passed around.

Template constants need to provide a list of available icon types, and we need to know what the known and accepted variants in the application (`normal`, `selected`) - this is already an available constant in templates.constants

We do not expect an icon to have a variant that is not `normal` or `selected`

If it is a known icon type, then we expect both variants to be available. It doesn't make sense for us to throw invalid arguments to the user as if they did something wrong, just because we can't find a file (this should be a 404)


In short, this should just be 

- is pinIcon in the list of known icons
- is variant in the list of known variants

backend/src/v5/processors/teamspaces/settings.js
```diff
@@ -1,5 +1,5 @@
 /**
- *  Copyright (C) 2022 3D Repo Ltd
+ *  Copyright (C) 2026 3D Repo Ltd
```
carmenfan: the date should not change here. it represents the year that this file was created.

backend/src/v5/processors/teamspaces/settings.tickets.js
carmenfan: at this point, we can assume the middleware has done the necessary check and all you need to do is read the file.

If we can't find the file, throw `templates.pinIconNotfound`. 

Dont' bother re-checking if it is a known icon/variant.

backend/src/v5/processors/teamspaces/settings.tickets.js
```diff
@@ -0,0 +1,49 @@
...
+TicketSettings.updateTicketTemplate = async (teamspace, id, data) => {
+	await updateTemplate(teamspace, id, data);
+	publish(events.TICKET_TEMPLATE_UPDATED, { teamspace, template: id, data });
+};
+TicketSettings.getTemplateList = (teamspace) => getAllTemplates(
+	teamspace, true, { _id: 1, name: 1, code: 1, deprecated: 1 },
+);
+TicketSettings.getPinIconNames = () => {
+	const iconDetails = getDefaultPinIconsDetails();
+	return Object.keys(iconDetails).sort();
```
carmenfan: I'm seeing you sorting this in various places... isn't it better to just sort it once when you establish the list in templates.constants?

Then you can just assume it is sorted there after?

backend/src/v5/schemas/tickets/templates.constants.js
```diff
@@ -248,6 +254,38 @@ TemplateConstants.supportedPatterns = createConstantMapping([
...
+const defaultPinIconsDetails = {};
+const iconVariants = ['normal', 'selected'];
+
+const iconFilenameRegex = new RegExp(
+	`^(?<name>.+)\\.(?<variant>${iconVariants.join('|')})\\.svg$`,
```
carmenfan: you'll need a `// eslint-disable-next-line <name of rule> `to stop the warning

backend/src/v5/schemas/tickets/templates.constants.js
```diff
@@ -248,6 +254,38 @@ TemplateConstants.supportedPatterns = createConstantMapping([
...
+);
+
+TemplateConstants.iconVariants = createConstantMapping(iconVariants);
+
+const setDefaultPinIcons = () => {
```
carmenfan: `set` traditionally implies we're doing a write... so I feel like this function could be named a bit better.

This is more like an investigative approach to figure out what the default pins are,  so perhaps `init ', `determine` or even `get` is more fitting?

backend/src/v5/schemas/tickets/templates.constants.js
```diff
@@ -248,6 +254,38 @@ TemplateConstants.supportedPatterns = createConstantMapping([
...
+
+		const match = entry.name.match(iconFilenameRegex);
+		if (!match) throw templates.pinIconNotFound;
+
+		const { name, variant } = match.groups;
+		if (!defaultPinIconsDetails[name]) {
+			defaultPinIconsDetails[name] = { [variant]: path.join(ICONS_DIR, entry.name) };
+			return;
+		}
+		defaultPinIconsDetails[name][variant] = path.join(ICONS_DIR, entry.name);
```
carmenfan: as mentioned in the other file. I think we just want ot store a pinIconTypes array with the name of the icons.

Once we've established an icon type has all the variants we require, we want to ditch this object.

backend/src/v5/schemas/tickets/templates.constants.js
```diff
@@ -248,6 +254,38 @@ TemplateConstants.supportedPatterns = createConstantMapping([
...
+	});
+};
+
+TemplateConstants.getDefaultPinIconsDetails = () => {
```
carmenfan: The name should change here due to the comments above, but I don't think we need this to be a function. Just let it process at import time and keep the static list.

```suggestion
TemplateConstants.getDefaultPinIconsDetails = setDefaultPinIcons();
```

backend/src/v5/schemas/tickets/templates.js
```diff
@@ -43,7 +43,8 @@ const defaultFalse = stripWhen(Yup.boolean().default(false), (v) => !v);
 const nameSchema = types.strings.title.min(1);
 
 const createPinConditionalSchema = (type) => {
-	const dataSchema = type === 'icon' ? Yup.string().oneOf(Object.values(pinIcons)) : types.color3Arr;
+	const pinIconNames = Object.keys(getDefaultPinIconsDetails()).sort();
```
carmenfan: can ditch the sort if you sort it in templates.constants. It's actually not even important whether it's sorted or not 😆

carmenfan (changes requested):

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/commons/tickets.js
Bogdan-Crn: What if someone uses the API and puts in the wrong icon name or a variant that does not exist wouldn't a 400 request be more appropriate?
I can do a check and see if the pinIcon is in the list of known icons and the same with the variant
But shouldn't the 404 be thrown by the processor if it's trying to read the file and is not there? Especially since we are looking at in the future storing custom pinIcons uploaded by users.

backend/src/v5/processors/teamspaces/settings.js
Bogdan-Crn: Yeah, don't really know how that got updated since I didn't touch that line, I'll revert it now.

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/models/commons/tickets.js
carmenfan: That's exactly what I'm trying to say:


The middleware check should be these 2:
- is pinIcon in the list of known icons
- is variant in the list of known variants

If either of them is false, it should throw invalid arguments.

But if both are fine, then processor throws 404

The main thing here is that we don't need the object to store which icon has which variants. we expect all icons to have all known variants. so we just need a list of icon names, and the constant for variants.

backend/src/v5/schemas/tickets/templates.constants.js
Bogdan-Crn: function name can change but we are writing the names of the icons in the cache so I thought it made sense.

backend/src/v5/processors/teamspaces/settings.tickets.js
```diff
@@ -0,0 +1,49 @@
...
+TicketSettings.getTemplateList = (teamspace) => getAllTemplates(
+	teamspace, true, { _id: 1, name: 1, code: 1, deprecated: 1 },
+);
+TicketSettings.getPinIconNames = getDefaultPinIconNames;
+
+TicketSettings.getPinIcon = async (iconName, variant) => {
+	const iconPath = path.join(iconsDir, `${iconName}.${variant}.svg`);
+
+	try {
+		return await readFile(iconPath);
```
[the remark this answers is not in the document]
Bogdan-Crn: Once it gets to that point we are sure that the iconName and variant are checked and nothing that we don't expect.

carmenfan:
@sebjf (as discussed in the meeting), could we please have a version of the viewer that lets us adjust the icon size in this branch? Thank you!

sebjf:
Hi @carmenfan, I have pushed a viewer with the ability to control the pin size. To do so call `UnityUtil.toUnity('SetPinScale',3,1.5)` The last parameter is the pin size. The current/default size is 1.5.

carmenfan (dismissed):
The backend looks fine now. 

We're waiting for a unity update to change the default size of the icon and potentially address some artefact problem that @sebjf's team is looking into

PR #6198, opened 2026-08-11
Issue 6117 - Clash info added to daily digest

### LINKED ISSUE #6117: Add clash run results into daily digest emails
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/998

Currently, we send daily digest emails to users who are subscribed to provide them information about tickets that are related to them.

In that daily digest, we would like to add additional information about clash runs that has happened within that day. A copy of a mock up of the email can be found in the product issue.

The clash run information should only be sent to the users/jobs who are listed in the clash configuration to receive the notification. If there are no one configured on the clash plan, no one should be notified. This logic should work similarly to the ticket notifications.

This means we will have to add an additional configuration within the clash plan:

````js
{
   name: "pipes vs walls",
   notify: ["user1", "user2", "job1"] //(This is optional)
   // ... other clash plan information
}
````

In the example above, user1, user2 and any users within job1 should be notified about the results of the clash run **provided they have access to the project**.

Similar to previously, email digest should only be sent if it's enabled within the teamspace.

### Goals
- [ ] As a clash manager, I want to be able to specify who should be notified about the results of the clash runs on my clash plan
- [ ] As a person who should be notified about clash runs, I want to see the information about the clash runs within the daily digest email.

### Tasks
- [x] Add the new property in clash plans (1 hr)
- [x] Add jobs and users validation  (0.5 day)
- [x] Add a new notification in the table for each new clash run added (0.5 day)
- [x] Amend the dailyDigest script and functions to include the clash data (4 days)
- [x] Amend the email template schema (0.5 day)
- [x] Amend the email template html (2 days)
- [x] Amend all unit and e2e tests (2 days)

### PR DESCRIPTION
**Pull request by ChristosTsiotsias**
This fixes #6117 

#### Description
Clash information has been added to the daily digest emails.
Clash Plans now include a notify property which is an array of jobs and users and defines which users should receive emails.
The clash information is regarding clash runs that have been executed during the past 24 hours and includes the time, the status and the results or error reason of the clash run.
Unit and e2e tests have been added.

#### Acceptance Criteria
Create a clash plan and populate the notify array.
Start some clash runs based on this plan.
Manually send a daily digest email for this teamspace.
The email should include clash data as long as any ticket data.

### CODE REVIEW

File: backend/src/v5/models/notifications.js [hunk 1]
```diff
@@ -103,41 +112,178 @@ const getGroupedNotificationsByQuery = (query) => {
...
+									data: '$data',
+								},
+							},
+						},
+					},
+					// Normalize the shape before merging both facets.
+					{
+						$project: {
+							_id: 1,
+							ticketData: 1,
```
carmenfan: if we rename both ticketData and clashData to just `data` then we don't really need this step?

backend/src/v5/models/notifications.js
```diff
@@ -103,41 +112,178 @@ const getGroupedNotificationsByQuery = (query) => {
...
+							runs: {
+								$push: {
+									status: '$data.status',
+									stats: '$data.results.stats',
+									error: '$data.results.error',
+									triggeredAt: '$data.triggeredAt',
+								},
+							},
+							planName: {
+								$first: '$data.planName',
```
carmenfan: plan name could've changed since teh notification so it's better if we pull the plan name from the plan document. (like we do with projectId and model Id)

backend/src/v5/models/notifications.js
```diff
@@ -103,41 +112,178 @@ const getGroupedNotificationsByQuery = (query) => {
...
+							_id: 1,
+							ticketData: 1,
+							clashData: { $literal: [] },
+						},
+					},
+				],
+				clashNotifications: [
+					// Only process completed clash-run notifications.
+					{
+						$match: { type: 'CLASH_RUN_COMPLETED' },
```
carmenfan: i think it's better to match data.plan exists

So If we have more clash notifications in the future we don't have to change this part.

backend/src/v5/models/notifications.js
```diff
@@ -103,41 +112,178 @@ const getGroupedNotificationsByQuery = (query) => {
...
+								teamspace: '$data.teamspace',
+								project: '$data.project',
+								plan: '$data.plan',
+							},
+							runs: {
+								$push: {
+									status: '$data.status',
+									stats: '$data.results.stats',
+									error: '$data.results.error',
+									triggeredAt: '$data.triggeredAt',
```
carmenfan: I think it's better to just push `data` in so it's not rigid (and more extendable with different data type if we expand it to more clash notifications.

backend/src/v5/models/notifications.js
```diff
@@ -103,41 +112,178 @@ const getGroupedNotificationsByQuery = (query) => {
...
+						$match: { type: 'CLASH_RUN_COMPLETED' },
+					},
+					// Group clash runs under each plan.
+					{
+						$group: {
+							_id: {
+								user: '$user',
+								teamspace: '$data.teamspace',
+								project: '$data.project',
+								plan: '$data.plan',
```
carmenfan: it's a little bit odd to me that we group by plans before we group by project.

Was there any significance to the order?

backend/src/v5/models/notifications.js
```diff
@@ -103,41 +112,178 @@ const getGroupedNotificationsByQuery = (query) => {
...
 		{
-			// group by user/teamspace/project/model/type, count up the tickets
-			$group: {
-				_id: {
-					user: '$user',
-					teamspace: '$data.teamspace',
-					project: '$data.project',
-					model: '$data.model',
-					type: '$type',
+			$facet: {
```
carmenfan: This is becoming very long and complex, and I'm wondering if it's necessary. 

A quick Google search tells me that Mongo can group things into properties where they may not exist for some objects. Could we have the pipeline to group by both planId and modelId and keep everything generic (with the digest logic handling the details)? 

then entries with planId we'd know is clash runs entries, and the ones with modelIds is ticket entries.

backend/src/v5/models/notifications.js
```diff
@@ -51,6 +51,15 @@ const generateNotification = (type, user, data) => ({
 	data,
 });
 
+Notifications.insertClashNotifications = async (teamspace, project, notificationData, recipients) => {
+	const records = recipients.map((recipient) => generateNotification(notificationTypes.CLASH_RUN_COMPLETED,
```
carmenfan: 2 thoughts here

- We should have different notifications for different states (Success, failed, aborted). 
- And they should have their own entry functions here.

The notification type would depict the status; you don't really need to pass the status down. This woud be more inline with how you expect notifications to work i think (you don't expect a failure and success to look the same if you have a UI for notifications)

backend/src/v5/models/notifications.js
carmenfan: I haven't 100% thought through this so don't take it as an instruction to change. Have a think and if you don't think it'll work, let's discuss it further.

backend/src/v5/processors/teamspaces/projects/clashes.js
```diff
@@ -268,6 +268,7 @@ Clashes.createRun = async (teamspace, project, plan, user) => {
 		selfIntersectionsCheck: plan.selfIntersectionsCheck,
 		selectionA: plan.selectionA,
 		selectionB: plan.selectionB,
+		notify: plan.notify,
```
carmenfan: We don't want this in the clash run version of the config.

Who we notify should be depicted by the latest version of the plan, not at the point of this run was triggered.

backend/src/v5/middleware/dataConverter/inputs/teamspaces/projects/clashes.js
```diff
@@ -60,6 +62,19 @@ const generatePlanSchema = (teamspace, project, user, isUpdate) => {
...
+		.test('jobs-and-users-validation', 'All jobs and users must have access to the project', async (values) => {
+			if (!values?.length) {
+				return true;
+			}
+
+			const usersWithAccess = await getUsersWithAccess(teamspace, project);
+			const jobsWithAccess = await getJobsByUsers(teamspace, usersWithAccess);
+			const jobsAndUsers = new Set([...usersWithAccess, ...jobsWithAccess]);
+
+			return values.every((value) => jobsAndUsers.has(value));
```
carmenfan: Instead of doing this on a test, can we just do [`oneOf`](https://github.com/jquense/yup#schemaoneofarrayofvalues-arrayany-message-string--function-schema-alias-equals)?

Given that we generate the schema on the fly anyway, it doesn't make much sense to run this test at runtime.

Can we just get it to give you a bespoke message if we don't want to return the general one with a long list of options?

backend/src/v5/services/eventsListener/components/clashEvents.js
```diff
@@ -139,6 +153,21 @@ const clashRunCompleted = async (payload) => {
 			payload,
 			error,
 		}, undefined, true);
+	} finally {
+		const { plan, results: completeResults, status, triggeredAt } = await getClashRunById(teamspace,
```
carmenfan: Let the notifications service handle the creation of notifications (see services/notifications/tickets.js)

We can listen for the same events or trigger new ones, whichever is appropriate.

You can see there's already a getUserList function in tickets.js in the notification service. If we refactor that into a utils file in services/notifications, we can also avoid duplicate code.

This would also get rid of the copilot comment about job users are getting notifications despite them not having access to the project.

backend/src/v5/services/eventsListener/components/clashEvents.js
```diff
@@ -139,6 +153,21 @@ const clashRunCompleted = async (payload) => {
...
+			project, runId, { plan: 1, results: 1, status: 1, triggeredAt: 1 });
+
+		if (plan.notify?.length) {
+			const recipients = await getNotificationRecipients(teamspace, plan.notify);
+			const { name } = await getPlanById(teamspace, project, plan._id, { name: 1 });
```
carmenfan: as mentioned in the other file, don't store the name.

we want to provide the latest name as depicted in the plan data when we send the email.

backend/src/v5/services/mailer/templates/dailyDigest.js
```diff
@@ -30,14 +31,32 @@ const dataSchema = Yup.object({
...
+				status: Yup.string().required(),
+				stats: Yup.object({
+					new: Yup.number().min(0),
+					active: Yup.number().min(0),
+					resolved: Yup.number().min(0),
+				}),
+				error: Yup.object({
+					reason: Yup.string(),
+				}),
+				triggeredAt: Yup.string().transform((_, originalValue) => new Date(originalValue)
```
carmenfan: If we're converting this to a locale string, we should check the user's country of origin and base it on that (same with timezone). and let the default be a server config (we don't really want to display the UK default on a CA deployment, for example)

carmenfan (changes requested):

carmenfan (approved):

PR #6233, opened 2026-08-27
ISSUE #6225 Closing and re-opening the group tab with an unsaved group is turning the title into a placeholder

### LINKED ISSUE #6225: Closing and re-opening the group tab with an unsaved group is turning the title into a placeholder
### Description
Closing and re-opening the group tab with an unsaved group is turnign the title into a placeholder

### Steps to replicate
- create a v4 group, with a modified title without saving
- hide the tab by clicking on the groups icon
- show the tab again

### Current Behaviour
- it is reappearing with the title greyed out and replacable like a placeholder

(note: you can also see there's a split second where it turns black and not place holder like when the tab is about to close...)

### Expected Behaviour
- As this is a title I have added, it should never appear as a placeholder.

Product ticket: https://github.com/3drepo/3D-Repo-Product-Team/issues/1013

The-Daniel:
This fixes #6225 

#### Description
Change TextField in PreviewDetails to be the newer v5 component. This means it no longer has the functionality where on first click the autogenerated title is erased so the user can write their own. This was the source of the bug as when reopening the groups tab there was no way of it knowing whether the title was the user's or if it were autogenerated. 

I also noticed that when typing in this field that redux actions were getting called on every keystroke. So I changed it to trigger on blur. This was also happening in a few other places so it is now fixed there too (groups description, issues/risks titles)

### DIFF
--- a/frontend/src/v4/routes/viewerGui/components/groups/components/groupDetails/groupDetailsForm.component.tsx
+++ b/frontend/src/v4/routes/viewerGui/components/groups/components/groupDetails/groupDetailsForm.component.tsx
@@ -113,7 +113,7 @@ export class GroupDetailsForm extends PureComponent<IProps, any> {
 					<Field name="desc" render={({ field, form }) => (
 						<Description
 							{...field}
-							onBlur={this.handleFieldChange(field.onChange, form)}
+							onChange={this.handleFieldChange(field.onChange, form)}
 							validationSchema={GroupSchema}
 							fullWidth
 							multiline
--- a/frontend/src/v4/routes/viewerGui/components/groups/components/groupDetails/groupDetailsForm.component.tsx
+++ b/frontend/src/v4/routes/viewerGui/components/groups/components/groupDetails/groupDetailsForm.component.tsx
@@ -113,7 +113,7 @@ export class GroupDetailsForm extends PureComponent<IProps, any> {
 					<Field name="desc" render={({ field, form }) => (
 						<Description
 							{...field}
-							onChange={this.handleFieldChange(field.onChange, form)}
+							onBlur={this.handleFieldChange(field.onChange, form)}
 							validationSchema={GroupSchema}
 							fullWidth
 							multiline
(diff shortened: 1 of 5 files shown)


frontend/src/v4/routes/viewerGui/components/groups/components/groupDetails/groupDetailsForm.component.tsx
```diff
@@ -113,7 +113,7 @@ export class GroupDetailsForm extends PureComponent<IProps, any> {
 					<Field name="desc" render={({ field, form }) => (
 						<Description
 							{...field}
-							onChange={this.handleFieldChange(field.onChange, form)}
+							onBlur={this.handleFieldChange(field.onChange, form)}
```
sanmont3drepo: Now if I change the description of the group the save button keeps disabled, so I can't be updated now.

sanmont3drepo: also its kind of weird that if I change the value it doesnt update the enable state of the button until blur. In this regard I think it was better before withh the onchange thing

frontend/src/v4/routes/viewerGui/components/issues/components/issueDetails/issueDetails.component.tsx
```diff
@@ -176,7 +176,7 @@ export class IssueDetails extends PureComponent<IProps, IState> {
 					key={this.issueData._id}
 					defaultExpanded={horizontal || expandDetails}
 					editable={this.canEditBasicProperty}
-					onNameChange={this.handleNameChange}
+					onNameBlur={this.handleNameBlur}
```
sanmont3drepo: I think the blur thing doesnt work well still with issues and risks because there's a save button in the bottom. It's different to how tickets work.

sanmont3drepo (changes requested):

sanmont3drepo (approved):

PR #6254, opened 2026-09-07
ISSUE #6253 - When recalibrating a drawing the 2D calibration arrow does not appear  

### LINKED ISSUE #6253: When recalibrating a drawing the 2D calibration arrow does not appear
### Description
When recalibrating a drawing to a 3D model we expect to see the calibration arrows for the current calibration. We do see this for the 3D calibration however for the 2D it is missing.

NOTE: This is similar to bug #6241 but is ultimately a different bug

### Steps to replicate
- Calibrate a drawing to a model
- Click recalibrate drawing and set it to the same model again

### Current Behaviour
- The 3D calibration arrow is visible
- The 2D calibration arrow is not visible
- The user can progress with the calibration without knowing what the 2D calibration vector is

### Expected Behaviour
- Both calibration arrows are visible

Image
Note that the continue button is enabled, so the 2D calibration does exist, it's just not being shown

The-Daniel:
This fixes #6253 

#### Description
The arrow was only being set by the value of the calibration on page load. If the details for the calibration were not fetched yet then it would never update the arrow once the information was fetched. I have changed it so it will now update when the calibration is fetched


sanmont3drepo:
Couldnt reproduce the issue

sanmont3drepo:
Found a way to reproduce it in staging:

- open a calibrated drawing  (should show on the the righthand side), but dont enter calibration
- refresh
- try to calibrate

### what happens
- the 2d vector is not showing

### what should happen
- the 2s vector is showing

sanmont3drepo (approved):

